PR #68, opened 2013-04-19
Count how long tasks take & add a --num-tasks option

rory:
Added a TaskStats class to count (a la RequestStats) how long tasks take to complete.

Lotsa refactoring here. Refactored RequestStats to be a subclass of (new) Stats class which does the actual counting. Lotsa s/min_request_time/min_time/ type changes needed. TaskStats just uses this class. This makes it easy to add new stats later. Task stats are printed & summarised at end.

Added a ``-t``/``--num-tasks`` option, like ``-r``/``-num-requests`` to stop after a certain number of requests are finished.

All tests pass, but I didn't add any for TaskStats. Didn't do anything with distributed taskstats, so it probably won't work. We haven't started using distributed testing yet.

The motivation behind this is that we're trying to test & measure a user flow that takes many steps. We want to know how long it takes to complete this flow. Also ``--num-tasks`` is more useful to us than ``--num-requests``.


rory:
I think I'll close this. I'm no longer actively looking at this.

PR #113, opened 2013-12-03
remove trailing whitespace and unused imports

sanga:
This kind of thing is sometimes a bit contentious but now that you guys have cleaned up the line endings I wonder might you be interested in a few more small cleanups?

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -6,7 +6,6 @@ monkey.patch_all(thread=False)
 from time import time
 import sys
 import random
-import warnings
 import traceback
 import logging
 
--- a/locust/rpc/socketrpc.py
+++ b/locust/rpc/socketrpc.py
@@ -3,7 +3,6 @@ import logging
 
 import gevent
 from gevent import socket
-from gevent import queue
 
 from locust.exception import LocustError
 from .protocol import Message
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1,6 +1,5 @@
 # coding=UTF-8
 import socket
-import traceback
 import warnings
 import random
 import logging
--- a/locust/web.py
+++ b/locust/web.py
@@ -13,7 +13,6 @@ import runners
 from runners import MasterLocustRunner
 from locust.stats import median_from_dict
 from locust import version
-import gevent
 
 import logging
 logger = logging.getLogger(__name__)
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 from setuptools import setup, find_packages, Command
-import sys, os
+import subprocess
 
 version = '0.7.0'
 
@@ -16,7 +16,6 @@ class Unit2Discover(Command):
         pass
 
     def run(self):
-        import sys, subprocess
         basecmd = ['unit2', 'discover']
         errno = subprocess.call(basecmd)
         raise SystemExit(errno)


acatton:
What the hell??!! This discussion is baffling me...

I use this project, and I wanted to do a pull request because there was some misunderstanding of how polymorphism work in the code. And I open the code, and my editor is all red every where because of trailing white spaces.

I was going do grep/sed the whole code to remove it because this is nonsense, it's useless and noisy. And the copy/pasting code is just such a bad argument, if you're testing by running code in your REPL, you're very likely doing it wrong.

 * http://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal
 * http://www.dinduks.com/why-are-trailing-whitespaces-bad/
 * https://blog.codinghorror.com/whitespace-the-silent-killer/

Is there any change in opinion here?

heyman:
Hi!

I think it's mainly an issue of taste and personal preference. And probably tightly coupled with the type of editor one uses. I - who have written a large part of the Locust code base -  happen to use one where it's beneficial to have spaces that match the current indentation on "empty" lines. The copy & paste thing is just a bonus.

Also, I don't like the idea of messing up the git history with one massive whitespace commit (I know one can ignore whitespace changes, but it's not done by default).

> http://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal

Didn't read the whole thread, but the accepted answer mainly talks about trailing whitespace on lines with non-whitespace characters, causing the caret to not be placed at the end when hitting the end button. This I don't think Locust has much of. It also happens to actually be one of the reasons why I like whitespace that match the current indentation level on "empty" lines. When I hit end (or CTRL-E in my case) I expect the caret to be placed at the current indentation level.

> http://www.dinduks.com/why-are-trailing-whitespaces-bad/

Basically the same argument (talks about trailing whitespace on lines with non-whitespace characters). Also mentions file sizes which I just find ridiculous.

> https://blog.codinghorror.com/whitespace-the-silent-killer/

Not convinced by him either.


This, like the tabs vs spaces, can probably be debated endlessly, but I think it's hard to argue that one way is objectively better than another.

acatton:
Hey,

Sorry for what might have seem to be aggressiveness. But I am genuinely astonished by this whole thread, and the fact that we're still having a discussion about it. I really thought that removing trailing white spaces was a given good practice in any programmer's mind. And I would expect it to be followed by any Python programmer, since [the pep 0008 highly recommends against them](https://www.python.org/dev/peps/pep-0008/#other-recommendations). 

> I - who have written a large part of the Locust code base - happen to use one where it's beneficial to have spaces that match the current indentation on "empty" lines

It sounds like your editor is misconfigured. I mean maybe your editor is not correctly configure, but you're using git right? And when looking at the diff before committing, if you have git colors enabled, you should see big red strips of white spaces everywhere, isn't that bothering you?

> Also, I don't like the idea of messing up the git history with one massive whitespace commit (I know one can ignore whitespace changes, but it's not done by default).

I understand that, but if a mistake has been made, it has to be fixed. It would be like you have some part of the code that works but has very bad structure (nested ifs with mutation and difficulty to track): by re-factoring this code you would change the code totally and claim git history, but you would keep the behavior/feature from the original author. I'm pretty sure you wouldn't accept the argument "well we don't want to mess up the git history, let's keep this legacy code".

And by not fixing this (2 years ago now) the problem just kept building up. Now I have to change my editor configuration, because when saving I remove all trailing white spaces on my files.

> Didn't read the whole thread, but the accepted answer mainly talks about trailing whitespace on lines with non-whitespace characters, causing the caret to not be placed at the end when hitting the end button. This I don't think Locust has much of.

It has, [the whole files that contains the class that holds statistics has them.](https://github.com/locustio/locust/blob/3cddb18621f2ca905cc9d6f239f0f7e95b230143/locust/stats.py#L46)

> It also happens to actually be one of the reasons why I like whitespace that match the current indentation level on "empty" lines. When I hit end (or CTRL-E in my case) I expect the caret to be placed at the current indentation level.

Again, I don't want to judge, but it sounds like a misconfiguration in your editor. It looks like you're missing some kind of "smart indentation" plugin or something like this. (I dunno what is your editor so I can't tell if it's a plugin or a configuration value.)

> This, like the tabs vs spaces, can probably be debated endlessly, but I think it's hard to argue that one way is objectively better than another.

The tab vs spaces I understand it. I know some people don't, but people have different opinions, but anybody can play with both IMHO. But the trailing white space thingy, is really noisy, all editor and source control manager highlights them. The PEP-0008 recommends against it, I don't know what to say...

I'm sorry. Still a little bit in chock.

mfilenko:
Sacrificing beauty for the sake of the history sounds reasonable, but Git [did it](https://github.com/git/git/commit/a6080a0a44d5ead84db3dabbbc80e82df838533d). Just saying ¯\\\_(ツ)_/¯

acatton:
@mfilenko how is this related to this ticket?

mfilenko:
@acatton,

Another two cents to reconsider.

PR #169, opened 2014-06-26
Added event hatch_starting for pre-simulation setup.

jharriman:
One of the use cases I wanted to use Locust for required that I do some initial setup before beginning the simulation. Two or more Locust classes needed to have knowledge of how many instances of another type of Locust class and that number could not be gathered from the hatch_complete event. So I added an event that fires right when STATE_HATCHING is entered so that initial setup could be done before running the simulation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -74,5 +74,5 @@ Available hooks
 The following event hooks are available under the **locust.events** module:
 
 .. automodule:: locust.events
-	:members: request_success, request_failure, locust_error, report_to_master, slave_report, hatch_complete, quitting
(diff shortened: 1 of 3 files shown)


tom-dalton-fanduel:
I'm also interested in additional event hooks, would be nice to have this merged.

mbarrien:
I know this is old, but this looks to me to be equivalent to the existing `events.locust_start_hatching`.

jharriman:
@mbarrien It does indeed look like my change was superseded by a different commit (bdafcd93f9e1a4d087a5ee452e96a15d2054157b). 

It is notable that this PR pre-dated that commit by a couple months. I would have enjoyed improving this commit then, but it doesn't appear that the owners have been that active with managing the incoming PRs for this project.

PR #175, opened 2014-07-09
Don't reset stats after hatching is complete

drewtunes:
The stats reset when hatching starts already and if you hatch slowly, there can be many tasks completed by the time hatching completes. Without this change, those completed tasks are removed from the stats as soon as hatching completes.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -40,14 +40,8 @@ class LocustRunner(object):
         
         def on_hatch_complete(user_count):
             self.state = STATE_RUNNING
-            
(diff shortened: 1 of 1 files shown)


spearsem:
Can this be made configurable? Some load tests may be used to see how throughput and latency vary as the number of users increases, up to an expected saturation point, and traffic prior to fully hatched point will still be meaningful (and will not suffer from unwarm caches, etc., by design).

heyman:
@spearsem It is configurable in Locust >= 0.8. Use the `—no-reset-stats` option.

PR #177, opened 2014-07-12
Include percentile response times in the requests stats csv

fordhurley:
This is the second part of PR #160:

> Include the percentile response times in the main "requests" CSV. This change could be a bit more controversial, but I don't see the reason to have to download two different CSVs containing the same set of rows, but with different data. The tricky part about this change is that it obviates the need for the percentile distribution CSV entirely, but removing it could break existing implementations. Also, there would no longer be a need both the "Max response time" and "100%" columns as these are duplicates. I left both in for backwards compatibility with any tools that analyze the CSVs directly.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -562,11 +562,20 @@ def requests_csv():
             '"Max response time"',
             '"Average Content Size"',
             '"Requests/s"',
+            '"50%"',
+            '"66%"',
+            '"75%"',
+            '"80%"',
+            '"90%"',
+            '"95%"',
+            '"98%"',
+            '"99%"',
+            '"100%"',
         ])
     ]
 
     for s in chain(sort_stats(runners.locust_runner.request_stats), [runners.locust_runner.stats.aggregated_stats("Total", full_request_history=True)]):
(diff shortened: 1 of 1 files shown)


mbeacom:
@fordhurley Do you still want to pursue this PR and discuss having it merged?

fordhurley:
I still think it's a good idea, but it definitely requires some thinking, testing, and documenting. At the moment, it's more of a suggestion with a little code for how it could be done.

I just rebased, but haven't tested the updated code at all. I don't have the time at the moment to get it running.

mbeacom:
@fordhurley Thank you!
@aldenpeterson-wf Do you happen to have any input on this?

aldenpeterson-wf:
I think this is probably a good idea longer term to simplify things.

My main concern would be making those files too large but that's.... going to be requiring an insane number of unique tasks being reported to really matter at all.

cyberw:
Hi! This PR has been untouched for a very long time and has conflicts so I am closing it. Let me know if this was a mistake and feel free to open it again.

PR #186, opened 2014-08-06
Show test hostname

hartfordfive:
Applied small modification so that the hostname being tested can be visible on the index page


thaffenden:
This was introduced in #447 so this can probably be closed.

PR #200, opened 2014-09-19
Add test duration no-web option

cava23:
This works, but I'm killing the thread incorrectly and it returns an error: "LoopExit: This operation would block forever". Perhaps someone can help me understand the correct way to shutdown when the timer expires?


mbeacom:
@cava23 Are you still interested in pursuing this PR?

cava23:
I am not.

Issue #135, opened 2014-03-20
Configure target host from web interface


dair-targ:
It will be nice to be able to change target host from web interface. My suggestion:
- Command-line `--host` argument value is used as default (initial) `host` value and become optional.
- Add a new field _Target host_ to _Start new Locust swarm_ dialog. This field should be used to change `host` value from web interface without need to restart locus server.
- `host` may not be changing while locust swarm is running.

justiniso:
See https://github.com/locustio/locust/pull/523

PR #250, opened 2015-03-04
add host cli arg to quickstart

groovecoder:
Document the need for `--host` argument in the quickstart.

The quickstart wasn't as quick as it could be for me, because I got errors about the missing host, and then an error about the host value missing a scheme. https://github.com/locustio/locust/issues/135 seems like the real fix, but this should at least help new locust users to get started even more quickly.

Note: do slave processes need a `--host` argument, or do they get it from the master?


heyman:
Thanks!

PR #258, opened 2015-03-17
Configurable precision

cpennington:
This allows the user to set `--precision X` on the commandline to store up to that many digits of precision in response times.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -159,21 +159,24 @@ class StatsEntry(object):
         self.min_response_time = min(self.min_response_time, response_time)
         self.max_response_time = max(self.max_response_time, response_time)
 
-        # to avoid to much data that has to be transfered to the master node when
-        # running in distributed mode, we save the response time rounded in a dict
-        # so that 147 becomes 150, 3432 becomes 3400 and 58760 becomes 59000
-        if response_time < 100:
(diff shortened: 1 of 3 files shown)


locust/stats.py
```diff
@@ -157,22 +169,25 @@ def _log_response_time(self, response_time):
...
-        # so that 147 becomes 150, 3432 becomes 3400 and 58760 becomes 59000
-        if response_time < 100:
-            rounded_response_time = response_time
-        elif response_time < 1000:
-            rounded_response_time = int(round(response_time, -1))
-        elif response_time < 10000:
-            rounded_response_time = int(round(response_time, -2))
-        else:
-            rounded_response_time = int(round(response_time, -3))
+        rounded_response_time = self._round_response_time(response_time)
```
BenjiLee: Our of curiosity, what was the reasoning of this doc? Why not just give the more precise number in the first place?

BenjiLee:
:+1:

cpennington:
@cgoldberg: I just rebased this, so it should be merge-able now.

cpennington:
Is this something that you just don't want? Or were you assuming since it was old it was abandoned? I'd be happy to rebase it to get it mergeable.

Issue #260, opened 2015-03-18
list supported Python versions in README and setup.py.


cgoldberg:
It would be helpful if the README displayed supported versions of Python.  The only place I could find the info was in .travis.yml.

Also, the following lines should be added to setup.py (classifiers) to specify versions supported:

Programming Language :: Python :: 2
Programming Language :: Python :: 2.6
Programming Language :: Python :: 2.7

frica:
I agree with @cgoldberg. Also it would be nice to have it mentioned in http://docs.locust.io/en/latest/installation.html.

cgoldberg:
I added this info in Pull Request #261

PR #261, opened 2015-03-19
list supported Python versions

cgoldberg:
fix for #260 

This branch documents versions of Python that locust requires.  Added info to:

* setup.py
* README.md
* installtion.rst


heyman:
Thanks!

Issue #236, opened 2015-01-27
confused by multiple requests in on_start


mblakele:
This is with 0.7.2 from pip, and also using the latest from github. It seems that `on_start` isn't expecting more than one request, and does strange things when this happens.

Use case: django apps often use a guard token that is the same in headers and form, and posting the form has to replay it correctly: `csrfmiddlewaretoken`. If a login form uses this token, then the setup phase must GET the form, parse out the token, and then POST the login request with the right token.

```
class UserBehavior(TaskSet):
    def on_start(self):
        """
        on_start is called when a Locust start before,
        any task is scheduled
        """
        self.login()

    def login(self):
        # Response from GET will include csrfmiddlewaretoken
        resp = self.client.get("/login")
        token = resp.cookies.get('csrftoken')
        self.client.post("/login",
                       {"csrfmiddlewaretoken":token,
                        "username":"joe.test",
                        "password":"test"})

    @task(1)
    def index(self):
        self.client.get("/")

class WebsiteUser(HttpLocust):
    task_set = UserBehavior
    min_wait =  500
    max_wait = 1500

```
If this locustfile is run with `n=1`, both login requests are visible in the server log. But locust only reports the GET, not the POST. I think this happens because it is counting requests, and expects the `on_start` to only contain one request.

If run with `n=2`, the final summary shows both `on_start` requests but the server log also shows `GET /`. With sufficiently large `n` I suppose this doesn't matter, but it would be more accurate to report on all the requests that were made, whatever value `n` has.

If the `GET /` task is removed and `n=2`, locust goes into a loop of `IndexError` exceptions.

```
[2015-01-26 16:34:13,705] localhost/ERROR/stderr: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/locust/core.py", line 264, in run
    self.schedule_task(self.get_next_task())
  File "/usr/local/lib/python2.7/site-packages/locust/core.py", line 325, in get_next_task
    return random.choice(self.tasks)
  File "/usr/local/Cellar/python/2.7.8_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/random.py", line 275, in choice
    return seq[int(self.random() * len(seq))]  # raises IndexError if seq is empty
IndexError: list index out of range
[2015-01-26 16:34:15,032] localhost/ERROR/stderr: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/locust/core.py", line 264, in run
[...]

PR #267, opened 2015-03-25
changed how request_meta["method"] is set

dantagg:
method was set against the request response method. In the case of a POST request being redirected to GET, common on success of POSTs for login and registration. The request is incorrectly logged against the GET method for the original URL.

Closes locustio/locust#236


heyman:
Thanks!

PR #284, opened 2015-05-29
Web auth

SandyChapman:


justiniso:
Hmm, I can see the use case for this, but I'm concerned it's adding feature bloat to address a problem that is better handled through other means. e.g. you can slap nginx with basic auth in front of it. There are a million other solutions to apply authentication and I feel like it's best left to users to address their own unique cases.

SandyChapman:
The reason we added this was to essentially support a shared secret. We had locust running in the cloud on AWS and wanted to ensure there was little ability to access to page without the credentials. It's fine if you do not want the change.

PR #285, opened 2015-05-29
Fixed a bug causing blank lines to be printed in stdout.

SandyChapman:

### DIFF
--- a/locust/log.py
+++ b/locust/log.py
@@ -23,6 +23,8 @@ class StdOutWrapper(object):
     Wrapper for stdout
     """
     def write(self, s):
+        if s == None or s.strip() == '':
+            return
         stdout_logger.info(s.strip())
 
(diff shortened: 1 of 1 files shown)


locust/log.py
```diff
@@ -23,6 +23,8 @@ class StdOutWrapper(object):
     Wrapper for stdout
     """
     def write(self, s):
+        if s == None or s.strip() == '':
```
justiniso: Please use `s is None`https://www.python.org/dev/peps/pep-0008/#programming-recommendations

locust/log.py
```diff
@@ -30,6 +32,8 @@ class StdErrWrapper(object):
     Wrapper for stderr
     """
     def write(self, s):
+        if s == None or s.strip() == '':
```
justiniso: Same comment, see above

PR #286, opened 2015-05-29
Added a handler that will allow locusts and task sets to be notified …

SandyChapman:
…when the locust is being killed. This can be useful for resource cleanup (i.e. such as killing threads).

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -112,6 +112,7 @@ class LocustRunner(object):
                     try:
                         locust().run()
                     except GreenletExit:
+                        instance.die()
                         pass
                 new_locust = self.locusts.spawn(start_locust, locust)
                 if len(self.locusts) % 10 == 0:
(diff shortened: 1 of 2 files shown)


locust/runners.py
```diff
@@ -112,6 +112,7 @@ def start_locust(_):
                     try:
                         locust().run()
                     except GreenletExit:
+                        instance.die()
```
justiniso: Wouldn't we want this to be in a `finally` block?

SandyChapman: I suppose if we wanted the clean up code to handle regardless we could put it there. I was thinking of it more as an exception handler. I can move if you'd like.

mbeacom:
@SandyChapman Are you still interested in making the requested changes to move forward on this?

cyberw:
Hi @SandyChapman  ! This has been idle for a long time and now has conflicts, so I am closing it. Feel free to reopen if this is still important to you!

PR #289, opened 2015-05-30
Make resetting stats after hatching configurable

### LINKED ISSUE #205: Option to prevent stats from being reset when all locusts are hatched

pglass:
This adds a global variable `locust.runners.RESET_STATS_AFTER_HATCHING` that you can use from a locust file, as well as a command line flag `--no-reset-stats`. Locust should still default to resetting the statistics and the "Reset Stats" link in the web interface should work regardless of this flag.

Resolves #205

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -191,6 +191,13 @@ def parse_options():
        default=False,
        help='Only print the summary stats'
     )
+
+    parser.add_option(
+        '--no-reset-stats',
+        action='store_true',
+        dest='no_reset_stats',
+        default=False,
+        help="Never reset statistics after hatching")
     
     # List locust commands found in loaded locust files/source files
     parser.add_option(
@@ -391,6 +398,11 @@ def main():
     if options.master and options.no_web:
         logger.error("Locust can not run distributed with the web interface disabled (do not use --no-web and --master together)")
         sys.exit(0)
+
+    if options.no_reset_stats:
+        runners.RESET_STATS_AFTER_HATCHING = False
+    if not runners.RESET_STATS_AFTER_HATCHING:
+        logger.info("NOT resetting stats after hatching")
 
     if not options.no_web and not options.slave:
         # spawn web greenlet
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -23,6 +23,7 @@ locust_runner = None
 
 STATE_INIT, STATE_HATCHING, STATE_RUNNING, STATE_STOPPED = ["ready", "hatching", "running", "stopped"]
 SLAVE_REPORT_INTERVAL = 3.0
+RESET_STATS_AFTER_HATCHING = True
 
 
 class LocustRunner(object):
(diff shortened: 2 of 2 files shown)


billibala:
thanks for this patch! saves me hours of debug time.

SecretObsession:
I just wanted to comment to support this PR. 

We have been using locust for load testing, but we have started to venture into determining if all insert records stemmed from the API are making it into our database. Unfortunately with the reset happening after everything is hatched, makes this task difficult, when it shouldn't be.

As an example, if I hatch 10000 users, there will be 8000-11000 requests prior to the stats being reset. Which means the counts themselves will be off. If we are going to compare the number of requests made by locust, to the number of records in the database, it will appear that the database has more records than it should.

marklynch:
Adding a vote to this - about to do some similar testing that @SecretObsession described so having this merged in would be great.

cgoldberg:
+1

nhoening:
+1

adionditsak:
+1

alexandrul:
+1

Slach:
+1

jjstellr:
+1

JordyvanDortmont:
+1

di:
+1

TimGerlach:
+1

swordmaster2k:
+1

shun-liang:
+1

partojkander:
+1

dimakr:
+1

titleistfour:
+1

Just starting using locust and ran into the same issue were requests did not match the database records.  Would be really great to get this finally merged in.

locust/main.py
```diff
@@ -392,6 +399,11 @@ def main():
         logger.error("Locust can not run distributed with the web interface disabled (do not use --no-web and --master together)")
         sys.exit(0)
 
+    if options.no_reset_stats:
+        runners.RESET_STATS_AFTER_HATCHING = False
+    if not runners.RESET_STATS_AFTER_HATCHING:
```
justiniso: This `if` condition is unnecessary in my opinion. Why not just log when setting the flag to False?

pglass: If I remember right (after 2 years), it was to handle the case where the module flag was set to false by anything other than the `--no-reset-stats` flag. For example, if the locust file imports the module and sets that global flag to false.

heyman:
Ok, this has now (finally) been merged. Thank you!

I cherry picked the commit, but changed the implementation slightly (removed the module level variable for setting the reset behaviour, and removed the log about the setting).

locust/main.py
heyman: I removed the log message all together, and changed so that we don't use the module level variable for controlling the reset behaviour. Let me know if you think that's a bad idea.

alvissd:
Hi, is this implemented in version 0.8a2? I have read the change log and download but it seems like nothing changed from 0.7.5

aldenpeterson-wf:
@alvissd it shows up in the v0.8a1 release commits:

https://github.com/locustio/locust/compare/v0.8a1...master

From feb 17th that commit is in there.

alvissd:
@aldenpeterson-wf thanks, I just found it in the source code. Somehow pip install locustio=0.8a2 doesn't install the newest.

wormhoudt:
I can confirm that this is not in 0.8a1 or 0.8a2 on PyPi. I can see that it is in the source code, it just isnt available when installing through pip.

heyman:
The first version it's in is 0.8a3 I think. v0.8a1...master will show commits that are *not* in v0.8a1 :).

0.8 is released on PyPI now.

Issue #270, opened 2015-04-07
Web UI should show what host is being used


thomasbiddle:
Whether I get to this or someone else:

The web UI needs to show what host is being targeted. I just spent a bunch of time debugging this only to find out supervisor wasn't reading my updated config and was still hitting a bad host - Had to hack around to get the logs to spew out what host it was looking at.

DataGreed:
+1 to that.

RollForReflex:
I have this implemented in my local fork. I'll spin off a PR if anyone else is still interested.

thomasbiddle:
@RollForReflex I see no reason not to :+1:

justiniso:
Closing as this has multiple open PR's already.

PR #295, opened 2015-06-19
adding visual representation of host in Locust UI

RollForReflex:
When running a locustfile via the web UI, it would be helpful to be able to see what host you're hitting. This thought is echoed in issue 270 (https://github.com/locustio/locust/issues/270).

I've added a little bit of HTML and CSS code to show the host name in the top left hand corner
host

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -40,7 +40,8 @@ def index():
         is_distributed=is_distributed,
         slave_count=slave_count,
         user_count=runners.locust_runner.user_count,
-        version=version
+        version=version,
+        host=runners.locust_runner.host or runners.locust_runner.locust_classes[0].host
     )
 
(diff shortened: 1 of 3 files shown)


locust/web.py
```diff
@@ -40,7 +40,8 @@ def index():
         is_distributed=is_distributed,
         slave_count=slave_count,
         user_count=runners.locust_runner.user_count,
-        version=version
+        version=version,
+        host=runners.locust_runner.host or runners.locust_runner.locust_classes[0].host
```
cgoldberg: I like this PR, but this line needs to be fixed.  It currently causes a test failure.  See travis-ci logs for details

thaffenden:
This change was introduced by #447, so this can probably be closed.

PR #298, opened 2015-06-22
Introduce docs for increasing the max number of open files limit

ericandrewlewis:


cgoldberg:
:+1:

heyman:
Great! Thanks!

aidylewis:
Should we also describe how the open file limit can be increased in Linux through /etc/sysctl.conf (system wide limit)? And should we also mention about increasing ephemeral ip ports?

aidylewis:
I've got this for Fedora based Linux

```sh
# increase system file descriptor limit
echo fs.file-max = 100000 | sudo tee /etc/sysctl.conf
# increase ulimit ( < system file descriptor )
printf '*       soft    nofile  99999 \n*       hard    nofile  99999' | sudo tee /etc/security/limits.conf
```
Also, need to reboot.

For Debian based, I think you also need to enable PAM user limits.

PR #329, opened 2015-08-26
Added websocket support and an example

normanb:
The websocket example requires Flask-Sockets==0.1. I have enabled the websockethandler in web.py

To run the example run ws.py as a locust file and test, then run ws.py from as separate terminal and it will connect to the websocket.

### DIFF
--- a/examples/ws.py
+++ b/examples/ws.py
@@ -1,0 +1,66 @@
+# encoding: utf-8
+
+"""
+This is an example of a locustfile that uses Locust's built in event hooks to
+push events on a websocket
+"""
+
+### WebSocket Client
+if __name__ == "__main__":
+    from ws4py.client.threadedclient import WebSocketClient
+
+    class DummyClient(WebSocketClient):
+        def received_message(self, m):
+            print m
+
+    try:
+        ws = DummyClient('ws://127.0.0.1:8089/ws')
+        ws.connect()
+        ws.run_forever()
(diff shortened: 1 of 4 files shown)


omazurashu:
Hi.
it 2017 now :) Any plans to merge this to master?

cgoldberg:
it needs to be rebased and have merge conflicts resolved

jlarocque88:
This would be very useful for me as well. Can this get taken care of?

heyman:
What is the use-case for this? From what I can make out of the code it adds the ability for the Locust web UI Flask server to act as a websocket server. Is that correct?

mbeacom:
What's everyone's thoughts on this? Do we want locust to be able to serve WS? If so, could we make this dependent on whether or not the user wants to run with WS support, possibly an `extras_require` entry with the additional dependencies, and additional documentation on installing and running with WS support? I assume we wouldn't want to run the websocket handler in all cases. 

Is this something you're still interested in having merged @normanb?

Thanks!

normanb:
My original need for this has gone, however my original intent was to use web sockets to do real time charting.

heyman:
I think that unless we have a clear need/gain by adding web socket support to the web UI, we shouldn't add it. It could make the the UI "more real-time", but then again, when running Locust distributed it won't be real-time anyway since the stats are reported to the master node every 2 seconds.

I'd say this is a close for now.

PR #330, opened 2015-08-26
include the full request history in the web UI stats, hence /stats/re…

marrotte:
…quests route


marrotte:
forget about this...

Might need something like  "total_rps": s.total_rps, added to the report and memoize removed.  looking into it.

Issue #196, opened 2014-09-14
Run the tests for the specified time


diegows:
I think it would be great to have an option to define how many time we want to run the tests from the web interface.

heyman:
Fixed by #656!

PR #333, opened 2015-09-03
Allow --no-web together with --master for automation

undera:
Summary:
- removed limitation to not have --no-web together with --master
- added --expect-slaves option to tell how many slaves expected to connect
- on all expected slaves are connected, master initiates hatching

Taurus mailing list discussion for reference: https://groups.google.com/forum/#!topic/codename-taurus/Gs3VdphRSjo

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -8,6 +8,7 @@ import signal
 import inspect
 import logging
 import socket
+import time
 from optparse import OptionParser
 
 import web
@@ -113,6 +114,15 @@ def parse_options():
         dest='master_bind_port',
         default=5557,
         help="Port that locust master should bind to. Only used when running with --master. Defaults to 5557. Note that Locust will also use this port + 1, so by default the master node will bind to 5557 and 5558."
+    )
+
+    parser.add_option(
+        '--expect-slaves',
+        action='store',
+        type='int',
+        dest='expect_slaves',
+        default=1,
+        help="How many slaves master should expect to connect before starting the test (only when --no-web used)."
     )
 
     # if we should print stats in the console
@@ -386,11 +396,6 @@ def main():
         }
         console_logger.info(dumps(task_data))
         sys.exit(0)
-    
-    # if --master is set, make sure --no-web isn't set
-    if options.master and options.no_web:
-        logger.error("Locust can not run distributed with the web interface disabled (do not use --no-web and --master together)")
-        sys.exit(0)
 
     if not options.no_web and not options.slave:
         # spawn web greenlet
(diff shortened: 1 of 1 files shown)


aidylewis:
Hi @undera Will this functionality allow us to run a master and multiple slaves using the Taurus config?

undera:
Yep. It's not 100% ready, but enough to get the idea.

heyman:
I like this functionality, though maybe I'd prefer the flag to be called `--expected-slave-count`. This, together with an option to specify a time limit on the test (#196) should be enough for letting people run Locust tests in their CI cycle without any hacks. Also, I'd be ready to drop the `-n/--num-request` flag, since I think it's a little weird stop condition, and setting a time limit should work just as well in almost every case.

Any thoughts on this?

undera:
From my experience, some people are willing to have exact control on number if requests made. For example, they doing benchmark of 100 000 requests and compare them to each other. Usually I'm implementing in my tools the logic "stop whatever end first, requests limit or time limit".

undera:
I'm fine with the name `--expected-slave-count`

locust/main.py
```diff
@@ -115,6 +116,15 @@ def parse_options():
         help="Port that locust master should bind to. Only used when running with --master. Defaults to 5557. Note that Locust will also use this port + 1, so by default the master node will bind to 5557 and 5558."
     )
 
+    parser.add_option(
+        '--expect-slaves',
```
justiniso: Personally I prefer `--min-slaves` as it better communicates what the behavior is. Since you can have greater than the number of expected slaves with this implementation. See https://github.com/locustio/locust/pull/372/files#diff-6f782ed63a6a642694db58c0d5cdd932R125

heyman: Wouldn't the current implementation start close to immediately once the min slave count has been reached? If so, how would I go about to start a test with more slaves than min_slaves (other than just spawning them all at the same time and hoping that they would connect before the the poll check)?

justiniso: Yes, it would but I think that's the behavior I would expect. I think most people using this feature would bring up the master process specifying `--min-slaves 10`, then bring up 10 slave processes to connect to the master. If I wanted 11 slaves, I would simply indicate `--min-slaves 11`. However if I _thought_ I wanted 10 slaves, but later realized I need 11 slaves to reach my desired throughput, then I could add another 1 slave after the original 10 slaves have already started swarming.

heyman: In that case I think i'd still prefer `--num-slaves` or `--expected-slave-count`. Though it might be bikeshedding, and I'm fine with `--min-slaves` as well :).

> but later realized I need 11 slaves to reach my desired throughput, then I could add another 1 slave after the original 10 slaves have already started swarming

That would require us to have some kind of automatic re-balancing of simulated users when new slaves connect.

danielward:
@heyman @justiniso Are there any updates on merging in this feature?

gruzewski:
👍

PR #341, opened 2015-10-02
updated gevent version to support El Capitan release for MacOS

SecretObsession:
I was running into issues installing gevent on Mac "El Capitan" release. This is the fix for the installation issue.

I bumped the gevent version to 1.1.b1 from 1.0.1


justiniso:
Thanks, the gevent version will be bumped to 1.1.0 for the Python 3 release.

PR #343, opened 2015-10-05
Enhancement: make the total row more easily visible

BenjamenMeyer:
Tests that have a lot of responses can make the browser get bogged down and make it harder to see the total column. By simply making the total column the first row it solves the issue.

FYI - I ran into this under Firefox on Ubuntu 15.04 where I had 150+ requests being displayed where we were using it to do performance testing of a service API. Keeping the total at the top makes it a lot easier to see if one needs to look at the failures or exceptions; it was otherwise getting really hard to see the total as I always had to keep it scrolled to the bottom and at times that was difficult to do where Firefox just didn't want to cooperate. Pretty sure other browsers could have the same issue.

### DIFF
--- a/locust/static/locust.js
+++ b/locust/static/locust.js
@@ -102,7 +102,7 @@ $(".stats_label").click(function(event) {
     alternate = false;
     totalRow = report.stats.pop()
     sortedStats = (report.stats).sort(sortBy(sortAttribute, desc))
-    sortedStats.push(totalRow)
+    sortedStats.unshift(totalRow)
     $('#stats tbody').jqoteapp(stats_tpl, sortedStats);
     alternate = false;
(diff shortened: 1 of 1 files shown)


locust/static/locust.js
```diff
@@ -102,7 +102,7 @@ $(".stats_label").click(function(event) {
     alternate = false;
     totalRow = report.stats.pop()
```
BenjamenMeyer: Need to fix this since totalRow is no longer at the end. I have it locally; waiting to test and then will update the PR.

mbeacom:
Thank you for your contribution.  It appears this issue has been previously resolved.  Is this something you'd still like to have merged? If so, can you verify that this issue still exists, as I can't seem to replicate it.

BenjamenMeyer:
@mbeacom it's been a while since I've used Locust and I don't have a project setup I can test it with; if you've already solved the problem another way than that is fine with me; feel free to close.

PR #352, opened 2015-10-26
Added full request history (available via report download)

SecretObsession:
Locust will now save the request statistics for each request into a "history". This history can be downloaded in CSV form. 

(This can help identify issues which happen intermittently in the run, such as the web server rubber banding on requests.)

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -33,6 +33,7 @@ class LocustRunner(object):
         self.num_requests = options.num_requests
         self.host = options.host
         self.locusts = Group()
+        self.locust_count = 0
         self.state = STATE_INIT
         self.hatching_greenlet = None
         self.exceptions = {}
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -33,7 +33,6 @@ class LocustRunner(object):
         self.num_requests = options.num_requests
         self.host = options.host
         self.locusts = Group()
-        self.locust_count = 0
         self.state = STATE_INIT
         self.hatching_greenlet = None
         self.exceptions = {}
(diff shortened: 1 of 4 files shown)


cgoldberg:
I'd be hesitant to merge this PR.  It appends a history line to a global list for *every* successful request.  In longer tests, would this exhaust memory?  I'd want to see memory usage over a long duration test at a high rps, to see how much memory it uses.

cgoldberg:
calling this feature "full request history" is misleading, as it only includes successful requests, not requests that failed.

locust/runners.py
```diff
@@ -146,6 +150,7 @@ def start_hatching(self, locust_count=None, hatch_rate=None, wait=False):
         if self.state != STATE_RUNNING and self.state != STATE_HATCHING:
             self.stats.clear_all()
             self.stats.start_time = time()
+            self.stats.locust_count = locust_count
```
cgoldberg: where does locust_count ever get incremented?  (sorry if I am missing this)

SecretObsession:
Thanks for the feedback @cgoldberg. I have some refactoring ahead of me.

I wasn't seeing any real memory issues at 1 million requests, but it's probably still a bad idea to make it default, so I'm going to add in a CLI argument to activate if needed. I'll run some additional testing to see what memory looks like with a few million requests. If memory starts to become an issue, I can just have it write out the data to a file directly, instead of holding it in memory. (File specified by command line, or defaults to a certain file naming convention.)

Originally I had written it to keep track of the requests during ramping, but I'm not sure how useful it is with the time between ramp being hardcoded. (The stats are also reset after hatching, which kinda defeats the purpose. Paired with the -no-reset-stats PR that is on here, it would make it pretty useful.) (I'll change this to allow changing the time between ramp values, but I'm probably going to submit it as a separate PR.)

I'm also going to add the returned http status codes. (For success and failure.)

Again, thanks for the feedback.

PR #358, opened 2015-10-29
Fixed failures percentage calculation in Total

mlo0352:
The percentage calculation was essentially comparing failures to the number of successes (confusingly referred to as reqs [requests]) - fixed the math to have it calculate percent failure as a percentage of all requests.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -21,7 +21,7 @@ class RequestStats(object):
         self.max_requests = None
         self.last_request_timestamp = None
         self.start_time = None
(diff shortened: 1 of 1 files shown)


mbeacom:
@justiniso Is this a verified issue?

raiyankamal:
I believe this PR is superseeded by https://github.com/locustio/locust/pull/939 where the logic for counting number of requests were corrected to include both succeeded and failed requests. @cgoldberg please check.

PR #361, opened 2015-11-19
doc fixes

gward:


docs/installation.rst
```diff
@@ -12,7 +12,7 @@ or::
     easy_install locustio
 
 When Locust is installed, a **locust** command should be available in your shell (if you're not using 
-virtualenv - which you should - make sure your python script directory is on your path).
+virtualenv—which you should—make sure your python script directory is on your path).
```
cgoldberg: em-dashes are nice.

I don't understand the need to mention virtualenv at all.  The way it is written implies that if you *are* using a virtualenv, you can run scripts from a directory that's not on your path.  I don't think that is correct.

cgoldberg:
nice :+1:

PR #391, opened 2016-03-13
zero out negative response time

marrotte:


marrotte:
It’s not a complete locust problem.  But, locust can defend against non-monotonic client timers -- that is clients that measure response time using time.time versus os.times.

sj2208:
With this change the min response time will appear as 0

marrotte:
@sj2208 how so?  The code sets the min response time to the smallest response time >=0.  Min will only be set to zero if response time is ever zero or negative.

mbeacom:
@marrotte Thanks for the contribution! Is this still an outstanding issue?  Can you elaborate on the details of this PR?

vivekaditya:
@mbeacom  
Is this still an outstanding issue ?  **Yes** 
Can you elaborate on the details of this PR ? 
The PR is pretty simple. Clients using non-monotonic timers can lead to negative response times, e.g. NTP corrections. So we do a max of min response time and 0.

cyberw:
Hi @marrotte ! This is a really old PR and now has conflicts, so I'm going to close it. Feel free to reopen if this is still an issue for you.

PR #394, opened 2016-03-17
main function accepts direct call.

FGtatsuro:
This PR allows main function to be called directly with command line arguments. Could you check and merge it if you think it's good feature?
## Motivation

I want to integrate locust testing to setuptools like [tox](http://tox.readthedocs.org/en/latest/example/basic.html#integration-with-setuptools-distribute-test-commands).

Current implementation uses optparse.OptionParser#parse_args(). This regards sys.argv[1:] as command line arguments automatically. Thus, this also gets arguments(like custom args of setuptools) not related to locust unexpectedly. 

```
# -a is custom arguments of setuptools for tox.
$ python setup.py test -a "-epy27"
```

For this PR, we can run locust tests with arbitrary arguments like this.

```
class Locust(TestCommand):

    user_options = [
        ('locust-args=', 'a', 'Arguments for Locust')
    ]

    def initialize_options(self):
        TestCommand.initialize_options(self)
        self.locust_args = []

    def finalize_options(self):
        TestCommand.finalize_options(self)
        self.test_args = []
        self.test_suite = True

    def run_tests(self):
        import shlex
        from locust.main import main

        args = None
        if self.locust_args:
            args = shlex.split(self.locust_args)
        main(args=args)
```

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -22,9 +22,11 @@ from .stats import (print_error_report, print_percentile_stats, print_stats,
 _internals = [Locust, HttpLocust]
 version = locust.__version__
 
-def parse_options():
+def parse_options(args=None):
     """
     Handle command-line options with optparse.OptionParser.
+    If args is given, parser regards them as command line arguments like ['--opt1=value1', '-o', value2'].
+    If not, sys.argv[1:] will be used.(default action of OptionParser)
 
     Return list of arguments, largely for use in `parse_arguments`.
     """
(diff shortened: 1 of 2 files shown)


mbeacom:
@FGtatsuro Thank you for the contribution! Would you still like to see this merged? Can you elaborate on the utility of this feature?

FGtatsuro:
@mbeacom 
Thank you for your comment. My motivation is still same of https://github.com/locustio/locust/pull/394#issue-141678578 (Motivation section).

I often use setup tools as common interface for several commands including tests. For this purpose, It's useful for custom command to get arbitrary arguments. (Locust command example in above issue comment).
If you think my usage as useful, could you consider this PR?

cyberw:
Hi @FGtatsuro ! I think your PR does make some sense, but it has been idle for very long and now has conflicts. Please resolve them if you want us to take another look! Closing for now.

PR #400, opened 2016-04-01
Master node to configure slaves before hatching.

alexeypavlenko:
Hello colleagues,

I was looking for a way to dynamically configure slave nodes before hatching new users, but failed to utilize any existing locust mechanism. So I came with two new events to send configuration data from master node to all attached slaves.

Changes:
- _configuration_to_slaves_ event configuration data from master node before hatching new users.
- _master_configuration_ event provides configuration data received from master node.
- slightly modified `examples/events.py` sample to illustrate the usage of new events
- update `api.rst` to include new properties

Please let me know if my approach is suitable, or if there is some better/more traditional solution.
Thanks in advance!

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -74,5 +74,5 @@ Available hooks
 The following event hooks are available under the **locust.events** module:
 
 .. automodule:: locust.events
-	:members: request_success, request_failure, locust_error, report_to_master, slave_report, hatch_complete, quitting
+	:members: request_success, request_failure, locust_error, report_to_master, slave_report, configuration_to_slaves, master_configuration, hatch_complete, quitting
 
--- a/examples/events.py
+++ b/examples/events.py
@@ -27,9 +27,10 @@ class WebsiteUser(HttpLocust):
 We need somewhere to store the stats.
 
 On the master node stats will contain the aggregated sum of all content-lengths,
-while on the slave nodes this will be the sum of the content-lengths since the 
+while on the slave nodes this will be the sum of the content-lengths since the
 last stats report was sent to the master
 """
+collect_content_length = False
 stats = {"content-length":0}
 
 def on_request_success(request_type, name, response_time, response_length):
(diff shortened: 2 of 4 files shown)


arjunrn:
Hi Alexey, we have been working on something like this but by wrapping locust. To make it more clear, could you add some code samples on how this could be used?

alexeypavlenko:
Hi Arjun,

In this PR I've slightly extended existing sample `example/events.py` to allow collecting `content-length` field using newly added events. By default collection is disabled, so it is master host now who overrides the setting and configures slaves to report `content-length` field.

Primary scenario addressed by the new functionality (PR) is as follows:
  - perform configuration of the test environment (e.g. create a limited set of users in the shared database)
  - distribute configuration data to slave nodes using capabilities of the locust framework (e.g. send each slave host a random subset of user ids)

Before implementing this PR I was trying to solve the task above for my current project. Obviously, with current facilities of the locust framework I would need to introduce some external tool (like, I guess, you do) to communicate with all slave hosts and perform necessary configuration. IMO, the latter approach would heavily duplicate locust functionality, therefore I've made this configuration a part of locust itself.

Speaking of the completeness of the PR, I think I've done minimal implementation. However, it can be further extended as follows: provide the master host information regarding both the slave host being configured and the variety of slave hosts (e.g. `host 1 of 5 is being configured`). But before writing more code I would like to get an idea first, if this PR is accepted by locust community.

Thanks.

pglass:
I'm +1 on locust moving toward parameterized tests, and replication of your locust configs out to slaves is a step in that direction.

I'm imagining a world where,

- I can add arbitrary config values in the web UI (or the http api) before starting the test
- I can specify reasonable defaults for unspecified values in my code

Right now, locust lets you pick the values of the hatch rate and number of users. But this would let you set the values of any config you make up. It would put locust in a place where we could write very parameterized load tests, which would make locust much more flexible. Other configs like the host and min/max wait times could be dynamically specified as well.

I have a use for this in my day-to-day work, where I need to run various load tests against several deployments of the same api. This requires a bunch of configs that I manage in files that get loaded when locust starts up. I actually have to redeploy the locust cluster with a different config file when I need to run against another environment (which is automated via other tools but is a pain to maintain).

cyberw:
Hi! This PR has been untouched for a very long time and has conflicts so I will close it. Let me know if this was a mistake.

If you use https://github.com/SvenskaSpel/locust-swarm to launch your test/slaves, any env vars named LOCUST_* will be forwarded to slaves.

PR #363, opened 2015-12-02
Python 3 support

pmdarrow:
Added support for Python 3 and running tests against all supported versions of Python with tox. 

Borrowed heavily from https://github.com/locustio/locust/pull/328 but cleaned things up & simplified some changes in the hopes that it will be more likely to be merged.

PR #408, opened 2016-04-28
Python3

will-mooney-3:


Slach:
Will this poll request merged ?? 
Where is merge policy for locust ?? Where i can read about it?
i really need Python3.5 support in official pypi package

justiniso:
@Slach this depends on gevent and we are testing a new version. The Python 3 discussion is happening over on https://github.com/locustio/locust/pull/363.

PR #410, opened 2016-05-02
Add starting events

xrmx:
I need a way to perform at task before the workers are started. I cannot use the current starting_hatching events because i need the host.

Starting event is fired before the runners are setup, options
are passed as parameter.

I'd like to get some feedback if the feature is useful before spending time adding tests.

### DIFF
--- a/locust/events.py
+++ b/locust/events.py
@@ -97,6 +97,14 @@ Event is fire with the following arguments:
 * *user_count*: Number of users that was hatched
 """
 
+starting = EventHook()
+"""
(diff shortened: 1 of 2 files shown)


cyberw:
Hi! This PR has been untouched for a very long time so I am closing it. Let me know if this was a mistake, and feel free to open it again.

xrmx:
@cyberw I don't know if it is a mistake closing this but I don't have a use for it anymore, so fine for me  :)

PR #416, opened 2016-05-13
Add host setting in HttpLocust initialization

sr-murthy:
This is a small change to allow `HttpLocust` to be initialised with a `host` attribute.

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -123,9 +123,10 @@ class HttpLocust(Locust):
     The client support cookies, and therefore keeps the session between HTTP requests.
     """
     
-    def __init__(self):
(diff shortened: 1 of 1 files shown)


heyman:
Hi! What's the reason you want to change this? Currently, the locust classes are instantiated by internal code in Locust, and one shouldn't have to instantiate them manually in load testing code.

justiniso:
I don't think this is a good idea at the moment. There are already multiple ways of specifying the host and active discussions around setting the host through the UI as well. Further, this is not a supported API for end users. Since OP hasn't responded to @heyman's question, I assume interest is gone.

sr-murthy:
That's fine.

Issue #407, opened 2016-04-28
v0.7.4


justiniso:
Since it's been quite a while since the last release, let's push a new patch version with current master. The primary fix would be for: https://github.com/locustio/locust/issues/273, but there are a bunch of copy edits to documentation. 

@cgbystrom @cgoldberg @pmdarrow @mthurlin I believe @heyman owns the pypi creds.

heyman:
Sure, I'd be happy to do that!

Someone should probably update `docs/changelog` as well.

justiniso:
Cool, I'll start working on a release branch.

PR #418, opened 2016-05-16
Release updates for v0.7.4

justiniso:
@heyman let me know if I'm missing anything. I tried to diff from the last version, but the tag was missing. 

re: https://github.com/locustio/locust/issues/407

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,14 @@
 Changelog
 ##########
 
+0.7.4
+=====
+
+* Use a newer version of requests, which fixed an issue for users with older versions of 
+  requests getting ConnectionErrors (https://github.com/locustio/locust/issues/273).
+* Various fixes to documentation.
+
+
 0.7.3
 =====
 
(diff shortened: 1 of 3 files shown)


heyman:
Great!

The version is duplicated in `locust/__init__.py` as well which I've updated in 6bbbeeeeac2732a786bd3c42ebf68d4de40b57b3.

0.7.4 has now been uploaded to PyPI!

justiniso:
Awesome thanks! Any reason the version is not imported?

heyman:
If I recall correctly it was because of an import error (I'm not sure exactly when it occurs). 

Looking at some other packages, it seems like one way to go about it is to manually parse the version from `__init__.py`. (https://github.com/kennethreitz/requests/blob/39d693548892057adad703fda630f925e61ee557/setup.py#L50-L55 and https://github.com/pallets/flask/blob/567fff9d0d20650743abd8fe17d3dc9d98fa19dc/setup.py#L50-L54)

PR #438, opened 2016-06-25
Fix an import error on Python 3

jleclanche:


justiniso:
LGTM 👍

swordmaster2k:
+1 for this to be merged

heyman:
Merged #456 which was exactly the same fix, but without whitespace changes. Thanks!

PR #445, opened 2016-07-11
Save stats to file with --statfile option

gunesmes:


justiniso:
There are a few ways of doing this being discussed. The question is whether we want to expose a file with summary stats at the end, all stats over time, or both.

locust/stats.py
```diff
@@ -512,8 +512,26 @@ def print_error_report():
...
 
+def store_stats(filename, stats):
+    with open(filename, "w") as f:
+        f.write("path,method,num_requests,num_failures,min_response_time,max_response_time,avg_response_time\n")
+        for k in stats:
+            r = stats[k]
+            # need to add try - except block if one of the value is None
+            try:
+                f.write("%s,%s,%d,%d,%d,%d,%d\n" % (r.name,r.method,r.num_requests,r.num_failures,r.min_response_time,r.max_response_time,r.avg_response_time));
+            except:
```
justiniso: This should not be a catch-all exception.

locust/main.py
```diff
@@ -416,6 +426,9 @@ def main():
     if not options.only_summary and (options.print_stats or (options.no_web and not options.slave)):
         # spawn stats printing greenlet
         gevent.spawn(stats_printer)
+
+    if options.statsfile:
+        gevent.spawn(stats_persist, options.statsfile)
```
jami3: Requires import stats_persist.

naveensrinivasan:
Anything I can do to move this along?

jvautour:
this would be swell.  looking forward to having it merged

AXington:
Would also really like to see this get merged.

aldenpeterson-wf:
It'd be nice to see this in the 0.8 release, we would definitely use this functionality too.

prince-mishra:
When running with stop_timeout, the statsfile is empty after the run. Essentially, towards the end of the run (when all locusts die due to the timeout), there are no stats to report and this an empty file. Is that expected?

aldenpeterson-wf:
I rewrote this PR into this - https://github.com/locustio/locust/pull/612

I was looking at modifying this to fix the issues @prince-mishra pointed out but eventually just decided it would be easier to rewrite it.

mbeacom:
Closed in favor of #612 (merged) - Thank you!

PR #447, opened 2016-07-13
Adds host name to the header

thaffenden:
Adds the host url to the header:

screen shot 2016-07-13 at 17 11 22

PR #295 does the same thing slightly differently but failed tests and is over a year old so created my own (with less css changes as displaying it with the existing header items keeps the UI more visually consistent).
The max width and word wrap have been set to prevent long urls ruining the layout of whole header.

### DIFF
--- a/locust/static/style.css
+++ b/locust/static/style.css
@@ -77,6 +77,8 @@ a:hover {
     color: #000;
     font-size: 16px;
     font-weight: bold;
+    max-width: 165px;
+    word-wrap: break-word;
 }
 .boxes .box_rps .value, .boxes .box_slaves .value {
     font-size: 32px;
@@ -128,6 +130,10 @@ a:hover {
     background: #4e4141;
 }
 
+#host_url {
+    font-size: 14px;
+    font-weight: normal;
+}
 .boxes .box_running {
     display: none;
 }
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -10,6 +10,12 @@
         <div class="top-content container">
             <img src="/static/img/logo.png?v={{ version }}" class="logo" />
             <div class="boxes">
+                <div class="top_box box_url">
+                    <div class="label">HOST</div>
+                    <div class="value" id="host_url">
+                        {{host}}
+                    </div>
+                </div>
                 <div class="top_box box_status">
                     <div class="label">STATUS</div>
                     <div class="value" id="status_text">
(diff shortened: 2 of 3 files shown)


thaffenden:
I've just seen an issue with this pull request where running in master with a large number of slaves and a large number of failures ruins the header layout. I'll update the css to address this, so wouldn't merge it until I have added another commit.

thaffenden:
I've set a max width to the reset stats box so the text now wraps with the words on different lines and reduced the padding for the boxes from 20px to 12 px. 
Updated screenshot (with some dummy data trying to show some longer strings):
screen shot 2016-07-15 at 10 29 50

heyman:
Hm, not sure about this since it won't work when setting the hostname through the `host` attribute of the Locust class.

thaffenden:
@heyman sorry for delay, only just had a chance to look at this again. 
I've added an `if` block to set the ui's host value based on where it's specified in your run. 
Let me know if you have any thoughts or concerns about it.

PayscaleNateW:
I included a similar change to my recent pull request, only putting it lower in the header. I too only looked at the runner's host, not the locusts. My concern was what if different locusts had different hosts? I think the host in the ui should only display if the runner itself has a host.

thaffenden:
@PayscaleNateW Awesome, you changes were on my todo list of thing's I'd like to improve in locust, so that saves me the work! 
Do you have a screenshot of your implementation for comparison?

With regards to getting the actual host data I agree if they are different it doesn't make sense to display one of them, but if you have to pick I assume you would want to show the one set by the master. Is that an actual use case that people do with locust?
@heyman @justiniso Any thoughts on this?

heyman:
@thaffenden Yeah, I guess the most sensible thing to do is to only show it if it's been set through the command line options, or in the host attribute of a single Locust class (like you've done).

Other than that, I think I would also like to decrease the font-size of the hostname a bit (maybe 14px and normal font-weight), just to make it a little less prominent, and work better with really long hostnames.

thaffenden:
@heyman I've updated the CSS to your recommended values.
Here's a screenshot with some dummy data long strings:
locustheader

justiniso:
@thaffenden you'll want to pull in the latest changes from master. There have been some style updates to the header.

thaffenden:
@justiniso rebased from master.
New screenshot:
locustheader2

thaffenden:
Rebased again due to other UI changes that have been merged to master.
New screenshot:
locust_header

thaffenden:
@justiniso @heyman is there anything else you wanted doing to this one or is it good to go?

justiniso:
Thanks @thaffenden !

PR #451, opened 2016-07-19
Changes to consolidate errors

schuSF:
These changes were made so Failures made during a Locust run can we easily viewable. 

More information on issue [#452](https://github.com/locustio/locust/issues/452)


heyman:
Thanks!

PR #456, opened 2016-07-28
Python 3 compatibility with --no-web option

mrsanders:
Fixes logging during locust hatching when using --no-web with Python 3.
(previously a "ImportError: No module named 'runners'" error is raised and no logging occurs)


justiniso:
Thanks @mrsanders, this is a duplicate of https://github.com/locustio/locust/pull/438, which is older so I'll keep that one around and close this.

heyman:
Thanks!

PR #460, opened 2016-08-11
Fix document bug concerning '--host' option.

d6e:
When this rst is converted to html, the "--host" becomes "–host". This change fixes that with an inline code block.


heyman:
Thanks!

PR #462, opened 2016-08-16
Update gevent==1.1.2

di:


cgoldberg:
did you run all tests with tox using gevent 1.1.2?

di:
@cgoldberg Are there tests that aren't run by CI?

cgoldberg:
@di oops.. my bad.  travis used to not support Python 3.5, and we had to run them locally... but your results show 3.5 in there.  so I am :+1:

PR #467, opened 2016-08-29
add 90% line support

scyiwei:
add 90% line stats


justiniso:
It's unclear to me what this adds. p90 is already covered in the stats.

PR #487, opened 2016-10-17
Remove unsupported language classifiers

powersjcb:
Until this project supports it, the classifiers for python 3 versions should be removed from setup.py


heyman:
The master branch supports Python 3. The latest on PyPI does not, but it also doesn't state that (https://pypi.python.org/pypi/locustio).

PR #514, opened 2017-01-11
Expose number of clients to TaskSet

ticosax:
Expose client_index to `TaskSet`

Useful in distributed mode, when we need to simulate users identified uniquely.
Otherwise a `TaskSet` doesn't know its range of ids to generate.

Example: 3 clients share the work of 3000 users.
    
client 1 will generate ids from 1 to 1000
client 2 will generate ids from 1001 to 2000
client 3 will generate ids from 2001 to 3000.

Otherwise all clients will generate ids from 1 to 1000 since they don't know how many they are.

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -96,9 +96,9 @@ class Locust(object):
     client = NoClientWarningRaiser()
     _catch_exceptions = True
     
-    def __init__(self):
-        super(Locust, self).__init__()
-    
+    def __init__(self, runner=None):
+        self.runner = runner
+
     def run(self):
         try:
             self.task_set(self).run()
@@ -125,8 +125,8 @@ class HttpLocust(Locust):
     The client support cookies, and therefore keeps the session between HTTP requests.
     """
     
-    def __init__(self):
-        super(HttpLocust, self).__init__()
+    def __init__(self, runner=None):
+        super(HttpLocust, self).__init__(runner=runner)
         if self.host is None:
             raise LocustError("You must specify the base host. Either in the host attribute in the Locust class, or on the command line using the --host option.")
         
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -28,6 +28,7 @@ SLAVE_REPORT_INTERVAL = 3.0
 
 
 class LocustRunner(object):
+    client_index = 0
     def __init__(self, locust_classes, options):
         self.options = options
         self.locust_classes = locust_classes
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -1,10 +1,10 @@
...
 [testenv]
 deps =
-    mock
     pyzmq
-    unittest2
+    pytest
+    py27: mock
 commands =
-    unit2 discover []
+    pytest {posargs}
```
cgoldberg: why are you switching to pytest?

cgoldberg:
there is some good stuff in here, but there are also many small unrelated changes.  Those should be moved into separate PRs.  (for example, I wouldn't expect the test runner to change in a PR for exposing the number of clients)

tox.ini
ticosax: pytest is really awesome.
Most important feature IMO, you can write concise tests that are easy to read and so to maintain.

ticosax:
I rebased it to keep only the commit that this pull request intent to bring.

tox.ini
cgoldberg: pytest is great... but nothing in this PR requires this change.   And more importantly... We don't just arbitrarily change the dev/test environment without discussion.

thanks for removing this in your rebase.

mbeacom:
@ticosax You could reopen this and the builds should pass.

alexandrul:
This PR passes all travis tests when rebased on current master, is it possible to merge it?

heyman:
I think it's a god idea to enumerate the slave runners, but I'm not sure we want to expose the whole runner instance to the Locust instances. Maybe we could set the `LocustRunner.client_index` but instead of injecting the runner instance into each Locust instance, one would have to manually import `locust.runners.locust_runner` in the test scripts to access it (we could document it as a semi-public/private API)?

I see a lot of potential problems if users starts to treat the runner instance as part of the public API.

Another possibility would be to inject the client_index into each LocustUser, but that will suddenly make them" aware" of how they are being run (which might be okay, but it's a conceptual change of how Locust users work now).

alexandrul:
@heyman actually, I have to use various workarounds to ensure that each locust gets a distinct configuration, so making them aware of the setup might not be a bad idea after all

I could surely use a variable for identifying each node and another one for each individual locust (something that can be used as a key for my configuration dicts).

cyberw:
Hi @ticosax ! This PR has been idle for a very long time and has conflicts, so I am closing it. Please feel free to reopen it and resolve the conflicts if you are still interested. There could be workarounds for your use case (for example, setting an environment variable identifying the client at startup)

PR #515, opened 2017-01-11
seems sane to not support py26 anymore

ticosax:


cgoldberg:
> _seems sane to not support py36 anymore_

I think you meant py26, not py36 :)

dropping Python 2.6 support seems reasonable at this point.  Could you please also remove all references to 2.6 from the following:
* `README.md`
*  `setup.py`.
* `/docs/installation.rst`

There are also some 2.6 specific bits in: `/locust/test/testcases.py`

ticosax:
@cgoldberg thank you for the feedback I will follow up.

chris-surge:
I agree, let's drop it, no reason someone can't use 2.7, especially for a library designed to load test.

cgoldberg:
is someone willing to provide an updated PR?  We need to remove mentions of 2.6 support from:
* /README.md
* /setup.py
* /docs/installation.rst
* /locust/test/testcases.py

cgoldberg:
sayonara Python 2.6!

PR #517, opened 2017-01-12
Unpin pyzmq

ticosax:


cgoldberg:
please include descriptions with PR's including the motivation for the change.

ticosax:
Sorry I though it was self explanatory.
Why pining pyzmq is bad ? because it want to be able to choose which version I want to use within my own tasks.

PR #523, opened 2017-01-25
Specify host in web ui

PayscaleNateW:
Add support for specifying the hostname to test in the web ui, as opposed to in the command line.

- Tests pass (had to make some changes there)
- Hostname is now displayed in the UI (similar to https://github.com/locustio/locust/pull/295)
- Updated docs to reflect that you no longer need to specify host in the command line


PayscaleNateW:
Hmm getting closer - now just python 2.6 is failing

PayscaleNateW:
Having a lot of trouble debugging that 2.6 issue... I get different issues in each environment I try. So far have run vagrant up to try to use the virtual environment, in which case running `export TOXENV=py26; tox` gets me:

```
making hard links in locustio-0.8a2...
hard linking MANIFEST.in -> locustio-0.8a2
error: Operation not permitted

ERROR: FAIL could not package project - v = InvocationError('/usr/bin/python /vagrant/setup.py sdist --formats=zip --dist-dir /vagrant/.tox/dist (see /vagrant/.tox/log/tox-0.log)', 1)
```

Also, on my mac, I get: 

```
'import site' failed; use -v for traceback
  File "/usr/local/lib/python3.5/site-packages/virtualenv.py", line 16, in <module>
    import base64
  File "/usr/local/lib/python2.6/base64.py", line 9, in <module>
    import struct
  File "/usr/local/lib/python2.6/struct.py", line 1, in <module>
    from _struct import *
ImportError: dlopen(/usr/local/lib/python2.6/lib-dynload/_struct.so, 2): Symbol not found: _PyExc_DeprecationWarning
```

cgoldberg:
there is an open PR to drop 2.6 support

cgoldberg:
@PayscaleNateW don't worry about Python 2.6.  I just merged a PR that drops support for it.  
(make sure to rebase your branch and fix any conflicts)

PayscaleNateW:
Awesome, thanks so much @cgoldberg !

PayscaleNateW:
Now tests are actually passing; conflicts resolved. Hopefully the pr looks ok

sgerhardt:
@PayscaleNateW I pulled your changes, and I'm wondering, does it make sense to default in any values for the host text box? Maybe by parsing the locust file, and/or by reading a `--host` flag?

For example, it's now an extra step to run basic.py in the examples. It used to default to the value in the file, but now we'd have to explicitly enter it into the text box. 

Otherwise, I like the idea of making this configurable from the command line.

PayscaleNateW:
Thanks for checking it out! Good call on including the default value. Just added a basic default. In terms of getting it from the locustfile, what would be the recommended way? Is it to set `runners.locust_runner.host` to `locust_classes[0].host` or something like that? That feels a bit off to me - what if you have multiple classes with different hosts? Then that would override the other classes' hosts when it ran. I know a similar problem emerges for displaying the host at the top in the ui. Seems like storing the host on individual classes creates some difficulties.

sgerhardt:
Fair point, trying to parse the files wouldn't work.  I think "Set host input default to command line value" is a good enough improvement.

PayscaleNateW:
Cool. Maybe I should remove the host attribute from the examples?

paravz:
Any help needed for this feature to merge? This feature would help install master+slaves once and POST different target hosts via web ui. Thanks @PayscaleNateW

PayscaleNateW:
Hi @paravz - definitely open to help! I sort of lost track of this PR and looks like it now conflicts. So I don't know how it would play with the code now. But definitely feel free to pick it up from here.

mbeacom:
@PayscaleNateW Conflicts resolved.  What other changes are outstanding here?

cgoldberg:
> ALTERNATIVE_HOST = 'http://localhost'
> SWARM_DATA_WITH_HOST = {'locust_count': 5, 'hatch_rate': 5, 'host': ALTERNATIVE_HOST}
> SWARM_DATA_WITH_NO_HOST = {'locust_count': 5, 'hatch_rate': 5}

these might be better declared within the tests since the data isn't shared globally

Acanthostega:
Hi,
it seems great. Nobody wants to merge it ?

mbeacom:
@Acanthostega If you'd like to resolve the conflicts and turn around the requested changes/feedback, we're definitely open to reviewing the PR.

cyberw:
@Acanthostega I'm so sorry that this PR has not gotten the attention it deserves. If you just resolve the conflicts I will test it out and merge it.

cyberw:
@heyman I would like to merge this, but I cant seem to fix the conflicts directly on GH (I think I can do it by adding a commit to master with one line to remove that diff completely).

heyman:
I've merged master into this branch. However the web UI needs to be updated before we can merge it. I also think it should have the label `Target Host (optional)` and be the third input box instead of first (since it's optional).

This is what it looks like currently:

image

cyberw:
I can fix the UI ordering once it is merged. Having this really old PR open is annoying to me :P

@heyman Is it ok if I make number of users to simulate and hatch rate default to 1 at the same time? I dont use the web ui much, but having to fill out that number just to be able to run is annoying to me :)

heyman:
@cyberw You can merge it into a separate branch and make the changes there. Or make the changes directly to this PR (the [PayscaleNateW:specify-host-in-web-ui](https://github.com/PayscaleNateW/locust/tree/specify-host-in-web-ui) branch). 

The rendering  bug (see the upper left corner below the logo in the screenshot from my previous message) needs to be fixed as well. It 's probably caused by changes to the DOM/CSS that has been made in master after this PR was created.

> Is it ok if I make number of users to simulate and hatch rate default to 1 at the same time?

I think it should be much more common that one wants to spawn >1 users than exactly 1 user, so I'd rather leave them empty I think.

cyberw:
I'll try to make the changes directly to the PR.

I can make it default to 1 and auto select the value when you click the field so there is no downside, even if you want to run >1 users...

cyberw:
Hmm.. actually, if there needs to be fixes to the DOM/CSS then I'm probably not the right person to fix it :)

heyman:
> Hmm.. actually, if there needs to be fixes to the DOM/CSS then I'm probably not the right person to fix it :)

I can take a look at it!

heyman:
I've fixed the issues, but while doing so I found another issue that we need to fix before merging (looking back in the thread, I can also see that @PayscaleNateW has already [mentioned](https://github.com/locustio/locust/pull/523#issuecomment-277895224) this issue):

When running a test with two (or more) different Locust classes, with different `host` attributes specified (an example could be a `WebUser` class that would use `http://example.com` as host and a `ApiUser` class that would use `http://api.example.com` as host), both classes' `host` attribute would get overridden by the value specified in the form.

I think the solution to this would be to only pre-fill the value picked up from the Locust classes if all of them have *the same* host attribute, while also adding some kind of caveat note that specifying a host through the Web UI will override the host for *all* locust classes.

heyman:
I *think* this should be mergeable now.

cyberw:
LGTM & tested it

PR #525, opened 2017-01-26
Update supported Python versions in docs

chuckus:
To match what it says in README.md


chuckus:
I see that only 0.8a1 and above support Python 3.x, so closing.

PR #536, opened 2017-02-15
Updated installation.rst

klis87:
Updated supported Python versions.


klis87:
Closing as I didn't notice that Python 3 support was added only in alpha release.

PR #537, opened 2017-02-15
Add units to table

benrudolph:
It wasn't clear to me what units were being displayed in the table. Assuming it was milliseconds and adding it the table so it's clear. Thoughts?


heyman:
Sounds good, but maybe it would be enough to add it to the `title` attribute  of the `<th>` in order to add less clutter the UI?

benrudolph:
👍 agree with that, will update

justiniso:
Awesome, thanks for this!

PR #541, opened 2017-02-17
Started working on a more modern and (hopefully) better looking design

heyman:
I've started to work on modernizing and improving the design of the web UI. Mainly making the design slightly flatter, adding some paddings, and brighten up the design slightly (mainly because the current background is so dark, that it's hard to use darker colors to emphasize other elements). I've also changed so that the page width can go up to 1200px if there's room (previously it was fixed to 1000px). 

Also, I updated the logo with one with text, but I've made it smaller. I think it looks better, and with the increased page width, and a slightly more narrow logo, it works pretty well. 

It's a work in progress. This is what it currently looks like:

image

Compared to:

image

There's still a couple of more things that I'd like to fix:

* [x] Change colors of the modal dialogs
* [x] Make the start and stop buttons slightly flatter, and retina (probably by not using an image)
* [x] Make the reset stats link into a button (that matches the stop button)


cgoldberg:
@heyman you are doing some awesome work! :1st_place_medal:

heyman:
I think this is ready to be merged now. There's definitely more tweaks that can be done, but overall I think it's an improvement. Once it's merged, I'll update #539 to make sure that the charts looks good in the new design as well.

Here are the latest changes:

image

image

PR #546, opened 2017-02-20
Prohibits the locustfile from being named 'locust.py'

### LINKED ISSUE #138: Test file can not be named locust.py (or any other name that is the same as an existing python package)
Thanks for writing the locust test tool. It's very interesting!

I follow the instruction on http://locust.io/,  first I installed 

```
pip install locustio
pip install pyzmq
```

Then I copied and pasted the sample "Simple example" to  locust.py in my local directory and run it against a server listening on another PC (10.3.0.3) as the following but got error:

```
 locust -f locust.py -H http://10.3.0.3
[2014-04-04 12:03:28,901] john-HP/ERROR/locust.main: No Locust class found!
```

Not sure what have gone wrong. 

Thanks.

cgoldberg:
Fixes #138

This PR prohibits naming your locustfile `locust.py`, which previously caused Locust to exit with a misleading message ("No Locust class found").

The new behavior is to exit with a more helpful error message explaining not to name the file `locust.py`. 

The issue was occuring because the locustfile gets imported as a Python module (with the .py extension removed).  So naming it `locust.py` caused a namespace clash with the internal `locust` module, which would get re-imported instead of the locustfile.

Note: This only fixes the common case where you have named it `locust.py`.  There are also other reserved names that can not be used as locustfile names.  We might want to add something to the docs to explain this.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -350,8 +350,13 @@ def main():
         sys.exit(0)
 
     locustfile = find_locustfile(options.locustfile)
+
     if not locustfile:
         logger.error("Could not find any locustfile! Ensure file ends in '.py' and see --help for available options.")
+        sys.exit(1)
+
+    if locustfile == "locust.py":
+        logger.error("The locustfile must not be named `locust.py`. Please rename the file and try again.")
         sys.exit(1)
 
     docstring, locusts = load_locustfile(locustfile)


locust/main.py
```diff
@@ -350,10 +350,15 @@ def main():
         sys.exit(0)
 
     locustfile = find_locustfile(options.locustfile)
+
     if not locustfile:
         logger.error("Could not find any locustfile! Ensure file ends in '.py' and see --help for available options.")
         sys.exit(1)
 
+    if locustfile == "locust.py":
+        logger.error("The locustfile must not be named `locust.py`. Please rename the file and try again.")
```
justiniso: Error message could be more helpful with: `Did you mean locustfile.py?`

cgoldberg: @justiniso thanks!
however, the file can be named something when you use the `-f` option.  do you think your message implies it must be named `locustfile.py`?

justiniso: No I mean if someone is naming their file `locust.py`, seems like they don't have many locustfiles and are just looking for whatever "default" name works, in which case `locustfile.py` suits that need. But I guess it can be misinterpreted. Let's leave it then.

PR #549, opened 2017-02-22
Refactored stats code and display median as well as 95% percentile response times in web UI's charts

heyman:
I've refactored the stats code to track the total request stats in it’s own separate StatsEntry instance under RequestStats.total instead of dynamically calculating the total stats from all other stats entries.

I believe this makes the code slightly more clear. The total stats is now accessible through `RequestStats.total` and there is no longer a need for `RequestStats.aggregated_stats()`. `StatsEntry.extend()` is now only used for merging stats from slave reports, and therefore the `full_request_history` argument is not longer needed.

The background for this is that I'm looking to display the 95% percentile response time in the response time graph (as mentioned in https://github.com/locustio/locust/pull/539#issuecomment-280798460). When I started to look into the best way of implementing this I realized that the way we currently store stats data, doesn't allow us to retrieve percentile response times for the last couple of seconds, only for the whole test run (since we use the `StatsEntry.response_times` dict to calculate it). To allow this, we need to store a snapshot of this dict every second for a number of seconds back. We probably don't want to do this for every single stats entry, but it should be okay to do it for the total stats entry. Now that the total stats is tracked in it's own `StatsEntry`, and isn't calculated dynamically on the fly, it'll be easier to do that.


heyman:
I think this is pretty much ready to be merged. Would love if someone could take a look and review it.

These are the main changes:

`RequestStats` now holds a `total` attribute of the type `StatsEntry` that represents the "total" row in the statistics. Before the total was calculated on the fly by aggregating the StatsEntry instances for each entry (row) in the statistics. This allowed me to simplify the `StatsEntry.extend()` by removing the `full_request_history` argument.

`StatsEntry` has gotten a cache that holds the `response_times` dict for the last 20 seconds. The cache is turned off by default, but it's turned on for `RequestStats.total`. This cache is used to calculate the *current* response time percentiles. The 50 and 95 percentiles are now graphed in the web UI.

locust/stats.py
```diff
@@ -117,6 +176,19 @@ class StatsEntry(object):
     This dict is used to calculate the median and percentile response times.
     """
     
+    use_response_times_cache = False
+    """
+    If set to True, the copy of the response_time dict will be stored in response_times_cache 
+    every second, and kept for 20 seconds. We can use this dict to calculate the *current*
```
aldenpeterson-wf: similar comment to my question about the 10 seconds sliding window time, I can see use cases for adjusting this (for example if your operations take a long time you may end up with wonky results here if the cache is only kept/reported for 20 seconds)

locust/stats.py
```diff
@@ -439,18 +539,21 @@ def median_from_dict(total, count):
...
 def on_request_success(request_type, name, response_time, response_length):
-    global_stats.get(name, request_type).log(response_time, response_length)
+    #global_stats.get(name, request_type).log(response_time, response_length)
+    global_stats.log_request(request_type, name, response_time, response_length)
     if global_stats.max_requests is not None and (global_stats.num_requests + global_stats.num_failures) >= global_stats.max_requests:
         raise StopLocust("Maximum number of requests reached")
 
 def on_request_failure(request_type, name, response_time, exception):
-    global_stats.get(name, request_type).log_error(exception)
+    #global_stats.get(name, request_type).log_error(exception)
```
aldenpeterson-wf: should this be deleted instead of commented?

locust/stats.py
```diff
@@ -439,18 +539,21 @@ def median_from_dict(total, count):
 """
 
 def on_request_success(request_type, name, response_time, response_length):
-    global_stats.get(name, request_type).log(response_time, response_length)
+    #global_stats.get(name, request_type).log(response_time, response_length)
```
aldenpeterson-wf: should this be deleted instead of commented?

heyman: Ah, yes :)!

locust/stats.py
```diff
@@ -20,20 +21,85 @@
...
+    num_requests: Number of request made (could be derived from response_times, 
+                  but we save some CPU cycles by using the value which we already store)
+    percent: The percentile we want to calculate. Specified in range: 0.0 - 1.0
+    """
+    num_of_request = int((num_requests * percent))
+
+    processed_count = 0
+    for response_time in sorted(six.iterkeys(response_times), reverse=True):
+        processed_count += response_times[response_time]
+        if((num_requests - processed_count) <= num_of_request):
```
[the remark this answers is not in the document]
heyman: 👍

locust/stats.py
```diff
@@ -341,20 +400,50 @@ def get_response_time_percentile(self, percent):
...
-            processed_count += self.response_times[response_time]
-            if((self.num_requests - processed_count) <= num_of_request):
-                return response_time
-
+        return calculate_response_time_percentile(self.response_times, self.num_requests, percent)
+    
+    def get_current_response_time_percentile(self, percent):
+        """
+        Calculate the *current* response time for a certain percentile. We use a sliding 
+        window of (approximately) the last 10 seconds when calculating this.
```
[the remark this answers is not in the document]
heyman: Yeah, a module variable should be good. Right now we only expose this to the users as the "current" response time percentile, which lets us get away with it being approximately the last 10 seconds. We also only use it in the very ephemeral web UI charts. If we were to expose it as an option, maybe people would expect it to be more exact.

Will fix!

locust/stats.py
heyman: I don't think it'll be more wonky than the other stats. Since the stats gets reported once a request is complete, if you have endpoints that takes really long (e.g. 60 seconds), the stats will feel kind of delayed.

locust/stats.py
```diff
@@ -366,6 +455,17 @@ def percentile(self, tpl=" %-" + str(STATS_NAME_WIDTH) + "s %8d %6d %6d %6d %6d
             self.get_response_time_percentile(0.99),
             self.max_response_time
         )
+    
+    def _cache_response_times(self, t):
+        self.response_times_cache[t] = CachedResponseTimes(
+            response_times=copy(self.response_times),
+            num_requests=self.num_requests,
+        )
+        if len(self.response_times_cache) > 20:
```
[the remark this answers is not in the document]
heyman: Since the cache might not hold response times for every second. We scan through the cache at time()-CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW +/- 10 seconds to find a cache entry which we'll use to calculate the delta in order to get the "current" response time percentile.

I've changed the cache size to be CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW + 10.

I guess we *could* make the number of seconds we scan configurable as well, but I really don't see a use-case where one would need to change this.

aldenpeterson-wf:
Looks good to me!

heyman:
Okay, great, thanks!

Issue #331, opened 2015-08-27
Percentiles rounding error


pior:
```
 Name                                                           # reqs    50%    66%    75%    80%    90%    95%    98%    99%   100%
--------------------------------------------------------------------------------------------------------------------------------------------
 POST /projects/<>/devices                                          50    230    240    250    270    320    340    600    600    596
```

Note that all latencies are multiple of 10. And 99% is higher than 100%.

jalan:
I noticed this as well. Seems to be deliberate: [locust/stats.py#L160](https://github.com/locustio/locust/blob/master/locust/stats.py#L160)

The numbers are accurate enough for my purposes, so no complaint here

cgoldberg:
@jalan 99% > 100% is confusing.  Perhaps the 100th percentile should be rounded also?

jalan:
@cgoldberg: Agreed. One option would be to round the min and max times just like all the other times:

```diff
diff --git a/locust/stats.py b/locust/stats.py
index 267dee4..79d71a5 100644
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -151,12 +151,6 @@ class StatsEntry(object):
     def _log_response_time(self, response_time):
         self.total_response_time += response_time
 
-        if self.min_response_time is None:
-            self.min_response_time = response_time
-
-        self.min_response_time = min(self.min_response_time, response_time)
-        self.max_response_time = max(self.max_response_time, response_time)
-
         # to avoid to much data that has to be transfered to the master node when
         # running in distributed mode, we save the response time rounded in a dict
         # so that 147 becomes 150, 3432 becomes 3400 and 58760 becomes 59000
@@ -173,6 +167,12 @@ class StatsEntry(object):
         self.response_times.setdefault(rounded_response_time, 0)
         self.response_times[rounded_response_time] += 1
 
+        # update min and max response times
+        if self.min_response_time is None:
+            self.min_response_time = rounded_response_time
+        self.min_response_time = min(self.min_response_time, rounded_response_time)
+        self.max_response_time = max(self.max_response_time, rounded_response_time)
+
     def log_error(self, error):
         self.num_failures += 1
         self.stats.num_failures += 1
```

Jonnymcc:
@pior Looks like a fix was merged for this. Can this bug be closed?

pior:
I can't really test that currently.
But from the discussion and the fix, it looks to me like the issue I reported is properly addressed.

/cc @Jonnymcc

PR #558, opened 2017-03-06
#331:  Use rounded_response_time for min/max/total response times

jude:
I was going to open a new ticket for these changes, but saw #331 already existed with almost the same changes that ended up working for me.  (I also use the rounded_response_time when computing the total response time)

I also added a few unit tests to cover the rounding behavior a little, and added a test_suite line to setup.py, so that 'python setup.py test' will find and run the tests.


heyman:
I agree that it's confusing that you might get 99 percentile that is larger than 100% (as discussed in #331). 

However, I don't think we should round the `total_response_time` as it's used to calculate the average response time. Rounding the min and max should be okay, but I'm also thinking that maybe it's even better to change so that we use `self.get_response_time_percentile()` for the 100% value as well (at this row: https://github.com/jude/locust/blob/a65de0fddc9bb6f4ca027325cbf510cfdbef7839/locust/stats.py#L358), and not rounding the min/max.

The rounding of the response times are a necessary evil that drastically shrinks the amount of data that needs to be sent from the worker nodes to the master. But that's not really needed for min and max.

jude:
Cool.  I've updated the PR as per the comment above.

I also updated the Vagrantfile to use xenial instead of precise, since it looks like precise has EOL'd since the the PR was opened.

heyman:
Great, thanks!

Issue #399, opened 2016-03-31
on_request_xxx checks exiting condition against the wrong number


yalinhuang:
`on_request_xxx()` checks exiting condition before incrementing the relevant variables; this should be done in the reverse order.

A test run exits upon `StopLocust` exception is raised. The condition is checked every time a request is sent and responded (or failed). However the condition checking looks at the wrong number; for example, in `on_request_success()` in [stats.py](https://github.com/locustio/locust/blob/master/locust/stats.py#L415)

```
def on_request_success(request_type, name, response_time, response_length):
    if global_stats.max_requests is not None and (global_stats.num_requests + global_stats.num_failures) >= global_stats.max_requests:
        raise StopLocust("Maximum number of requests reached")
    global_stats.get(name, request_type).log(response_time, response_length)
```

`global_stats.num_requests` is increased by one in [the `log()` function](https://github.com/locustio/locust/blob/master/locust/stats.py#L135) after the condition is checked. This leads to sending one more requests than being asked to.

Was wondering if this design is intended? or?

PR #566, opened 2017-03-22
Fix off by 1 error in stats.py resulting in additional request always being off

aldenpeterson-wf:
Address #399, which is the same issue we just hit in our usage.

The counter is incremented in `log` but it was being checked for the end condition before that point was hit.
This meant that if you ran 3 requests using the command line mode:

```
locust --no-web -c 1 -r 1 -n 3 -f tornado_nats_locust.py
```
you would end with Locust actually executing 4 tasks.

FYI @cgoldberg @heyman @seangerhardt-wf


heyman:
Thanks!

PR #570, opened 2017-04-03
Add slack signup link to readme

aldenpeterson-wf:
I don't think there are nice "badges" for slack the way there are for gitter.

I didn't remove that link either, is the goal here to replace gitter with Slack?

@heyman


heyman:
Yeah, I think we should remove the Gitter link and stick with just one channel for chatting.

heyman:
👍

PR #577, opened 2017-04-26
Fix formatting issue combining double dashes into single dash

swoodford:
Copying and pasting the command to get help into the console fails because of a formatting issue that combines the double dashes into a single dash.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -105,7 +105,7 @@ host defaults to 127.0.0.1)::
 
     To see all available options type
 
-        locust --help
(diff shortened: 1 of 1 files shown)


cgoldberg:
you can leave it unquoted... just add a double-colon to the end of the previous line so it renders in a code block.  That will keep it more consistent with how nearby code looks. 

```
To see all available options, type::

        locust --help
```
if you push a change for this, I'd be happy to merge it for you.

cgoldberg:
LGTM

PR #579, opened 2017-04-27
Add description of RPS breakdown between multiple locusts.

markfickett:
I'm guessing this is how it works. But, I'm trying to put my question / guesses in the form of doc updates in case that's useful for posterity.

If you have multiple Locusts, by default it looks like it runs all of them (but not base classes; maybe it looks for a task_set)?

If you hatch N total Locusts via the UI, is that the total across all Locust classes? I assume, given the weighting, it wouldn't be N of each.


cyberw:
Hi! This PR has been untouched for a very long time so I am closing it. Let me know if this was a mistake and feel free to open it again.

PR #588, opened 2017-05-17
Added failures per second as a series in the chart

alercunha:
As a user of locust it's very useful to know how many requests are failing in the chart. The failure % is useful but when pushing heavy loads against an API in my case it's common the autoscaling can't cope up with the demand and a lot of fails start to happen. Having those represented in the chart is very valuable to understand how the scaling overall is working.

image

### DIFF
--- a/locust/static/locust.js
+++ b/locust/static/locust.js
@@ -116,7 +116,7 @@ $(".stats_label").click(function(event) {
 });
 
 // init charts
-var rpsChart = new LocustLineChart($(".charts-container"), "Total Requests per Second", ["RPS"], "reqs/s");
+var rpsChart = new LocustLineChart($(".charts-container"), "Total Requests per Second", ["RPS", "FPS"], "reqs/s");
 var responseTimeChart = new LocustLineChart($(".charts-container"), "Average Response Time", ["Average Response Time"], "ms");
 var usersChart = new LocustLineChart($(".charts-container"), "Number of Users", ["Users"], "users");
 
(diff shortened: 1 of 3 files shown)


jrouaix:
Too sad this PR is not merged since 2 years :'(

alercunha:
> Too sad this PR is not merged since 2 years :'(

My default installation of locust still points to this branch. I find it much more useful that way....

cyberw:
@heyman I think this is a very useful contribution (although I dont use the web ui myself). Ok to merge, if someone fixes the conflicts?

cyberw:
Would be nice if we could make the line red.

alercunha:
@cyberw PR is too old. I'll drop this one and create a new PR fresh from master in a couple days.

cyberw:
@alercunha Awesome!

heyman:
I'm +1. I wonder if it's best to have the failure line in the same graph, or if it's better to have a separate graph for failures (as suggested in #952)? I also think that the color of the line should be red.

cyberw:
@alercunha Is your new PR on the way? :)

alercunha:
@cyberw sorry I've been quite busy in the past few days. I'll try to get this done during the weekend.

alercunha:
@cyberw @heyman I posted a [new PR](https://github.com/locustio/locust/pull/1140). @heyman based on my experience having the failure line in the same graph makes more sense. Being able to see the correlation between the two lines in the same chart helped me figuring out the root cause of degradation of my service some years ago.

PR #592, opened 2017-05-25
Add sequential-execute tasks support for <TaskSet> class, if want to …

lailongwei:
…use sequential-execute property, Add random_execute = False in your <TaskSet> class, and specific every tasks order at @task() decorator

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -19,7 +19,7 @@ from .exception import LocustError, InterruptTaskSet, RescheduleTask, Reschedule
 logger = logging.getLogger(__name__)
 
 
-def task(weight=1):
+def task(weight=1, order=-sys.maxint - 1):
     """
     Used as a convenience decorator to be able to declare tasks for a TaskSet 
     inline in the class. Example::
@@ -33,9 +33,9 @@ def task(weight=1):
             def create_thread(self):
                 pass
     """
-    
+
     def decorator_func(func):
-        func.locust_task_weight = weight
+        func.locust_task_info = {"weight": weight, "order": order}
         return func
     
     """
@@ -158,12 +158,18 @@ class TaskSetMeta(type):
                     new_tasks.append(task)
         
         for item in six.itervalues(classDict):
-            if hasattr(item, "locust_task_weight"):
-                for i in xrange(0, item.locust_task_weight):
+            if hasattr(item, "locust_task_info"):
+                for i in xrange(0, item.locust_task_info["weight"]):
                     new_tasks.append(item)
-        
+
+        if "random_execute" not in classDict:
+            classDict["random_execute"] = True
+        elif not classDict['random_execute']:
+            new_tasks = sorted(
+                new_tasks, cmp=lambda left, right: cmp(left.locust_task_info["order"], right.locust_task_info["order"]))
+
         classDict["tasks"] = new_tasks
-        
+
         return type.__new__(mcs, classname, bases, classDict)
 
 @six.add_metaclass(TaskSetMeta)
@@ -241,6 +247,8 @@ class TaskSet(object):
             self.min_wait = self.locust.min_wait
         if not self.max_wait:
             self.max_wait = self.locust.max_wait
+
+        self.last_execute_task_index = 0
 
     def run(self, *args, **kwargs):
         self.args = args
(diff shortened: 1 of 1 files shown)


lailongwei:
Add sequential-execute tasks support for <TaskSet> class, if want to yse sequential-execute property, Add random_execute = False in your <TaskSet> class, and specific every tasks order at @task() decorator

lailongwei:
### Add sequential-Execute support on **`TaskSet`**
Add sequential-Execute support on TaskSet, e.g:

```python
class MyNestedTask(TaskSet):
    @task(order=1)
    def nest_task_1(self):
        print('Hello, I am nest task 1')

    @task(order=2)
    def nest_task_2(self):
        print('Hello, I am nest task 2')
        self.interrupt()

class MyTask(TaskSet):
    tasks = {
        # Nested class, value can be number or dict, if is number, is weight, if value is dict, you can specific 'order' and 'weight' setting, default order value is intmax(2147483647), default weight value is 1.
        MyNestTask: {'order': 3, 'weight': 1},
    }

    @task(order=1)
    def task_1(self): # task_1 executes first over task_2
        print('Hello, I am task 1')

    @task(order=2)
    def task_2(self): #task_2 executes after task_1
        print('Hello, I am task 2')

    random_execute = False # Tell Locust do no random execute tasks(default is True, for compatible with old versions)
```

---
### Add new event **`load_success`**
Add new event define `load_success`, when all locusts load successfully, will fire this event, e.g:

```python
def OnPriorInit(options, locusts):
    print("locust load success, options:\n{}'.format(options))
    print("locusts(count: {}):".format(len(locusts)))
    for locust_name, locust_cls in locusts.iteritems():
        print('    {}: {}'.format(locust_name, locust_cls))
```

locust/core.py
```diff
@@ -19,7 +19,7 @@
 logger = logging.getLogger(__name__)
 
 
-def task(weight=1):
+def task(weight=1, order=2147483647):
     """
```
cgoldberg: might be nicer to declare a MAXINT constant and use that instead of 2147483647 directly

cgoldberg:
also note that the Pythyon 3.x builds are currently failing on Travis for this PR

locust/core.py
```diff
@@ -199,21 +238,24 @@ class TaskSet(object):
...
     max_wait = None
     """
     Maximum waiting time between the execution of locust tasks. Can be used to override 
     the max_wait defined in the root Locust class, which will be used if not set on the 
     TaskSet.
     """
-    
+
+    random_execute = None
+    """random execute falg"""
```
cgoldberg: s/falg/flag
:)

locust/main.py
```diff
@@ -462,7 +474,21 @@ def sig_term_handler():
...
     gevent.signal(signal.SIGTERM, sig_term_handler)
-    
+
+    # interval updater
+    def do_interval_update(update_interval):
+        update_times = 1
+        while True:
+            try:
+                events.locust_update.fire(times=update_times)
+            except Exception, e:
```
cgoldberg: change `except Exception, e:`  to `except Exception as e:`
only the newer exception syntax is valid in Python3

locust/core.py
```diff
@@ -137,35 +141,70 @@ class TaskSetMeta(type):
...
         for item in six.itervalues(classDict):
-            if hasattr(item, "locust_task_weight"):
-                for i in xrange(0, item.locust_task_weight):
+            if hasattr(item, "locust_task_info"):
+                for i in xrange(0, item.locust_task_info["weight"]):
                     new_tasks.append(item)
-        
+
+        new_tasks = sorted(
+            new_tasks, cmp=lambda left, right: cmp(left.locust_task_info["order"], right.locust_task_info["order"]))
```
cgoldberg: the `cmp()` function, and the `cmp` parameter to `sorted()` are no longer available in Python3.  You need to use the `key` parameter instead.

locust/core.py
```diff
@@ -73,34 +73,38 @@ class Locust(object):
...
-        
+
     client = NoClientWarningRaiser()
     _catch_exceptions = True
-    
+
     def __init__(self):
         super(Locust, self).__init__()
-    
+        self.task_set_instance = self.task_set(self)
```
cgoldberg: `self.task_set_instance = self.task_set(self)` is failing in Python 2.7 with `TypeError: 'NoneType' object is not callable`

cgoldberg:
another test failure under Python 2.7:
```
FAIL: test_exception_is_catched (locust.test.test_runners.TestMasterRunner)
Test that exceptions are stored, and execution continues
```

thomas-phillips-nz:
Is it possible to have the commit messages in English? I don't know if the locust project has rules about commit messages, but I suspect they should be in English

locust/core.py
lailongwei: 👍 Yep

locust/core.py
lailongwei: unnnnnnnnnn....My mistake, haha

Issue #598, opened 2017-06-07
bump gevent version


ed1d1a8d:
When using gevent 1.1.2 with Python 3.6, we get the following bug: https://github.com/gevent/gevent/issues/903

This is fixed if one updates to gevent 1.2.2. I propose we bump the resource here: https://github.com/locustio/locust/blob/master/setup.py#L43

stefanopretto:
How did you make Locust accept gevent 1.2.2? I updated it manually, but when I try to run Locust I get "The 'gevent==1.1.2' distribution was not found and is required by locustio".

aldenpeterson-wf:
@stefanopretto I believe the latest tag does not have that constraint  https://github.com/locustio/locust/releases/tag/v0.8a1

stefanopretto:
@aldenpeterson-wf I am using 0.8a1 but I get the error mentioned above, after manually updating gevent.

cgoldberg:
updated  in master... see #599

PR #599, opened 2017-06-07
Bump gevent version

ed1d1a8d:
Addresses https://github.com/locustio/locust/issues/598


aldenpeterson-wf:
+1 I've had to do this before too, but I guess I never thought to PR it back..

PR #609, opened 2017-06-21
find locustfile in the root directory

arthurdarcet:
if the `locustfile.py` file is located in the filesystem root, and the `locust` command is launched from the fs root too, then the file is not found.

(to work around this issue, `-f /locustfile` has to be explicitly passed to the `locust` command)

having the locustfile in the fs root is a common use case when running in docker


heyman:
Looks good! Thanks!

PR #612, opened 2017-06-26
Add ability to write csv stats files  

### LINKED ISSUE #598: bump gevent version

aldenpeterson-wf:
I started reviewing https://github.com/locustio/locust/pull/445 but found that it had enough problems that it'd be just easier to rewrite the entire PR.

This:

 - Ability to write csv files for both the requests and distribution csv as a command line flag
- Refactored the logic in the web handler to reuse this csv generation for better portability

I also fixed #598 because that issue has been annoying me for a while.

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -15,8 +15,7 @@ from flask import Flask, make_response, request, render_template
 from . import runners
 from .cache import memoize
 from .runners import MasterLocustRunner
-from .stats import sort_stats
-from locust.stats import median_from_dict, requests_csv, distribution_csv
+from .stats import sort_stats, median_from_dict, requests_csv, distribution_csv
 from locust import __version__ as version
 
 import logging
--- a/locust/web.py
+++ b/locust/web.py
@@ -15,7 +15,8 @@ from flask import Flask, make_response, request, render_template
 from . import runners
 from .cache import memoize
 from .runners import MasterLocustRunner
-from locust.stats import median_from_dict
+from .stats import sort_stats
+from locust.stats import median_from_dict, requests_csv, distribution_csv
 from locust import __version__ as version
 
 import logging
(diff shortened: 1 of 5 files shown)


mambusskruj:
Hi, as I understand one row will write in csv per 2 seconds? Why not per one second? It will be great for some TSDB like InfluxDB.

aldenpeterson-wf:
@mambusskruj I copied the interval here: https://github.com/locustio/locust/blob/master/locust/stats.py#L519

I'd actually be more inclined to _decrease_ the frequency given that it's blocking IO for locust and on slower systems that might affect your loadtest performance.

mambusskruj:
@aldenpeterson-wf well, if locust used with master-slave mod, this is not a big issue I think.

aldenpeterson-wf:
@mambusskruj I could make it a user defined option but I'm not sure that it's worth the added complexity.

I'd be curious what other people think. I could also define the time at the module level so anyone who really wanted to could set the timeframe for both the printing/csv save.

locust/web.py
```diff
@@ -15,7 +15,8 @@
 from . import runners
 from .cache import memoize
 from .runners import MasterLocustRunner
-from locust.stats import median_from_dict
+from .stats import sort_stats
```
cgoldberg: why the 2 different imports from `.stats` and then `locust.stats`?

cgoldberg:
>  I could make it a user defined option but I'm not sure that it's worth the added complexity.

I think 2 secs for a writing interval sounds good for the initial implementation.  Once we gauge impact of this change on performance, we can revisit or make it configurable.

First pass at a review LGTM, but I haven't tested it.  Also, can something be added to the docs or README that describes stats output?

locust/web.py
aldenpeterson-wf: .. no reason, will fix this

aldenpeterson-wf:
@cgoldberg will update the docs.

I also might just make the sleep time to be constants in that file, so someone can change them it if they want since.. python.

aldenpeterson-wf:
I'm not 100% sure where to put this in the docs. There's not really a good section detailing all the flags but I added them, given activity on GH it seems an option like this would be very commonly used.

Also @mambusskruj I added the ability to easily customize how frequently the stats CSV is written.

aldenpeterson-wf:
@cgoldberg this is updated and ready for review

aldenpeterson-wf:
@cgoldberg bump :)

PR #614, opened 2017-06-27
Add GH issue template and update readme

aldenpeterson-wf:
Update the text seen when creating an issue to make them more useful.

Also update README slightly to indicate this change.


mbeacom:
Is this ready to merge or do we need to determine the `TBD` entries?

aldenpeterson-wf:
The idea is that it should be filled in when someone writes an issue. A lot of the issues submitted here are.. unclear/vague and it'd be nice to have more information about how to reproduce them.

PR #616, opened 2017-07-04
Add order argument

ilmesi:
This approach is similar to #592 but it allows you to use order and ratio together and avoids the need to use a MAXINT flag.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -7,7 +7,7 @@ Locust class
 ============
 
 .. autoclass:: locust.core.Locust
-	:members: min_wait, max_wait, task_set, weight
+	:members: min_wait, max_wait, task_set, weight, order
 
 HttpLocust class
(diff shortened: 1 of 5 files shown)


javivdm:
Hi, I'm new to Locust (and to GitHub actually), but I am very interested in this functionality and the approach you took. Is this going to be merged at some point or how does this work?

Thanks!

cyberw:
Hi @ilmesi ! This PR has not been touched in a very long time and now has merge conflicts, so I'm closing it. Feel free to reopen it if you are still interested in changing this!

ilmesi:
Hi @cyberw! Same as the other one, I needed this feature to perform a bunch of tests a few years ago. I don't why it never was reviewed like #592, since it is clearly much more simple and safer approach and had some interest for some users.

PR #617, opened 2017-07-05
History & get_next_task refactor

ilmesi:
It would be nice if locust could handle dependent tasks through get_next_task. To allow this feature in a extensible way, the following PR contains three complementary changes:

- `TaskSet._task_history` persist each of the tasks that goes through `execute_next_task`
- `task_callable` is optional for schedule_task
- `get_next_task` receives an extra argument called last_task

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -225,6 +225,7 @@ class TaskSet(object):
 
     def __init__(self, parent):
         self._task_queue = []
+        self._task_history = []
         self._time_start = time()
         
         if isinstance(parent, TaskSet):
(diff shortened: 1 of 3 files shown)


heyman:
Hi!

From what I can tell (I've only skimmed the code though), there's a possible memory issue with the task history list growing huge for long running tests. 

Could you give some examples of when this would be needed? Could this be achieved by subclassing `TaskSet` in the locust test files instead?

cyberw:
Hi! This PR has been untouched for a very long time so I am closing it. Let me know if this was a mistake and feel free to open it again.

ilmesi:
@cyberw Hi! It has been a few years since I don't use Locust, so that's why I didn't follow up this PR (sorry). FYI: I ended up using this code in production and I didn't have any issues.

I think that you should keep it closed if no one needs this feature or you think it cannot be improved. As @heyman noted there's a possible memory issue if the task history list is not controlled (somehow).

The kind of tests I needed to perform were quite complex and require to do different flows depending the scenario (different responses and states). Defining a new TaskSet subclass was not an good option since I have to remove a large amount of code from run() and schedule_task()

Issue #34, opened 2012-09-19
Support for plugins


heyman:
I'd like locust to have a plugin system that would provide the ability to extend locust with different functionality. We could then probably refactor out some of the core functionality into plugins that are loaded by default and resides in a contrib package.

I think we could get a long way if we would implement a plugin system with the following features:
- A way for plugins to add custom tabs to the web UI
- An event that's triggered when a web UI client retrieves new data through AJAX, where a plugin can add data
- A client side event when new data has been retrieved, where plugin code can process/present any data that was added in the server side event
- A (documented) way of adding endpoints/routes to the web UI from within a plugin
- A convention on how to structure plugins
- A way to specify what plugins to load in the locust test file(s).
- Go through the current internal events and make sure they provide as much data as possible
- Provide a good way of writing unit tests for plugins
- Good documentation on how to write plugins

I believe the above features would cover a majority of the different plugin use cases one might have. I also think it's good to start small and keeping the plugin API as small as possible, since there's less risk of making bad API decisions and they will be easier to correct :).

adrianlewis:
I am also interested in locust pushing data to graphite.

sanga:
@heyman I'd forgotten about this completely but now I recall the reason why I didn't just start the graphite logging greenlet in our scenario. Because you only want that running on the master not the slaves. I could have changed only the master scenario to include the logger greenlet but because of the way we distribute the scenarios it was more convenient to just write a 5 line patch for locust.

venth:
Any progress on plugin concept? I'm interested in percentiles

arthurlogilab:
Would be interested in a statsd plugin

gerryhernandez:
I'm very interested in a plugin architecture. I love the underpinnings of Locust, but I'd like to add some _opinionated_ functionality that some people wouldn't agree with. I don't want to implement these features and force them upon everyone. :)

Say I were to implement the plugin functionality; do any of the primary contributors have any particular design considerations?

The first thing I would like to implement is a way to implement additional HTTP routes/handlers into the main/master process. This would open many doors.

Thanks!

FrancisJen:
I also hope locust can open a world for plugin, It is possible to be greater for testing engineer who only know python, because they can use python to do automation testing on selenium, but for performance testing. Jmeter is very popular because its support for plugin, if locust can support plugin, It will be more popular.

mbeacom:
Maybe we could start a GH project for this repo and scope a plugin system?
[...]

PR #622, opened 2017-07-17
added a prometheus metrics endpoint

RyanHartje:
I've added a /metrics route that allows prometheus to scrape locust so that you can cross compare the effects of load testing with other metrics gathered by prometheus.

The use case would be for people who want to see how their benchmarks affect systems performance.

### DIFF
--- a/locust/test/test_web.py
+++ b/locust/test/test_web.py
@@ -37,6 +37,9 @@ class TestWebUI(LocustTestCase):
     
     def test_index(self):
         self.assertEqual(200, requests.get("http://127.0.0.1:%i/" % self.web_port).status_code)
+
+    def test_metrics(self):
+        self.assertEqual(200, requests.get("http://127.0.0.1:%i/metrics" % self.web_port).status_code)
     
     def test_stats_no_data(self):
         self.assertEqual(200, requests.get("http://127.0.0.1:%i/stats/requests" % self.web_port).status_code)
--- a/locust/web.py
+++ b/locust/web.py
@@ -10,7 +10,7 @@ from six.moves import StringIO, xrange
 import six
 
 from gevent import wsgi
-from flask import Flask, make_response, request, render_template
+from flask import Flask, jsonify, make_response, request, render_template
 
 from . import runners
 from .cache import memoize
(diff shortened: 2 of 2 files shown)


lambertjosh:
This would be really helpful for us, too.

kmhagan:
Total response time and total content length would also be helpful metrics to log into prometheus. This way we can calculate the averages from the last x minutes and not just the average from the total load test.

heyman:
At the moment, I think I'm leaning towards not adding this feature into locust itself, since it's quite easy to add it within your test files:

```python
from locust import web

@web.app.route("/metrics")
def prometheus_metrics():
    # do the stuff
    # ...
```

aldenpeterson-wf:
I wonder if adding some documentation around this would be better, I'm inclined to agree with @heyman here because there are a ton of different stats utilities which could be included.

heyman:
@aldenpeterson-wf We do have [this](https://docs.locust.io/en/latest/extending-locust.html#adding-web-routes), but maybe it can be made more detailed/prominent.

kmhagan:
I was able to add the functionality by creating a separate file and importing it into my tests. That way I only need to do a one line change to the tests.

```
from locust import HttpLocust, TaskSet
import locust_prometheus
# ...
```
It would be nice if the prometheus functionality could be added as an official plugin, which I assume would require #34.

aidylewis:
Should we be using a Prometheus exporter to scrape the csv files (<server>:8089/stats/requests/csv & <server>:8089/stats/distribution/csv)?

aidylewis:
Having a bit of a hard time consuming the Locust csv metrics with haproxy_exporter. Are Locust metrics exposed in any other format? Or is was thinking of using Prometheus to scrape the web metrics without the use of Locust stats.

aidylewis:
I am currently using a mongdb_exporter with the Locust Java load injector. I am using Prometheus as a  Grafana datasource. It would be nice to also include the Locust metrics.

aidylewis:
Maybe we could use statsd and the statsd_exporter for Prometheus to pull Locust metrics?

myzhan:
If you are looking for an example of prometheus_exporter, try this

https://github.com/myzhan/boomer/blob/master/prometheus_exporter.py

As @heyman put it, we can add it within test files, and keep locust simple.

OleksiiHryhorian:
Hi guys, 
Do I correctly understand that provided solution for prometheus_exporter from locust is working only within WEB_UI running mode of locust?
What if we run locust in a headless mode and it needs to configure prometheus exporter for further setting up locust-prometheus-grafana without any UI.
Is there any easy way?

Issue #649, opened 2017-09-08
Libev over libevent and gevent suggested wsgi?


mbeacom:
### Description of feature request

Gevent suggests using `libev` versus the older `libevent` libraries and using the `gevent.pywsgi.WSGIServer` over the legacy `gevent.wsgi.WSGIServer`

PR #650, opened 2017-09-08
Modify gevent wsgi and libev dependencies

mbeacom:
The goal of this PR is to migrate from `libevent` to `libev` and from gevent's `wsgi` module to `pywsgi` .

- Addresses #649


aldenpeterson-wf:
+1, this worked for me when installing locally

Issue #512, opened 2017-01-09
num-requests bug


gathanase:
locust behaves strangely sometimes when using "num-requests". For some reason the test does not stop after num-request is reached, but it starts again from the beginning (with fewer clients).
My guess is that all clients are not created when the num-request is reached.

```
"locust --no-web --hatch-rate=10 --clients=80 --num-requests=200 --host=myhost MyUsecase"


 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 Total                                                              0     0(0.00%)                                       0.00

 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                       14     0(0.00%)      52      20      98  |      58    0.00
 GET /status                                                       15     0(0.00%)      51       3     109  |      49    0.00
 Total                                                             29     0(0.00%)                                       0.00

 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                       48     0(0.00%)      52      20     158  |      49    6.00
 GET /status                                                       50     0(0.00%)      27       3     109  |       5    7.50
 Total                                                             98     0(0.00%)                                      13.50

 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                       86     0(0.00%)      52      20     158  |      37    9.00
 GET /status                                                      114     0(0.00%)      20       2     109  |       5    9.67
 Total                                                            200     0(0.00%)                                      18.67
=> it should end there
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                       86     0(0.00%)      52      20     158  |      37    9.00
 GET /status                                                      114     0(0.00%)      20       2     109  |       5    9.67
 Total                                                            200     0(0.00%)                                      18.67
=> or it should end there
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                        1     0(0.00%)     100     100     100  |     100    0.00
 GET /status                                                        1     0(0.00%)      10      10      10  |      10    0.00
 Total                                                              2     0(0.00%)                                       0.00
=> it restarts...
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
 GET /health                                                        2     0(0.00%)      65      30     100  |      30    1.00
 GET /status                                                        2     0(0.00%)      10      10      11  |      10    0.00
 Total                                                              4     0(0.00%)                                       1.00
```

[...]

PR #656, opened 2017-09-18
Remove --num-requests/-n in favor of --run-time/-t

heyman:
This adds the `--run-time`/`-t` option that can be used, together with `--no-web`, to specify the run time of a test. It also removes the `-n`/`--num-requests` option as discussed in #196 and #512 (which can both be closed once this is merged).


locust/util/time.py
```diff
@@ -0,0 +1,25 @@
...
+        raise ValueError("Invalid time span format")
+    
+    if re.match(r'^\d+$', time_str):
+        # if an int is specified we assume they want seconds
+        return int(time_str)
+    
+    timespan_regex = re.compile(r'((?P<hours>\d+?)h)?((?P<minutes>\d+?)m)?((?P<seconds>\d+?)s)?')
+    parts = timespan_regex.match(time_str)
+    if not parts:
+        raise ValueError("Invalid time span format")
```
aldenpeterson-wf: small nit: it'd be nice if this message showed the valid or expected formats from the docstring

aldenpeterson-wf:
This should probably be added to the docs too, but this looks great and something we will definitely want to use 👍

locust/util/time.py
heyman: Fixed!

heyman:
Thanks! I've now added info to the documentation (very good point :)).

I actually ended up restructuring the docs a bit, by moving out info on how to run Locust without the web UI, as well as how to retrieve CSV stats, into their own pages. It didn't really belong on the Quickstart page, and now it'll be easier to find directly from the docs index page.

I'll go ahead and merge this.

locust/main.py
```diff
@@ -423,6 +423,22 @@ def main():
         }
         console_logger.info(dumps(task_data))
         sys.exit(0)
+    
+    if options.run_time:
+        if not options.no_web:
+            logger.error("The --run-time argument can only be used together with --no-web")
```
waberc: Why does it forbid this option if we are working from the web console ? 

In my use-case I would find it super useful to be able to **start the test from the web**, and have it stopping automatically even when I am disconnected / in a meeting **after some time** or when I hit the **stop button**. This would avoid me forgetting to stop the test and act as security to avoid a test running for ever...

Am I the only one who would like this feature not restricted to the ``--no-web`` ? Should I create an issue / feature request ?

diegotmorais: +1

rk4n3:
Why was the -n/--num-requests removed ?  Both mechanisms have distinct value/usage-scenarios, and I am one that really needs the -n/--num-requests capability.

In short, a particular test (with plenty of sophistication in its design/operation) is needed for both typical load testing scenario, AND a separate scenario where a single execution must execute and exit as quickly as possible.

Can we restore the lost functionality ?

PraveenWhatfix:
Yes, this would be a good feature to have. If we want to test performance for a particular number of requests rather than run-time.

locust/main.py
eduardogmisiuk: We would want this feature to be used with the web interface too, would be very useful.

heyman:
> Why was the -n/--num-requests removed ? Both mechanisms have distinct value/usage-scenarios, and I am one that really needs the -n/--num-requests capability.

One problem with having the `--num-requests` option is how it's supposed to work when running Locust distributed. Either we would make the master runner send out a stop message to all the worker nodes when the number of requests have been reached. However this method would result in a situation where more requests than the specified would be fired. This could be confusing for someone who assumes that the specified `--num-requests` would be respected. Alternatively the master node could - at start time - direct each node to only perform a certain number of requests, but that could cause some workers to stop generating load earlier than others, resulting in a RPS drop at the end of the test.

albertleadata:
> > Why was the -n/--num-requests removed ? Both mechanisms have distinct value/usage-scenarios, and I am one that really needs the -n/--num-requests capability.
> 
> One problem with having the `--num-requests` option is how it's supposed to work when running Locust distributed.

It would be perfectly reasonable (and to me acceptable) if this were addressed by making the -n/--num-requests option supported for non-distributed operation only.  It could even be asserted that it only makes sense in smaller-scale testing scenarios, which would naturally fall outside distributed operation anyway.

In my case, the need for it is specific to very small-scale tests that have to be executed at high iteration - the tests need to execute and completely exit as fast as possible, with only a specific and small number (often only 1) requests made.

heyman:
> It would be perfectly reasonable (and to me acceptable) if this were addressed by making the -n/--num-requests option supported for non-distributed operation only. It could even be asserted that it only makes sense in smaller-scale testing scenarios, which would naturally fall outside distributed operation anyway.
> 
> In my case, the need for it is specific to very small-scale tests that have to be executed at high iteration - the tests need to execute and completely exit as fast as possible, with only a specific and small number (often only 1) requests made.

I think everything that works non-distributed should work when running Locust distributed as well. We could possibly accept slightly different behaviour (e.g. the exact number of requests aren't guaranteed), but I'm not keen on having totally different feature sets.

Could you describe your use-case in more detail? To me, it doesn't really sound like it's a load test that you're doing.

albertleadata:
> Could you describe your use-case in more detail? To me, it doesn't really sound like it's a load test that you're doing.

Its not a load test - its what would be called a "synthetic health-check" ... its a test of arbitrary complexity that can benefit from (even require) all the functionality present in a load-testing tool like JMeter or Locust, but run to completion as quickly as possible with one iteration and 1 user.

vignesh-18:
Is there any way to stop the execution after specific number of requests are done?

TimorMuzi:
> Is there any way to stop the execution after specific number of requests are done?

Can you exit by throwing an StopLocus exception?
Or,You can try to judge in the code, when the number is reached, execute runners.locust_ runner.quit ()

Issue #59, opened 2013-03-12
Setup/teardown hooks


klausbrunner:
It'd be extremely useful to have dedicated setup and teardown functionality in Locust (or if there is something like this already, to have it documented). 

My rough idea would be:
- Setup is called exactly once, and must complete, before the test locusts are spawned. Teardown is called once after all of them have stopped.
- Setup should be fail-fast (if anything in setup fails, abort the test).
- Setup/teardown is realised using a Locust class like the others, so that state can be kept from setup to teardown if need be. Tasks are marked with setup and teardown annotations as appropriate.
- Setup/teardown is not included in any statistics.

Thoughts? (Have I missed something that already exists?)

shawngustaw:
This issue is pretty old and I'm looking for something along the lines of what's been discussed. Has there been any progress?

samjiks:
any progress on this, would be great.

mohanraj-r:
The [on_start()](http://docs.locust.io/en/latest/writing-a-locustfile.html#the-on-start-function) can be used as a setup I guess? And not sure if the (events.quitting)[http://docs.locust.io/en/latest/api.html#locust.events.quitting] can be used to create a hook that acts as a teardown? Would be nice to have a `on_stop()` that can be defined similarly to `on_start()`.

swordmaster2k:
+1 for an on_stop() feature. I have some custom websockets started on their on greenlet and having an on_stop handler would enable me to tear them down gracefully.

rmandar16:
+1 for on_stop() feature...Have some common tear-down tasks, to be executed.

cainj13:
Another +1 for on_stop!  would be immensely helpful

JDiPierro:
In place of an `on_stop` I'm overriding the `run` function on `Locust`, catching `GreenletExit`, executing my stop logic, and then re-raising. Seems to work with no downsides.

```
class MyLocust(Locust):
    def run(self, *args, **kwargs):
        try:
            super(MyLocust, self).run(*args, **kwargs)
        except GreenletExit:
            self.client.disconnect()
            raise
```

Jim-Lambert-Bose:
Another +1 for on_stop()

jdabello:
+1000

ad34:
+1

ad34:
awesome :) I am curently testing a websocket based title and it will help a lot because stopping the test dont close the websockets

PR #658, opened 2017-09-20
Adds support for setup, teardown, and on_stop methods

DeepHorizons:
If a TaskSet has an `on_stop` method, it will be run before the process exits. It will also look for a property on the class named `always_run_on_stop`, which if True will run regardless if the `on_start` method finished.

Looking to solve #59 

Any comments on this functionality?

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -12,6 +12,9 @@ Below is a quick little example of a simple **locustfile.py**::
     def login(l):
         l.client.post("/login", {"username":"ellen_key", "password":"education"})
     
+    def logout(l):
+        l.client.post("/logout", {"username":"ellen_key", "password":"education"})
+    
     def index(l):
         l.client.get("/")
     
@@ -23,6 +26,9 @@ Below is a quick little example of a simple **locustfile.py**::
         
         def on_start(self):
             login(self)
+        
+        def on_stop(self):
+            logout(self)
     
     class WebsiteUser(HttpLocust):
         task_set = UserBehavior
@@ -52,8 +58,15 @@ Another way we could declare tasks, which is usually more convenient, is to use 
             """ on_start is called when a Locust start before any task is scheduled """
             self.login()
         
+        def on_stop(self):
+            """ on_stop is called when the TaskSet is stopping """
+            self.logout()
+        
         def login(self):
             self.client.post("/login", {"username":"ellen_key", "password":"education"})
+        
+        def logout(self):
+            self.client.post("/logout", {"username":"ellen_key", "password":"education"})
         
         @task(2)
         def index(self):
--- a/docs/writing-a-locustfile.rst
+++ b/docs/writing-a-locustfile.rst
@@ -249,19 +249,19 @@ and :py:class:`TaskSet <locust.core.TaskSet>` :py:meth:`on_start <locust.core.Ta
 Setups and Teardowns
 --------------------
 
-:py:meth:`setup <locust.core.Locust.setup>` and :py:meth:`teardown <locust.core.Locust.teardown>`, whether it's run on :py:class:`Locust <locust.core.Locust>` or :py:class:`TaskSet <locust.core.TaskSet>`, are functions that are run only once.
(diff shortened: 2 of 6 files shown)


heyman:
Hi! I've looked through the changes and here are my initial thoughts:

The `on_start` method of a TaskSet is run every time a Locust "enters" that TaskSet which might be many times during a test run (since it's common to have nested TaskSets). I do think it could be a good idea to add the ability to have an `on_stop` method on TaskSets, but then I think it should run within the Locust greenlet whenever a TaskSet is interrupted (so that on_stop is symmetrically opposite to on_start).

If we want to provide some hook that is executed for each Locust user when a test is stopped, I think it should be a method under the Locust class, and I also think it should be executed within the Locust greenlet (by catching GreenletExit, and we need to make sure that the greenlets are actually killed properly when running locust locally with `--no-web` and pressing CTRL-C).

DeepHorizons:
That makes way more sense than what I was thinking/doing. `on_stop` should be a one to one with `on_start`, while the setup and teardown should be just once.

I re-worked the Locust run method to catch that exception, then run `on_stop` if it exists.  I also added the `locust_runner.quit()` during shutdown which should clean up those greenlets. I'm not sure if you use them anywhere else though.

Here is my idea for the setup and teardown. The Locust class and the TaskSet class can have a `setup` and `teardown` method. They will be class methods, not tied to any particular instance. There will also be some class variable to hold the state to see if they have been run already. Order of running would be `setup` on Locust, `setup` on TaskSet, `teardown` on TaskSet, `teardown` on Locust. The `__init__` methods will be modified to check to see if the `setup` method have been run. If not, run them and update the state. As for the teardown, I'll probably do something similar to what I was doing adding a handler to the `quitting` event.

j616:
I just thought I'd feed back that I've been using this branch for a few weeks for the `on_stop` functionality in task sets. It works exactly as I'd expect it to and feels almost essential for allowing my locustfiles to clean up after themselves. Hoping this get's accepted into the main upstream. It's too useful to just sit in someone's fork. Thanks for putting in the dev work, @DeepHorizons

DeepHorizons:
@j616 Thanks for letting me know! Let me add some documentation, then I'll remove the [WIP]

swordmaster2k:
I'm also actively using the functionality provided by this branch, the on_stop functions exactly how I'd expect it to. I have a websocket associated with each locust running on their on greenlets so I needed some way to close the websockets when the controlling locust was stopped. This branch provides that ability, would be nice to see it merged into the official branch.

aldenpeterson-wf:
this LGTM.

@heyman @cgbystrom @mbeacom any of you have thoughts on this? It'd be nice to get this into the 0.9.0 release.

mbeacom:
@aldenpeterson-wf Unless there are any objections, lgtm.  We'll wait a little and if no one speaks up, maybe merge it this week?

cgoldberg:
sure.. might as well be consistent

cgoldberg:
@DeepHorizons thanks for the contribution!

aldenpeterson-wf:
@heyman you want to cut another release? This is pretty great functionality and combined with the time based run worthy of releasing imo.

cgoldberg:
+1 for release.  We also need to add a nice description of this to `/docs/changelog.rst`.

migueleliasweb:
**Waiting anxiously**

migueleliasweb:
@cgoldberg How's things ? Do you think that it will take long before releasing this ?

cgoldberg:
@migueleliasweb 
we are going to cut a release candidate (hopefully soon) ... but not sure the timeline yet for the official 0.9 release.

heyman:
Hey! I realize that this comment is way late, but as far as I can tell from the code, the setup/teardown will run multiple times when running Locust distributed. I think we either need to make sure this is not the case (perhaps by only running the setup & teardown in the master node), or at least update the documentation which currently states that the setup/teardown is run only once. Though I think it's much preferable if Locust works the same wether it's being run distributed or not.

stefanbie:
As far as I can tell Locust teardown is not working in 0.9.0.

stefanbie:
Sorry, I was wrong, it actually works, great work guys!!! :)

AbigailMcP:
@stefanbie - I'm having trouble getting Locust teardown to work in 0.9.0 (it doesn't appear to run at all?) could you share your solution?

PR #660, opened 2017-09-28
clarify that the CSV is only written to once

adamknox-wf:
The CSV is only written to at the end of the run, so it is a bit misleading to say the CSV is written to periodically.


aldenpeterson-wf:
It does write the CSV based on [this code](https://github.com/locustio/locust/blob/master/locust/stats.py#L657-L661) triggered from [here](https://github.com/locustio/locust/blob/master/locust/main.py#L483-L484).

PR #665, opened 2017-10-16
Exit with code 1 on greenlet exceptions

jnschulze:
Locust exits with code 0 even if an exception in the greenlet occurs. This tiny fix makes it exit with 1 accordingly.


aldenpeterson-wf:
This lgtm, not sure if there are weird edge cases this might falsely flag - @heyman any ideas on that? I assume any gevent exception should be an actual "failure" here too, but not 100% sure.

cyberw:
@jnschulze Could you resolve the conflicts and (optionally) add a test case for this?

heyman:
I was too hasty when merging this. `main_greenlet` is a `gevent.pool.Group` instance when running Locust as a master or slave node, and `Group` doesn't have the exception attribute. I've reverted the merge since the current change causes an exception on shutdown of slave and master nodes.

PR #669, opened 2017-10-25
Switch to pytest

sanga:
Pytest has a few nice features (if you've never used it before). Notably:

- bare assertions that fail in a nice way (see: https://docs.pytest.org/en/latest/assert.html)
- and I think the the test fixture handling is a lot nicer too


thomas-phillips-nz:
@sanga There are conflicts to resolve now :disappointed:

sanga:
I could rebase this if it's wanted. Do the maintainers have any opinion? @cgoldberg ?

cgoldberg:
I like pytest, but I'm -1 on changing the test framework right now.  If we had more active development I'd be more inclined.

PR #674, opened 2017-10-31
Do not reset statistics on hatch complete 

### LINKED ISSUE #672: --no-reset-stats should be on by default
Please make this on by default.  It is what most people will want in a realistic load test.  

Case in point. We are gradually hatching about 1000 locusts that all need to authenticate with a login server. We hatch them slowly to avoid overwhelming the login server. After each locust logs in,  it immediately starts performing our regular workflow. In our first test runs, we lost the stats that got collected during the hatch because we didn't know `--no-reset-stats` existed. Watching your stats reset to 0 is extremely baffling. We had to scour the interweb to learn about this setting, and then spent a few hours figuring out how the switch works in a master/slave scenario

Making this on by default begs the question of how someone would turn it back off if they really do want to reset their stats when the hatch completes. Some options:

1. Ditch `--no-reset-stats` and replace it with a switch that has the opposite polarity : `--reset-stats-on-hatch-complete`
2. Just let me hook `on_hatch_complete` and reset the stats on my own if that is the behavior I want. Assumes that I can do this in my locust class or task set.

hhowe29:
Do not reset statistics on hatch complete unless the user specifically chooses to. Added a `--reset-stats` option to force a reset, which inverts the meaning of the older `--no-reset-stats` option. Deprecated the `--no-reset-stats` option, but continue to allow it for backwards compatibility.

Closes #672

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -216,9 +216,15 @@ def parse_options():
     parser.add_option(
         '--no-reset-stats',
         action='store_true',
-        dest='no_reset_stats',
+        help="[DEPRECATED] Do not reset statistics once hatching has been completed. This is now the default behavior. See --reset-stats to disable",
+    )
+
+    parser.add_option(
+        '--reset-stats',
+        action='store_true',
+        dest='reset_stats',
         default=False,
-        help="Do not reset statistics once hatching has been completed",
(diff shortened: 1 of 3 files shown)


aldenpeterson-wf:
@heyman this lgtm

heyman:
Great job!

PR #676, opened 2017-11-06
Small python 3 syntax improvement in docs

miki725:
While browsing docs noticed in one doc small Python3 incompatibility. Switched to using `print()` as function call vs as statement


aldenpeterson-wf:
+1

Issue #345, opened 2015-10-12
Support for distributing arbitrary arguments to Locust [Proposal]


mlromramse:
Since it now is possible to write your own Locusts, which we have started to do it would be appreciated if the command line arguments not directed to locust.main or HttpLocust could be passed along into the Locust derivate of choice.

If you for instance create a Locust that doesn't use Http the host argument is not needed. Still this is the only argument you can pass to your Locust class.

My suggestion is that `*args` and `**kwargs` is taken from the command line arguments and added to the `Locust.__init__`.

In this way you can build your own locusts and control them from the command line alone.

cyberw:
Closing due to lack of activity. Let me know if this is still an issue.

PR #686, opened 2017-11-14
Allow passing additional args to locusts from the command line

bugsduggan:
ref: #345

This now allows you to specify arbitrary args to pass to locust instances e.g.
`locust -f locustfile.py --locust-arg foo --locust-arg bar`

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -99,7 +99,7 @@ class Locust(object):
     client = NoClientWarningRaiser()
     _catch_exceptions = True
     
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         super(Locust, self).__init__()
     
     def run(self):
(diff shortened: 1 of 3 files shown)


thinksource:
I just need that one for my configure files of my script

thomas-phillips-nz:
This looks good. I usually use environment variables, to control my locust instances, but this looks a bit more explicit.

aldenpeterson-wf:
This might benefit from some doc updates showing how to pass both a regular arg as well as kwarg.

heyman:
I'm not against the idea of being able to pass in arguments to Locust classes. However I'm not a fan of using positional arguments and I think we would need to come up with another API. We would also have to consider if the arguments should be passed along automatically to the worker nodes when running Locust distributed, add tests and documentations, etc.

cyberw:
Env vars solve this issue in a more "standard" way. I vote to close. Possibly we could add some documentation on how to use env vars.

cyberw:
Closing this for now. A different PR along the lines suggested by @heyman might get accepted, but I think env vars should be enough for most ppl...

PR #692, opened 2017-11-21
Make UI URL links relative

### LINKED ISSUE #149: URL_PREFIX feature for web UI?
Hi,

Is it possible to add a --url-prefix option for the web ui, so that proxy setups can run the web ui under a sub-url? We want to run locust proxied behind our https server, which proxies other services like @Jahaja 's ps-dash, and to do this requires any assets being loaded to be fetched either relative to "./" or full-pathed under a known prefix.

Thanks,
Jeff

karol-brejna-i:
Fixes #149

When using proxy or exposing Locust UI in Kubernetes (via ingress) the URLs get prefixed (for example, served under http://domain.com/locust instead of http://domain.com).

On the other hand, links in html template and JS (rest calls) are absolute (starting with '/') which makes the UI inoperable.

This PR makes the links relative - they work fine both with and without a prefix.

To validate:
```
git clone https://github.com/karol-brejna-i/locust.git
git checkout origin/149_url-prefix-for-web-ui
pip3.6 uninstall locustio -y 
pip3.6 install ./locust/ 
locust -f locust/examples/basic.py --host=http://localhost:8089
```

When looking at HTML page sources you should see `<link rel="stylesheet" type="text/css" href="./static/style.css?v=0.8" media="screen">` somewhere at the beginning of the doc.

Verify if the UI works: starting, stopping the tests, browsing the stats produce valid rest calls.
(Be aware of browser caching; sometimes "old" HTML and JS are served.)

### DIFF
--- a/locust/static/locust.js
+++ b/locust/static/locust.js
@@ -123,7 +123,7 @@ var responseTimeChart = new LocustLineChart($(".charts-container"), "Response Ti
 var usersChart = new LocustLineChart($(".charts-container"), "Number of Users", ["Users"], "users");
 
 function updateStats() {
-    $.get('/stats/requests', function (report) {
+    $.get('./stats/requests', function (report) {
         $("#total_rps").html(Math.round(report.total_rps*100)/100);
         //$("#fail_ratio").html(Math.round(report.fail_ratio*10000)/100);
         $("#fail_ratio").html(Math.round(report.fail_ratio*100));
@@ -164,7 +164,7 @@ function updateStats() {
 updateStats();
 
 function updateExceptions() {
-    $.get('/exceptions', function (data) {
+    $.get('./exceptions', function (data) {
         $('#exceptions tbody').empty();
         $('#exceptions tbody').jqoteapp(exceptions_tpl, data.exceptions);
         setTimeout(updateExceptions, 5000);
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -2,13 +2,13 @@
 <html>
 <head>
     <title>Locust</title>
-    <link rel="stylesheet" type="text/css" href="/static/style.css?v={{ version }}" media="screen">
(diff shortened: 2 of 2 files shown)


karol-brejna-i:
Rebased against latest master.

kasturichavan:
@cgoldberg Im facing this issue even now. 
I am on locustio - 0.11.0 
 But im seeing the same issue when running locust in kubernetes and when the path is is changed in kubernetes ingress from `/` to `/locust`. Everything should work right ? Considering this bug is fixed.

w32-blaster:
> @cgoldberg Im facing this issue even now.
> I am on locustio - 0.11.0
> But im seeing the same issue when running locust in kubernetes and when the path is is changed in kubernetes ingress from `/` to `/locust`. Everything should work right ? Considering this bug is fixed.

Did you find any solution for this? We would like to expose locust on different paths in k8s using the nginx ingress rewrite annotation feature. Looks like static files are still served from wrong path.

dchourasia:
@karol-brejna-i  does this PR make url-prefix configurable for the web-ui... did not see any such parameter in locust documentation.. not sure if I am missing something. if you can help please.

awalker125:
If anyone comes here looking for the annotations to make this work with k8s ingress-nginx then here is what I used:
```yaml
- apiVersion: networking.k8s.io/v1
  kind: Ingress
  metadata:
    annotations:
      nginx.ingress.kubernetes.io/rewrite-target: /$2
      nginx.ingress.kubernetes.io/use-regex: "true"
    name: locust
    namespace: locust
  spec:
    ingressClassName: nginx
    rules:
    - http:
        paths:
        - backend:
            service:
              name: locust-master
              port:
                number: 8089
          path: /locust(/|$)(.*)
          pathType: Prefix

```

This is based on https://kubernetes.github.io/ingress-nginx/examples/rewrite/

PR #694, opened 2017-11-21
Allow the option to run a single Locust through the UI

jeffrand:
I would like to be able to choose which Locust to run through the UI. This would make it very easy to support having a large corpus of tests and have the option to only run a single one when it suits the user. 

The core of this feature relies on the Locust class's `__name__` property. The `LocustRunner` holds a dict of `class.__name__` to `class` in `locust_classes_by_name` and provides a method to building that dict and a way of picking one with `filter_locusts_by_name`.

Right now this only supports running a single Locust or all of them. The web UI has a select that defaults to "All":

image

And has a list of Locusts to run:

image

The URL `/swarm` accepts `locust_class` in the POST body but defaults it to `all`. `all` in this case will be converted to `None` when passed to the `LocustRunner`.

Let me know what you think, I'd be happy to make any changes you recommend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ dist/**
 build/
 .coverage
 .tox/
+/.cache/
+.DS_Store
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import collections
 import logging
 import random
 import socket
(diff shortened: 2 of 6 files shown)


jeffrand:
Closing in favor of https://github.com/locustio/locust/pull/695 (a fork from my github rather than my employers)

PR #695, opened 2017-11-22
Allow the option to run a single Locust through the UI

jeffrand:
I would like to be able to choose which Locust to run through the UI. This would make it very easy to support having a large corpus of tests and have the option to only run a single one when it suits the user. 

The core of this feature relies on the Locust class's `__name__` property. The `LocustRunner` holds a dict of `class.__name__` to `class` in `locust_classes_by_name` and provides a method to building that dict and a way of picking one with `filter_locusts_by_name`.

Right now this only supports running a single Locust or all of them. The web UI has a select that defaults to "All":

image

And has a list of Locusts to run:

image

The URL `/swarm` accepts `locust_class` in the POST body but defaults it to `all`. `all` in this case will be converted to `None` when passed to the `LocustRunner`.

Let me know what you think, I'd be happy to make any changes you recommend.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import collections
 import logging
 import random
 import socket
@@ -31,6 +32,7 @@ class LocustRunner(object):
     def __init__(self, locust_classes, options):
         self.options = options
         self.locust_classes = locust_classes
+        self.locust_classes_by_name = self.build_locust_classes_by_name(self.locust_classes)
         self.hatch_rate = options.hatch_rate
         self.num_clients = options.num_clients
         self.host = options.host
@@ -49,6 +51,27 @@ class LocustRunner(object):
                 self.stats.reset_all()
         events.hatch_complete += on_hatch_complete
 
+    @staticmethod
+    def build_locust_classes_by_name(locust_classes):
+        """
+        Returns a dictionary of the form {locust_class_name: locust_class}
+        """
+        if isinstance(locust_classes, type):
+            return {locust_classes.__name__: locust_classes}
+        elif isinstance(locust_classes, collections.Iterable):
+            return {locust.__name__: locust for locust in locust_classes}
+        else:
+            return {}
+
+    def filter_locusts_by_name(self, locust_class=None):
+        """
+        By default, get all locusts. If locust_class is supplied, return a list of
+        only a locust with that class name. If it doesn't exist, return all locusts
+        """
+        if locust_class and locust_class in self.locust_classes_by_name:
+            return [self.locust_classes_by_name[locust_class]]
+        return self.locust_classes
+
     @property
     def request_stats(self):
         return self.stats.entries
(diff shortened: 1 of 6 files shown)


locust/runners.py
```diff
@@ -49,6 +51,27 @@ def on_hatch_complete(user_count):
...
+    def build_locust_classes_by_name(locust_classes):
+        """
+        Returns a dictionary of the form {locust_class_name: locust_class}
+        """
+        if isinstance(locust_classes, type):
+            return {locust_classes.__name__: locust_classes}
+        elif isinstance(locust_classes, collections.Iterable):
+            return {locust.__name__: locust for locust in locust_classes}
+        else:
+            return {}
```
aldenpeterson-wf: won't this case become an error at the UI level because there will be no locusts to select? I'm trying to imagine a case where it hits here, though, because iirc there's validation prior to this point.

locust/runners.py
```diff
@@ -49,6 +51,27 @@ def on_hatch_complete(user_count):
...
+
+    def filter_locusts_by_name(self, locust_class=None):
+        """
+        By default, get all locusts. If locust_class is supplied, return a list of
+        only a locust with that class name. If it doesn't exist, return all locusts
+        """
+        if locust_class and locust_class in self.locust_classes_by_name:
```
aldenpeterson-wf: would it be better to throw an exception if someone passes an invalid locust_class name rather than silently returning all?

locust/templates/index.html
```diff
@@ -57,6 +57,13 @@
             <div class="padder">
                 <h2>Start new Locust swarm</h2>
                 <form action="/swarm" method="POST" id="swarm_form">
+                    <label for="locust_class">Locust to run</label>
```
aldenpeterson-wf: I wonder if it would be nice for this to be a checkbox list or something that let you pick multiples but not all?

locust/test/test_runners.py
```diff
@@ -38,6 +38,66 @@ def send(self, message):
...
+        """ Assert that locust_classes_by_name functions with None """
+        runner = LocustRunner(None, self.options)
+        self.assertEqual({}, runner.locust_classes_by_name)
+
+    def test_filter_locusts_by_name(self):
+        """ Assert that filter_locusts_by_name functions correctly"""
+        runner = LocustRunner(self.locust_classes, self.options)
+        self.assertEqual(self.locust_classes, runner.filter_locusts_by_name(None))
+        self.assertEqual([self.locust_classes[0]], runner.filter_locusts_by_name(self.locust_classes[0].__name__))
+        self.assertEqual(self.locust_classes, runner.filter_locusts_by_name('foo'))
```
aldenpeterson-wf: this is the condition that seems like it should generate an exception to me, thoughts?

locust/runners.py
jeffrand: That's a fair point, I can raise an exception here, and catch the exception in the runner tests.

locust/runners.py
jeffrand: Yeah, I'll do that.

oesah:
Can I take this over?

cgoldberg:
go for it

oesah:
@cgoldberg Anything specific you want in there? I would like to give it a try with a checkboxed table, in which you choose your classes.

cgoldberg:
fixing the merge conflicts and the makjng changes to address the suggestions from comments would be a good start

cyberw:
@oesah Have you had time to take a look at this?

cyberw:
Closing due to inactivity.

cyberw:
@heyman Maybe I was too quick to close this? It kind of relates to #926

heyman:
@cyberw Yes, it's related, but I don't think an implementation with only the ability to select either a single locust class, or all classes, through the UI is good enough to merge. Even if this PR changes so that multiple locust classes can be selected, then comes the question if one should also be able to set the weight dynamically? There are a couple of open issues that are somewhat related to this, and I think that we should consider all of them when we implement a solution.

PR #693, opened 2017-11-21
graph error rates and make charts downloadable

craig-day:
First, thanks for the amazing tool! It was super simple to deploy and get running.

~I found myself wanting to see more detailed error information and I believe this is useful for other users as well. This just adds another chart that simply shows the total error count over time.~ I also wanted to be able to save the full chart data so I could reproduce the graphs later, so I added links to the downloads page for each carts exact data points. Hopefully others find this useful.

/cc @cgbystrom @heyman @mbeacom 

Screenshots of the ~new charts page and the~ download page. 

screen shot 2017-11-21 at 11 01 59 am

PR #702, opened 2017-12-09
chart download

yivanus:
add download toolbox to charts.
给图形添加下载的功能，下载为png图片

### DIFF
--- a/locust/static/chart.js
+++ b/locust/static/chart.js
@@ -31,6 +31,16 @@
                     x: 10,
                     y: 10,
                 },
+                toolbox:{
+                	show:true,
+                	feature:{
+                		saveAsImage:{
+                			type:'png',
+                			title:"download",
+                			pixelRatio:1
+                		}
+                	}
+                },
                 tooltip: {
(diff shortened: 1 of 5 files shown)


yivanus:
chartdown

cgoldberg:
why is code from #703 included here?

thomas-phillips-nz:
This looks to be a nicer solution than #693, if it's tidied up.

locust/static/chart.js
```diff
@@ -31,6 +31,16 @@
                     x: 10,
                     y: 10,
                 },
+                toolbox:{
+                	show:true,
+                	feature:{
+                		saveAsImage:{
```
craig-day: does this just give you an image of the graph? If so, it serves a different purpose than #693 which actually gives you all of the data points in a CSV file

yivanus: yes。only the graph。

cyberw:
Hi @yivanus ! Is this still relevant? If so then please resolve the conflicts and we can have a look.

PR #704, opened 2017-12-09
Added link to an Ansible role as a 3rd party tool.

tinx:
I did a thing, and I hope it's useful for people other than myself.

The role is able to install Locust, but can also start and restart Locust master, slave and stand-alone, while preserving the parameters. More than one Locust "instance" can exist on the same host in parallel.

The role can copy a directory with files alongside installing Locust, so it's also a convenient way to put your locustfile, classes and required data onto the host in a single step.

Finally, it can be used as an easy way to build docker images with Locust and the desired locustfiles already baked in.


aldenpeterson-wf:
This is probably good, I get questions often from folks wondering how to effectively deploy locust via docker.

cyberw:
Looks nice. Sorry for the super slow response @tinx . Will you resolve the conflicts? Then I can merge.

Issue #683, opened 2017-11-13
Provide a way to specify locust counts and hatch rate per locust class


hyperair:
### Description of issue / feature request

Currently, when running locust with multiple Locust classes, there isn't a way to specify how many instances of each locust class we want to hatch from the web interface, whether by UI widget or by API call. As a result, if we want to run tests with multiple combinations of `FooLocust` and `BarLocust` counts, we need to mess with the weightage of each locust class in the codebase of each locust slave.

### Expected behavior

Ideally, there would be a way to specify the exact counts of each locust type during runtime without requiring code changes, e.g.

```json
{
    "FooLocust": {"locust_count": 123, "hatch_rate": 12},
    "BarLocust": {"locust_count": 234, "hatch_rate": 34}
}
```

thomas-phillips-nz:
You could use environment variables to change the weighting of each class as a work-around for now, rather than changing the codebase.

cgoldberg:
send a pull request

hyperair:
@cgoldberg Yeah, working on it.

@thomas-phillips-nz That's what I'm doing now, but the slaves need to be restarted with different environment variables set, which makes it a bit clunky.

hyperair:
@cgoldberg PR submitted: ^

hyperair:
Oops, it's all red, hang on.

hyperair:
Yay, it's all green now

cyberw:
Hi! Is this something you still need? See my comment on the PR...

hyperair:
Hey @cyberw, thanks for getting back to me. I don't think I'll need it in the near future, but I'll keep my branch around in case I do. I still do think that the feature is desirable though.

cyberw:
Closing this for now. Again, sorry for the slow respone on your PR :)

PR #712, opened 2017-12-19
Per locust hatching

hyperair:
Fixes: #683 

This PR changes `LocustRunner` to use per-locust structures internally and extends the `/swarm` interface to accept an `application/json` of the following format, in addition to the old `locust_count=1234&hatch_rate=10` style for backward compatibility:

```json
{
    "FooLocust": { "locust_count": 1234, "hatch_rate": 10.0 },
    "BarLocust": { "locust_count": 123, "hatch_rate": 5.5 }
}
```

In `LocustRunner`, the following attributes have been renamed, with properties added for the old names to maintain backward compatibility with existing locustfiles that use these attributes:

 - `hatch_rate` (`float``) -> `hatch_rates` (`dict` of classes -> `float`)
 - `num_clients` (`int`) -> `num_clients_by_class` (`dict` of classes -> `int`)
 - `locust_classes` (`list`) -> `locust_classes_by_name` (`dict` of `str` -> locust class)

The following `LocustRunner` methods have been extended to accept `dict` of classes -> `float/int` for their locust count and hatch rate parameters. They'll still work with the old `int` or `float` parameters though.

 - `spawn_locusts`
 - `start_hatching`
 - `kill_locusts`

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -444,10 +444,25 @@ class MasterLocustRunner(DistributedLocustRunner):
                            "Please connect slaves prior to swarming.")
             return
 
-        self.num_clients = locust_count
-        slave_num_clients = locust_count // (num_slaves or 1)
-        slave_hatch_rate = float(hatch_rate) / (num_slaves or 1)
-        remaining = locust_count % num_slaves
+        locust_count = self._preprocess_locust_count(locust_count)
+        hatch_rate = self._preprocess_hatch_rate(hatch_rate)
+
+        if set(six.iterkeys(locust_count)) != set(six.iterkeys(hatch_rate)):
+            raise ValueError("locust_count and hatch_rate do not have matching"
+                             " locust classes")
+
+        self.num_clients_by_class.clear()
+        self.num_clients_by_class.update(locust_count)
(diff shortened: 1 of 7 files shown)


thomas-phillips-nz:
It might be easier to rebase and squash some of your commits together. It's a lot to review

hyperair:
@thomas-phillips-nz There, it's 9 commits now. Does this work for you?

cyberw:
Wow, this PR has been around for a long time. Do we want to merge it @heyman ? @hyperair , would you have time to resolve the conflicts if we would accept it?

heyman:
I'm  currently -0 on this one I think. I can see that there might be cases where it could be useful, but I think the current implementation of having a total number of users and Locust class weights should work for most cases. If we were to provide the ability to specify the absolute number of users for the individual locust classes, I still don't think there should be an option to set individual hatch rates. The use-case for that seems too narrow to me.

cyberw:
Closing this. We can revisit it later if someone really needs it, but I agree with @heyman that the use case is a little narrow and doesnt really justify the added complexity.

PR #713, opened 2017-12-19
Added FastHttpLocust with HTTP client that uses geventhttpclient

heyman:
I've implemented a custom Locust client that uses geventhttpclient (https://github.com/gwik/geventhttpclient/) instead of python-requests. The main advantage is better performance, to a cost of a worse developer API. I did some quick and dirty benchmarks on my local machine (see below), and they seem to indicate a 5x-6x performance increase. 

I've added the new `FastHttpLocust` under a new package `locust.contrib.fasthttp`. Even though I think it's a better to have python-requests be the default HTTP-client, I think it makes sense to include a more performant client as an alternative for people who are doing really large load tests.

This PR should be seen as a proposal & work in progress, and a place to discuss if people think this is a good idea.

Limitations of the current implementation:

* ~~Currently, the `catch_response` argument doesn't work with the FastHttpClient. This can probably be fixed.~~
* ~~Doesn't support basic auth~~
* The API of geventhttpclient isn't the same as Requests, so therefore the FastHttpLocust isn't a drop-in replacement of HttpLocust. I think this is fine, and we should have a separate documentation page about FastHttpLocust

# Benchmarks

Here are results from a quick and dirty benchmark ran on my laptop, on a single core, against nginx.

## HttpLocust

image

~240 reqs/s with the CPU almost maxed out.

## FastHttpLocust

image

~1400 reqs/s

## Code used

```python
from locust import task, TaskSet, HttpLocust
from locust.contrib.fasthttp import FastHttpLocust


class BenchLocust(FastHttpLocust): # or HttpLocust
    min_wait = 0
    max_wait = 0
    host = "http://127.0.0.1:8001"
    
    class task_set(TaskSet):
        @task
        def index(self):
            self.client.get("/")
```


locust/contrib/geventhttpclient.py
```diff
@@ -62,7 +75,7 @@ def request(self, method, path, name=None, **kwargs):
         
         try:
             response = self.client.urlopen(url, method=method, **kwargs)
-        except (ConnectionError, ConnectionRefusedError) as e:
+        except (ConnectionError, ConnectionRefusedError, socket.error) as e:
             # record the consumed time
             request_meta["response_time"] = int((time() - request_meta["start_time"]) * 1000)
```
cgoldberg: you can use `timeit.default_timer()` instead of `time.time()` to get the most accurate timer on the current platform (time.time() is not accurate on Windows, and better timers were introduced in Python 3)

heyman: Nice, fixed!

heyman:
I renamed the class to `FastHttpLocust` (from `GeventHttpLocust`) and updated the PR description. I think it's more descriptive. Having a class called GeventHttpLocust feels odd when Locust as a whole uses gevent, and the underlying implementation of the client doesn't really matter to users of Locust.

locust/contrib/geventhttpclient.py
```diff
@@ -61,23 +66,31 @@ class GeventHttpSession(object):
...
-        self.client = UserAgent(max_retries=1, cookiejar=self.cookiejar)
-        
+        self.client = LocustUserAgent(max_retries=1, cookiejar=self.cookiejar)
+    
+    def _build_url(self, path):
+        """ prepend url with hostname unless it's already an absolute URL """
+        if absolute_http_url_regexp.match(path):
+            return path
+        else:
+            return "%s%s" % (self.base_url, path)
```
cgoldberg: - do you need a regex here, or is `path.startswith('http')` sufficient?
- see `urllib.parse.urljoin()` (or `urlparse.urljoin()` in py2).  It handles edge cases (like extra slashes) better than just concatenating the pieces

locust/contrib/geventhttpclient.py
```diff
@@ -61,23 +66,31 @@ class GeventHttpSession(object):
...
+    
     def request(self, method, path, name=None, **kwargs):
-        url = self.base_url + path
+        # prepend url with hostname unless it's already an absolute URL
+        url = self._build_url(path)
```
cgoldberg: you can probably simplify or remove the `_build_url()` method and just use something like:

`url = path if path.startswith('http') else urljoin(base_url, path)`

locust/contrib/geventhttpclient.py
```diff
@@ -61,23 +66,31 @@ class GeventHttpSession(object):
...
```
cgoldberg: - you can probably simplify or remove the `_build_url` method and just use something like:

`url = path if path.startswith('http') else urljoin(base_url, path)`

heyman: Hmm, changing to just check `path.startswith('http')` and using `urljoin()` would create a situation where this would works:

`self.client.get("some-path") # <-- no slash in beginning`

while this would not work:

`self.client.get("http-protocol")`

Maybe just switch to `urljoin()` but stick with the regexp to check if a full URL was specified?

cgoldberg: you could try parsing the string with `urllib.parse.urlparse()` ... but regex works too

heyman: I opted to instead validate that the `host` doesn't contain a trailing slash, and just use string concatination, because `urljoin()` seems a bit overkill for this use-case since it parses the whole URL. Maybe it's premature optimization, but for this specific case I think it might not be.

```python
>>> from timeit import timeit
>>> timeit("url = urljoin(base, path)", setup='base="http://127.0.0.1"; path="/test"; from urllib.parse import urljoin;', number=100000)
1.4690891270292923
>>> timeit("url = base + path", setup='base="http://127.0.0.1"; path="/test";', number=100000)
0.012773031019605696
```

locust/test/test_fasthttp.py
```diff
@@ -0,0 +1,370 @@
...
+            host = "http://127.0.0.1:1"
+        l = MyLocust()
+        with l.client.get("/", catch_response=True) as r:
+            self.assertEqual(r.status_code, 0)
+            self.assertEqual(None, r.content)
+            r.success()
+        self.assertEqual(1, self.num_success)
+        self.assertEqual(0, self.num_failures)
+    
+    def test_catch_response_connection_error_fail(self):
```
cgoldberg: this test for `test_catch_response_connection_error_fail` is identical to `test_catch_response_connection_error_success`.  I think you forgot to make it fail.

docs/increase-performance.rst
```diff
@@ -0,0 +1,67 @@
...
+  support the :ref:`catch-response <catch-response>` argument.
+* Basic auth is currently not supported.
+
+
+How to use FastHttpLocust
+===========================
+
+First, you need to install the geventhttplocust python package::
+
+    pip install geventhttpclient
```
cgoldberg: Do you think we should skip this step and just add `geventhttpclient` as a required dependency in setup.py (rather than only installing it in `tests_require`)?  That way we could also pin a version number.

locust/contrib/fasthttp.py
```diff
@@ -0,0 +1,325 @@
...
+    
+    def raise_for_status(self):
+        """Raise any connection errors that occured during the request"""
+        if hasattr(self, 'error') and self.error:
+            raise self.error
+    
+    @property
+    def status_code(self):
+        """
+        We override status_code in order to return None if bo valid response was
```
cgoldberg: s/bo/no/

locust/test/test_fasthttp.py
heyman: Oh, good catch! I just copied the tests for the current HTTP client. I will update both tests!

docs/increase-performance.rst
heyman: Yeah, I think that's a good idea if we can either get the maintainers of geventhttpclient to upload python wheels (for linux, osx & windows), or worst-case create a separate PyPI repository where we upload wheels for it ourselves. But if we do the job of fixing an automated way of building wheels, I see no reason why they wouldn't accept it :).

Without wheels it will otherwise become much harder to install Locust (especially on windows), since it'll require that one has a C compiler and build environment set up.

locust/contrib/fasthttp.py
heyman: Fixed 👍

docs/increase-performance.rst
heyman: I've opened an issue at the geventhttpclient project asking about this: https://github.com/gwik/geventhttpclient/issues/93

heyman: And now a pull-request as well: https://github.com/gwik/geventhttpclient/pull/94

locust/contrib/fasthttp.py
```diff
@@ -66,10 +77,25 @@ def __init__(self):
...
+        if parsed_url.username and parsed_url.password:
+            netloc = parsed_url.hostname
+            if parsed_url.port:
+                netloc += ":%d" % parsed_url.port
+            
+            # remove username and password from the base_url
+            self.base_url = urlunparse((parsed_url.scheme, netloc, parsed_url.path, parsed_url.params, parsed_url.query, parsed_url.fragment))
+            # store authentication header (we construct this by using _basic_auth_str() function from requests.auth)
+            self.auth_header = _construct_basic_auth_str(parsed_url.username, parsed_url.password)
+            #self.auth = HTTPBasicAuth(parsed_url.username, parsed_url.password)
```
cgoldberg: comment leftover?

docs/increase-performance.rst
mbeacom: Per https://github.com/gwik/geventhttpclient/issues/93#issuecomment-353781687 - maybe we  should consider adding a `locustio/geventhttpclient` fork and build our own wheels from that? Or we could check the validity of the wheels built at:  https://pypi.python.org/pypi/geventhttpclient-wheels/ and use those?  The latest pypi push was on 12-23 and seems to reflect the latest `geventhttpclient` master.

docs/increase-performance.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Locust's default HTTP client uses `python-requests <http://www.python-requests.org/>`_. 
+The reason for this is that requests is a very well-maintained python package, that 
+provides a really nice API, that many python developers are familiar with. Therefore, 
+in many cases, we recommend that you use the default :py:class:`HttpLocust <locust.core.HttpLocust>` 
+which uses requests. However, if your're planning to run really large scale scale tests, 
+Locust comes with an alternative HTTP client, 
+:py:class:`FastHttpLocust <locust.contrib.fasthttp.FastHttpLocust>` which 
+uses `geventhttpclient <https://github.com/gwik/geventhttpclient/>`_ instead of requests.
+This client is significantly faster, and we've seen 5x-6x performance increases for
```
cgoldberg: I'm hesitant to make any specific performance claims since so many factors are in play.  I'd probably reword to something like:
"This client is significantly faster, and we've seen significant performance improvements for many workloads"

docs/increase-performance.rst
heyman: The geventhttpclient-wheels is my package :). I created it while I was putting together this PR to the geventhttpclient project: https://github.com/gwik/geventhttpclient/pull/94

Hopefully the owner will show up and merge it! Otherwise we'll have to keep a fork which can automatically upload wheels.

locust/contrib/fasthttp.py
heyman: Ah, yes, will fix!

docs/increase-performance.rst
heyman: I've made it clearer (I think) in: 42bfffb3d8eaef8205ccc4f7cdead9dd9a22c2ae

docs/changelog.rst
```diff
@@ -14,7 +14,7 @@ Changelog
 
 * Python 3 support
 * Dropped support for Python 2.6
-* Added `--no-reset-stats` option for controling if the statistics should be reset once 
+* Added :code:`--no-reset-stats` option for controling if the statistics should be reset once
```
aldenpeterson-wf: I think it might be worth trying to update this more often... I like this and it's a pain having to try to figure out everything that went into a change log.

docs/increase-performance.rst
```diff
@@ -0,0 +1,70 @@
...
+
+==============================================================
+Increase Locust's performance with a faster HTTP client
+==============================================================
+
+Locust's default HTTP client uses `python-requests <http://www.python-requests.org/>`_. 
+The reason for this is that requests is a very well-maintained python package, that 
+provides a really nice API, that many python developers are familiar with. Therefore, 
+in many cases, we recommend that you use the default :py:class:`HttpLocust <locust.core.HttpLocust>` 
+which uses requests. However, if your're planning to run really large scale scale tests,
```
aldenpeterson-wf: nit: s/your're/you're

docs/increase-performance.rst
aldenpeterson-wf: Could also make it an extras package so it's `pip install locustio[geventhttp]` and not installed all the time.

docs/increase-performance.rst
```diff
@@ -0,0 +1,70 @@
...
+    
+    class MyLocust(FastHttpLocust):
+        task_set = MyTaskSet
+        min_wait = 1000
+        max_wait = 60000
+
+
+.. note::
+
+    FastHttpLocust uses a whole other HTTP client implementation, with a different API, compared to
```
aldenpeterson-wf: if I was python BDFL or BDFaD (a day) I'd hunt down every HTTP library builder and force them to pick the same API to avoid this....

docs/running-locust-distributed.rst
```diff
@@ -85,3 +85,10 @@ Running Locust distributed without the web UI
 =============================================
 
 See :ref:`running-locust-distributed-without-web-ui`
+
+
+Increase Locust's performance
```
aldenpeterson-wf: could also link boomer here, I suspect that has higher overall RPS throughput than any python code ever will

locust/contrib/fasthttp.py
```diff
@@ -0,0 +1,383 @@
...
+from geventhttpclient.response import HTTPConnectionClosed
+
+from locust import events
+from locust.core import Locust
+from locust.exception import LocustError, CatchResponseError, ResponseError
+
+
+# Monkey patch geventhttpclient.useragent.CompatRequest so that Cookiejar works with Python >= 3.3
+# More info: https://github.com/requests/requests/pull/871
+CompatRequest.unverifiable = False
```
aldenpeterson-wf: should this only be done for python 3 in the `else` above?

docs/increase-performance.rst
```diff
@@ -0,0 +1,70 @@
...
+the load testing script does. However, if your locust scripts are spending most of their 
+CPU time in making HTTP-requests, you are likely to see signifant performance gains.
+
+
+Known limitations
+=================
+
+* :py:class:`FastHttpLocust <locust.contrib.fasthttp.FastHttpLocust>` does not 
+  support the :ref:`catch-response <catch-response>` argument.
+* Basic auth is currently not supported.
```
aldenpeterson-wf: isn't this what is `_construct_basic_auth_str`?

locust/contrib/fasthttp.py
```diff
@@ -0,0 +1,383 @@
...
+    auth_header = None
+    
+    def __init__(self, base_url):
+        self.base_url = base_url
+        self.cookiejar = CookieJar()
+        self.client = LocustUserAgent(max_retries=1, cookiejar=self.cookiejar)
+        
+        # Check for basic authentication
+        parsed_url = urlparse(self.base_url)
+        if parsed_url.username and parsed_url.password:
```
aldenpeterson-wf: is it worth warning if only one was provided?

locust/contrib/fasthttp.py
```diff
@@ -0,0 +1,383 @@
...
+            self.base_url = urlunparse((parsed_url.scheme, netloc, parsed_url.path, parsed_url.params, parsed_url.query, parsed_url.fragment))
+            # store authentication header (we construct this by using _basic_auth_str() function from requests.auth)
+            self.auth_header = _construct_basic_auth_str(parsed_url.username, parsed_url.password)
+    
+    def _build_url(self, path):
+        """ prepend url with hostname unless it's already an absolute URL """
+        if absolute_http_url_regexp.match(path):
+            return path
+        else:
+            return "%s%s" % (self.base_url, path)
```
aldenpeterson-wf: this might be a bit confusing if you have a something like:
```
base_url = http://localhost.com
path = myservice
```

locust/contrib/fasthttp.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+class ErrorResponse(object):
+    """
+    This is used as a dummy response object when geventhttpclient raises an error 
+    that doesn't have a real Response object attached. E.g. a socket error or similar
+    """
+    headers = None
+    content = None
+    status_code = 0
```
aldenpeterson-wf: will a status code of 0 confuse people?

locust/contrib/fasthttp.py
heyman: Doesn't really matter. I've also submitted a PR that fixes this is geventhttpclient, which has been merged, so we should be able to remove this altogether once they release a new version.

docs/increase-performance.rst
heyman: Yes! Both limitations are now fixed, so I've removed the whole section.

locust/contrib/fasthttp.py
heyman: Yeah, maybe. Shouldn't be too hard to figure out the exception though. We currently have the same potential problem with the default HTTP client as well.

locust/contrib/fasthttp.py
heyman: The reason for setting status_code to 0 is that there are some case where Requests does this (I don't remember exactly when), and I wanted to keep the API as similar as possible.

heyman:
I've now set up a repository (https://github.com/locustio/geventhttpclient/) with a fork of geventhttpclient that automatically will build wheels and upload them to the geventhttpclient-wheels PyPI package (https://pypi.python.org/pypi/geventhttpclient-wheels) when a new git tag is pushed.

We can now let Locust depend on the geventhttpclient-wheels package.

coderlifter:
Hi @heyman! Is there any chance this is getting merged and released any time soon?

I have built it locally and the performance boost is astonishing :)

sj2208:
@heyman  - 
Hey this is really some great stuff. 
I have issues when i am using catch_response.   I am not able to get the content out of the response. 
Can you please help here


from locust import task, TaskSet, HttpLocust
from locust.contrib.fasthttp import FastHttpLocust

class BenchLocust(FastHttpLocust): # or HttpLocust
    min_wait = 0
    max_wait = 0
    host = "http://127.0.0.1:8001"
    
    class task_set(TaskSet):
        @task
        def index(self):
            response = self.client.get("/", catch_response=True)
            if response.status_code == 200:
                 jsonBody = json.loads(response.content)
                 // Do some stuff here with jsonBody

vamega:
Now that the merge conflicts have been resolved, is there anything that prevents this from being merged?

guilatrova:
Why is it not merged yet?

coderlifter:
I have had the same problem with other Pull Requests for a while. I don't think the gatekeepers are active anymore @guilatrova

mkusz:
@cgoldberg What is the reason for this PR is not being merged?

dalbani:
I've been personally very satisfied with the performance of Locust with this branch. Without it I would have to ditch Locust for another solution, given the performance limitation with the standard HTTP client.

wuyanfeng1981:
image

cgoldberg:
@wuyanfeng1981,
What's a screenshot of an error supposed to tell us?

what code are you running?  how are you running it?  what's your environment?  how does this relate to this PR?  etc.. etc... etc...

(also, it's never a good idea to post screenshots of code or logs.  just paste the actual text it contains... that way the text is readable, searchable, and cut/pasteable)

aldenpeterson-wf:
IMO @heyman @cgoldberg  we should merge this after the current release is cut and then rerelease this as a separate release as it's a reasonably sizable feature

tvhphuong:
Hello guys, we also need to use FastHttpLocust , can you help with any information about when it will be merged. Thanks!

wmorgan6796:
Also waiting for this to be merged. Any update would be appreciated. Thanks!

tylerturk:
In speaking with @cgoldberg on Slack, it seems this PR is being replaced by #838 which fixes some merge conflict issues but currently has some other issues. I'm going to see if I can find some time today to true those up and get this ready to be merged.

aldenpeterson-wf:
Closing in favor of https://github.com/locustio/locust/pull/838

PR #717, opened 2017-12-28
fix error message on invalid time format

cgoldberg:
error message referred to `--time-limit`.  This PR changes it to `--run-time`.


heyman:
👍

PR #718, opened 2018-01-04
Add failure rate chart

thomas-phillips-nz:
Adds a failure rate chart. Also corrected 95% to 95th


thomas-phillips-nz:
Not correct

PR #719, opened 2018-01-04
Add failure rate chart to charts

thomas-phillips-nz:
Added a chart of failure ratio over time.


thomas-phillips-nz:
This with #693 looks to be getting to too many charts :worried:

Maybe someone with good JS skills should add a drop-down list of possible charts??

PR #720, opened 2018-01-05
Add failure rate chart to charts

thomas-phillips-nz:
Added a chart of failure ratio over time.

This with #693 looks to be getting to too many charts :worried:

Maybe someone with good JS skills should add a drop-down list of possible charts??


brianorwhatever:
Would love to see this merged in. The jenkins build looks like it only failed on the connection to pypi. A rerun of the build should clear it up.

etiennechabert:
This would be so good to see this feature alive

cyberw:
@thomas-phillips-nz Could you look into the conflicts? If you resolve them I'd be happy to merge this.

cyberw:
This will be superseeded by https://github.com/locustio/locust/pull/588 (or rather, its successor), as having the failure rate in the same graph as the success rate is much better imho :)

PR #725, opened 2018-01-23
Use flask.jsonify for json responses

hyperair:
This ensures that all json responses have the correct mimetype without needing
the extra boilerplate to set the content-type header.

This helps let extensions like JSONViewer do the right thing when examining JSON responses like `/stats/requests`


heyman:
Looks good, thanks!

PR #731, opened 2018-02-09
Update installation.rst

Fyb3roptik:
Fixing the install script for MacOS. Changing name from OS X to MacOS


aldenpeterson-wf:
I kind of think the preferred way should be using a virtual environment tbh, rather than installing "globally for that user"0.

cgoldberg:
hmm.. @aldenpeterson-wf is right

Fyb3roptik:
Well can we get an updated Docu for that? Currently following the instructions results in a non working install

Issue #163, opened 2014-05-28
log.py's StdErrWrapper swallows fatal stacktraces


jpotter:
In the condition where there's a python error in the locust file, the stack trace is swallowed by the StdErrWrapper.

In main.py, the code imports the locust file:
    imported = **import**(os.path.splitext(locustfile)[0])
without a try/catch block.

One possibility is to try/catch around this, and logging the stack trace there, but the try block would have to explicitly list things like IOError, KeyError, etc as Error exceptions are not caught by an unbounded "except" clause. The other possibility is to not wrap STDERR, and leave any prints to sys.stderr as something to be handled by the supervisor that's managing the process.

This is actually a really annoyance: anytime we have an issue in any library of ours, there's no way to see the stacktrace; locust just instantly exits 1 with no info as to the cause.

What's the suggested way to fix this?

Thanks,
Jeff

Exide:
@heyman @jpotter This issue is fairly old but I ran into it also and submitted #738 to side step the ```setup_logging``` call with a CLI option.

heyman:
This should now be avoidable by using `--skip-log-setup` merged in #738.

PR #738, opened 2018-02-22
Allow skipping the logging setup

Exide:
New CLI option:

    --skip-log-setup

This would allow people to setup their own logging configuration. It would also provide a workaround for #163.


Exide:
I've not used Codecov before so I'm finding it difficult to understand what threshold I need to meet. Let me know what is needed and I'll rectify it.

soulrebel:
I was trying to use pudb to do some simple debugging, but it fails to open the console interface due to the wrapping of standard error/output

Exide:
@soulrebel This PR would allow you to skip the call to ```log.setup_logging``` which is where that wrapper is established.

SpencerPinegar:
This should be merged with the code base -- vital when testing code or else the unittestpy.py with error out with -- 'StdOutWrapper' object has no attribute 'getvalue' -- if this functionality could be incorporated into the wrapper that would be cool, but it looked like the getvalue function had other dependencies within the unittestpy.py module.

aldenpeterson-wf:
This LGTM. 

Anyone have objections to me merging this?

tally38:
Also ran into this issue, and would like this to be merged!

PR #743, opened 2018-03-05
Change Locust website url to https

iassal:
http urls are not working


cgoldberg:
merged.  @iassal , thanks!

PR #746, opened 2018-03-07
Introduce **kwargs to request_success/failure parameter list

### LINKED ISSUE #745: Default implementation of on_request_success and _failure lacks **kwargs declaration
### Description of issue / feature request
When implementing new (custom) client or trying to inject more metadata into results the user will usually need to introduce dedicated success/failure handlers that may extend the default list of parameters and/or fire success/failure events manually (probably with additional arguments).

Take this for example: 
https://gist.github.com/karol-brejna-i/da6295a4d7bf25c2796fa95c622e7d80

### Expected behavior
Users should be able to introduce additional request handlers that include extended parameter lists.

### Actual behavior
Because of the way the default request handlers are implemented (https://github.com/locustio/locust/blob/5e1ccc578e0984692890df455b00a94e06559321/locust/stats.py#L557), Locust will fail in this case with something similar to:
https://gist.github.com/karol-brejna-i/4ab6b0a283a5cc56e4013acef31622f8

### Environment settings (for bug reports)
- OS: doesn't affect the results; Docker, tested on Ubuntu 16.04, Windows 10
- Python version: 3.6
- Locust version: 0.8.1

Define additional handlers, fire request_success with additional arguments as described in https://medium.com/p/183d2ae4a4c2

karol-brejna-i:
Fixes #745

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -554,10 +554,10 @@ global_stats = RequestStats()
 A global instance for holding the statistics. Should be removed eventually.
 """
 
-def on_request_success(request_type, name, response_time, response_length):
+def on_request_success(request_type, name, response_time, response_length, **kwargs):
     global_stats.log_request(request_type, name, response_time, response_length)
 
-def on_request_failure(request_type, name, response_time, exception):
+def on_request_failure(request_type, name, response_time, exception, **kwargs):
     global_stats.log_error(request_type, name, exception)
 
 def on_report_to_master(client_id, data):


ovanes:
Can you please explain how this code is supposed to work in distributed Locust scenario: Master + Multiple Slave. How the `kwargs` are transferred from a slave to the master node.

karol-brejna-i:
@ovanes Sorry for the late answer. 

As described here: https://github.com/locustio/locust/issues/745 - current implementation prevent users from introducing their own (additional) on_request_success/failure. This patch is simply about enabling users to do so, and do whatever they want in the additional handler.

Reporting to master is left to the default handlers. 

In this case you have:
* exact the same Locust mechanics for collecting the results
* ability for the users to capture/handle additional metadata.

PR #747, opened 2018-03-09
Docs: update test statistics page with example responses

aldenpeterson-wf:
Provide an example in the docs based on  [this Stack Overflow question](https://stackoverflow.com/questions/49197829/locust-test-statistics-meta-data/49198304#49198304).

### DIFF
--- a/docs/retrieving-stats.rst
+++ b/docs/retrieving-stats.rst
@@ -16,14 +16,14 @@ or for v0.8 (where the ``-t`` option isn't available)::
     locust -f examples/basic.py --csv=example --no-web -n10 -c10
 
 
-The files will be named `example_distribution.csv` and `example_requests.csv` (when using `--csv=example`) and mirror Locust's built in stat pages.
+The files will be named ``example_distribution.csv`` and ``example_requests.csv`` (when using ``--csv=example``) and mirror Locust's built in stat pages.
 
 You can also customize how frequently this is written if you desire faster (or slower) writing::
 
     import locust.stats
     locust.stats.CSV_STATS_INTERVAL_SEC = 5 # default is 2 seconds
 
-This data will write two files with `_distribution.csv` and `_requests.csv` added to the name you give::
+This data will write two files with ``_distribution.csv`` and ``_requests.csv`` added to the name you give::
 
     $cat example_distribution.csv
     "Name","# requests","50%","66%","75%","80%","90%","95%","98%","99%","100%"
--- a/docs/retrieving-stats.rst
+++ b/docs/retrieving-stats.rst
@@ -9,11 +9,8 @@ First, when running Locust with the web UI, you can retrieve CSV files under the
 Secondly, you can run Locust with a flag which will periodically save two CSV files. This is particularly useful
 if you plan on running Locust in an automated way with the ``--no-web`` flag::
 
-    locust -f examples/basic.py --csv=example --no-web -n10 -t10m
+    locust -f examples/basic.py --csv=example --no-web -t10m
 
-or for v0.8 (where the ``-t`` option isn't available)::
-
-    locust -f examples/basic.py --csv=example --no-web -n10 -c10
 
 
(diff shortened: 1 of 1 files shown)


docs/retrieving-stats.rst
```diff
@@ -9,13 +9,31 @@ First, when running Locust with the web UI, you can retrieve CSV files under the
...
 You can also customize how frequently this is written if you desire faster (or slower) writing::
 
     import locust.stats
     locust.stats.CSV_STATS_INTERVAL_SEC = 5 # default is 2 seconds
+
+This data will look like::
+
+    $cat example_distribution.csv
+    "Name","# requests","50%","66%","75%","80%","90%","95%","98%","99%","100%"
+    "GET /",31,4,4,4,4,4,4,4,4,4
```
cgoldberg: isn't the "Method" field header missing here?

docs/retrieving-stats.rst
```diff
@@ -9,13 +9,31 @@ First, when running Locust with the web UI, you can retrieve CSV files under the
 Secondly, you can run Locust with a flag which will periodically save the CSV file. This is particularly useful
 if you plan on running Locust in an automated way with the ``--no-web`` flag::
 
-    locust -f locust_files/my_locust_file.py --csv=foobar --no-web -n10 -t10m
+    locust -f examples/basic.py --csv=example --no-web -n10 -t10m
```
cgoldberg: is `example` the full name of the csv file?  Might want to name it `example.csv`

docs/retrieving-stats.rst
```diff
@@ -9,13 +9,31 @@ First, when running Locust with the web UI, you can retrieve CSV files under the
...
 
 or for v0.8 (where the ``-t`` option isn't available)::
 
-    locust -f locust_files/my_locust_file.py --csv=foobar --no-web -n10 -c10
+    locust -f examples/basic.py --csv=example --no-web -n10 -c10
```
cgoldberg: same as above

docs/retrieving-stats.rst
aldenpeterson-wf: I'm not sure, it's identical to what is generated by locust for the stats pages currently.

docs/retrieving-stats.rst
aldenpeterson-wf: it gets prepended with the type of file, so providing `example` results in `example_distribution.csv` and `example_requests.csv`

cgoldberg: gotcha.  that's confusing.  maybe note that in the docs?

docs/retrieving-stats.rst
cgoldberg: oh I see.  Method and Name are merged into a "Name" column.  This seems like a bug to me (though unrelated to your PR

docs/retrieving-stats.rst
aldenpeterson-wf: I added clarification there

docs/retrieving-stats.rst
```diff
@@ -6,16 +6,37 @@ You may wish to consume your Locust results via a CSV file. In this case, there
...
+Secondly, you can run Locust with a flag which will periodically save two CSV files. This is particularly useful
 if you plan on running Locust in an automated way with the ``--no-web`` flag::
 
-    locust -f locust_files/my_locust_file.py --csv=foobar --no-web -n10 -t10m
+    locust -f examples/basic.py --csv=example --no-web -n10 -t10m
 
 or for v0.8 (where the ``-t`` option isn't available)::
 
-    locust -f locust_files/my_locust_file.py --csv=foobar --no-web -n10 -c10
+    locust -f examples/basic.py --csv=example --no-web -n10 -c10
```
karol-brejna-i: is `-n` switch working? If not (which I suspect) maybe we should either remove it from the example or reintroduce it to the code.

aldenpeterson-wf: good catch, will update - it worked when I _originally_ wrote this functionality :)

cgoldberg: @aldenpeterson-wf can you update ^^ ?  then let's get this merged.

PR #749, opened 2018-03-16
Removed unused parameters in web.start method

sourcepirate:
Removed unused classes parameter from web.start method.


aldenpeterson-wf:
Is there more context about why you want to do this?

It would break folks that are directly invoking that.

sourcepirate:
@aldenpeterson-wf Even invoked directly the `start` method and `web` module doesn't use locust_classes in the parameter.
Is there any specific reason why locust_classes are passed in to the parameter ?

cyberw:
Hi! This PR has been untouched for a very long time and has conflicts so I am closing it.

Let me know if this was a mistake and feel free to open it again.

PR #755, opened 2018-03-23
Remove test dependency unittest2

jdufresne:
As locust no longer support Python 2.6, there is no need to install unittest2. The unittest shipped with the Python 2.7+ stdlib is the same as unittest2.

Results in fewer dependencies and slightly faster installation during testing.

Python support was removed in version 0.8.


cgoldberg:
@jdufresne thanks!

PR #759, opened 2018-03-25
Trim trailing white space throughout the project

jdufresne:
Many editors clean up trailing white space on save. By removing it all in one go, it helps keep future diffs cleaner by avoiding spurious white space changes on unrelated lines.


aldenpeterson-wf:
Hah is one way to get a lot of lines of code contributions to a project.

LGTM though.

cgoldberg:
I'll see your trimmed trailing whitespace, and raise you full pep8 whitespace compliance!

```
$ autopep8 --in-place --recursive locust/
```

----

** https://pypi.python.org/pypi/autopep8

jdufresne:
Thanks for the suggestion. I ran the suggested autopep8 command for other style fixes.

heyman:
I'm -0 on this. It obfuscates the history when using git blame and git log which has a real impact when trying to understand the history of the source code. And the actual gains are small/imaginary. (I do know one can get around that by adding `-w`/`--ignore-space-at-eol` to git commands, and `?w=1` to Github URLs, but still, it makes it more complicated)

jdufresne:
> It obfuscates the history when using git blame and git log which has a real impact when trying to understand the history of the source code. 

I understand this concern.

For myself, using Emacs to view git blame allows one to quickly jump to the previous reviouse  of a line. So if one wants to see the blame information before the white space cleanup, one would simply hit the "a" key on that line. So for me, this isn't an issue in practice. I recognize people use different tools or may not be familiar with such a feature.

> And the actual gains are small/imaginary.

I think there are real gains.

tysonclugg:
I'm +1 on this.  Addressing a couple of the concerns expressed by others:
 - Big diff / git blame fallout: This is a tooling issue, not an issue with the code in question.  Tooling issues should not dictate best practice, instead either learn to use the tools or help get the tools fixed.
 - Editors are removing trailing whitespace by default:  Create an [`.editorconfig`](http://editorconfig.org/) file to specify editor options for all project developers, including options for controlling trailing space.

While we're on the topic of editor configuration, it might be a good idea to align with current best practice of [expecting developers keep individual editor specific configuration outside of the repo](https://github.com/django/django/blob/master/.gitignore).

jdufresne:
I think adding an `.editorconfig` file is a good suggestion. I have included it in the latest revision.

jdufresne:
Rebased to resolve merge conflicts.

jdufresne:
Rebased :slightly_smiling_face:

cyberw:
@jdufresne Lets do this when there are fewer open PRs, as the merge conflicts would be terrible (I also prefer Black over autopep8, but that is more of a personal preference :) )

heyman:
I'm still not convinced of this change. I really don't see the real benefit, but I see some real drawbacks.

`.editorconfig` sounds like a good idea though, and we could set the `trim_trailing_whitespace`in `.editorconfig` to `false`. Hopefully that would make it less of a problem for people who use editors that by default trim all whitespace at end of lines.

PR #762, opened 2018-03-25
Fix bytes/text confusion with response objects

jdufresne:
`response.content` is a byte string. It should be compared against other byte strings. To examine the response content as text, use the `response.text` attribute.

Fixes warnings when Python is passed the `-b` argument (Issue warnings about `str(bytes_instance)`, `str(bytearray_instance)` and comparing bytes/bytearray with str.)

```
BytesWarning: str() on a bytes instance
```

For additional information on the `-b` argument, see:

https://docs.python.org/3/using/cmdline.html#cmdoption-b

For additional information on response.text, see:

http://docs.python-requests.org/en/master/api/#requests.Response.text

### DIFF
--- a/docs/writing-a-locustfile.rst
+++ b/docs/writing-a-locustfile.rst
@@ -342,7 +342,7 @@ class::
 
     response = self.client.get("/about")
     print("Response status code:", response.status_code)
-    print("Response content:", response.content)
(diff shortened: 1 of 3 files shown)


cgoldberg:
thanks!

PR #769, opened 2018-04-02
Switch from using optparse to argparse for command line arguments

jdufresne:
The Python module optparse is deprecated and is no longer receiving updates. From the [Python documentation](https://docs.python.org/3/library/optparse.html):

> Deprecated since version 3.2: The optparse module is deprecated and will not be developed further; development will continue with the argparse module.

Can take advantage of the argpares API by making the following cleanups:

---

Remove `default=None`. It is the default.

https://docs.python.org/3/library/argparse.html#default

> The default keyword argument of add_argument(), whose value defaults to None, ...

---

Remove `dest='...'`, when it matches the first long option, which is the argparse default.

https://docs.python.org/3/library/argparse.html#dest

> For optional argument actions, the value of `dest` is normally inferred from the option strings. `ArgumentParser` generates the value of `dest` by taking the first long option string and stripping away the initial `--` string. [...] Any internal `-` characters will be converted to `_` characters to make sure the string is a valid attribute name.

---

Remove default=False for type='store_true' as it is an implied default:

https://docs.python.org/3/library/argparse.html#action

> `'store_true'` and `'store_false'` - These are special cases of `'store_const'` used for storing the values `True` and `False` respectively. In addition, they create default values of `False` and `True` respectively.

Use argparse builtin version support:

https://docs.python.org/3/library/argparse.html#action

> `'version'` - This expects a `version=` keyword argument in the `add_argument()` call, and prints version information and exits when invoked

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -6,7 +6,7 @@ import signal
 import socket
 import sys
 import time
-from optparse import OptionParser
+import argparse
 
 import gevent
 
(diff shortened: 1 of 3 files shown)


cyberw:
LGTM. @jdufresne , if you fix the conflicts I think we should merge this. @heyman , do you agree?

heyman:
Yes, sounds good!

jdufresne:
Rebased :slightly_smiling_face:

heyman:
Great, thanks!

PR #779, opened 2018-04-21
expand and re-order documentation index

cgoldberg:
this PR enhances the docs by:

* expands the index page entries in the Documentation
* re-orders the index so changelog is after third party tools

expanding the index is a huge improvement (IMO).

docs


aldenpeterson-wf:
👍

PR #783, opened 2018-04-25
Adding default dummy locustfile so masters can be run more easily

migueleliasweb:
With this running master in distributed mode becomes easier:
```bash
$ locust --master
```

Instead of:

```bash
$ locust --master -f /path/to/my/dummy_locustfile.py
```

When the user does not need to change any configurations to the master running in distributed mode, we should provide a simple dummy file.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -46,7 +46,7 @@ def parse_options():
         default="",
         help="Host to bind the web interface to. Defaults to '' (all interfaces)"
     )
-    
+
     parser.add_option(
         '-P', '--port', '--web-port',
         type="int",
@@ -54,7 +54,7 @@ def parse_options():
         default=8089,
         help="Port on which to run web host"
     )
-    
+
     parser.add_option(
         '-f', '--locustfile',
         dest='locustfile',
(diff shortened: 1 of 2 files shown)


mangatmodi:
Any decision on this one? Because
1. It doesn't make sense to give a locust file for Master. Master can be a aggregated single point of test log collection and coordination.
2. Adding a dummy file will at least make it standard and easy config for everyone.

cyberw:
I think it does make a lot of sense to be able to run master w/o a locustfile.

But for me it doesnt actually work unless I'm actually standing in the locust checkout, hardly what was intended, right? And the biggest issue is that it will also default to load this file when *not* running --master !

heyman:
On the one hand I do agree. It's not really needed currently. On the other hand removing it as a required parameter for the master, would make it impossible (without re-adding the `-f` requirement) for us to add functionality that requires that the master has knowledge of what code that the slaves are running.

cyberw:
Closing this PR. While the issue does deserve a fix, this solution doesnt really work.

Issue #18, opened 2012-01-13
Define wait times by function instead of variable


cgbystrom:
The current way of specifying the wait time for a Locust makes a rather large assumption on how a developers will want to do this. It assumes that wait time should be randomly distributed in a given interval of milliseconds.

With the current API,  a locust with random wait times between 2000 and 5000 ms looks like this:

``` python
class User(Locust):
    tasks = [index, stats]
    min_wait = 2000
    max_wait = 5000
```

However, if we were to replace those hard constants with a function instead, determining that wait time would be a lot easier to both communicate and implement. It would make no assumption on how the developer would like to define his/her wait times and at the same time provide a small, isolated place were this can take place (inside a function). Since randomly distributed wait times within an interval arguably would be the most common, we naturally would ship with some default ones.

Here's an example:

``` python
class User(Locust):
    tasks = [index, stats]
    wait_time = random_between(2, 5)
```

As you can see, wait_time is assigned a function returned by the `random_between` function. Very straight forward and pluggable. Even simpler than the current implementation as the Locust core would not even need to be aware of how to calculate wait times - just call the wait function to determine the wait time.

Two examples of wait functions that we could ship:

``` python
def random_between(from_time, to_time):
    def wait():
        return random.uniform(from_time, to_time)
    return wait

def constant(time):
    def wait():
        return time
    return wait
```

As with any API changes, discussing them here before diving in. What are your thoughts?

PR #785, opened 2018-04-27
Added user-defined wait_function to locust and TaskSet

ps-george:
Most real-world inter-arrival times are exponentially distributed so it would be nice to be able to model that using locust. I've added the possibility for a user-defined wait_function, so *any* desired inter-arrival time can be modelled. Added some [examples](https://github.com/ps-george/locust/blob/df89db25a3444bc41a5c00b252f581156e80f405/examples/custom_wait_function.py) of exponential wait times I'd like to be able to model. If `wait_function` is not set, it will default to exactly as before, i.e. `random.randint(self.min_wait,self.max_wait)`

Change-log needs updating if merging this.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -7,20 +7,20 @@ Locust class
 ============
 
 .. autoclass:: locust.core.Locust
-	:members: min_wait, max_wait, task_set, weight
+	:members: min_wait, max_wait, wait_function, task_set, weight
 
 HttpLocust class
 ================
 
 .. autoclass:: locust.core.HttpLocust
-	:members: min_wait, max_wait, task_set, client
+	:members: min_wait, max_wait, wait_function, task_set, client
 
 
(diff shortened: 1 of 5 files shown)


heyman:
Thanks for the PR!

This is currently possible to achieve by overriding the `wait()` method in the `TaskSet` class. However I think this is a nicer API. This would also resolve #18 :).

I'm +1

Does anyone else has thoughts on this?

aldenpeterson-wf:
+1 as well. I wish it was easier to figure out what sections of docs to update short of having them all memorized.

Also this is a great PR example too. A small, concise, clear feature request with a good API... 🏆

ps-george:
Thanks for the fast and welcoming response!

aldenpeterson-wf:
@ps-george realistically it's easy in a PR like this ;-)

PR #789, opened 2018-05-09
Adding more Percentiles to request stats web endpoint

albertowar:
At work, we use Locust for Performance testing.

We poll the request stats web endpoint to publish metrics to Grafana and we would like to have more information about response times (percentiles).

### DIFF
--- a/locust/test/test_runners.py
+++ b/locust/test/test_runners.py
@@ -156,23 +156,31 @@ class TestMasterRunner(LocustTestCase):
                     "user_count": 2,
                 }, "fake_client"))
                 mocked_time.return_value += 4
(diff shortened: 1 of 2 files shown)


cyberw:
Hi @albertowar ! Do you actually need the 10th and 25th percentiles? I've hardly ever heard anyone ask for anything below 50 :) Calculating percentiles is actually relatively expensive (as it requires sorting all the requests) so I'm -1 for this...

albertowar:
It has been a while indeed since I wrote this one and now I share the same sentiment. If anything, it would be nice to have a P99, since that's what people are asking for these days.

Will close for now, since I would like to revisit stats in general in a separate PR.

Issue #565, opened 2017-03-21
Median response times off


steventang2013:
```
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
----------------------------------------------------------------------------------------------------
 POST long                                                        1     0(0.00%)    6067    6067    6067  |    6100    0.00
 POST short                                                       1     0(0.00%)    1239    1239    1239  |    1200    0.20
----------------------------------------------------------------------------------------------------
 Total                                                           2     0(0.00%)                                       0.20
```
```
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s
--------------------------------------------------------------------------------------------------------------------------------------------
 POST medium                                                       7     0(0.00%)     468     199    1006  |     360    
 POST short                                                        2     0(0.00%)    1063     574    1553  |     570    0.33
----------------------------------------------------------------------------------------------------
 Total                                                              9     0(0.00%)                                       1.00
```

This only seems to be an intermittent issue, but notice how the median response times are not within the bounds of min and max. 
Anyone else ran into a similar issue?

giantryansaul:
Yes! I had someone at the office show me a situation where they ran 10 requests, the average was way outside of the bounds of what the logs showed. So probably not just an issue with the median, seems to be an issue with reporting. It also seems to be intermittent for us too so I haven't really gone too deep on it yet.

heyman:
In order to calculate the median response time, as well as response times for specific percentiles, without storing the response time for every single request, we keep a dict of the following format: {response_time: numer_of_requests}. In order to save memory we round the response time to only use two digits of precision (so that 6067 becomes 6100, 1239 -> 1200, 574 -> 570, and so on). Since that dict is used to calculate the median response times, while the exact response times are used when calculating min/max, it could happen that the median ends up outside of the min/max boundaries (especially for few requests).

I guess it would probably be good to mention that we only use two digits precision for median & percentile response times, in the docs and the web UI.

jredrejo:
I would like to know the status if this issue. I understand is a precision problem because only 2 digits are taken, but I don't understand why this is not improved. Having values of median over max and min is not what it's expected in a tool that's used to measure performance. Wouldn't it be possible to use more accuracy, or at least, round it to the max or min value in this cases?

cgoldberg:
> Having values of median over max and min is not what it's expected 
> in a tool that's used to measure performance

[...]

PR #790, opened 2018-05-10
Ensure median does not have impossible values

jredrejo:
This is a fix to limit the scope of the problem described in issue #565
Rounding still exists, but at least median won't show impossible values.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -294,8 +294,12 @@ class StatsEntry(object):
     def median_response_time(self):
         if not self.response_times:
             return 0
-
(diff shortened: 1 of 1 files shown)


cyberw:
@jredrejo  I think this PR makes sense. If you resolve the conflict (or file a completely new one) I will merge it.

heyman:
I merged this and added a test and comment: https://github.com/locustio/locust/compare/b16cf497305e83b271065700bc995ec18a6d1d73...551859352f76e116e1186300a6979463d500e9ab

Thanks!

Issue #795, opened 2018-05-16
branch geventhttpclient need a update


ityoung:
## Description

I think I meet a bottleneck of CPU in my testing machine when RPS reach 1700, while using Jmeter can reach about 2500. And I found a issue #727 which just like my situation. 

I think maybe I can try the `geventhttpclient` branch to raise the RPS, but the branch seems haven't update, and I need `setup` function which was added by version 0.8.1. @heyman 

when I run locust and RPS raise more than 1700, I got the error message on my slaves:

```
[2018-05-16 09:35:42,434] test/ERROR/stderr: During handling of the above exception, another exception occurred:
[2018-05-16 09:35:42,434] test/ERROR/stderr:
[2018-05-16 09:35:42,434] test/ERROR/stderr: Traceback (most recent call last):
[2018-05-16 09:35:42,434] test/ERROR/stderr:
[2018-05-16 09:35:42,434] test/ERROR/^C
    body=body, headers=headers)[2018-05-16 09:36:11,459] test/ERROR/stderr: Traceback (most recent call last):
[2018-05-16 09:36:11,460] test/ERROR/stderr:
[2018-05-16 09:36:11,460] test/ERROR/stderr: File "/usr/lib/python3/dist-packages/urllib3/connection.py", line 137, in _new_conn
    (self.host, self.port), self.timeout, **extra_kw)
[2018-05-16 09:36:11,460] test/ERROR/stderr:
[2018-05-16 09:36:11,460] test/ERROR/stderr: File "/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 91, in create_connection
    raise err
[2018-05-16 09:36:11,461] test/ERROR/stderr:
[2018-05-16 09:36:11,461] test/ERROR/stderr: File "/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 81, in create_connection
    sock.connect(sa)
[2018-05-16 09:36:11,461] test/ERROR/stderr:
[2018-05-16 09:36:11,461] test/ERROR/stderr: File "/usr/local/lib/python3.5/dist-packages/gevent/_socket3.py", line 336, in connect
    raise error(err, strerror(err))
[2018-05-16 09:36:11,461] test/ERROR/stderr:
[2018-05-16 09:36:11,461] test/ERROR/stderr: TimeoutError: [Errno 110] Connection timed out
```

## Locust Script Info

master + 4 slave

I have 7 api to test , and this is a snippet of my test script:

```
[...]

PR #796, opened 2018-05-16
update geventhttpclient

ityoung:
I create an issue #795 before, and found that can be merge straightly with no conflict.. awkward >_<

hope to be merged. @heyman


ityoung:
@heyman Hi, I merged this on my own PC and using `FastHttpLocut` to replace `HttpLocust`, it runs fast but, when I use `with` keyword try to set a checkpoint, the `response.content`(or`response.text`) is always an empty bytearray(b''). 

```
        with self.client.get(my_url, catch_response=True) as response:
            res = response.content
            response.failure(res)    # use this failure to output response.content
```

And if I use `HttpLocust` with `with` keyword, it response exactly what I expect.

And, when I'm not using `with` keyword, both of `FastHttpLocust` and `HttpLocust` work perfectly.

```
        res = self.client.get(my_url).text
        assert len(res) == 32, res    # use this to output response.text
```

I suspect that it may be unfinished of the request? I check the source code, but I really didn't know much about geventhttpclient. Hope you can help!

cgoldberg:
Our workflow has generally used Pull Requests to merge into master, not for merging master onto other dev branches on behalf of someone else... so this PR kinda confused me.

@heyman can you update your geventhttpclient branch with latest from master?  (either via this PR or just merging your local and pushing directly)

@ityoung for reference, we have #713 open for merging the geventhttpclient branch to master once ready.

ityoung:
@cgoldberg Yes I found #713 , and I found the branch `geventhttpclient` has 2 month didn't update, so I raise this PR. Thanks for telling me the usage of PR, I might need to give a issue instead, right?

@heyman I have merged from master on my local computer, and found some problem, hope you can fix that, it's a real good work!

PR #798, opened 2018-05-18
Changes to allow running Locust programmatically

ps-george:


ps-george:
Meant to merge this locally first..

PR #799, opened 2018-05-18
Allow running tests programmatically

### LINKED ISSUE #222: How to run test programatically
I would like to run these load testing through clicking a button, or something else. How to do this ? and also how to grab those result.

ps-george:
Fixes #222 

- Added main code into a function called `run_locust` that takes OptionParser object and additional arguments.
- Moved parser code into a file called `parser.py`
- Added tests to `test_main.py` for new changes.
- Added example file to demonstrate new functionality
- Added new docs for running locust programmatically
- Changed `from gevent import wsgi` in `test_web.py` to `from gevent import pywsgi` because it was throwing an error (also all uses of wsgi to pywsgi)
- Added Napoleon to Sphinx conf.py for sane parsing of function docstrings.
- Added functions `is_taskset` and `load_tasksetfile` since would be useful and minimal effort to add.
- Added create_options function to create default options for new mode

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ matrix:
       env: TOXENV=py35
     - python: 3.4
       env: TOXENV=py34
-    - python: 3.3
-      env: TOXENV=py33
     - python: 2.7
       env: TOXENV=py27
 addons:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,6 +9,8 @@
 # serve to show the default value.
 
 import os
+import sys
+sys.path.insert(0, os.path.abspath('../'))
 
 # The default replacements for |version| and |release|, also used in various
 # other places throughout the built documents.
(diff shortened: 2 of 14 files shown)


ps-george:
> Changed from gevent import wsgi in test_web.py to from gevent import pywsgi because it was throwing an error (also all uses of wsgi to pywsgi)
Note that before this change all Travis builds are failing (see https://travis-ci.org/locustio/locust/jobs/380659365)

ps-george:
Open to any alternative suggestions to dealing with Travis CI no longer supporting python3.3 other than removing it from tox.ini and .travis.yml if anyone has any ideas.

tysonclugg:
https://www.python.org/dev/peps/pep-0398/#lifespan
> As of 2017-09-29, Python 3.3.x reached end-of-life status.

I'd drop support for 3.3.

tysonclugg:
I've had a look through the changes, and I can't see any issues.  Especially since the tests pass on this branch, and tests fail on master.

cgoldberg:
I think dropping 3.3 support is fine, but not in this PR.. please propose that separately.   also... *all* of the other changes unrelated to this PR (whitespace, refactoring, rearranging, doc theming, etc) really need to be moved into individual PR's.

cgoldberg:
> it was just difficult to submit a PR that failed the CI

you can add your PR to fix CI first

PR #801, opened 2018-05-22
Added a log handler to write metrics for all the requests into a specific file.

yannlv:
We use Locust at mappy.fr for performance testing.

For monitoring or dataviz purposes, Locust should be able to write the usual metrics in a specific file for *every* requests sent against the target server (option --resplogfile=).

The resplog impements the following format:
`[YYYY-mm-DD HH:MM:SS,sss]   hostname/loglevel/resplog   NAME   HTTP_METHOD   HTTP_CODE   URL   RESP_TIME   CONTENT_LENGTH`

We're aware that this feature could be easily implemented within the locust files through an event listener. But we are actually working on a Locust plugin for [Yandex Tank](https://yandextank.readthedocs.io/en/latest/) which requires the feature (with the objective of having all the tech teams in the company use this solution).
Having that PR merged will make things much easier for further maintenance and distribution across the teams + it would be a nice opportunity for Yandex Tank users to discover Locust when the plugin is complete.

### DIFF
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -116,7 +116,9 @@ class HttpSession(requests.Session):
         
         # record the consumed time
         request_meta["response_time"] = int((time.time() - request_meta["start_time"]) * 1000)
-        
+       
+        # record the HTP status code
+        request_meta["status_code"]=response.status_code
     
         request_meta["name"] = name or (response.history and response.history[0] or response).request.path_url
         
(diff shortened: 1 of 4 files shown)


vamega:
I don't like the hardcoding of a particular log format here.
Could that format be made configurable?

yannlv:
Hi @vamega,

The log format for the resplog handler is actually following the initial locust log format.
Anyway, you may add a _resplog_format_ argument to _setup_resplogging()_ in _locust/log.py_ and pass the the desired log format to the function from an option exposed in _locust/main.py_ .

Hope it helps.

Yann

cyberw:
Hi @yannlv ! Is this still of interest to you? The PR has not moved in a really long time and now has merge conflicts. I also think we meed to be careful with some of these things, not to break backwards compatibility.

heyman:
This is currently a -1 from me. The use-case sounds too specific to include it within Locust itself. Also the PR seems to add HTTP specific arguments to the request success & failure events which would make Locust less system agnostic.

cyberw:
Closing this. Perhaps a solution could be made for your specific use case that doesnt require modifying locust itself?

PR #804, opened 2018-05-28
Drop Python 3.3 support

ps-george:
* Dropping support for python 3.3 as discussed in #799 
* Fixing test_web.py, wsgi has been superceded with pywsgi


cgoldberg:
:+1: 
thanks!

PR #805, opened 2018-05-28
Run locust programmatically

### LINKED ISSUE #222: How to run test programatically

ps-george:
Fixes #222, fewer changes than #799


ps-george:
CI will pass after #804 has been merged.

ps-george:
Hi, could someone tell me what is delaying this PR or give a detailed breakdown of exactly what I need to do to get this merged?

vamega:
So I merged this into my own branch, and I think there's a few things that need to be changed before this should be merged.

This PR does a great job in providing an alternative interface to launching locust, but I think there is still some work to be done in accessing the results programatically.

[Inovkust](https://github.com/FutureSharks/invokust) shows a way of allowing access to the results in process rather than having that be controlled through an option in the command line.

I'll look into it more to see if there's a good way to combine the benefits of `invokust` with the benefits of this PR.

ps-george:
@vamega Ok let me know how you get on and if you need any help. Keen to have this functionality merged.

joaonc:
Would love to see this get merged!  @ps-george would you mind fixing the conflicts and contact the project owner(s) to have it merged?

ps-george:
@joaonc Me too, conflicts resolved.

It's an important but also quite big change, I'm not sure the project owners have the time to properly review.

I'm sure if we get some thumbs up and support for this PR, they will be more willing to look into it and help us merge this.

locust/main.py
```diff
@@ -262,13 +262,124 @@ def parse_options():
         default=False,
         help="show program's version number and exit"
     )
+    return parser
+
+
+def create_options(
+        locustfile='locustfile',
```
aldenpeterson-wf: I wonder if there's a more reader friendly way to do this - or if the general use case will be ok with a massive list of kwargs here.

locust/main.py
```diff
@@ -210,7 +210,7 @@ def parse_options():
        action='store_true',
        dest='only_summary',
        default=False,
-       help='Only print the summary stats'
+       help="Only print the summary stats"
```
aldenpeterson-wf: super nitpicky but why this change?

examples/run_programmatically.py
```diff
@@ -0,0 +1,28 @@
...
+class WebsiteUser(HttpLocust):
+    """
+    Locust user class that does requests to the locust web server running on localhost
+    """
+    host = "http://127.0.0.1:8089"
+    min_wait = 2000
+    max_wait = 5000
+    task_set = UserTasks
+
+options = create_options(locust_classes = [WebsiteUser])
```
aldenpeterson-wf: I'd add a comment here saying that all command line options are also available as kwargs to this function

locust/main.py
```diff
@@ -367,45 +480,74 @@ def load_locustfile(path):
...
+        logger.error(message)
+        if not cli_mode:
+            raise err_type(message)
+        sys.exit(exit_type)
     
     if options.show_version:
-        print("Locust %s" % (version,))
+        version_text = "Locust %s" % (version,)
+        print(version_text)
+        if not cli_mode:
```
aldenpeterson-wf: I'm trying to think of a valid use case for this type of function call from a non-cli mode - is there a reason you'd ever intentionally hit this codepath programmatically?

docs/running-locust-programmatically.rst
```diff
@@ -0,0 +1,17 @@
+.. _running-locust-programmatically:
+
+===============================
+Running Locust programmatically
```
aldenpeterson-wf: Nit: you could link to the example here

aldenpeterson-wf:
This looks pretty good to me, hopefully a few other folks can chime in - most of my comments are pretty minor and documentation related.

I think this is a great add to Locust. There's still likely opportunities to ensure that the results can be reported meaningfully but this is a good start  (I think you could just access them directly at `runners.locust_runner.request_stats` right now though).

locust/main.py
```diff
@@ -262,13 +262,124 @@ def parse_options():
...
+        only_summary=False,
+        show_task_ratio=False,
+        show_task_ratio_json=False,
+        logfile=None,
+        loglevel='INFO',
+        show_version=False,
+        list_commands=False):
+    """Create options objects for passing to `run_locust` when running Locust programmatically.
+    
+    Keyword Arguments:
```
SpencerPinegar: I like how all of the options are listed here, but could the most important ones be isolated out and set as kwargs in run_locust call with appropriate defaults. Other options could be grouped by their functionality like no-web options, distributed options, etc..

SpencerPinegar:
One other thing to note, if locust is run pragmatically by the multiprocessing module the response times are not accurate (they are largely inflated) - when you run them in the subprocess module the response times appear to be accurate - if this could be noted explicitly that would help avoid a lot of confusion.

joaonc:
@SpencerPinegar RE response times, you think issue #930 would take care of it?

SpencerPinegar:
While I cannot say definitively, I think that this fix works great for all clients using http requests (and it seems it would be the most accurate way to benchmark) but this wont work for non http-clients so using it as the way to monitor all response times isn't the options (unless I am missing something - kinda hope I am because it seems accurate).

That being said for most http clients it would probably fix the multiprocessing issue.

locust/main.py
ps-george: The way I was finding use from this was being able to `import locust` in Python and being able to interact with all the statistics directly on the class, rather than reading from the CLI interface.

locust/main.py
ps-george: Good comment; can probably apply some functional programming concepts here to reduce the complexity.

locust/main.py
ps-george: Looking back, I have no idea. I think it was to be consist with some code elsewhere.

locust/main.py
ps-george: Unsure, open to any suggestions.

Essentially, with the CLI tool, you're basically doing this same thing - loads of default arguments but the ability to change all of them via CLI arguments. But have put defaults in so unless the user who's calling this code only needs to put in the ones they're changing.

locust/main.py
ps-george: Ok, will group these up with comments explaining the grouping.

cyberw:
@ps-george Hi! sorry for the lack of "official" response on this PR. @heyman what is your opinion on this? The diff is reasonably small at least...

heyman:
I think it's a good idea to be able to run Locust programatically (and be able to use it more as a a lib). I'm thinking that the best way to do this would be to expose a larger part of Locust's internal code and classes (runners, stats, etc.) as part of the public API. However, this would definitely require some refactoring/re-work of some of Locust's code. Both to make sure we expose a  good API, and also because once we add something to the public API, we want to try our best to not do breaking changes. I've been thinking about this quite a bit, but haven't taken the time to concretize the ideas into a Github issue yet. 

For this reason, I'm a bit hesitant  to merge this pull request. I understand that this PR aims to introduce this feature with the least obtrusive changes as possible. That is usually a great goal for PRs to have, but in this case I think it might be better to take a step back and do some re-factoring of the existing code instead.

cyberw:
@heyman Ok to close/decline?

PR #808, opened 2018-05-29
Update third-party-tools.rst

anhldbk:
Added Java Locust Swarm client


anhldbk:
@ps-george I'm sorry. What do you think about the idea of grouping clients based on their languages? If we don't need that, then I add another heading for my client

ps-george:
I think it's a good idea to group clients by languages :+1:

anhldbk:
@ps-george ps I updated

aldenpeterson-wf:
LGTM

Issue #311, opened 2015-08-03
run each task in sequence


aidylewis:
Is there a way to run each task in a sequence, instead of being chosen randomly?

vividhsv:
You can't have tasks run in a sequence. However, you can make multiple requests within the same task which would be run sequentially.

inexion:
Does anyone know if there's a tool like Locust that allows for sequential task running? That would be hugely useful...

sp1rs:
@vividhsv why cannot we not add functionality like sequence task ? .. 
like weight attribute , we can also have sequence number like attribute ?
I think it can be possible..

giantryansaul:
I have had similar issues in the past where the sequence of events needs to be correct to accurately measure performance for a business case. To make tasks more readable, you can break them down into separate methods without the @task decorator and have a single task control the sequence of events. Locust still reports each request made, whether you put those requests into single tasks or not.

inexion:
@heyman @HeyHugo @cgbystrom @Jahaja can we get this added as a feature? 

Would this be difficult to just accomplish on a fork @giantryansaul @vividhsv @sp1rs ?

giantryansaul:
I think I meant more that your task can define the entire workflow and not just a single request. You would not need a fork of locust to accomplish this. Here is an example:

    class UserWorkflow(TaskSet):
        def on_start(self):
            self.login()

        def login(self):
            self.client.post("/login", {"username":"jack_sparrow", "password":"blackpearl"})

        @task(1)
        def user_workflow(self):
            self.load_home_page()
            self.get_list_of_posts()
            self.get_single_post(123)

        def load_home_page(self):
            self.client.get("/")

        def get_list_of_posts(self):
            self.client.get("/posts")

        def get_single_post(self, post_id):
            self.client.get("/posts/{}".format(post_id))

aidylewis:
I raised this issue, and as far as I am concerned @giantryansaul offers a reasonable solution.

javivdm:
When you put multiple requests under the same task as showed above, is the min and max wait still applied to each request or does it only apply to each task?

giantryansaul:
@javivdm The waits are applied to the tasks, so you can use sleep functions to specify waits between the requests.

wessel-techtribe-nl:
self.tasks.sort returns a `TypeError: '<' not supported between instances of 'function' and 'function'` in python3.6 but works fine in python2.7

[...]

PR #827, opened 2018-06-23
Add tasks sequence support

Ramshell:
This pr is related to [this issue](https://github.com/locustio/locust/issues/311).

As I said In slack, I know there is an easy workaround of defining only one task in the TaskSet and defining all the behavior there. The problem with the workaround is that it is a bit hacky, since you have to do `self.wait()` whenever it's needed instead of letting the framework handle it.

# If accepted this pr will
Introduce a new class: `TaskSequence` which accepts a new decorator @seq_task(int) to define the order of the tasks.

For example:
```python
class BrowseDocumentationSequence(TaskSquence):
    def on_start(self):
        self.urls_on_current_page = self.toc_urls

    # assume all users arrive at the index page
    @seq_task(1)
    def index_page(self):
        r = self.client.get("/")
        pq = PyQuery(r.content)
        link_elements = pq(".toctree-wrapper a.internal")
        self.toc_urls = [
            l.attrib["href"] for l in link_elements
        ]

    @seq_task(2)
    @task(50)
    def load_page(self, url=None):
        url = random.choice(self.toc_urls)
        r = self.client.get(url)
        pq = PyQuery(r.content)
        link_elements = pq("a.internal")
        self.urls_on_current_page = [
            l.attrib["href"] for l in link_elements
        ]

    @seq_task(3)
    @task(30)
    def load_sub_page(self):
        url = random.choice(self.urls_on_current_page)
        r = self.client.get(url)
```

In the example above, index_page/1 will be executed first, then load_page/1 fifty times, and then load_sub_page/1 thirty times.

`@seq_task(int)` doesn't have a default parameter because I found that if I set the same order value to every task, the final order ends up being kind of random.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -27,6 +27,16 @@ task decorator
 
 .. autofunction:: locust.core.task
 
+TaskSequence class
+==================
+
+.. autoclass:: locust.core.TaskSequence
+	:members: locust, parent, min_wait, max_wait, wait_function, client, tasks, interrupt, schedule_task
+
+seq_task decorator
+==============
+
+.. autofunction:: locust.core.seq_task
 
 HttpSession class
 =================
--- a/docs/writing-a-locustfile.rst
+++ b/docs/writing-a-locustfile.rst
@@ -255,6 +255,31 @@ its Locust instance, and the attribute :py:attr:`parent <locust.core.TaskSet.par
 parent TaskSet (it will point to the Locust instance, in the base TaskSet).
 
 
+TaskSequence class
+==================
+
+TaskSequence class is a TaskSet but its tasks will be executed in order.
+To define this order you should do the following:
+
+.. code-block:: python
+
+    class MyTaskSequence(TaskSequence):
+        @seq_task(1)
+        def first_task(self):
+            pass
+
+        @seq_task(2)
+        def second_task(self):
+            pass
+
+        @seq_task(3)
+        @task(10)
+        def third_task(self):
+            pass
+
+In the above example, the order is defined to execute first_task, then second_task and lastly the third_task for 10 times.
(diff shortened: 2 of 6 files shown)


cgoldberg:
can you also include an update to the docs in your PR if we decide to merge it?

Ramshell:
@cgoldberg Yeah, It's okay if I add and example in `quick start` and an example with explanation in `writing a locust file`?

Ramshell:
Well, finally I decided not to change the quick start because it's supposed to be short. Instead I've updated the `api.rst` and `writing-locustfile.rst` files. If there is anything else needed, just tell me :)

ityoung:
That's a real good work! I try to do that but met some trouble, thanks for your work!

cgoldberg:
I'm +1 on merging this.  Anyone else?
/cc @locustio/committers

ityoung:
I think you should add `TaskSequence` and `seq_task` to file `locust/__init__.py`, so that we can use `from locust import seq_task`.

Ramshell:
Adding TaskSequence and seq_task to file locust/__init__.py sounds like something useful. @cgoldberg what do you say?

cgoldberg:
agreed.

aldenpeterson-wf:
This LGTM

Ramshell:
If there is anything else you think I could add, just tell me! :)

cgoldberg:
let's leave this for another day or 2 to see if any other maintainers comment, then I'll merge it.

cgoldberg:
looks like consensus for this PR is a 👍 ... merging now.

cgoldberg:
thanks very much for the contribution!

Ramshell:
My pleasure!

Issue #849, opened 2018-07-24
Provide an official Docker image


JDiPierro:
It'd be great if Locust provided an official Docker image on Dockerhub. Right now the most used image is [hakobera/locust](https://hub.docker.com/r/hakobera/locust/) but it hasn't been updated in 3 years.

JDiPierro:
The image would only need to contain locust and pyzmq. I agree it's a simple image and the hakobera/locust Dockerfile is pretty good, it just needs to be rebuilt with a newer version of Locust; right now it's missing a bunch of features. It is set up as an automatic build so any github repo activity would trigger a new build but that's tied to the [hakobera/docker-locust](https://github.com/hakobera/docker-locust) repo so it's only triggered when the docker setup gets changed and like you said: it's a simple image, the config doesn't need to be updated often. And the dev (ping, @hakobera) seems to be gone from Github, no activity in over a year.

cgoldberg:
We really need someone to volunteer to do this.  A Pull Request with an updated dockerfile and some documentation would be a great start :) (hint)  We would also want to update the travis config so we can build/test it regularly.

colutti:
I just created one with the latest version of locust as well as python 3.7.

https://store.docker.com/community/images/colutti/locust

I will work on the its documentation

cgoldberg:
@colutti see #850

spayeur207:
@cgoldberg I think this can now be closed

sekka1:
Can we get an official image up to Docker hub or another repository that this project maintains?  I have this PR for a Helm Chart to use Locust in Kubernetes but I am pointing it to an image in my Docker Hub repo.  I would rather point it to an "official image".

https://github.com/helm/charts/pull/7970/files#diff-c9789cc33518b31f9230b6284b2b6116R4

sekka1:
ping ping

I can help but not sure what would need to be done in order to get a Locust image into Docker Hub.

raiyankamal:
@sekka1 As far as I know, there is no official docker hub account for locustio. Creating an official account seems to be the obvious first step. Next, the maintainer of that account should upload the docker image. Ideally, a new docker image should be uploaded for each release of Locust.

In future, several images may be uploaded for each version e.g. based on different versions of ubuntu and alpine linux.

sekka1:
@raiyankamal I am happy to help here but not sure how.  If you point me in the right direction i can take something on to help make this happen.

swarupdonepudi:
[...]

PR #850, opened 2018-07-25
Official Docker image and documentation

JDiPierro:
For #849 :)

Adds a basic `Dockerfile` that starts from `python:3.6` and a includes custom startup script that allows you to start in any mode.

Includes a first attempt at documentation.

I added a Travis stage to build the image but for releasing I think the better approach is for someone to create a DockerHub account for locustio and set up an [automated build](https://docs.docker.com/docker-hub/builds/) for the repository. I believe they do security scanning on the resulting image as well.

Let me know if there's any docs changes, extra features, etc you'd like to see or if I can help with the DockerHub setup. I'm hanging around the Slack server as @JDiPierro as well :D

### DIFF
--- a/docker_start.sh
+++ b/docker_start.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [ -z "${TARGET_URL}" ]; then
+  echo "ERROR: TARGET_URL not configured" >&2
+  exit 1
+fi
+
+LOCUST_MODE="${LOCUST_MODE:=standalone}"
+LOCUST_OPTS="-f ${LOCUSTFILE_PATH:-/locustfile.py} -H ${TARGET_URL}"
+
+if [ "${LOCUST_MODE}" = "master" ]; then
+    LOCUST_OPTS="${LOCUST_OPTS} --master"
+elif [ "${LOCUST_MODE}" = "slave" ]; then
+    if [ -z "${LOCUST_MASTER_HOST}" ]; then
+        echo "ERROR: MASTER_HOST is empty. Slave mode requires a master" >&2
+        exit 1
+    fi
+
+    LOCUST_OPTS="${LOCUST_OPTS} --slave --master-host=${LOCUST_MASTER_HOST} --master-port=${LOCUST_MASTER_PORT:-5557}"
+fi
+
+echo "Starting Locust..."
+echo "$ locust ${LOCUST_OPTS}"
+
+locust ${LOCUST_OPTS}
--- a/docs/running-locust-docker.rst
+++ b/docs/running-locust-docker.rst
@@ -41,17 +41,3 @@ ADD locustfile.py locustfile.py
 You'll need to push the built image to a Docker repository such as Dockerhub, AWS ECR, or GCR in order for
 distributed infrastructure to be able to pull the image. See your chosen repository's documentation on how
 to authenticate with the repository to pull the image.
-
-
-Running in Docker Compose
----------------------------------------------
-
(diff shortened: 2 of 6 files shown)


cgoldberg:
thanks for this!  
I'd like a few maintainers to give this a review and try it out, so we are able to maintain it and answer questions that come up.
@aldenpeterson-wf @mbeacom can you guys give this a look and try it out when you have a chance?

heyman:
Hey! I think having an official Docker image for Locust would be great.

Do we really need the `docker_start.sh` shell-script? It introduces a new API and interface that you have to "learn" to use Locust with Docker. Couldn't we just set the Docker entrypoint to the locust binary?

docker_start.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [ -z "${TARGET_URL}" ]; then
```
mbeacom: While I agree we should handle env vars for all of these settings, I think we should handle these at the app leveland simply take whatever the `$@` entrypoint is, defaulting to `locust`.  This allows the user to modify their entry as necessary and simply pass the envvars to be used.

docs/running-locust-docker.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Add your tests
+---------------------------------------------
+
+The easiest way to get your tests running is to build an image with your test file built in. Once you've 
+written your locustfile you can bake it into a Docker image with a simple ``Dockerfile``:
+
+```
+FROM locustio/locust
+ADD locustfile.py locustfile.py
```
mbeacom: Maybe we should discourage the usage of `ADD` for the user's `locustfile.py` and document mounting a directory with the `locustfile.py` instead.

examples/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3.4"
```
mbeacom: I'm all for using the latest `docker-compose` syntax version, but maybe we could pin at `3` to increase availability/adoption.

sekka1:
I have tried the docker image as is in this PR branch and it works great!  There was a Locust Helm chart in the stable repository that hasn't been updated in about a year.  I made all of the changes necessary to do the update to v0.9.0 and also added a bunch of other Kubernetes items in there to make it work a little nicer like nodeSelectors/Taints and an Ingress.  

I have this PR for the Helm Chart update but I should probably wait for an official image instead of using the image in my Docker Hub repo.  Here is the updated Chart: https://github.com/helm/charts/compare/master...sekka1:with-v0.9.0?expand=1#diff-c9789cc33518b31f9230b6284b2b6116R4

The last response to the updates was about a month ago.  I'm happy to continue on if @JDiPierro is too busy now.  

Thoughs @mbeacom

cgoldberg:
> I'm happy to continue on if @JDiPierro is too busy now.

@sekka1 this PR was superseded by #882 , which has already been merged... so Docker stuff is in master branch now.   Could you give it a try?

I am going to close this one now.

sekka1:
@cgoldberg Is there any plans for an official Locust Docker hub image?

quorak:
It would be great to release the image to an official account. This account looks like scam https://hub.docker.com/r/locustio/locust , though not sure if it actually is.

https://hub.docker.com/r/_/locust would be great !

swarupdonepudi:
Is this docker image built and pushed to a public registry yet?

swarupdonepudi:
> Is this docker image built and pushed to a public registry yet?

Nevermind. There is an ongoing discussion around this topic - https://github.com/locustio/locust/issues/849

PR #856, opened 2018-07-30
Specify seconds for the `timeout` parameter

hoylemd:
Tiny doc change - The Requests module's docstrings mention that `timeout` should be in seconds, but it's unclear here.


aldenpeterson-wf:
+1

Issue #848, opened 2018-07-20
Tests fail to run when providing a small number of clients


saif-momin:
Hello,

My current setup in locustfile.py currently looks something like this:
```python
class VolumeTests(object):
      host = 'http://example.com'
      min_wait = 5000
      max_wait = 15000

class ExampleTest(VolumeTests, HttpLocust):
      task_set = MyTaskSet
```
I have 7 of those ExampleTests at the moment, all different types of tests.
In the command line, when I run `locust -c 3 --no-web`, the testing quits and I get the following message:
"Hatching and swarming 0 clients at the rate 1 clients/s..."

However, when I run `locust -c 4 --no-web`, changing the number of clients from 3 to 4, the tests run successfully. 

After some experimenting, it seems like the number of clients you provide must be at least half of the number of tests (i.e. ExampleTest).

Is there something in the locust documentation that I am missing? Or is there something that I am doing wrong? Thanks.

plioi:
+1 What's the relationship between number of clients and number of tasks/task sets? I find myself having to essentially guess a big enough number as my test suite grows.

PR #860, opened 2018-08-05
Fix uneven distribution of locust classes between running locusts

Umqra:
This PR fix bug related to the issue: #848 
### Bug description:
There was a function responsible for calculating the locust classes that will be spawned/killed after startup/changing of configuration. The algorithm for spawning `n` locusts work as following:
1. Calculate percentage `p_i` for each locust class based on the class weight
2. Calculate the number of spawned locusts per class by formula: `round(p_i * n)`

Of course, this algorithm doesn't guarantee that exactly `n` locusts will be spawned because of rounding. This behavior can cause small, but annoying bugs (one such bug described in the related issue)

### Pull request description:
This pull request introduces simple and separate unit `LocustsCollection`, that is responsible for even distribution of locust classes between current running locust executors. This class implements a simple algorithm that maintains next invariant: 
> For each locust class the amount of running locust executors differs by no more than one from the value `p_i * n`.

Also, this PR includes small refactorings and little improvements in the `runners.py` file and tests.


LukeCarrier:
@Umqra thanks for this -- I just identified the same problem with one of our suites and you've saved me a of debugging!

locust/runners.py
```diff
@@ -61,65 +63,54 @@ def errors(self):
...
             if not locust.task_set:
-                warnings.warn("Notice: Found Locust class (%s) got no task_set. Skipping..." % locust.__name__)
-                continue
+                warnings.warn("Notice: Found Locust class (%s.%s) got no task_set. Skipping..." % (locust.__module__, locust.__name__))
+            else:
+                true_locust_classes.append(locust)
+        return true_locust_classes
 
+    def update_locusts_settings(self, locusts, stop_timeout=None):
+        for locust in self.locusts:
```
LukeCarrier: Should this be `locusts` rather than `self.locusts`? Without this change I get a `LocustError` here when assigning the base URL via the `--host` option.

Umqra: Yeah, this is a bug. I fix it in the recent commit.

Umqra:
@LukeCarrier Are there anything else that I should do to merge this pull request?

LukeCarrier:
@Umqra I'm just another Locust user I'm afraid -- I'm not knowledgeable enough about Locust's internals to give a quality review. Hopefully a maintainer will check in soon.

cyberw:
Hi @umqra ! First of all, sorry for the lack of response from maintainers. 

@heyman, do you think this makes sense? The main downside I can think of is that it increases complexity a little, but it does solve a real issue.

heyman:
I'm not sure about this one. I think the problem should be very uncommon in real world scenarios. To me it doesn't sound like you're doing a *load test* if you're affected by rounding errors in the number of spawned  users.

Still, the behaviour of this PR is better than the  current behaviour, but I'm not sure it justifies the extra code and complexity. I wonder if it wouldn't be possible to instead add some extra logic to `LocustRunner.weight_locusts()` and achieve the same thing?

heyman:
I took a stab at it in #1113.

cyberw:
Closing this in favour of #1113

PR #875, opened 2018-09-05
Fix Sphinx build warnings

jdufresne:
WARNING: Title underline too short.
WARNING: Title overline too short.


cgoldberg:
thanks!

PR #882, opened 2018-09-13
Official Docker image and documentation V2

spayeur207:
After talking with @JDiPierro I'm taking over this initiative. However I can't actually interact with the old PR so this is being created. His original work has been rebased onto latest master.

Referencing #849 

Adds a basic `Dockerfile` with a base of `python:3.7-alpine` putting final image size at 291MB. 
I believe most if not all of the changes requested by @mbeacom have been addressed in this as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3.6
+
+RUN pip install locustio pyzmq
+
+ADD docker_start.sh /usr/bin/docker_start
+RUN chmod 755 /usr/bin/docker_start
+
+EXPOSE 8089 5557 5558
+
+CMD ["docker_start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM python:3.6-alpine
+FROM python:3.6.6-alpine3.8
 
-RUN apk --no-cache add g++
-RUN pip install locustio pyzmq
+RUN apk --no-cache add g++ \ 
+      && pip install locustio pyzmq
 
 EXPOSE 8089 5557 5558
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.6-alpine
 
 RUN apk --no-cache add g++
 RUN pip install locustio pyzmq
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6-alpine
```
mcdoker18: > Adds a basic Dockerfile with a base of python:3.7-alpine

I see the 3.6 python.
Anyway I think is good idea to specify full version. For example: `3.6.6-alpine3.8`

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+RUN apk --no-cache add g++
```
mcdoker18: Please, group the build dependencies and delete their at the end of Dockerfile

spayeur207: Grouping build dependencies yes. Deleting them I'm not sold on. 

Yes it shrinks the image by quite a bit, but it also reduces usability by others. This isn't ever supposed to be "production" grade simply due to it being a testing lib. The extra ~140MB of g++ and associated packages seems a reasonable price for facilitating the use of Locust for an audience that isn't as familiar with Docker or Alpine. Hell by that measure I should probably toss in bash and eat the extra 40MB.

Dockerfile
spayeur207: Good point. Updated in 974f861

Dockerfile
xoffey: I appreciate @spayeur207's comment, but cutting the size in half would be great for those who would like to widely distribute workers. Maybe you could make it delete the build-only items, and add a note on how devs could comment-out that behavior. Then it would be nice for non-devs and (arguably) still good enough for advanced devs.

quorak:
unfortunately there is no Docker image under https://hub.docker.com/r/locustio/locust as described in the docs.
I would also suggest to use the official branch https://docs.docker.com/docker-hub/official_images/

SBSolera:
Please add an official image for locustio on docker hub!

heyman:
I'm thinking of removing the `docker_start.sh` shell script for Locust 1.0, in favor of just having `locust` as the Docker CMD. If anyone sees any problem with that please let me know in #1247.

Issue #145, opened 2014-04-25
nitpick: "# requests" should be "# successful requests"?


sanga:
in the main webui, there is "# requests" and "# fails". A colleague of mine asked me "how is it that the # of fails can be greater than the number of requests?". He, I guess assumed # requests => total number of requests, which seems like a kinda natural assumption to me. As it is, it's perhaps _slightly_ misleading or could be clearer. For eg HTTP 500 is counted as a fail, whereas it is a request, just not a successful request.

=> "# successes"? or "# successful requests?"

Or maybe it's just late Friday afternoon and this is clear to everyone else...

rootAvish:
@heyman I ran into the same confusion as well, after a good 4 years since the inception of this issue. Have you guys decided against changing this?

manibharathitn:
So, for now, its logically "# success"

cgoldberg:
total = #requests + #fails

cyberw:
I think this was fixed in #939

PR #884, opened 2018-09-16
Change # requests to # successes

cuongtranx:
Fixing what was discussed here: https://github.com/locustio/locust/issues/145

Also, full transparency, I use a white-space removal inside my text-editor. So if you don't want to see the white-space diff, either use Github UI, or add a `w=1` query param into the PR's URL

### DIFF
--- a/docs/retrieving-stats.rst
+++ b/docs/retrieving-stats.rst
@@ -4,7 +4,7 @@ Retrieve test statistics in CSV format
 
 You may wish to consume your Locust results via a CSV file. In this case, there are two ways to do this.
 
-First, when running Locust with the web UI, you can retrieve CSV files under the Download Data tab. 
+First, when running Locust with the web UI, you can retrieve CSV files under the Download Data tab.
 
 Secondly, you can run Locust with a flag which will periodically save two CSV files. This is particularly useful
 if you plan on running Locust in an automated way with the ``--no-web`` flag:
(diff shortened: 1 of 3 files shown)


cgoldberg:
> So if you don't want to see the white-space diff,

Hiding whitespace changes to make review easier is helpful... but they still exist in the diff.  To make this PR less intrusive, can you remove them?

cuongtranx:
I added back the whitespace for the sake of reviewing this PR, CI passed for some reasons :/

cgoldberg:
thanks.

I'm -1 on this...because I think "successes" is the wrong metric to display.  I'd prefer we changed it to display total requests (in this case, the label could remain "requests").   Total request throughput seems more useful than success-only throughput.

anyone else have thoughts?

cuongtranx:
The current way of calculating/showing number of successful requests (even if it's wrong) hasn't been changed for the last 4 years. I think it's fair to update the label to reflect the correct behaviour so new users don't get confused over this.

Besides, updating the label is trivial, while changing the whole successes to total requests is not.

sanga:
The UI is currently demonstrably wrong. I agree that it'd be nicer to show the total request metric rather than just the number of successes, but that shouldn't stop this change going in. This change stops the UI from being incorrect. A future PR could improve the UI to be more helpful (show total requests)

SpencerPinegar:
As someone using Locust my use case is I need to be able handle x amount of requests with a failure rate less than y. While I feel the locust UI should not have incorrect UI (if it's an easy fix I would like it fixed) but the total amount of requests has more practical significance when running performance tests.

raiyankamal:
I agree with @sanga that this PR will correct the UI and a future PR should come to correctly calculate # requests. In that case we may end up with two metrics, one of which is actually a derived value. In that sense it can be considered redundant.

@cgoldberg is your objection due to redundancy of the # success as a metric or due to effort not going toward solving the underlying issue i.e. how # requests is calculated.

cgoldberg:
@raiyankamal I'd like to see this PR amended or superseded with a proper fix instead of simply changing the UI.  It's pretty low effort to fix.

raiyankamal:
@cgoldberg I created a new PR to superseed this one. The logic for counting # requests is corrected instead of the UI. Please check https://github.com/locustio/locust/pull/939

PR #895, opened 2018-09-27
Use uuid4 to generate slave client_id

### LINKED ISSUE #894: slave client_id collisions in large environments
I'm running 64 locust slave processes per host, and getting client_id collisions regularly. 

https://github.com/locustio/locust/blob/4bde2ca90f28bde649ff4ce2c86e36a6e6260856/locust/runners.py#L355

The randomint here probably _increases_ the chances of collisions, since it's added to a millisecond scale time function.

Recommend using `uuid4()` or `uuid5(namespace, name)` using the hostname as a namespace.

mattbailey:
Fixes #894 

Current implementation collides easily in environments with large number of slave processes on the same host starting at the same time.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -4,7 +4,7 @@ import random
 import socket
 import traceback
 import warnings
-from hashlib import md5
+from uuid import uuid4
 from time import time
 
 import gevent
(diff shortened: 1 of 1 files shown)


cgoldberg:
@mattbailey thanks for contributing to Locust!
:+1:

PR #899, opened 2018-10-06
fix #145

### LINKED ISSUE #145: nitpick: "# requests" should be "# successful requests"?

sanga:
Change web UI to correctly show "# successes" as opposed to "# requests"

### DIFF
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -104,7 +104,7 @@
                             <tr>
                                 <th class="stats_label" href="#" data-sortkey="method">Type</th>
                                 <th class="stats_label" href="#" data-sortkey="name">Name</th>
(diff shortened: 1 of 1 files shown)


cgoldberg:
see #884

PR #910, opened 2018-11-07
Fix Dockerfile add docker_start.sh ENTRYPOINT

blackstorm:


spayeur207:
This is a fairly stark departure from the existing Dockerfile. Is there a reason you want to change the base image and add the docker_start.sh back as the entrypoint?

blackstorm:
> This is a fairly stark departure from the existing Dockerfile. Is there a reason you want to change the base image and add the docker_start.sh back as the entrypoint?

@spayeur207 

The cause is that the ```docker-compose.yml``` in the ```example``` folder can't work.

### About Dockerfile change: 
Add docker_start.sh because ```/usr/local/bin/locust``` can't get ```TARGET_URL``` or ```LOCUSTFILE_PATH``` environment 

And  i changed the base image to ```python:3.6```. Because ```python:3.6.6-alpine3.8``` image ```bash``` environment not supported

My English is poor.

PR #925, opened 2018-12-04
clarifying locust class usage when no classes specified on CLI

smadness:
Based on a simple test, it looks like when no locust class is specified on CLI execution, spawns pull from the full set of defined locusts.  This PR just clarifies that behavior.


aldenpeterson-wf:
👍

PR #927, opened 2018-12-04
Add heartbeat to detect down slaves

Jonnymcc:
I have noticed some issues bringing up master and slave nodes within Kubernetes. I noticed in some other github issues for Locust that some of the issues may be solved by adding a heartbeat between the nodes.

The changes include the addition of a heartbeat thread on the master. The slaves check in with the master by means of their own heartbeat worker thread. There is also a new state (missing) that will show in the UI if a slave fails to check in with the heartbeat to the master.

I also think that it prevents some stale state by having the slaves also send their current state to the master. Reason being that if a slave fails to check in and goes missing, how else would the master know what the slave is doing after it continues to check in.

I got my inspiration from the ZMQ guide, specifically this pattern...
http://zguide.zeromq.org/page:all#Robust-Reliable-Queuing-Paranoid-Pirate-Pattern

Thoughts, comments?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [run]
 branch = True
 source = locust
+concurrency = gevent
 
 [report]
 exclude_lines =
--- a/locust/main.py
+++ b/locust/main.py
@@ -128,6 +128,20 @@ def parse_options():
     )
 
     parser.add_option(
+        '--heartbeat-liveness',
+        action='store',
+        type='int',
+        help="set number of seconds before failed heartbeat from slave"
+    )
+
+    parser.add_option(
+        '--heartbeat-interval',
+        action='store',
+        type='int',
+        help="set number of seconds delay between slave heartbeats to master"
+    )
+
+    parser.add_option(
         '--expect-slaves',
         action='store',
         type='int',
--- a/locust/main.py
+++ b/locust/main.py
@@ -131,6 +131,7 @@ def parse_options():
         '--heartbeat-liveness',
         action='store',
         type='int',
+        default=3,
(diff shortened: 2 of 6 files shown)


myzhan:
Hi, @Jonnymcc 

It's a very useful feature. So far, the master can not detect slaves that exit unexpectedly.

But, can you keep compatible with previous versions of locust and [third party tools](https://docs.locust.io/en/stable/third-party-tools.html)?

What about binding to another port?

Jonnymcc:
@myzhan I think we can support previous versions. Let me know if I understand what is needed though.

1. Removing the second port(master_port + 1) is not important.  If it is important (collecting stats from slaves) why can't the master bind port be used?
2. The issue stems from the changes to the locust zmq server/client, specifically the addition of the multipart send and receive.
3. Supporting third party tools can be accomplished by maintaining the same send and receive methods of the zmq server/client and returning the Message type and not returning the multiple frames (a list).

By compatible with previous versions are you implying running a master/client with a master/client of another version? That sounds like asking for bugs regardless of the changes made in the PR.

cgoldberg:
@Jonnymcc  :+1:  this is really nice!
sorry for taking so long to review the PR... I just merged it to master.
thanks for the contribution to Locust!

mangatmodi:
@cgoldberg Any plans for release? I checked last release was in September, 2018. This is quite a major fix, do you think we should have a release now?

aldenpeterson-wf:
@mangatmodi this is now released 🎉

tsykora-verimatrix:
I guess the fix to stop the test after hitting the stop button in the UI did not make it to the 0.11.0 release

PR #935, opened 2018-12-18
Speed up task sequence tests

Jonnymcc:
The default min and max wait time between tasks result in these two tests taking a combined ~6 seconds of run time. With these new values the total run time for the entire test suite is reduced by about 68%.


aldenpeterson-wf:
LGTM

PR #936, opened 2018-12-18
Remove repeated imports of mock

Jonnymcc:
I am not sure why mock is imported multiple times but the tests pass even without them. If there is a good reason to keep them then this could be closed. Otherwise for readability I think they should be removed.


aldenpeterson-wf:
I'm fine with this, I'd guess what happened was:

1. A single test got written with the inline import
2. That test got copied everywhere
3. Someone realized "hey we can global import"
4. Step 2 never got cleaned up

cgoldberg:
LGTM

PR #938, opened 2018-12-28
Support horizontal scrolling for the stats table

mingrammer:
I can't see the last column in the stat table when the name is long. So I think locust web should support the horizontal scrolling on the stats table.

### DIFF
--- a/locust/static/style.css
+++ b/locust/static/style.css
@@ -258,7 +258,9 @@ a:hover {
 }
 
 .status table {
+    display: block;
     width: 100%;
+    overflow-x: auto;
     border-radius: 3px;
     border: 2px solid #11251c;
     border-spacing: 0;


myzhan:
This change makes the stats table look like this, it's weird. Can we make it look better?

image

cgoldberg:
@myzhan can you PR an improvement?  I'd like to get it looking better.

I also don't like the Type column

myzhan:
I'm trying to revert the previous change, and added this.

```css
#stats {
    table-layout: fixed;
}

#stats td {
    word-wrap: break-word;
}
```

Long text will be broke into multiple lines and you can use browser's window scrolling to see the whole table.

shot

What do you think? @cgoldberg @mingrammer

myzhan:
And @cgoldberg , I can commit to the same branch if you are willing to merge https://github.com/locustio/locust/pull/945

Or I will checkout a new branch.

mingrammer:
@myzhan I agree! That would be better.

cgoldberg:
create a new PR for the changes and we can review there

myzhan:
@cgoldberg please have a look at https://github.com/locustio/locust/pull/962

Issue #940, opened 2019-01-04
filenames with several dots fails


mdtrooper:
### Description of issue / feature request

filenames with several dots fails

### Expected behavior

Runs fine

### Actual behavior

Fails

### Environment settings (for bug reports)

```
unuser@miguel-pc:/$ uname -a
Linux miguel-pc 4.15.0-43-generic #46~16.04.1-Ubuntu SMP Fri Dec 7 13:31:08 UTC 2018 x86_64 GNU/Linux
unuser@miguel-pc:/$ python --version
Python 3.5.6
unuser@miguel-pc:/$ pip list | grep locustio
locustio     0.9.0     
unuser@miguel-pc:/$ 
```

### Steps to reproduce (for bug reports)

```
unuser@miguel-pc:/$ locust -f /tests/login_and_minimal_navigation.test.py -H http://127.0.0.1:10000 --no-web -c 10 -r 2 -t 30s --only-summary
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: Traceback (most recent call last):
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: File "/usr/local/bin/locust", line 11, in <module>
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: 
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: sys.exit(main())
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: 
[2019-01-04 09:32:05,137] miguel-pc/ERROR/stderr: File "/usr/local/lib/python3.5/site-packages/locust/main.py", line 391, in main
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: 
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: docstring, locusts = load_locustfile(locustfile)
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: 
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: File "/usr/local/lib/python3.5/site-packages/locust/main.py", line 358, in load_locustfile
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: 
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: imported = __import__(os.path.splitext(locustfile)[0])
[2019-01-04 09:32:05,138] miguel-pc/ERROR/stderr: 
[...]

PR #941, opened 2019-01-07
Allow loading of a locustfile with multiple dots in filename

raiyankamal:
Fix for https://github.com/locustio/locust/issues/940
Uses methods from the `imp` module to allow loading of files with multiple dots in the filename.


raiyankamal:
Closing because incomplete.

raiyankamal:
Reopening PR after adding unit tests and correcting code for loading locustfile.

PR #960, opened 2019-02-14
Update change logs & release 0.10.0

aldenpeterson-wf:
Should be more straightforward this time.

The auto changelog reordered some things but left them all present.

@cgbystrom @heyman @cgoldberg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,76 @@
 # Change Log
 
+## [Unreleased](https://github.com/Locustio/Locust/tree/HEAD)
 
-## [v0.9.0](https://github.com/Locustio/Locust/tree/v0.9.0) (2018-08-20)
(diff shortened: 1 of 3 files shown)


jmirc:
When do you plan to release

aldenpeterson-wf:
@jmirc sometime in 2019, most likely.

jmirc:
We are using locust with kubernetes and we are interesting by the "Add heartbeat to detect down slaves" feature. Do you have a doc to create a release on our side? We could be able to push it inside a private docker repository

aldenpeterson-wf:
My comment was perhaps a bit sarcastic but realistically, once someone with pypi push credentials reviews this and tags as needed  (I don't have these creds or I would).

Release process is documented here - https://github.com/locustio/locust/blob/master/.github/CONTRIBUTING.md

PR #968, opened 2019-02-27
updated the client listener

IgalKorb:
now it checks that locusts started spawning (user_count > 0) and checks when they all finished hence the execution can quit (user_count == 0)


cyberw:
Hi @IgalKorb ! Maybe I am confused, but what problem does this PR solve?

PR #970, opened 2019-03-01
balance/recover the load distribution when new slave joins

delulu:
With Locust master and slave agents running in Kubernetes, Kubernetes will guarantee the availability of Locust agents. 

But when a slave agent crashes and restarts, it will have a different client id and it has no idea of the user load that master assigned to it previously. And the total number of running locusts will not be as many as expected. 

So it might be a better way to balance the user load when new client joins, and the total number of running locusts will still be the same as we specified in the swarm request. 

also this PR fixes some issue I noticed when running in Python 3 with web mode, it turns out to be the inconsistency introduced in recv_from_client and send_to_client

Any thought or comment?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.6.6-alpine3.8
 
 RUN apk --no-cache add g++ \ 
+      && apk --no-cache add zeromq-dev \
       && pip install locustio pyzmq
 
 EXPOSE 8089 5557 5558
--- a/locust/main.py
+++ b/locust/main.py
@@ -132,6 +132,7 @@ def parse_options():
         '--heartbeat-liveness',
         action='store',
         type='int',
+        dest='heartbeat_liveness',
         default=3,
         help="set number of seconds before failed heartbeat from slave"
     )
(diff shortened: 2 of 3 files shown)


delulu:
@Jonnymcc for awareness

cgoldberg:
> also this PR fixes some issue I noticed when running in Python 3 with web mode

can you move those to a separate PR?

locust/test/test_zmqrpc.py
```diff
@@ -17,12 +17,12 @@ def tearDown(self):
...
         self.client.send(Message('test', 'message', 'identity'))
         addr, msg = self.server.recv_from_client()
-        self.assertEqual(addr, b'identity')
+        self.assertEqual(addr, 'identity')
         self.assertEqual(msg.type, 'test')
         self.assertEqual(msg.data, 'message')
 
     def test_client_recv(self):
-        sleep(0.01)
+        sleep(0.1)
```
Jonnymcc: Was the sleep not long enough?

delulu: no, not long enough at my side, besides there's no harm to set a longer time here.

delulu:
> 
> 
> > also this PR fixes some issue I noticed when running in Python 3 with web mode
> 
> can you move those to a separate PR?

sure, here is the separate PR: https://github.com/locustio/locust/pull/972

delulu:
@cgoldberg please help check this pr and merge into master, let me know if any concern, thx!

PR #973, opened 2019-03-05
add retry in zmqrpc

delulu:
When I'm running Locust in distributed mode, I've noticed an exception thrown from slave agent "AssertionError: Only one greenlet can be waiting on this event".

On investigating this issue, I find out that there're multiple greenlets in SlaveLocustRunner as shown in https://github.com/locustio/locust/blob/87a9aa1cb937c68e4456cafd23f3c42f25023931/locust/runners.py#L384-L388

and they're talking to Locust master with the same zmq socket, so there might be concurrent calls to self.client.send, then the exception above will be thrown.

We can leverage semaphore to avoid the concurrent calls, but there will be a performance cost. Since these concurrent situations are not very likely, it's better to simply add a retry for socket communication and makes Locust more robust.

### DIFF
--- a/locust/rpc/zmqrpc.py
+++ b/locust/rpc/zmqrpc.py
@@ -1,24 +1,28 @@
 import zmq.green as zmq
 
 from .protocol import Message
-
+from locust.util.exception_handler import retry
 
 class BaseSocket(object):
     def __init__(self, sock_type):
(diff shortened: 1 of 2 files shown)


delulu:
@cgoldberg for awareness, let me know if any concern.

Issue #889, opened 2018-09-19
Strange behavior of "Total Requests per Second" chart


hoangphucITJP:
### Description of issue

Total Requests per Second's RPS drop to below 1 after the test start a while despite the RPS number on top keep rising

### Expected behavior

The RPS of Total Requests per Second chart  should keep updating according to the RPS number on top

### Actual behavior

The RPS of Total Requests per Second chart value does not match the RPS number on top after the test start a while

### Environment settings (for bug reports)

- OS: Ubuntu 16.04.3 LTS (GNU/Linux 4.4.0-78-generic x86_64)
- Python version: 2.7.12
- Locust version: 0.9.0

### Steps to reproduce (for bug reports)

I don't know what to put here but feel free to ask for any extra information. I can't provide the code though

### Screenshot

![imgpsh_fullsize](https://user-images.githubusercontent.com/14938527/45725860-f873a200-bbe6-11e8-8fcf-3487d0afe8cc.png)

P/s: Sorry for ignored the template.

klloveall:
I have also encountered this issue. I dug into the responses and it's happening because the `/stats/request` endpoint stops returning the `"Total"` stat block at the end of the `stats` list. This then causes the graph to show a RPS from a random path instead of from the total like it should. Interestingly, as is shown in the screenshot from @hoangphucITJP, the RPS in the header is still correct.

I tried to dig into the Python code and figure out what's causing it there, but haven't had much success there yet (mostly b/c I'm only running this in AWS, not locally so my debugging options are somewhat limited).

Please let me know if there's any suggestions as to how to solve this or any other debugging information I can provide.

supermasita:
@cgoldberg I think my simple change should work as a fix (it has solved the issue for me). Thanks.

supermasita:
@hoangphucITJP @klloveall - let me know if manually applying my change fixes your issue. I think my PR is correct, but it is not building for unrelated reasons. Thanks.

sweetib:
we are also facing this issue, when is the PR supposed to be merged?

victor61zhong:
any update for this issue ?

canstudios-simonm:
I also see the RPS drop off as shown in the graph every time we run Locust. It makes tracking the average RPS value over a run very hard without it working.

cyberw:
I cant reproduce this (on MacOS, using locust 0.12.1). Is this only an issue on Linux? Or is it something related to the test plan?

[...]

PR #977, opened 2019-03-09
Fix RPS graph

supermasita:
Use "report.total_rps" instead of "total.current_rps", to get the same value for RPS as displayed on header. "total.current_rps" is not always available.

#889

### DIFF
--- a/locust/static/locust.js
+++ b/locust/static/locust.js
@@ -153,7 +153,7 @@ function updateStats() {
             // get total stats row
             var total = report.stats[report.stats.length-1];
             // update charts
-            rpsChart.addValue([total.current_rps]);
+            rpsChart.addValue([report.total_rps]);
             responseTimeChart.addValue([report.current_response_time_percentile_50, report.current_response_time_percentile_95]);
             usersChart.addValue([report.user_count]);
         }


cgoldberg:
Now that I look at this, I question if "Total RPS" is really what we want to display?  I think the moving average like the stats display is more useful.  thoughts?

supermasita:
I think the graph timeline with the RPS is valid and needed. **I always want to match the RPS to what the servers and apps where doing while handling them** - compare RPS timeline VS server metrics, for example.

This is how it looks with the change applied - what you would expect to see:


We can still think about new graphs or improvements to the current, but I wouldn't change/remove this one.

Thanks!

cgoldberg:
> I think the graph timeline with the RPS is valid and needed

I wasn't suggesting removing RPS.  My point was that "Total RPS" is displaying RPS based on values since the beginning of the run, while the values on the Stats tab are using a moving average.  I think a moving average RPS is a better metric to watch.  Why would you want a graph of RPS value that is calculated from total data since the run began?

supermasita:
> My point was that "Total RPS" is displaying RPS based on values since the beginning of the run
What you are saying is that "report.total_rps" != "total.current_rps" ? Or they are the same but you are proposing to change the behavior of this graph? 

Im trying to understand if my fix its wrong or if we are using this opportunity to discuss a possible enhacement.

Thanks!

cgoldberg:
`report.total_rps != total.current_rps`
`current_rps` is based on most recent stats entry only.

supermasita:
Well, when I check "stats/requests" I get the same values on "total.current_rps" and in "report.total_rps", until a minute or so later, when "total.current_rps" disappears completely, as [pointed out](https://github.com/locustio/locust/issues/889#issuecomment-461299089) by @klloveall

 
```
    {
      "avg_content_length": 1472.1518578352182,
      "avg_response_time": 194.64605863722153,
      "current_rps": 137.0,
      "max_response_time": 1093.0635929107666,
      "median_response_time": 170,
      "method": null,
      "min_response_time": 4.691839218139648,
      "name": "Total",
      "num_failures": 0,
      "num_requests": 1238
    }
  ],
  "total_rps": 137.0,
  "user_count": 69
```

```
$ while true; do curl  http://192.168.0.179:8089/stats/requests --silent | grep Total -C 5 | grep rps ; sleep 1; done
      "current_rps": 0.0,
  "total_rps": 0.0,
      "current_rps": 0.0,
  "total_rps": 0.0,
      "current_rps": 0.0,
  "total_rps": 0.0,
      "current_rps": 0.0,
  "total_rps": 0.0,
      "current_rps": 135.0,
  "total_rps": 135.0,
      "current_rps": 134.75,
  "total_rps": 134.75,
      "current_rps": 134.75,
  "total_rps": 134.75,
      "current_rps": 142.16666666666666,
  "total_rps": 142.16666666666666,
      "current_rps": 142.16666666666666,
  "total_rps": 142.16666666666666,
      "current_rps": 142.5,
  "total_rps": 142.5,
      "current_rps": 142.5,
  "total_rps": 142.5,
      "current_rps": 142.1,
  "total_rps": 142.1,
      "current_rps": 142.1,
  "total_rps": 142.1,
      "current_rps": 157.1,
  "total_rps": 157.1,
      "current_rps": 157.1,
  "total_rps": 157.1,
^C
```

ivoilic:
If this change isn't going to happen it would be great to have an option to switch between the two. Currently the RPS is useless to me and others because of this issue.

cyberw:
I agree with @cgoldberg on this. We should report and graph the current rps, not the rps over the whole test. We should fix the calculation of current_rps instead.

Applying this PR would just hide the problem, and add confusion.

Just to be clear, I do agree with you that the current situation is really bad (I dont use the web ui so I havent seen this myself), but it needs a different fix.

heyman:
The  code *does* use the current RPS. The`total` var in the locust.js file (https://github.com/locustio/locust/blob/master/locust/static/locust.js#L151-L154) refers to the summary row that includes all endpoints/URLs, and not to the `total_rps` which refers to the total average RPS since the start of the test. I do think it's a bit confusing that they are both use "total" but are referring to different things.

Issue #981, opened 2019-03-15
Unable to stop load from web UI with 0.11.0


solowalker27:
### Description of issue

Since upgrading to Locust 0.11.0 I often have difficulty stopping Locust from the web UI when running distributed. I click the Stop button, but the load doesn't actually stop. At best, it dips briefly but then resumes.

### Expected behavior

Load stops almost immediately.

### Actual behavior

It seems slaves stop their load and then rejoin the pool and then have load immediately sent to them again. Example output from master, showing repeated attempts to stop the load:

```[2019-03-15 22:01:51,128] 6f9c3d75aaee/INFO/locust.main: Starting web monitor at *:8089
[2019-03-15 22:01:51,129] 6f9c3d75aaee/INFO/locust.main: Starting Locust 0.11.0
[2019-03-15 22:02:45,732] 6f9c3d75aaee/INFO/locust.runners: Client 'e4b5991156b6_1163c8d3946a47a2954a3a19ef2ae9e0' reported as ready. Currently 1 clients ready to swarm.
[2019-03-15 22:02:45,863] 6f9c3d75aaee/INFO/locust.runners: Client '78a2ff78ec5f_6d6cced06e1140839ee19adbda4e68d3' reported as ready. Currently 2 clients ready to swarm.
[2019-03-15 22:02:45,901] 6f9c3d75aaee/INFO/locust.runners: Client '18a9ef01593d_74673a7f91a74b88ae64c796ef12acd9' reported as ready. Currently 3 clients ready to swarm.
[2019-03-15 22:02:45,932] 6f9c3d75aaee/INFO/locust.runners: Client 'eb8785101cf9_493118373e79416f8e5a76eb93b11c12' reported as ready. Currently 4 clients ready to swarm.
[2019-03-15 22:02:46,134] 6f9c3d75aaee/INFO/locust.runners: Client '5b213bf39f9d_43174c9cafcc415b98ded4e41fdff56e' reported as ready. Currently 5 clients ready to swarm.
[2019-03-15 22:02:46,146] 6f9c3d75aaee/INFO/locust.runners: Client '08e26b61419d_2dd8bd9aa5c745948c818b6b529c0df5' reported as ready. Currently 6 clients ready to swarm.
[2019-03-15 22:02:46,177] 6f9c3d75aaee/INFO/locust.runners: Client 'cc15e8e95ace_748c65bf793746048fbed3c2b613b694' reported as ready. Currently 7 clients ready to swarm.
[...]

PR #982, opened 2019-03-18
Add "stopping" state.

solowalker27:
Fixes https://github.com/locustio/locust/issues/981

### DIFF
--- a/locust/__init__.py
+++ b/locust/__init__.py
@@ -1,4 +1,4 @@
 from .core import HttpLocust, Locust, TaskSet, TaskSequence, task, seq_task
 from .exception import InterruptTaskSet, ResponseError, RescheduleTaskImmediately
 
-__version__ = "0.11.0"
+__version__ = "0.11.1"
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -23,7 +23,7 @@ logger = logging.getLogger(__name__)
 # global locust runner singleton
 locust_runner = None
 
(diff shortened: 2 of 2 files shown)


Eadword:
Looks good to me. Tested working on my setup.

mangatmodi:
I saw the code and it should work, but I am trying to think why this is introduced.

here - https://github.com/locustio/locust/blob/68ac9d3eef6d43f76451b53f00acc1a2a5ab9de9/locust/runners.py#L436
Client on being stop sends `client_ready` message. On receiving `client_ready` now we hatch the client
https://github.com/locustio/locust/blob/68ac9d3eef6d43f76451b53f00acc1a2a5ab9de9/locust/runners.py#L346
Which was added in recent commit - https://github.com/locustio/locust/commit/15499297cd04b90275d3ee4a59a1068ad739b53b

Client is now in `HATCHING` state - https://github.com/locustio/locust/blob/68ac9d3eef6d43f76451b53f00acc1a2a5ab9de9/locust/runners.py#L426

With a bit of bad luck(or enough number of runs), server might receive one of the client as hatching and hence the condition is always false - 
https://github.com/locustio/locust/blob/68ac9d3eef6d43f76451b53f00acc1a2a5ab9de9/locust/runners.py#L374

Adding  `STOPPING` works as it will avoid hatching the clients, no matter the status. It seems good enough, but I see the problem when a client is stopped, it shouldn't be ready.

mangatmodi:
I think it is ready because master can send another test request, After stopping one. That also means the heartbeat should continue even when client is stopped :)

wincus:
validated that this fixed the issue on our setup as well. Thanks!

tsykora-verimatrix:
glad to see this fixed, good job, is it going to be part of 0.12.0 or how it will be released?

Issue #989, opened 2019-04-03
Run locust with 1 concurrent users on 3 Locust clients exit without warnings/errors


ztou:
### Description of issue

when there are 3 locust clients, but run with only 1 concurrent Locust users, locust exit without error or warning.

https://github.com/locustio/locust/blob/8b17c286034178ba31c6d48b93f29202514f5659/locust/runners.py#L83

`num_locusts` could be 0

### Expected behavior

num_locusts should be at least 1

### Actual behavior

num_locusts is 0

### Steps to reproduce (for bug reports)

when there are 3 locust clients, but run with only 1 concurrent Locust users like below, locust exit without error or warning.

```
locust -f l.py --host=http://127.0.0.1:5000 --no-web -c 1 -r 1
```

l.py as below:

```
class T1(TaskSet):
    @task
    def test1(self):
        self.client.get("/test1")

class T2(TaskSet):
    @task
    def test2(self):
        self.client.get("/test2")

class T3(TaskSet):
    @task
    def test3(self):
        self.client.get("/test3")

class User1(HttpLocust):
    task_set = T1

class User2(HttpLocust):
    task_set = T2

class User3(HttpLocust):
    task_set = T3
```

ztou:
add a simple fix for your reference - https://github.com/locustio/locust/pull/990

delulu:
you haven't consider the situation where zero locust is spawn, so there's one test failure test_spawn_zero_locusts.

heyman:
I believe this was fixed by #1113

PR #990, opened 2019-04-03
Fix the issue that num_locusts could be 0

ztou:
When running locust with -c 1 on 3 locust clients, locust exit silently.

Not sure about the original design, so just add a simple fix to set
num_locusts to 1 when it was 0 on round()

ISSUE - https://github.com/locustio/locust/issues/989


cyberw:
@ztou If you fix the failing tests I might be able to merge this PR.

heyman:
I believe this was fixed by #1113

PR #992, opened 2019-04-06
Add command line argument to specify exit code on response errors

### LINKED ISSUE #560: Locust exits with 1 on timeouts / errors
Depending on how others are deploying this tool from the command line it may be a non-issue, but it's causing problems for us. Returning non-zero with the same code that we use when there are misconfigurations for, say, one 500 response code out of thousands is not subtle enough.

By the time we exit in this case, the entire test was already completed and requests were sent. Whatever downstream metrics systems will have valid data, etc. If we retry based on a non-zero exit code, we run the test twice. Users of the tool should probably be doing something more intelligent downstream to determine if the test was a failure or not than if there was at least 1 error. So why exit with an error code?

Two initial ideas that could improve the behavior (either or both would work):
1) Use a return code that's unique to this kind of "failure but not really", so it can be handled and ignored / logged with the correct criticality
2) Add a new command line parameter that is a threshold for failure (percent using stats.fail_ratio and/or absolute compared against num_requests)

Is this bothering anyone else?
([ref](https://github.com/locustio/locust/blob/v0.8a1/locust/main.py#L443))

Stateford:
Fixes #560. Give an option to disable it. It's probably too late to change the functionality to an opt-in, but exit codes should only post if there's actually a problem with the binary, not because there was an error in a test.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -282,12 +282,14 @@ def parse_options():
         help="show program's version number and exit"
     )
 
+    # set the exit code to post on errors
     parser.add_option(
-        '--exit-code-on-fail',
-        action='store_false',
-        dest='exit_code_on_fail',
-        default=True,
-        help="post exit code 1 on error"
+        '--exit-code-on-error',
+        action='store',
+        type="int",
+        dest='exit_code_on_error',
+        default=1,
+        help="sets the exit code to post on error"
     )
 
     # Finalize
@@ -561,8 +563,8 @@ def main():
         logger.info("Starting Locust %s" % version)
         main_greenlet.join()
         code = 0
-        if len(runners.locust_runner.errors) and options.exit_code_on_fail:
-            code = 1
+        if len(runners.locust_runner.errors):
+            code = options.exit_code_on_error
         shutdown(code=code)
     except KeyboardInterrupt as e:
         shutdown(0)
--- a/locust/main.py
+++ b/locust/main.py
@@ -282,6 +282,14 @@ def parse_options():
         help="show program's version number and exit"
     )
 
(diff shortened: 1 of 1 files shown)


cgoldberg:
I think it would be better if the option was truthy... so --exit-code-on-fail defaulting to True instead of --no-exit-code-on-fail.

Stateford:
I agree, that is a much better way to handle the argument.

locust/main.py
```diff
@@ -553,7 +561,7 @@ def sig_term_handler():
         logger.info("Starting Locust %s" % version)
         main_greenlet.join()
         code = 0
-        if len(runners.locust_runner.errors):
+        if len(runners.locust_runner.errors) and options.exit_code_on_fail:
```
aldenpeterson-wf: This may be confusing as running locust with `--exit-code-on-fail=2` will actually exit 1 right (it'll still pass this check)?

Stateford: Should the option be an int, and post what ever exit code is specified on error?

Default to 1 (current behavior), but allow the user to select which exit code to post on errors.

aldenpeterson-wf: I'm actually not sure. That seems to make sense though - you could give it `0` then for the situation you want here.

aldenpeterson-wf:
this LGTM but I want at least one more person's input on what is "intuitive" before declaring this more intuitive :)

Stateford:
Bump.

aldenpeterson-wf:
@cgoldberg @mbeacom what do you think on this?

cgoldberg:
:+1:  thanks!

Issue #675, opened 2017-11-02
Add support for downloading all failures to a CSV


hhowe29:
CSV should look about like the table in the Failures page.  Proposed columns

```
# Need these for sure
method
name
url
status_code
error_text

# these would be nice to have, but might be unwieldy to cram into a CSV
response_body  
request_headers
response_headers
```

sanyco92:
Any updates on this issue?

cgoldberg:
no PR for this has been submitted

gboorse:
Hi @hhowe29 check the issue I made #864 

There is a workaround there that might help you.

aldenpeterson-wf:
Addressed in #999

PR #999, opened 2019-04-14
Add failures csv download

ajt89:
#675 Download the "failures" tab from the web ui as a csv file


cgoldberg:
nice feature.  thanks!

Issue #891, opened 2018-09-21
TaskSet min_wait and max_wait are ignored


dblakemore:
### Description of issue / feature request

`min_wait` / `max_wait` overrides in `TaskSet` class not working correctly. When setting `min_wait` and `max_wait` values in a `TaskSet` class, the values seem to be ignored and the default values of 1000 (from the `Locust` class) are used instead.

I've done some digging into this and I _think_ I know where the problem lies. The `TaskSet`'s `wait_function` is automatically set to the locust's `wait_function` if undefined, but the locust's `wait_function` was defined as a lambda using its own `min_wait` and `max_wait` values. As such, the `TaskSet`'s `min_wait` and `max_wait` value have no effect.

You can get around this by assigning a new lambda`wait_function` to the `TaskSet`, but this defeats the purpose of being able to override the `min_wait` and `max_wait` values.

### Expected behavior

When setting a min_wait and max_wait value of 10000 in a TaskSet class, tasks for each Locust should be spawned 10 seconds apart

### Actual behavior

When setting a min_wait and max_wait value of 10000 in a TaskSet class, tasks for each Locust are still spawned 1 second apart (which is the default value)

### Environment settings (for bug reports)

- OS: Alpine Linux (docker image)
- Python version: 2.7
- Locust version: 0.9.0

### Steps to reproduce (for bug reports)

* Create a TaskSet class which inherits from the base TaskSet, with a min_value and max_value of 10000
* Create a single task on the TaskSet class
* Create a Locust class which inherits from the base HttpLocust, with its task_set property set to the TaskSet class created in step one
* Start swarming with 1 user. You would expect the task to spawn every 10 seconds, but it spawns every 1 second.

aldenpeterson-wf:
Ah, good catch, looks like https://github.com/locustio/locust/blob/30275af9f2acd7f01a5ea04382610b4f3c2f7daa/locust/core.py#L320 should also be:

   lambda self: random.randint(self.min_wait,self.max_wait) 

in both the TaskSet as well as Locust. 

@cgoldberg @heyman @mbeacom anyone have thoughts on this? Fixing it _will_ change behavior for anyone who has defined TaskSet min/max. I'm unsure if this is enough of an issue to warrant not fixing this though.

devmonkey22:
[...]

PR #1005, opened 2019-04-28
TaskSet min_wait and max_wait not ignored

ajt89:
#891


aldenpeterson-wf:
Like I said on the issue:

@cgoldberg @heyman @mbeacom anyone have thoughts on this? Fixing it will change behavior for anyone who has defined TaskSet min/max. I'm unsure if this is enough of an issue to warrant not fixing this though.


I'm _probably_ ok with this change. It's just something to be aware of.

cyberw:
Made unnecessary by #1118

PR #1013, opened 2019-05-16
Update setup.py

caapi:
updating few points in Setup.py file

---Updated By Santanu Panda


cgoldberg:
wth

genev:
@caapi, would you like to explain what you were trying to accomplish here?

PR #1016, opened 2019-05-22
Add grpc compatibility.

Youki:
Add grpc compatibility so that custom load tests work with grpc.

Found that locust can't work with grpc because gevent is not compatible with grpc. This pull request is to add grpc compatibility to custom locust test. More information can be found [here](https://github.com/grpc/grpc/issues/4629)

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -16,6 +16,9 @@ from six.moves import xrange
 # See: https://github.com/requests/requests/issues/3752#issuecomment-294608002
 monkey.patch_all()
 
+import grpc.experimental.gevent as grpc_gevent
+grpc_gevent.init_gevent()
+
 from . import events
 from .clients import HttpSession
 from .exception import (InterruptTaskSet, LocustError, RescheduleTask,
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
     packages=find_packages(exclude=['examples', 'tests']),
     include_package_data=True,
     zip_safe=False,
-    install_requires=["gevent>=1.2.2", "flask>=0.10.1", "requests>=2.9.1", "msgpack>=0.4.2", "six>=1.10.0", "pyzmq>=16.0.2"],
+    install_requires=["gevent>=1.2.2", "flask>=0.10.1", "requests>=2.9.1", "msgpack>=0.4.2", "six>=1.10.0", "pyzmq>=16.0.2", "grpc>=1.12.0"],
     test_suite="locust.test",
     tests_require=['mock'],
     entry_points={
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
     packages=find_packages(exclude=['examples', 'tests']),
     include_package_data=True,
     zip_safe=False,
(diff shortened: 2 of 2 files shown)


cgoldberg:
what's the use case for this?

Youki:
I want to do load test on a machine learning inference server (e.g. TensorFlow Model Server) from multiple nodes, and the server provides grpc interface for prediction requests. I am trying to do above test with Locust, but request per sec was not increased (limited to 6 RPS) even though I increased the number of requests (the number of users hatched). After adding 2 lines of code to core.py, I got around 50 RPS with 10 users.

cgoldberg:
I guess grpc and gevent fight for the event loop.
interesting, but I don't think we should merge this into Locust for general use.

piero:
+1
I would find it really useful for general use.

heyman:
> I would find it really useful for general use.

Isn't it possible to add that line in your own test script? If not I think we should provide some hook for it, but I don't think it makes sense to add `grpc` as a dependency for Locust.

max-rocket-internet:
You could do it like this, no?

```python
from locust import events

start_time = time.time()

# Do some GRPC stuff here

if success:
    events.request_success.fire(request_type='sqs', name='new order', response_time=time.time() - start_time, response_length=0)
else:
    events. request_failure.fire(...)
```

maximveksler:
I too would appreciate gRPC as a first-class citizen of locust. For exactly the same reason @yaboo-oyabu  mentioned.

maximveksler:
Support for gevent was specifically merged into gRPC to allow it's reusability on gevent based event loops. https://github.com/grpc/grpc/issues/4629#issuecomment-376962677

heyman:
@maximveksler Then you should be able to just put the call to `init_gevent()` in your load testing scripts (after you import locust). E.g:

```python
from locust import User

import grpc.experimental.gevent as grpc_gevent
grpc_gevent.init_gevent()

...
```

arsoni20:
hey @heyman could you please share a snippet to load test gRPC service via client. I have my client ready but I am not able to piece it together to run a locust load test on it. If you could share some pointers, that'd be awesome!

max-rocket-internet:
gRPC is super popular these days and growing (for good reason). It would be good to have one or more solid examples in the repo.

arsoni20:
I made it work.. how do I contribute with my example, could anyone please let me know?

max-rocket-internet:
> I made it work..

Nice!

> how do I contribute with my example

Create a PR to add a file in here: https://github.com/locustio/locust/tree/master/examples

It shouldn't be a complete copy/paste of your locustfile. Try to make it as minimal and generic as possible to show clearly how it works and how it could be adapted by other users 🙂

loukotal:
I tried implementing the grpc client/wrapper inspired by the docs - this works for my use case, could be useful for someone else too :) 

https://github.com/loukotal/locust-grpc-client

salliewalecka:
@arsoni20 was there any additional things you needed to do aside from what @heyman recommended? I tried to add those 2 lines after importing locust, but it did not help to resolve the issue of low rps despite many users as far as I could tell.

cyberw:
@loukotal Looks cool. Will you consider making it a PR to locust-plugins?

PR #1020, opened 2019-05-29
Adding TCP Keep Alive to guarantee master-slave communication after idle periods

albertowar:
Hi Locust folks!

This PR is a revive of a previous [PR 740](https://github.com/locustio/locust/pull/740) from about a year ago.

For context, the cloud infrastructure my company uses has a firewall that closes idle connections after 5 minutes. For us, that means that if we leave a test running for a while and come back to it, we are unable to control the slaves through the master UI.

If you look through the comments [PR 740](https://github.com/locustio/locust/pull/740), you can see that this is easily fixed by adding keep alive at the ZMQ socket level.

We thought that [PR 927](https://github.com/locustio/locust/pull/927) would make the change redundant but, after testing Locust 0.11.0 (which includes the change) we were able to reproduce the problem again pretty consistently.

The change got reviewed by @heyman and I believe his concerns are addressed.

Thanks for your time and sorry for the confusion!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: python
 cache: pip
 matrix:
@@ -6,7 +5,6 @@ matrix:
     - python: 3.7
       env: TOXENV=py37
       dist: xenial
-      sudo: true
     - python: 3.6
       env: TOXENV=py36
     - python: 3.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ matrix:
       env: TOXENV=py36
     - python: 3.5
       env: TOXENV=py35
-    - python: 3.4
-      env: TOXENV=py34
     - python: 2.7
       env: TOXENV=py27
     - stage: Verify Docker image builds
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
+FROM python:3.6.6-alpine3.8 as builder
+
+RUN apk --no-cache add g++ zeromq-dev
+RUN pip install locustio pyzmq
+
 FROM python:3.6.6-alpine3.8
 
-RUN apk --no-cache add g++ \
-      && apk --no-cache add zeromq-dev \
-      && pip install locustio pyzmq
-
+RUN apk --no-cache add zeromq-dev
+COPY --from=builder /usr/local/lib/python3.6/site-packages /usr/local/lib/python3.6/site-packages
+COPY --from=builder /usr/local/bin/locust /usr/local/bin/locust
 COPY docker_start.sh docker_start.sh
 RUN chmod +x docker_start.sh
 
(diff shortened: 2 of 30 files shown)


albertowar:
Pinging @Jonnymcc for visibility, since he made the heartbeat contribution

didip:
Our private cloud also religiously closes connections. Would be nice to have this feature as well.

Jonnymcc:
@albertowar Is this statement from the ZMQ guide not accurate then? 
"Do heartbeating on the same socket you use for messages, so your heartbeats also act as a keep-alive to stop the network connection from going stale (some firewalls can be unkind to silent connections)."
Could it be that heartbeats are only being sent to the master and nothing besides hatch message get sent from the master?

albertowar:
I'd say the recommendation from the guide (http://zguide.zeromq.org/php:chapter4) is good. However, the current implementation of heartbeats in Locust is not cutting it.

> Could it be that heartbeats are only being sent to the master and nothing besides hatch message get sent from the master?
I suspect that could be it because the issues I experienced on the UI occur when I try to send an order to the slaves (increase users, start, stop, etc).

Adding TCP Keep Alive seems trivial enough and I personally have been operating on my fork (which includes this change) for over a year without any issues.

albertowar:
@Jonnymcc, after reading the code a bit more, I think we can confirm what you suspected: https://github.com/locustio/locust/blob/master/locust/runners.py#L420

The slaves are reporting heartbeats to the master but the master is not reporting heartbeats to the slaves. Given that the communication occurs in different ports, one of the connection is kept alive by the heartbeats but the other one is being closed.

Jonnymcc:
So add a pong to the slave's ping and all is well?

albertowar:
That's definitely an approach to solve it and it could work (I would have to test it to be 100% sure). However, I am not sure if that's the most optimal way to do it.

IIRC, the current `slaves` heartbeats were added to improve the distribution of load between the slaves when some of them are missing. As a side effect, those heartbeats help to maintain one of the connections alive.

This scenario is a bit different because I **only** want to ensure the connection `master-slave` is healthy.

Implementing this at the application level (with heartbeats) is a CPU overhead (another thread doing something in the background) that is not **yet** required. It could be handy if we ever wanted the slaves to react to a dead master, but that's not a feature people have interest in (as far as I know).

If we use TCP Keep Alive, it will be handled at the transport layer, which should be more efficient CPU-wise.

albertowar:
@mbeacom, if you have time, would you mind reviewing this one as well?

mbeacom:
@albertowar It appears you have a wonky rebase.  Any chance you could squash commits here or cherry pick to another branch?

albertowar:
Will do that in a separate PR then!

albertowar:
@mbeacom I will close this one in favor of: https://github.com/locustio/locust/pull/1101

PR #1024, opened 2019-06-09
Fix grammatical error in statistic reporting

MaxVanDeursen:
Changed all spelling off `occurences` to `occurrences`


cgoldberg:
thanks!

PR #1025, opened 2019-06-11
Ensure that the last samples get sent by slave and received by master.

cyberw:


locust/runners.py
```diff
@@ -321,6 +321,7 @@ def stop(self):
     def quit(self):
         for client in self.clients.all:
             self.server.send_to_client(Message("quit", None, client.id))
+        gevent.sleep(0.5) # wait for final stats report from all slaves
```
aldenpeterson-wf: This timing seems kind of arbitrary, is there a reason you went with 0.5?

cgoldberg: that was my thought also.

cyberw: Yes, it is kind of arbitrary, I'm assuming that the latency between master & slave is < .5s. 

I tried explicitly waiting for one more event, but ran into issues. It could definitely be done better, but I didnt find a way to do it without adding a significant amount of logic.

cyberw: @aldenpeterson-wf @cgoldberg Any more thoughts? The current behaviour is really bad (missed samples on master), so it needs some kind of fix...

heyman: I guess it could track  which worker nodes have sent their reports, and then some timeout. But I think this is "good enough" and definitely better than the previous behaviour!

PR #1032, opened 2019-06-21
Add required --host option to examples

ramitmittal:
While the examples in the page focus on the master/slave concept, not providing the --host option will result in failure to start the test. It should be added, so that any user that copies the command from the documentation does not have to decipher an error message before getting the test to run.

### DIFF
--- a/docs/running-locust-distributed.rst
+++ b/docs/running-locust-distributed.rst
@@ -26,11 +26,11 @@ Example
 
 To start locust in master mode::
 
-    locust -f my_locustfile.py --master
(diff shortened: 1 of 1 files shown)


cyberw:
Hmm, I see what you are geting at, but that example is for things that are specific to running distributed. Using the --host parameter isnt specifically for distributed runs, so I dont feel it really belongs in this example.

heyman:
I also understand the issue, but I'd also like to keep the examples minimal.

Instead of adding the `--host` argument to the example, I added a short note about it on the same documentation page: 40f64e0594d04b883bd68fefa294a4f849d6879c

PR #1038, opened 2019-06-26
Use print() function in both Python 2 and Python 3

cclauss:
Discovered via: __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__

Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.


cgoldberg:
thanks!

PR #1039, opened 2019-06-26
Tox: Add flake8 tests to find Python syntax errors and undefined names

cclauss:
Blocked by #1038  https://travis-ci.org/locustio/locust/jobs/550766994#L274

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,8 @@ envlist = py27, py34, py35, py36, py37
 [testenv]
 deps =
     codecov
+    flake8
     mock
 commands =
+    flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
     coverage run -m unittest discover []
--- a/tox.ini
+++ b/tox.ini
@@ -7,5 +7,5 @@ deps =
     flake8
     mock
 commands =
-    flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
+    flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
     coverage run -m unittest discover []


cyberw:
I love linters (although personally I primarily use pylint). 

@heyman , any reason we shouldnt merge this (other than maybe using pylint instead :) )

At some later point we should (imho) standardize on using an auto-formatter. My favourite there is Black (https://github.com/psf/black)

cclauss:
@cyberw https://lukeplant.me.uk/blog/posts/pylint-false-positives/ matches my experience pretty well.

cyberw:
> @cyberw https://lukeplant.me.uk/blog/posts/pylint-false-positives/ matches my experience pretty well.

Sure, but he was starting with a huge repo which he had already used flake8 on, so it is a bit of a strange comparison. Personally I use pylint, but I have turned off quite a few warnings and tuned some values (like line length) to match my preferences. Sort of the opposite to what I really like about Black - maybe it doesnt make a lot of sense :)

Anyway, I'd be happy to merge this, unless someone objects within the next couple of days :)

heyman:
To be honest I have little experience with linters in python, but the article posted by @cclauss also matches my experience when trying it out on existing projects. Also, I guess I'm not really bothered by having small formatting differences, or trailing whitespaces, as long as the code is clear and readable. I'm not very keen on obfuscating git history with big formatting changes.

With that said I'm absolutely +1 on merging this. However when I ran it right now I got the following error:
```
./locust/contrib/fasthttp.py:18:11: F821 undefined name 'unicode'
    str = unicode
          ^
```
which refers to these lines: https://github.com/locustio/locust/blob/master/locust/contrib/fasthttp.py#L12-L18

That code is fine since it should only run on python2. Not sure how to suppress that though?

cyberw:
The nice thing about Black (other than the fact that it lets you focus on coding rather than formatting - for me it made using python *so* much more enjoyable) is that it also makes diffs smaller and more easy to track (for example, by always using a trailing comma, even for the last parameter / dictionary entry). 

Of course, the first run might obfuscate the git history a little, but in the long run it is very beneficial. Anyways, later discussion :)

cclauss:
`str = unicode` is just a bad idea.  It changes basic assumptions that code readers would normally depend on.

The following is more commonly used and flake8 will not complain about it.
```python
try:
    unicode  # Python 2
except NameError:
    unicode = str  # Python 3
```

Issue #1059, opened 2019-08-05
Requests Per Second Plot Breaks When There are too Many Unique URLs


williamlhunter:
### Description of issue

Plotting requests per second breaks when locust has visited around 500 unique URLs

### Expected behavior

Total requests per second works as long as needed

### Actual behavior

It seems to break once you hit too many unique URLs.  I believe this is because locust gets the stats for every URL every time it updates the web ui.

### Environment settings

- OS: Windows
- Python version: 3.7.3
- Locust version: 0.10.0

### Steps to reproduce (for bug reports)

I am using locust to test a retail point of sale rest api.  The api uses document SIDs in the url, so everytime I make a new transaction, I get 3-4 new urls.  Not only does this make analyzing the results harder, as I need to group the urls together with regular expressions in pandas, but i believe that it is breaking the plotting of requests per second as well.
I can hammer the api with requests that don't use the document SID in the url forever, but I can only run a few hundred transactions before the plotting breaks.
To make sure, I made 15 new receipts per second for about 30 mins and everything worked correctly.  Afterword, I changed my locust file to make a blank receipt and then request it's contents (generating a request on a new url) and was only able to run the load test for 2 mins before plotting broke.  (at around 500 unique urls)

### Possible fix

I would love to be able to set the field that requests are grouped by to a string.  Something like:
`response = self.client.post("/v1/rest/document" headers=self.header, json=payload, tag="creating a document")`
but I'm not incredibly experienced, so this may not be as easy to implement as I'm hoping.
Also this implies that I'm correct as to why the requests per second plot is breaking.  Furthermore, this would save me (and people doing similar testing) from having to group api calls after running a test.

cgoldberg:
> before the plotting breaks

what about it is broken?

williamlhunter:
Sorry, I was in a bit of a rush and totally forgot to add that.  Here's an example:

![screenshot](https://i.imgur.com/uQqKdgs.png)

As you can see, the plot shows no data for each point in the second half, even though the rps at the top was constantly changing.

cgoldberg:
please give a clear description of the issue,  including details of what you are expecting and what is actually happening.   Please exclude all other information, conjectures, or ideas for possible fixes.  This issue report is packed with information that is completely irrelevant... which makes it difficult to follow.  

(For example... I can't even tell if this is a bug report, or a feature request for tagging responses ...  it seems like both)

williamlhunter:
[...]

PR #1060, opened 2019-08-06
Fixed issue with Total Requests Per Second plot

williamlhunter:
Fixes the plot when there are too many unique urls.

#1059 
#889 

The total stats block, which is usually located at the end of the list of stats, was getting truncated whenever there were too many URLs that were being requested.  Here I add the total stats back to the end of the report.

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -123,6 +123,8 @@ def request_stats():
     # Truncate the total number of stats and errors displayed since a large number of rows will cause the app
     # to render extremely slowly. Aggregate stats should be preserved.
     report = {"stats": stats[:500], "errors": errors[:500]}
+    if len(stats) > 500:
+        report["stats"] += [stats[-1]]
 
     if stats:
         report["total_rps"] = stats[len(stats)-1]["current_rps"]


jonyoder:
We've run into this bug while testing, too. Our load tests use a smart randomized URL which can result in thousands of unique URLs. If this fix is correct, I'd love to see it merged.

cgoldberg:
>  If this fix is correct, I'd love to see it merged.

I don't think it is, and the original issue is still much too vague.

williamlhunter:
I tested it, and it does fix the issue (at least for me).

Furthermore, let me know how I'm being vague and I'll elaborate.

RyanW89:
I left this comment on the original thread, i believe you need to add name="" to your request. The requests will be aggregated on this, rather that the URL.

williamlhunter:
> I left this comment on the original thread, you need to add name="RequestName" to your request. The requests will be aggregated on this, rather that the URL.

This did fix my problem, but I still feel that the fix is warranted for users who have more than 500 categories.

cyberw:
Thanks @williamlhunter !

heyman:
Nice!

I also think it could be nice to display some kind of warning notice in the Web UI if the stats have been truncated. Both to inform the user that it has happended, and also mentioning that it might affect performance with very large amount of stats entries.

PR #1075, opened 2019-08-27
Fix fail percentage

### LINKED ISSUE #1074: Logged stats show incorrect failure percentage
### Description of issue
The failure percentage stats printed to stdout are calculated incorrectly.
[Seen here](https://github.com/locustio/locust/blob/master/locust/stats.py#L388)
`fail_percent = (self.num_failures/float(self.num_requests + self.num_failures))*100`

### Expected behavior
If something fails every request made, the failure percentage should show 100%.
The calculation should be
`fail_percent = float(self.num_failures/self.num_requests)*100`

### Actual behavior
It shows 50% if a request fails every attempt.
### Environment settings

- OS: Fedora 29
- Python version: 3.6.8
- Locust version: latest source

Create a locust file with any failing requests. You can see the number of failures don't match up with the failure percentage.

SquirtsMacintosh:
Fixes #1074 

The fail percentage was calculated incorrectly. Something that failed
all of the time would be reported as failing 50% total.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -385,7 +385,7 @@ class StatsEntry(object):
 
     def __str__(self):
         try:
-            fail_percent = (self.num_failures/float(self.num_requests + self.num_failures))*100
+            fail_percent = float(self.num_failures/self.num_requests)*100
         except ZeroDivisionError:
             fail_percent = 0
         


cyberw:
@aldenpeterson-wf @SquirtsMacintosh This was fixed previously, but only in the in fail_ratio() method (https://github.com/locustio/locust/commit/ed542b0233d901896d2025eda22e647866ef408f#diff-5d5f310549d6d596beaa43a1282ec49e). This particular line never called that, so it needed a fix of its own. Probably, a better fix would be
`fail_percent = self.fail_ratio()*100`

raiyankamal:
I sent another PR with a commit on top of this one after seeing the suggestion by @cyberw 
https://github.com/locustio/locust/pull/1082

cgoldberg:
superceded by #1082

PR #1083, opened 2019-09-09
Allow options to be passed to locust in docker, update docs

max-rocket-internet:
In this PR I am:

- Adding a new env var so users can pass options to `locust` in docker image. This means `standalone` mode with no web is now possible. Great for running as k8s `Job`.
- Updating docker documentation with new `LOCUST_OPTS` env var and a couple extra examples for how to run your tests.

Does it make sense? Is that OK?


max-rocket-internet:
@mbeacom, would you mind reviewing this?

Issue #1081, opened 2019-09-05
response time has too many decimal places in the web statistics page


solxnp:
### Description of issue

response time has too many decimal places in the web statistics page

### Expected behavior

TBD

### Actual behavior

![image](https://user-images.githubusercontent.com/18416053/64319375-c5404380-cfee-11e9-8aff-b06b7fc635ea.png)

### Environment settings

- OS: centos
- Python version: 3.7.4
- Locust version: 0.11.0

### Steps to reproduce (for bug reports)

TBD - [please provide example code](https://stackoverflow.com/help/mcve)

aldenpeterson-wf:
You don't measure 20 significant digits worth of time when it comes to tracking request latency? 

:)

cyberw:
Ok to close? Or do we only close on release?

cgoldberg:
fixed by #1089

PR #1089, opened 2019-09-23
Min and max response times rounded to nearest int in web view

ajt89:
#1081

### DIFF
--- a/locust/test/test_util.py
+++ b/locust/test/test_util.py
@@ -1,5 +1,6 @@
 import unittest
 from locust.util.time import parse_timespan
+from locust.util.rounding import proper_round
 
 
 class TestParseTimespan(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


cgoldberg:
I'm not sure `round()` is sufficient here for both Python2 and Python3, since they behave differently...

For example:

**Python 2**:

```
>>> round(0.5)
1.0
>>> round(1.5)
2.0

```

**Python 3**:

```
>>> round(0.5)
0
>>> round(1.5)
2
```

thoughts?

ajt89:
@cgoldberg Good point. Should I create a rounding function in the util folder similar to this? https://stackoverflow.com/questions/31818050/round-number-to-nearest-integer

cgoldberg:
:+1:  nice...  thanks

PR #1090, opened 2019-09-25
Add an option to allow tasks to finish running their iteration before exiting

cyberw:
I ran into a case where I wanted to guarantee that task are allowed to finish (to finish things like logout, return of test data etc). I think this would be useful to other people as well. 

Adding the option also has the nice benefit of making it more clear that the default behaviour is that tasks can exit (almost) at any point.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -292,6 +292,15 @@ def parse_options():
         help="sets the exit code to post on error"
     )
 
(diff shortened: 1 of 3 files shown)


cyberw:
This doesnt actually work atm. I'll withdraw this and file a new PR once I get it working.

PR #1093, opened 2019-09-26
Improvements for Dockerfile

coderanger:
Installs from the local code rather than a release version, creates a non-root user, and switches from entrypoint to cmd for better customization.


max-rocket-internet:
> Installs from the local code rather than a release version

Nice! I was going to make a PR for this also.

If we resolve https://github.com/locustio/locust/issues/849 and merge https://github.com/locustio/locust/pull/1083 then we might finally have nice official docker option for locust.

Issue #1049, opened 2019-07-15
Locust Execution Ends intermittently without completing all tasks


TanmayNakhate:
### Description of issue

I have written locust code with python in such a way that POST and GET requests are handled in sequential manner using @seq_task decorator and GET Method execution should be completed for every POST Method.

    class NcsoTest(TaskSequence):

    @seq_task(1)
    def create_ics(self):
        self.client.post_ics_service() #POST Request

    @seq_task(2)
    def parse_results(self):
        self.client.parse_service_results() #GET Request

    def on_stop(self):
        pass

 
I am using following params to run my locust code during UT
 
 --min_wait_time 15000 --max_wait_time 15000 --num_clients 1 --hatch_rate 1 --test_time 2m

in summary reports i am seeing following behavior 

[![Locust Test Summary - Actual ][1]][1]

What I expect is post and get request count should be same.
Is there anything I should tweak?

### Expected behavior
[![Locust Test Summary - Expected][2]][2]

### Actual behavior

 [![Locust Test Summary - Actual ][1]][1]

### Environment settings

- OS: Centos7
- Python version: v2.7.16:413a49145e
- Locust version: Locust 0.11.0

### Steps to reproduce (for bug reports)

    class NcsoTest(TaskSequence):

    @seq_task(1)
    def create_ics(self):
        self.client.post_ics_service() #POST Request

    @seq_task(2)
    def parse_results(self):
        self.client.parse_service_results() #GET Request

    def on_stop(self):
        pass

Note: Please ignore the Avg, Min, Max and Median - they vary. What is imp is that count for Post and Get is same.

  [1]: https://i.stack.imgur.com/JdhmB.png
  [2]: https://i.stack.imgur.com/kOrjj.png

ernestom:
Perhaps related to https://github.com/locustio/locust/pull/1025.

TanmayNakhate:
> Perhaps related to #1025.

Seems like it. Cant wait to try.

cyberw:
> > Perhaps related to #1025.
> 
> Seems like it. Cant wait to try.

#1025 only relates to distributed runs. I think locust can terminate at any time between tasks, there is no guarantee that an entire TaskSet completes, so the behaviour you are seeing is "normal". 

Perhaps you could just use a single task?

```

class NcsoTest(TaskSet):

@task
def ics(self):
    self.client.post_ics_service() #POST Request
    self.client.parse_service_results() #GET Request

```

cyberw:
[...]

PR #1099, opened 2019-09-30
Add an option (--stop-timeout) to allow tasks to finish running their iteration before exiting 

cyberw:
Solves #1049


heyman:
I think the feature makes sense. With the current implementation (if my understanding of it is correct), the runner would very often wait much longer than actually needed, since all the locust users that are currently sleeping would have to wake up, and even do an extra task execution, before raising GreenletExit.

I'm wondering if it would perhaps make sense to introduce a `state` property on the Locust class which could be something like either `running`, `waiting`, or `stopping` (which could then replace `exit_at_end_of_iteration`)? Then the runner's stop method could immediately kill all Locust which are in waiting state, and then wait for the others to stop. 

Also I think the check for the stop condition should be done at the beginning of the TaskSet's main loop (for example here: https://github.com/locustio/locust/blob/644496f5548ceb5a9ff4220aa808ec25a4c9a143/locust/core.py#L365), because otherwise a sub-TaskSet calling `interrupt()` wouldn't cause the TaskSet to exit (instead another task would be executed since wait() isn't called by default after an interrupt).

cyberw:
@heyman Is there a way for me to get a reference to the actual locust instance from LocustRunner.stop() ? I can get a reference to the *class* using self.locusts[i].args[0], but I cant check the state of the instance...

heyman:
Hmm, from the look of the code:

https://github.com/locustio/locust/blob/d6d87b4b3cb29b9cec1190bed7586f8b87bb492b/locust/runners.py#L128-L132

we don't seem to store a reference to the instance (`locust()`). So we would need to start tracking them. Maybe a dict that maps the locust instances to the greenlet running the locust (since we would need a reference  to the greenlet to be able to kill specific locust instances).

cyberw:
Could we just spawn the greenlet with a Locust object instead of a locust class as argument? All the tests pass, and everything seems to work if I change those lines to:

```                new_locust = locust()
                def start_locust(_):
                    try:
                        new_locust.run(runner=self)
                    except GreenletExit:
                        pass
                self.locusts.spawn(start_locust, new_locust)```

I understand that it is kind of a sneaky change, but it does seem to work, and to me it does make more sense to spawn a greenlet with a reference to an object rather than to a class anyway...

**Comment by heyman** [comment]
> Could we just spawn the greenlet with a Locust object instead of a locust class as argument?

Ah, yes! That's much better. Just make sure to update the `kill_locusts` method accordingly :).

**Comment by cyberw** [comment]
@heyman Looks like it works now. I didnt see any need for modification in kill_locusts, but maybe I'm missing something?

> Also I think the check for the stop condition should be done at the beginning of the TaskSet's main loop ...

I dont think that is actually necessary, now that we kill sleeping locusts. Either the locust is sleeping when stop is run, and then it will be killed immediately, or it is running, in which case we set, exit_on_end_of_iteration and it will be killed as soon as the iteration is finished (it will never get to the beginning of the main loop)

**Comment by cyberw** [comment]
I dont understand the code coverage report. There's lots of diffs in there that shouldn't be there, and merging with master didnt make any difference...

**Comment by heyman** [comment]
> I didnt see any need for modification in kill_locusts, but maybe I'm missing something?

It's definitely not super clear, but if I'm not mistaken a small change is needed.

Full function for reference:
https://github.com/locustio/locust/blob/b1c347bff76d0668d9932f22dbef86cf3e817d93/locust/runners.py#L144-L161

`g.args[0]` on this line:

https://github.com/locustio/locust/blob/f7170f2dd7e236bd0d513742734697f62d7afed3/locust/runners.py#L154

currently refers to a Locust class, and this change:

https://github.com/locustio/locust/blob/b1c347bff76d0668d9932f22dbef86cf3e817d93/locust/runners.py#L134

changes so that `g.args[0]` will be a Locust class *instance*, so you'd have to change the line to `if l == type(g.args[0]):`

**Comment by heyman** [comment]
I just pushed a test to master which I believe should fail for this PR. I think the fix should be the oneline described in my previous comment though.

> I dont understand the code coverage report. 

Me neither.

**Comment by heyman** [comment]
Also, this line shouldn't be needed (also not from this PR, but if you want you can remove it):
https://github.com/locustio/locust/blob/f7170f2dd7e236bd0d513742734697f62d7afed3/locust/runners.py#L156

**Comment by heyman** [comment]
Oops, my test had a Python 2.7 incompatibility 🤦‍♀. Fixed now!

**Comment by cyberw** [comment]
> Also, this line shouldn't be needed (also not from this PR, but if you want you can remove it):
> https://github.com/locustio/locust/blob/f7170f2dd7e236bd0d513742734697f62d7afed3/locust/runners.py#L156

I guess it was there to serve as optimizing the loop? But I agree it is not needed.

**Comment by heyman** [comment]
> I guess it was there to optimize the loop? But I agree it is not needed.

Oh, yes, you're correct. I misinterpreted the code. It does make sense to keep it.

**Comment by cyberw** [comment]
Force push ftw :)

**Comment by heyman** [comment]
> in which case we set, exit_on_end_of_iteration and it will be killed as soon as the iteration is finished (it will never get to the beginning of the main loop)

Actually (unless I'm mistaken) if `TaskSet.interrupt()` is called with the default argument `reschedule=True`, the execution will end in the current TaskSet, and then a new task will be immediately executed in the parent TaskSet, and therefore the `wait` method will not be called. 

That won't happen if the check is at the beginning of the main loop. Also I still like the idea of using a `Locust._state` attribute (that can  be either STATE_RUNNING, STATE_WAITING or STATE_STOPPING) instead of the `_waiting` attribute. Then there is no need for the `exit_at_end_of_iteration` attribute.

**Comment by cyberw** [comment]
@heyman LGTY? If you want I can file a new PR, in a new branch so that the changes get the right name (stop timeout instead of task finish wait time), and also squash the changes.

File: locust/main.py [hunk 1]
```diff
@@ -242,6 +242,15 @@ def parse_options():
         help="sets the exit code to post on error"
     )
 
+    parser.add_argument(
+        '-s', '--stop-timeout',
+        action='store',
+        type=int,
+        dest='stop_timeout',
+        default=None,
+        help="number of seconds to wait for a taskset to complete an iteration before exiting. default is to terminate immediately."
```
heyman: Alternative copy suggestion:
`Number of seconds to wait for a simulated user to complete any executing task before exiting. Default is to terminate immediately.`

heyman:
It's nitty but I think that if you place the `if self.locust._state == LOCUST_STATE_STOPPING:` check immediately after this line:
https://github.com/locustio/locust/blob/1e6eab70863797490098113dc0c887b8b55e06b7/locust/core.py#L363
you don't need the check it in two places.

Other than that (and my suggested copy for the help text) I think it looks great! I don't mind merging this PR (I changed the title of it) :).

cyberw:
@heyman Ready to merge, IMO :)

PR #1112, opened 2019-10-19
Add __main__.py file

jdufresne:
Allows executing locust as a module using the command:

    $ python -m locust

https://docs.python.org/3/library/__main__.html


heyman:
Thanks!

PR #1113, opened 2019-10-20
Fix rounding error when spawning users from multiple locust classes

### LINKED ISSUE #848: Tests fail to run when providing a small number of clients

heyman:
This should fix #848 and is an alternative implementation to #860. It's simpler but less sophisticated.

With this PR one could still get an uneven distribution  of Locust classes when killing locusts. For example if one has multiple Locust classes and then kills a single user one by one by changing the number of locusts multiple times and decreasing the number by one each time. I don't think this is a big issue. It could be fixed later by rewriting the logic of the `LocustRunner.kill_locusts()` method to - instead of using `weight_locusts()` to get a list of classes to remove - use `weight_locusts()` to  get a list of the new distribution and then diffing the current distribution to the new one and killing the locusts needed to achieve the new distribution.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -68,6 +68,7 @@ class LocustRunner(object):
         """
         bucket = []
         weight_sum = sum((locust.weight for locust in self.locust_classes if locust.task_set))
+        residuals = {}
         for locust in self.locust_classes:
             if not locust.task_set:
                 warnings.warn("Notice: Found Locust class (%s) got no task_set. Skipping..." % locust.__name__)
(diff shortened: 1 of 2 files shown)


cyberw:
LGTM. Not loving the variable name "amount", but that was there before the PR so whatever :)

heyman:
> Not loving the variable name "amount", but that was there before the PR so whatever :)

I agree.

PR #1114, opened 2019-10-21
Table layout fix to use available space better

heyman:
This PR removes the `table-layout: fixed` CSS property from the  stats table in the web UI, in order to make better use of the available space.

The `table-layout:fixed` property was added by #962 (which is a PR for #938), so this PR will partly revert those changes. However I'm still able to see the whole table by scrolling the page horizontally after these changes.

**Before:**
image

**After:**
image

### DIFF
--- a/locust/static/style.css
+++ b/locust/static/style.css
@@ -290,6 +290,9 @@ a:hover {
 .status th.numeric {
     text-align: right;
 }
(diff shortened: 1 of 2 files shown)


max-rocket-internet:
Nice!

Issue #646, opened 2017-08-28
Allow a fixed RPS rate


etc0de:
### Description of issue / feature request

Please allow specifying a fixed RPS rate.

While the current model is nice for modelling users, it is not very useful for modelling more complex modern web applications which often will exhibit an exact known waiting behavior rather than something controlled by a more unpredictable distribution.

Currently, I see no good way modelling this with locust and we are having huge trouble in our current project working around by guessing settings needlessly to roughly get to the actual RPS the web app knowingly sets off.

### Expected behavior

RPS setting is available as alternative (including a way to guarantee e.g. exactly 1 request per second even if the actual request itself has a varying duration, e.g. a 200ms request response and a 500ms request response won't lead to strong variations of the interval)

### Actual behavior

I can't find an RPS setting, and requests taking longer seems to make users wait more instead of allowing to have some sort of predictable fixed interval behavior (as is unrealistic for real users of course, but not unrealistic for many automated web clients).

aldenpeterson-wf:
> RPS setting is available as alternative (including a way to guarantee e.g. exactly 1 request per second even if the actual request itself has a varying duration, e.g. a 200ms request response and a 500ms request response won't lead to strong variations of the interval)



Have you looked at [min and max wait](http://docs.locust.io/en/latest/writing-a-locustfile.html#the-min-wait-and-max-wait-attributes)? Setting them to both be 1000 (ms) would result in locust making 1 request/sec per client which seems to be exactly what you are looking to do.

JonasT:
@aldenpeterson-wf are you sure it won't result in `1000/(1000ms + request time milliseconds)` requests per second? Because that is what we seem to be seeing

mambusskruj:
@JonasT you're right! this is `1000/(1000ms + request time milliseconds)` exactly.

JonasT:
@mambusskruj yea and that's not 1 RPS but rather something lower.. so not what we're looking for

mambusskruj:
@JonasT the problem is to calculate different wait time based on response time. So, the right calculating is: `1000/(x wait time ms + request time milliseconds)` where x will calculating all the time during test process.

JonasT:
@mambusskruj 

It's not really hard to do:
```python
import time

request_ts = time.monotonic()
RPS = 1.0
while True:
    do_request()
    request_ts += (1.0/RPS)
    now = time.monotonic()
    if now < request_ts:
        time.sleep(request_ts - now)
```
[...]

PR #1118, opened 2019-10-22
New API for specifying wait time

heyman:
This is a proposal for changing the API for how the `Locust` and `TaskSet` wait time is specified. 

Here's an example of the current API:

```python
from locust import Locust
class User(Locust):
    min_wait = 1000
    max_wait = 25000
```
This would be the equivalent example with the proposed API:
```python
from locust import Locust, between
class User(Locust):
    wait_time = between(1, 25)
```

I've also created two other wait_time functions besides `between`: 

* `constant` - Always wait the specified (constant) time
* `constant_pacing` - Try to sleep the amount of time required to make the time between task executions constant (so if a task takes 2 seconds to execute, the wait time will be 2 seconds lower). This is a requested feature in #646.

**Some things to note**

* These changes are backwards compatible, so specifying `min_wait` and `max_wait` still *works* in this PR (though currently not `wait_function` but that could be fixed), so we could add deprecation warnings and keep that functionality for another version or two.
* The wait time is specified in seconds instead of milliseconds. That's what builtin python functions such as `time.sleep` expects, and I think it's more common that the wait time is multiple seconds rather than fractions of a second, so I think that change makes sense to do if we're changing the API.
* I have only updated the most important documentation pages, since I'd like some feedback before spending too much time on updating the docs.
* I haven't updated all existing tests but if we were to do this change we would have to update the old tests to the new API, and add some new tests that checks the backwards compatibility of `min_wait` and `max_wait`.

The reason I started working on this, was because I was working on fixing the bug in #891 and realised that the fix would result in a somewhat messy solution (https://github.com/locustio/locust/compare/master...fix-891) that would involve taking a bound method on the Locust instance, and re-binding it on the TaskSet instance. So besides giving us a (IMO) nicer API, it would also help avoid that messy solution  :).

What do you think?  Is there something I haven't thought of that would break by this?


cyberw:
Tbh, I think we should only have one mechanism for setting wait times. I'm +1 for at least deprecating (log a warning) if someone is using min_wait / max_wait.

heyman:
> Personally I favor pacing the total throughput for all running locusts, instead of individually

I can see that there might be use-cases for it, but I don't think it fits very well with the idea of defining User behaviour and then simulating those users (if the wait time of a user is dependent on the number of - and actions of - other users). This PR shouldn't break `TaskSetRPS` in any way though.

> I think we should change the default wait times to zero. 

In this PR I've changed so that a `MissingWaitTimeError` exception is raised if no `wait_time` is specified (or `min_wait + max_wait` or `wait_function`). I considered letting no specified wait_time result in `0`, but I think it's better to explicitly raise an exception. For example if someone forget to specify the wait time, I think it's better that they get an error instead of risk not noticing it.

heyman:
I've now also fixed backwards compatibility for `wait_function`, and added DeprecationWarnings when any of `min_wait/max_wait` or `wait_function` is used.

I've also updated all the tests and added som new tests for the old API.

This PR should actually be mergeable pretty soon.

HeyHugo:
Sorry if this is coming too late, but an alternative api could be to try using python primitives instead, that way you could skip importing any extra functions:
```python
# Instead of constant(1) interpret any numeric type as constant
wait_time = 1

# instead of between(1, 10) interpret tuples as between
wait_time = (1, 10)

# instead of constant_pacing(5) interpret a function taking first positional arg as time spent on last request
wait_time = lambda t: 5 - t

```

heyman:
> Sorry if this is coming too late, but an alternative api could be to try using python primitives instead, that way you could skip importing any extra functions

I think the proposed API is more clear, and less "magical", and it also allow users to define their own wait_time functions (that have access to the Locust instance). The third of your examples feels very unintuitive to me.

heyman:
We could possibly change so that just specifying a number as `wait_time` would result in `constant()` and  specifying a tuple would result in `between()`, though I'm not sure I think that it's worth it (making the code harder to understand to save an import).

cyberw:
Shouldnt the wait functions be in their own package? Maybe, also maybe "between" as a distribution description is a little vague? It could be "uniform" instead...

cyberw:
Also (I think I've talked about this earlier) I still think there is a reasonable default sleep time, which should be used if the user doesnt specify anything: zero.

cyberw:
I still think we could use zero as the default time (it is the default in every other load testing tool I've used :)

heyman:
> Shouldnt the wait functions be in their own package? Maybe, also maybe "between" as a distribution description is a little vague? It could be "uniform" instead...

I agree that `uniform` would perhaps be a more scientifically "correct" naming. However I think `between` *reads* better ("Wait time between x, y"), and I find it unlikely that anyone would assume any other distribution than uniform for a wait function named `between`.

> I still think we could use zero as the default time (it is the default in every other load testing tool I've used :)

This PR currently raises as `MissingWaitTimeError` if wait time is missing. I could see us defaulting to zero instead. However, I think it's quite rare that one actually want to have 0 as the wait_time which makes it a strange default. Many other load tests are just about cramming out as many reqs/s as possible, while Locust's main advantage is the ability to create realistic load by defining user behaviour in code.

heyman:
> Shouldnt the wait functions be in their own package?

Do you mean a totally separate python package? I think it make sense to include them (at least `between` and `constant`) in Locust. 

They are located in a separate module (`locust.wait_time`) within Locust and I was really going back and fourth between importing them into the main locust namespace. The python programmer in me wanted people to have to import them from `locust.wait_time` but at the same time I realise that Locust has some users who are not super familiar with python code, and figured it was worth to pollute the `locust.*` namespace with some default wait_time functions, in order to save people from having to do an additional import.

cyberw:
> > Shouldnt the wait functions be in their own package?
> 
> Do you mean a totally separate python package? I think it make sense to include them (at least `between` and `constant`) in Locust.
> 
> They are located in a separate module (`locust.wait_time`) within Locust and I was really going back and fourth between importing them into the main locust namespace. The python programmer in me wanted people to have to import them from `locust.wait_time` but at the same time I realise that Locust has some users who are not super familiar with python code, and figured it was worth to pollute the `locust.*` namespace with some default wait_time functions, in order to save people from having to do an additional import.

Yea, I meant separate module/namespace, not package. Sure, another import is not so nice, but finding the available wait_time functions would be easier if they were in a separate namespace. the main locust namespace is kind of polluted as is.

cyberw:
> > Maybe, also maybe "between" as a distribution description is a little vague? It could be "uniform" instead...
> 
> I agree that `uniform` would perhaps be a more scientifically "correct" naming. However I think `between` _reads_ better ("Wait time between x, y"), and I find it unlikely that anyone would assume any other distribution than uniform for a wait function named `between`.
> 
> > I still think we could use zero as the default time (it is the default in every other load testing tool I've used :)
> 
> This PR currently raises as `MissingWaitTimeError` if wait time is missing. I could see us defaulting to zero instead. However, I think it's quite rare that one actually want to have 0 as the wait_time which makes it a strange default. Many other load tests are just about cramming out as many reqs/s as possible, while Locust's main advantage is the ability to create realistic load by defining user behaviour in code.

"between" is fine in isolation, but looks weird next to "constant" and "constant_pacing". I guess it is a matter of taste.

heyman:
> Sure, another import is not so nice, but finding the available wait_time functions would be easier if they were in a separate namespace. the main locust namespace is kind of polluted as is.

It's still possible to import the wait_time functions from `locust.wait_time`. I don't feel strongly about this one, but I think I'm leaning more towards including it, for convenience.

> In my experience having zero wait times is not rare or strange.

I don't feel strongly about making `wait_time` mandatory VS having a default of zero, either. 

Does anyone else want to chime in?

cyberw:
Does this remove the need for #1005 ?

heyman:
> Does this remove the need for #1005 ?

Yes, it does.

Issue #374, opened 2016-01-11
URL names in dashboard are not HTML escaped


kumar303:
URL names (for stats) are not HTML-escaped in the dashboard. This makes names with angle brackets disappear. For example:

```
self.client.get(url, name='/some-resource/upload/<uuid>')
```

would show up as `/some-resource/upload/` instead of `/some-resource/upload/<uuid>` which is confusing. There could be some XSS security concerns here as well.

cyberw:
Closing due to lack of activity. Let me know if this is still an issue.

heyman:
I just tested this. It's still an issue, and I think we should fix it.

peterdemin:
Can I take it?

cyberw:
@peterdemin By all means!

peterdemin:
Alright, where do I start? I'm experienced with Python, but new to Locust.

cyberw:
Uhm, if you've never used locust then you would need to start using it first, enough that you can at least reproduce the error.

I'm not even sure where the fix should be applied. Maybe in request_stats() in web.py, on line 111.

peterdemin:
I'm using locust here and there for a few years already :-) Thanks for the guidance.

cyberw:
Cool, I was confused and thought maybe the hacktoberfest tag had attracted some kind of "drive by contributors" who didnt know what Locust was even for :P

peterdemin:
No, I just felt the urge to contribute back to the project I enjoy and used the Hacktoberfest label to filter low-hanging fruits.

peterdemin:
Fixed in #1119

peterdemin:
This can be closed, I believe.

cyberw:
Thx!

PR #1119, opened 2019-10-23
Escape HTML entities in endpoint names #374

peterdemin:
URL names (for stats) were not HTML-escaped in the dashboard.
This made names with angle brackets disappear. For example:

```
self.client.get(url, name='/some-resource/upload/<uuid>')
```

would show up as `/some-resource/upload/` instead of `/some-resource/upload/<uuid>` which is confusing.

I added new key to /stats/request - "safe_name" - so that escaping
is on the server side and javascript has less logic.

### DIFF
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -219,7 +219,7 @@
         <![CDATA[
         <tr class="<%=(alternate ? "dark" : "")%> <%=(this.name == "Total" ? "total" : "")%>">
             <td><%= (this.method ? this.method : "") %></td>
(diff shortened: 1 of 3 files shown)


cyberw:
Awesome! Looks like it has issues in Python 2.7 though... "ImportError: No module named html"

peterdemin:
Ah, I forgot to check if Python 2.7 is still supported. BTW, what are the plans on dropping Python 2.7 support?

cyberw:
Hopefully soon, IMHO :) thanks for your contribution!

PR #1126, opened 2019-10-28
Let's stop calling the package beta!

cyberw:
Locust has been widely used for many years now, it deserves to be considered production ready in PyPi! 

It will always have bugs, and even (rarely) breaking changes, but that is part of normal development (and handled by versioning, not project status in pypi anyway).


heyman:
Sounds good!

PR #1127, opened 2019-10-30
Fix and add test for when locusts fail to exit at end of iteration during stop timeout.

cyberw:
I discovered a bug introduced in the last commit on stop-timeout. Adding a test case for it (and once you have seen that it actually fails), I will revert that commit to show that it now works :)

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -358,15 +358,14 @@ class TaskSet(object):
         
         while (True):
             try:
-                if self.locust._state == LOCUST_STATE_STOPPING:
-                    raise GreenletExit()
-        
                 if not self._task_queue:
                     self.schedule_task(self.get_next_task())
                 
                 try:
                     self.execute_next_task()
                 except RescheduleTaskImmediately:
+                    if self.locust._state == LOCUST_STATE_STOPPING:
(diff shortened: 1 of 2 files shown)


heyman:
Ah, I see, my bad! Could we move the check from the `wait()` method to immediately after `self.execute_next_task()` instead?

cyberw:
> Ah, I see, my bad! Could we move the check from the `wait()` method to immediately after `self.execute_next_task()` instead?

You mean after line 365? That doesnt work, because it wont be called on interrupts.

cyberw:
> > Ah, I see, my bad! Could we move the check from the `wait()` method to immediately after `self.execute_next_task()` instead?
> 
> You mean after line 365? That doesnt work, because it wont be called on interrupts.

Wait, I get it now, you meant move the *other* check. Yes, that works, and arguably makes more sense. Commiting it now...

cyberw:
Always running one iteration when using stop-timeout should be fine imo (maybe even desirable). I'll can definitely add it, but I dont think it is worth bloating the code.

heyman:
Nice! LGTM?

Issue #1133, opened 2019-11-08
To improve build_url method in Locust HttpSession


vstepanov-lohika-tix:
### Is your feature request related to a problem? Please describe.
In Locust **clients.py** there is a custom **_build_url** method for HttpSession that returns:
`return "%s%s" % (self.base_url, path)`

It leads to errors if additional path is not starting from slash "/"

### Describe the solution you'd like
It will be more correct to use good and well-tried **urljoin** method from urllib.parse:
`return urljoin(self.base_url, path)`
So it will work properly with and without leading slash in path.

vstepanov-lohika-tix:
PR: https://github.com/locustio/locust/pull/1134

PR #1134, opened 2019-11-08
update build_url function in Locust HttpSession 

vstepanov-lohika-tix:
minor update for build_url function in HttpSession 
related to issue:
https://github.com/locustio/locust/issues/1133


heyman:
Thanks!

PR #1088, opened 2019-09-22
Allow None response time for requests

cyberw:
fixes #1087

PR #1144, opened 2019-11-13
Add response_length to request_failure event

cyberw:
Like with request_success... This breaks compatibility with existing tests that use event handlers with positional arguments. Maybe it is worth a bumped minor version, to indicate that to users.

### DIFF
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -143,6 +143,7 @@ class HttpSession(requests.Session):
                     request_type=request_meta["method"], 
                     name=request_meta["name"], 
                     response_time=request_meta["response_time"], 
+                    response_length=request_meta["content_size"],
                     exception=e, 
                 )
             else:
@@ -251,6 +252,7 @@ class ResponseContextManager(LocustResponse):
             request_type=self.locust_request_meta["method"],
             name=self.locust_request_meta["name"],
             response_time=self.locust_request_meta["response_time"],
+            response_length=self.locust_request_meta["content_size"],
             exception=exc,
         )
         self._is_reported = True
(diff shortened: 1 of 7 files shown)


heyman:
Nice! The `response_length` should also be passed along to the `log_error()` call in `on_request_failure` on this line:

https://github.com/locustio/locust/blob/795b5a14dd5b0991fec5a7f96f0d6491ce19e3d0/locust/stats.py#L616

> Maybe it is worth a bumped minor version, to indicate that to users.

Yeah, I think that the next version we release should be 0.13 (the new wait time API have now also been merged).

heyman:
Hmm, we might also need something like #1088, but for response_length

cyberw:
> Hmm, we might also need something like #1088, but for response_length

Responses that have an undefined response_length? Maybe... but I think we're ok with counting them as zero for now. Or do you feel this relates to the PR? As we dont calculate average response length, it doesnt really hurt that undefined response length cannot be separated from zero response times.

heyman:
> As we dont calculate average response length

We do calculate the average response length:

image

We've been using 0 for failures up until now (since it wasn't available in the event params), so it wouldn't make it *more* wrong to keep doing it, though maybe it would be worth fixing it now when it *is* available?

cyberw:
Oh, I didnt see it in the StatsEntry object, so I thought it wasnt used. Maybe I should start using the web UI more :)

But it looks like it is already fixed by my changes?

image

heyman:
> But it looks like it is already fixed by my changes?

Ah, nice!

Then I guess we can merge this, and perhaps create a separate issue for excluding failures with None as response_length, from the average response size.

PR #1148, opened 2019-11-15
Revert "update build_url function in Locust HttpSession "

vstepanov-lohika-tix:
Reverts locustio/locust#1134
Unfortunately, I've found a big disadvantage of this change:
 `urljoin` from `urllib` has a surprising behavior: it cuts everything in base url after "/", if the second part of url to join starts from "/".
For example, `urljoin("http://mysite.com/api/v0", "/path/to/join")" returns a result:
`http://mysite.com/path/to/join`

It will lead to a lot of issues and errors in all existing locust tests where combined base url is used.
The only option I see is to revert this change. Sorry for inconvenience.

### DIFF
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -8,7 +8,7 @@ from requests.auth import HTTPBasicAuth
 from requests.exceptions import (InvalidSchema, InvalidURL, MissingSchema,
                                  RequestException)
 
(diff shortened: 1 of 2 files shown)


heyman:
Ah, that's unfortunate. Good that you catched it!

Issue #1149, opened 2019-11-15
Be able to install pip packages using the docker image


max-rocket-internet:
The current docker image uses a non-root user: https://github.com/locustio/locust/blob/master/Dockerfile#L18

That's great but it means it's not possible to install pip packages using this docker image:

```
$ docker run -it locustio/locust:0.13.0 sh -c "pip install --user boto"

WARNING: The directory '/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
ERROR: Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/cli/base_command.py", line 153, in _main
    status = self.run(options, args)
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/commands/install.py", line 338, in run
    wheel_cache = WheelCache(options.cache_dir, options.format_control)
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/cache.py", line 219, in __init__
    self._ephem_cache = EphemWheelCache(format_control)
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/cache.py", line 195, in __init__
    self._temp_dir = TempDirectory(kind="ephem-wheel-cache")
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/utils/temp_dir.py", line 57, in __init__
    path = self._create(kind)
  File "/usr/local/lib/python3.6/site-packages/pip/_internal/utils/temp_dir.py", line 90, in _create
    tempfile.mkdtemp(prefix="pip-{}-".format(kind))
  File "/usr/local/lib/python3.6/tempfile.py", line 360, in mkdtemp
    prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
  File "/usr/local/lib/python3.6/tempfile.py", line 130, in _sanitize_params
    dir = gettempdir()
  File "/usr/local/lib/python3.6/tempfile.py", line 298, in gettempdir
    tempdir = _get_default_tempdir()
  File "/usr/local/lib/python3.6/tempfile.py", line 233, in _get_default_tempdir
    dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
[...]

PR #1150, opened 2019-11-15
Docker: add home directory for locust user to install pip packages

max-rocket-internet:
If the user doesn't have a home directory then `pip install --user xxx` fails as there's nowhere to write to.

Resolves https://github.com/locustio/locust/issues/1149


max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/1149

heyman:
Nice!

PR #1155, opened 2019-11-17
fix spelling error

sherry255:


heyman:
From https://www.grammar.com/behavior_vs._behaviour:

image

PR #1159, opened 2019-11-18
Allow locust to get SIGTERM(aka Ctrl+C) messages.

turgayozgur:
exec used to start so from now, it will be the top process and can handle the SIGTERM messages.


heyman:
Thanks!

PR #1165, opened 2019-11-19
Add toolbox control for for downloading chart as png

skivis:
I find myself taking screenshots of the charts quite often, this adds a small button in the top right corner of the charts for downloading them as PNG images.

download-button


cyberw:
nice!

RayRay10460:
Hi, can I download chart with http API?

LucianLiu6:
> Hi, can I download chart with http API?

I don't think so, this PR just add a config item of [echartsjs](https://echarts.apache.org), which just works on the frontend

Issue #1166, opened 2019-11-20
Set locust parameters via env vars & config file


cyberw:
It would be useful to be able to set locust's command line parameters via env vars (particularly to be able to provide defaults, to shorten the command that you have to type).

I suggest we use LOCUST_<long param name> for env vars.

Command line should override env vars of course.

I can make a PR if this sounds good to everyone. I'd probably use https://pypi.org/project/ConfigArgParse/ to do it.

heyman:
I can't think of any drawbacks so I'm +1/+0.

cyberw:
PR:ed. What does +0 mean? :)

heyman:
> What does +0 mean? :)

I believe it's the Django project that uses it (or at least used to use it) to indicate that you are positive to something, but don't feel strongly about it (it wouldn't matter so much to you that you're willing to champion it). I think they havethe policy that a -0 (or -1 obviously) from any core developer would have to be converted to a +0 or +1 before it can get implemented.

cyberw:
Fixed!

PR #1167, opened 2019-11-20
Use ConfigArgParse instead of argparse, to support getting parameters from config file and/or env vars.

cyberw:
Solves #1166

### DIFF
--- a/locust/test/test_parser.py
+++ b/locust/test/test_parser.py
@@ -1,5 +1,6 @@
 import unittest
 import os
+import tempfile
 
 from locust.main import parse_options
 
@@ -35,14 +36,13 @@ class TestParser(unittest.TestCase):
         self.assertEqual(opts.skip_log_setup, True)
 
     def test_parameter_parsing(self):
-        conf_file = '/tmp/locust.conf'
-        os.environ['LOCUST_LOCUSTFILE'] = "locustfile_from_env"
-        with open(conf_file, 'w') as file:
-            file.write("host host_from_config\nweb-host webhost_from_config\n")
-        parser, _ = parse_options(default_config_files=['/tmp/locust.conf'])
-        options = parser.parse_args(['-H','host_from_args'])
-        del os.environ['LOCUST_LOCUSTFILE'] # remove this so it doesnt impact later tests
-        os.remove(conf_file)
+        with tempfile.NamedTemporaryFile(mode='w') as file:
+            os.environ['LOCUST_LOCUSTFILE'] = "locustfile_from_env"
+            file.write("host host_from_config\nweb-host webhost_from_config")
+            file.flush()
+            parser, _ = parse_options(default_config_files=[file.name])
+            options = parser.parse_args(['-H','host_from_args'])
+            del os.environ['LOCUST_LOCUSTFILE']
         self.assertEqual(options.web_host, 'webhost_from_config')
         self.assertEqual(options.locustfile, 'locustfile_from_env')
         self.assertEqual(options.host, 'host_from_args') # overridden
--- a/locust/test/test_parser.py
+++ b/locust/test/test_parser.py
@@ -1,11 +1,12 @@
 import unittest
+import os
 
 from locust.main import parse_options
 
 
 class TestParser(unittest.TestCase):
     def setUp(self):
-        self.parser, _ = parse_options()
+        self.parser, _ = parse_options(default_config_files=[])
 
     def test_default(self):
         opts = self.parser.parse_args([])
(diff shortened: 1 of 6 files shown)


cyberw:
I like how this change only required me to change 3 lines of code (or really just one line of code, and two lines regarding imports)

:heart: Python

heyman:
Great!

It'd be nice to have a simple test for the arg parsing, and one that checks that  the environment variable settings work. We would probably have to make `main.parse_options()` take an additional optional `args` argument which should be passed along to `parser.parse_args`, in order to achive that.

Also, maybe it's worth adding a short example to the Documentation as well? For someone unfamiliar with ConfigArgParse I imagine that [this section](https://github.com/bw2/ConfigArgParse#config-file-syntax) can seem a bit dense.

cyberw:
Hmm... I tried adding a test case but ended up breaking parse_options instead. I dont think we need a test, because we're only relying on normal ConfigArgParse functionality.

There is a short example to the docs, do you think I should add more? I dont want to bloat the "quickstart" section. The link goes to the right place for configargparse on desktop (which should be readable to a programmer), but doesnt work on mobile. Maybe that's what happened?

cyberw:
What I could do (without changing the method signature & breaking stuff) is test reading from config & env vars (but not overriding on command line)

heyman:
> I dont think we need a test, because we're only relying on normal ConfigArgParse functionality.

As long as ConfigArgParse doesn't change their API or break in some future version, it should be fine. But I still think it would be nice to have an integration test for that. 

I've added a test for `parse_options()` to master in 009d104d9f2f689e4f4cbda335dd0aed57030380. You can merge in master and then make another test for overriding some argument in env variables.

> There is a short example to the docs, do you think I should add more?

Ah, nice! For some reason I had missed the environment variable example. 

> The link goes to the right place (config file syntax explanation) for configargparse on desktop (which should be readable to a programmer)

To me that page feels much more targeted towards users of ConfigArgParse (which makes sense). The first paragraph talks about implementation details of ConfigArgParse. Also it mentions two different file syntaxes for config files. I assume we use the default one, but maybe not clear to Locust end users? Though we could always improve the documentation later.

cyberw:
> > I dont think we need a test, because we're only relying on normal ConfigArgParse functionality.
> 
> As long as ConfigArgParse doesn't change their API or break in some future version, it should be fine. But I still think it would be nice to have an integration test for that.
> 
> I've added a test for `parse_options()` to master in [009d104](https://github.com/locustio/locust/commit/009d104d9f2f689e4f4cbda335dd0aed57030380). You can merge in master and then make another test for overriding some argument in env variables.

Will do! I figured out that the issue I was having was with config files, not with adding an argument to parse: My main config file, in ~/.locust.conf) was interfering with the tests! 

I'm looking at a workaround, either by temporarily changing $HOME in the tests (ugly) or detecting that we are running a test in parse_options (also ugly)

> > The link goes to the right place (config file syntax explanation) for configargparse on desktop (which should be readable to a programmer)
> 
> To me that page feels much more targeted towards users of ConfigArgParse (which makes sense). The first paragraph talks about implementation details of ConfigArgParse. Also it mentions two different file syntaxes for config files. I assume we use the default one, but maybe not clear to Locust end users? Though we could always improve the documentation later.

Users can use any of the different syntaxes, ConfigArgParse detects which one you are using.

heyman:
Maybe you could add a `default_config_files` argument to `parse_options()` and use that to specify some temporary file in the test?

cyberw:
> Maybe you could add a `default_config_files` argument to `parse_options()` and use that to specify some temporary file in the test?

Good idea. I'll try that.

cyberw:
LGTY?

On a slightly unrelated topic, I'd like to make the following change:
image

... because the tox-installed python annoys MacOS's software firewall (every time I run I get a dialogue asking for permission to open the port). A unit test has no business opening public ports anyway :)

locust/test/test_parser.py
```diff
@@ -32,3 +33,16 @@ def test_skip_log_setup(self):
         ]
         opts = self.parser.parse_args(args)
         self.assertEqual(opts.skip_log_setup, True)
+
+    def test_parameter_parsing(self):
+        conf_file = '/tmp/locust.conf'
```
heyman: Can we use NamedTemporaryFile here? (https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile)

cyberw: Sure, I'll change that!

heyman:
> A unit test has no business opening public ports anyway :)

Agreed. The second change you took a screenshot of should also use 0 as port, instead of hardcode 8888.

cyberw:
@heyman LGTY?

locust/test/test_zmqrpc.py
```diff
@@ -40,7 +40,8 @@ def test_client_recv(self):
         self.assertEqual(msg.node_id, 'identity')
 
     def test_client_retry(self):
-        server = zmqrpc.Server('0.0.0.0', 8888)
+        # use non-zero port this time, just to increase code coverage
+        server = zmqrpc.Server('127.0.0.1', 8888)
```
heyman: We should change the hardcoded port `8888` to `0`.

heyman:
Looks good!

locust/test/test_zmqrpc.py
cyberw: i changed it and then changed it back, because codecov mentioned that it reduces test coverage.

heyman: That's odd. It's still something we should change I think. To avoid false negative failed tests when port 8888 is already in use.

cyberw: Done. Pushed to master.

Issue #220, opened 2014-12-16
Alternative terminology for "slave"


clarkbreyman:
Please consider alternative terminology for "slave" (similar to https://github.com/django/django/pull/2692).  "Collaborator" seems a possible choice, albeit long. "Worker" is also a possible choice.

cyberw:
Closing this for now. There was a (not very good) PR & some discussion a while back to adress this, but I think we wont change this any time soon. If there is ever a new terminology that is as widely known & understood as master-slave then we should have a look again.

heyman:
Let's change the terminology to Master-Worker - or perhaps Primary-Worker - for 1.0.

max-rocket-internet:
> change the terminology to Master-Worker

Perfect. It's not too different to the current terminology, it's not offensive, I think `master/worker` is good 👍

heyman:
Fixed by #1303

PR #1173, opened 2019-11-27
renamed slave to drone

transformer11:
# Motivation
Be more inclusive and support better open source practices remove the offensive term 'slave' most often used to reference human rights abuses. See previous examples of similar name changes by other leading open source projects: Python (https://bugs.python.org/issue34605), Drupal (https://www.drupal.org/node/2275877), Django (https://github.com/django/django/pull/2692), Jenkins (https://issues.jenkins-ci.org/browse/JENKINS-29522) et al.

# Proposed Resolution
Rename word 'slave' to 'drone' as seems to match Locust's use case. Industry recommendations for alternate terms: https://www.theserverside.com/opinion/Master-slave-terminology-alternatives-you-can-use-right-now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 
 **Fixed bugs:**
 
-- autoscaling slaves resets users [\#1143](https://github.com/locustio/locust/issues/1143)
+- autoscaling drones resets users [\#1143](https://github.com/locustio/locust/issues/1143)
 - Repeated secure requests with FastHttpLocust crashes in cookie management [\#1138](https://github.com/locustio/locust/issues/1138)
 - FastHttpLocust gives ssl error with let's encrypt certs [\#1137](https://github.com/locustio/locust/issues/1137)
 - stop\_timeout defined in Locust class takes precedence over --run-time option [\#1117](https://github.com/locustio/locust/issues/1117)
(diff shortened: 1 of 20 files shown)


cyberw:
I think "drone" would be terrible name for what is currently called a "slave". Locust terminology is confusing enough with its "clients/locusts" and "swarming". Would anyone instinctively know what is the difference between a drone and a locust, for instance?

Maybe "worker" would be ok, but right now I think we should just not change it.

max-rocket-internet:
I like the PR.

> I think "drone" would be terrible name for what is currently called a "slave"

I think it's fine. I did a quick search but couldn't find something better. Perhaps `primary/replica` but that implies a different relationship. All the other options sound off to me.

> Locust terminology is confusing enough with its "clients/locusts" and "swarming"

1000% this. It's very confusing.

heyman:
There's an existing issue for this: #220. I've reopened it (I guess it got closed because of lack of activity).

I don't like `Drone`, but I think `Worker` would be fine.

heyman:
Note that this would break a lot of backwards compatibility with existing test environments. It would be perfect to do before a 1.0 release or something, but perhaps it could be acceptable in a new 0.x release as well?

cyberw:
I think we should do it as part of 1.0. I wouldnt mind doing a general clean up of terminology at the same time (favouring more standard terminology like "start" instead of "hatch", "ramp up" instead of "hatch rate" etc).

This PR also does some weird things like alter the change log. It looks a lot like just a global search-replace...

cyberw:
Also, if we're going to pollute the git history/blame of pretty much every file we might as well start to do auto-formatting at the same time :)

heyman:
Ok, let's do this properly for 1.0. I'll close this PR and we can continue the discussion in #220.

PR #1189, opened 2019-12-08
Set names of Locust classes to run from web or from master to slaves

mjstrasser:
# Summary

Set the names of Locust classes to run, either from the web interface or from the master command line.

This change means that slaves can be started without specifying which Locust classes they will run. Instead they are instructed to hatch only certain Locust classes by the master. The list of classes can come from the master command line or from the web interface.

## Web mode

A set of slaves is started with a locustfile that contains a number of Locust classes. The master is started in web mode with the same locustfile. The web form includes a text box with a comma-separated list of available class names.

If a subset of Locust class names is entered in the box, only those classes are run by all the slaves. Unrecognised strings are ignored.

A subset of Locust classes can also be run in local mode.

## Non-web mode

In non-web mode, slaves quit when the master quits. 

But slaves are instructed which Locust classes to run by the master so they can be started with the same command line. For example, a slave can be started with:
```bash
locust --slave --master-host=$MASTER_HOST -f locustfile.py
```
and will run different Locust classes in these two cases:
```bash
locust --master --host=$TARGET_HOST -c 10 -r 1 -t 20m -f locustfile.py Locust1
locust --master --host=$TARGET_HOST -c 10 -r 1 -t 20m -f locustfile.py Locust2
```

This makes it possible to have self-restarting slaves (e.g. in a script loop).

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -245,7 +245,7 @@ class LocalLocustRunner(LocustRunner):
             self.log_exception("local", str(exception), formatted_tb)
         events.locust_error += on_locust_error
 
-    def start_hatching(self, locust_count=None, hatch_rate=None, wait=False, class_names=False):
+    def start_hatching(self, locust_count=None, hatch_rate=None, wait=False, class_names=None):
         self.hatching_greenlet = gevent.spawn(lambda: super(LocalLocustRunner, self).start_hatching(locust_count, hatch_rate, wait=wait, class_names=class_names))
         self.greenlet = self.hatching_greenlet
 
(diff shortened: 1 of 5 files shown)


cyberw:
Hi! 

Thanks for your contribution! Unfortunately this seems to me like a too narrow/rare use case, especially since the slaves will still need to be aware of all the locust classes at startup. 

I'm -1 for this, it just doesn't motivate the increased complexity.

If you need an easy way to spawn slaves with the correct locust then maybe have a look at https://github.com/SvenskaSpel/locust-swarm

mjstrasser:
Interesting that this is considered a rare use case. 

In our usage, we create one master and 20 slaves in AWS Fargate. They all load the same locustfile with a set of 5 different Locust classes. We can use this same setup to select different load scenarios based on Locust classes, without having to redeploy masters and slaves. We will continue doing this using our own fork of the code.

I had not heard of locust-swarm or locust-plugins before. Perhaps you could add links to the Locust README.

PR #1199, opened 2019-12-18
Allow min_wait and max_wait times of 0

Aresius423:
I know this is a deprecated feature, but here's a fix anyways.


cyberw:
Thanks for your contribution!

PR #1204, opened 2019-12-22
Allow users to override encoding when decoding requests (maybe using chardet, as was the default before)

### LINKED ISSUE #1193: FastHttpLocust is very slow at returning the response text
Getting resp.text when using FastHttpLocust does a string decode (fasthttp.py:257) and it is crazy slow for large responses (~12s for a 1.5MB payload).

        with self.client.request(method, f"https://{host}{url}", name, catch_response=True) as resp:
            t = resp.text: # this takes ages
            ....
            return resp

cyberw:
maybe @mbeacom or @skivis has feedback on this? Particulary regarding the original change that was made to fix #1193

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -246,28 +246,20 @@ class FastResponse(CompatResponse):
     
     _response = None
     
+    encoding = None
+    """In some cases setting the encoding explicitly is needed. If so, do it before calling .text"""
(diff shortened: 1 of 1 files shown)


cyberw:
I dont want to wait any longer, so I'm merging this. If anyone (@mbeacom @skivis or someone else) has comments, let me know.

PR #1212, opened 2020-01-02
Update running-locust-distributed.rst

Neamar:
Fix a typo in documentation

(not updating the changelog and not creating a new version since it doesn't seem important enough)


cyberw:
thanks!

PR #1216, opened 2020-01-06
Remove leading backslashs from URLs.

TrevisGordan:
In code Sample:
```python
@task(1)
def profile(self):
    self.client.get("/profile") -> will curently result in ..host.com//profile
    # self.client.get("profile") will do ..host.com/profile
```
therefore removed leading backslashes from code samples.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -13,16 +13,16 @@ Below is a quick little example of a simple **locustfile.py**:
     from locust import HttpLocust, TaskSet, between
 
     def login(l):
-        l.client.post("/login", {"username":"ellen_key", "password":"education"})
+        l.client.post("login", {"username":"ellen_key", "password":"education"})
 
     def logout(l):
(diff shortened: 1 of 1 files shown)


cgoldberg:
I think the issue is that you are specifying a host with a trailing slash (i.e. `https://example.com/` rather than `https://example.com`).  Locust builds the url by blindly concatenating the Host and URL, which can lead to issues like this.

Rather than changing the example code, I think it would be better to make Locust tolerant of trailing slashes in the specified host.  AFAICT, this is a one line code change... I'll submit a Pull Request in a few mins to fix this.  Once fixed, we can kindly reject this PR.  Thanks for pointing this out!

cgoldberg:
fix in #1217 supersedes this PR.

TrevisGordan:
upsi , yes. I figured this just now too. 
and yes it would be better to fix it like that.

PR #1219, opened 2020-01-07
Fix link formatting for the Helm chart

TBBle:
The link was in Markdown, rather than reStructuredText.

### DIFF
--- a/docs/running-locust-docker.rst
+++ b/docs/running-locust-docker.rst
@@ -56,4 +56,4 @@ To run in standalone mode without the web UI, you can use the ``LOCUST_OPTS`` en
     docker run --volume $PWD/dir/of/locustfile:/mnt/locust -e LOCUSTFILE_PATH=/mnt/locust/locustfile.py -e TARGET_URL=https://abc.com -e LOCUST_OPTS="--clients=10 --no-web --run-time=600" locustio/locust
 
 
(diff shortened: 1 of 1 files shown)


docs/running-locust-docker.rst
```diff
@@ -56,4 +56,4 @@ To run in standalone mode without the web UI, you can use the ``LOCUST_OPTS`` en
     docker run --volume $PWD/dir/of/locustfile:/mnt/locust -e LOCUSTFILE_PATH=/mnt/locust/locustfile.py -e TARGET_URL=https://abc.com -e LOCUST_OPTS="--clients=10 --no-web --run-time=600" locustio/locust
 
 
-If you are Kubernetes user, you can use the [Helm chart](https://github.com/helm/charts/tree/master/stable/locust) to scale and run locust.
+If you are Kubernetes user, you can use the `Helm chart <https://github.com/helm/charts/tree/master/stable/locust>`_ to scale and run locust.
```
max-rocket-internet: Is it supposed to have an underscore at the end?

TBBle: Yes, that's what marks it as a reference, instead of "random thing that the document processor will interpret".

See the second example at https://docutils.sourceforge.io/docs/user/rst/quickref.html#external-hyperlink-targets.

PR #1221, opened 2020-01-10
Add "Running Locust with Docker" to the TOC

TBBle:
Without this, the only way to find the page is by either search, or happening on the link at the bottom of "Running Locust distributed".

Luckily, Google had indexed the page despite not being in the TOC, or I'd have never seen it.


heyman:
Thanks!

PR #1229, opened 2020-01-16
Disable codecov patch analysis, because it fails all the time

cyberw:


cyberw:
Works. Only filed a PR to see that it did :)

PR #1230, opened 2020-01-16
allow 1 percent codecov degradation before failing build (because it is flaky)

cyberw:


cyberw:
Great, we had just the right amount of flakyness in the build to validate my change :P

PR #1235, opened 2020-01-19
Update docs for running locust in Step Load Mode

delulu:
Change in this PR:

1. Update docs for running locust in Step Load Mode

2. Fix step load bug in local run

    When running locust in standalone way, self.greenlet will point to self.locusts. So we can't use this gevent Group to spawn the stepload_greenlet, otherwise it will be wrongly taken as the running locust.


cyberw:
Thanks!

Issue #1161, opened 2019-11-19
Locust should warn if CPU usage is too high


cyberw:
If locust runs out of CPU, that may limit the number of requests sent and impact the measurement of response times.

Using psutil we can detect high CPU usage and warn the user (in the web gui and on command line) if CPU usage exceeds a certain threshold (and suggest that the user run multiple processes or examine their test plan).

I suggest we warn if CPU exceeds 90% over a 10 second timespan. 

On the command line we can warn only twice: when the threshold was exceeded and at the end of the test.

In web gui we could have two different indicators: "currently above threshold" and "threshold was exceeded at some point during the test".

But we can build a command line-only version first.

heyman:
I think it's a good idea!

> In web gui we could have two different indicators: "currently above threshold" and "threshold was exceeded at some point during the test".

I think it's better to just output the usage number somewhere, and then the usage number for each slave node in "Slaves" tab. We could make it red (or highlight it in some other way) if it exceeds some threshold.

heyman:
We should also consider doing this for memory usage as well.

PR #1236, opened 2020-01-20
Warn if CPU usage is too high (>90%) #1161

cyberw:


cyberw:
In the future we could (should):
* make the threshold configurable
* add GUI support

But lets leave that for nother day...

locust/runners.py
```diff
@@ -180,11 +184,23 @@ def kill_locust_greenlets(self, greenlets):
...
+        
+    def cpucheck(self):
+        process = psutil.Process()
+        while True:
+            if not self.cpu_threshold_exceeded: 
+                current_cpu = process.cpu_percent()
+                if current_cpu > 90:
+                    self.cpu_threshold_exceeded = True
+                    logging.warning("Loadgen CPU usage above 90%! This may constrain your throughput and even give inconsistent response time measurements! See https://docs.locust.io/en/stable/running-locust-distributed.html for how to distribute the load over multiple CPU cores or machines")
+            gevent.sleep(2.0)
```
heyman: Maybe we should increase the interval to something like 10 seconds? That would prevent a warning if you have a single task that does something really CPU heavy for a few seconds, and it shouldn't really matter if you get a warning after 2 or 10 seconds. 

Also, do we know how much overhead (if any) `Process.cpu_percent()` introduces?

cyberw: I was thinking about the interval, but decided against increasing it, because even a short spike can cause incorrect measurements. I was thinking it is better to warn a few times too many than a few times too few...

I dont know exactly how much overhead it introduces, but I did a few tests and the call itself never took more than 0.2ms to run on my 2018 MacBook pro. I think it just reading a counter somewhere...

If you want, I can bump the interval to 5 seconds.

heyman: Ok, five seconds sounds good!

heyman:
When implementing this in the web UI, I think it would make sense to add the current CPU usage for each node to the "Slaves" tab. Therefore, maybe it would make sense to add `cpu_usage` to `locust.runners.SlaveNode`, and then have all the logic for emitting warnings run in the master?

cyberw:
> When implementing this in the web UI, I think it would make sense to add the current CPU usage for each node to the "Slaves" tab. Therefore, maybe it would make sense to add `cpu_usage` to `locust.runners.SlaveNode`, and then have all the logic for emitting warnings run in the master?

Hmm... yes, having the logic on the master side makes sense. 

The only problem is that if we want to log a value instead of just a flag of whether the threshold was exceeded then we may need to synchronize sending with measuring (or we might detect high usage but never send that particular metric, and thus miss a spike).

To fix this (without adding complex logic) could move the cpu checking for slaves to the heartbeat method and only run cpu_monitoring_greenlet on master + standalone?

I can make it so that we only update the cpu usage on every 5th heartbeat.

heyman:
Since the heartbeat interval is less than the measurement interval (at least if we go with 5 seconds), I think it should be fine to have the `cpu_monitoring_greenlet` store the value on the runner instance, and then include that value in the heartbeat message.

cyberw:
ok, I'll do it that way!

cyberw:
LGTY? I havent re-tested distributed yet, but I'll do it before I merge..

locust/runners.py
```diff
@@ -40,6 +41,10 @@ def __init__(self, locust_classes, options):
         self.state = STATE_INIT
         self.hatching_greenlet = None
         self.stepload_greenlet = None
+        self.current_cpu_usage = 0
+        self.cpu_threshold_exceeded = False
+        self.slave_cpu_threshold_exceeded = False
+        gevent.spawn(self.monitor_cpu)
```
heyman: Hmm, since we don't store any reference to this greenlet, it won't get killed. We should probably change so that the `LocalLocustRunner.greenlet` is a `gevent.pool.Group` instance (just like `MasterLocustRunner` and `SlaveLocustRunner`) and then make sure the CPU monitor greenlet is spawned from this group.

locust/runners.py
```diff
@@ -293,6 +309,7 @@ def __init__(self, id, state=STATE_INIT, heartbeat_liveness=3):
         self.state = state
         self.user_count = 0
         self.heartbeat = heartbeat_liveness
+        self.cpu_threshold_exceeded = False
```
heyman: I think it would be better to store the CPU usage value on the `SlaveNode` and track on the runner if we've emitted a warning. Then it'll be very easy to display CPU usage on the slave nodes tab in the web UI.

locust/runners.py
```diff
@@ -40,6 +41,10 @@ def __init__(self, locust_classes, options):
         self.state = STATE_INIT
         self.hatching_greenlet = None
         self.stepload_greenlet = None
+        self.current_cpu_usage = 0
+        self.cpu_threshold_exceeded = False
+        self.slave_cpu_threshold_exceeded = False
```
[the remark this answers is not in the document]
heyman: Hmm, for some reason Github has highlighted three rows, though I thought I only seleted the last one.

locust/runners.py
cyberw: Hmm... I'm not sure how to do that. LocustRunner is supposed to be a singleton, so it shouldnt really matter, right? (not that we want to be sloppy :)

If you think this is important, would you mind taking a look yourself?

heyman: At the moment, it's a singleton when started normally (through main.py). But we do create multiple runner instances within the tests.

Also, it's been proposed, and I think it's a good idea, to work towards an API where one can run Locust programatically, in which case I think the design will be much cleaner if we group the spawned greenlets together (that one can join) and make sure they are killed together.

I can definitely take a look!

cyberw: Thanks!

heyman: I've now changed so that we spawn all greenlets through the runner instance's `greenlet` attribute which is a `gevent.pool.Group` instance.

I also added a test for the CPU warning (and changed so that we use a constant for the montoring interval, so that I could decrease the run time of the test).

PR #1240, opened 2020-01-22
Add some sleeps to fix intermittent unit test errors in test_web.py

cyberw:
Add gevent.sleep(0.01) before each call to /stats/requests to make sure the stats have been recorded (this was causing intermittent issues on travis)

### DIFF
--- a/locust/test/test_web.py
+++ b/locust/test/test_web.py
@@ -50,6 +50,7 @@ class TestWebUI(LocustTestCase):
     
     def test_stats(self):
         stats.global_stats.log_request("GET", "/<html>", 120, 5612)
+        gevent.sleep(0.01)
         response = requests.get("http://127.0.0.1:%i/stats/requests" % self.web_port)
         self.assertEqual(200, response.status_code)
         
@@ -66,6 +67,7 @@ class TestWebUI(LocustTestCase):
         
     def test_stats_cache(self):
         stats.global_stats.log_request("GET", "/test", 120, 5612)
+        gevent.sleep(0.01)
         response = requests.get("http://127.0.0.1:%i/stats/requests" % self.web_port)
         self.assertEqual(200, response.status_code)
         data = json.loads(response.text)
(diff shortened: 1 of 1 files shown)


cyberw:
@heyman I really think this will solve the intermittent issues (they happened again in the IPS PR just now). If we merge this and they happen again then we can remove the sleeps again.

heyman:
I'd like to first understand *why* they errors are happening, before doing something like this. I want to be sure that there isn't some underlying real issue that we would just hide by adding a bunch of sleep calls.

heyman:
> (they happened again in the IPS PR just now)

The last couple of failed tests on the IPS branch on travis seems different from the ones that this PR aims to fix though, if I'm not mistaken?

cyberw:
> > (they happened again in the IPS PR just now)
> 
> The last couple of failed tests on the IPS branch on travis seems different from the ones that this PR aims to fix though, if I'm not mistaken?

There were some other (random) errors as well, but the last one was the one this PR aims to fix.

cyberw:
> I'd like to first understand _why_ they errors are happening, before doing something like this. I want to be sure that there isn't some underlying real issue that we would just hide by adding a bunch of sleep calls.

Sure, if you have the time to do a deeper analysis then feel free. I re-ran this build a bunch of times so does seem to fix the problem, but if you think it could be something deeper than just a little latency between logging the request and the request being available in the web gui, then by all means... To me it seems like a very probable cause though...

heyman:
I'd like to do some digging before going down the nuclear route of adding sleeps everywhere :).

It seems like its almost always the `test_stats` test that fails.

On a side note, I imagine that a `sleep(0)` could work just as well.

heyman:
> if you think it could be something deeper than just a little latency between logging the request and the request being available in the web gui

Thing is that the web UI *should* be reading from the variable that we have *already* changed. So I find the error really strange. Though I might be missing something.

heyman:
I'm not able to reproduce this on Travis. I made a branch which made the tests run over and over again until they fail and wasn't able to reproduce the failing test once, even though I ran one of the tests that have been failing more than 5k times, and the whole test_web.py > 1k times, and the whole test suite > 250 times.

It's possible that either 3b5dab7bb84974878ad5b7f48921df439bf3307f or e1e2b3d7825f9ea27078c53694beb0eb802c2ebf fixed the problem, or it's Travis that is flaky and sometimes spawns containers with an environment that is different and that somehow causes the failing tests.

I suggest we wait and see if the problem persists.

Issue #1120, opened 2019-10-23
Remove support for Python 2.7 & 3.5


cyberw:
Python 2 is no longer maintained (or at least it wont be at the end of this year).

A significant effort is required for any new code to be made 2.7 compatible, and we cant use new awesome features in 3.6 (f-strings ftw! :)

@aldenpeterson raised this in chat a while back, and compatibility was an issue (again) in a recent PR (https://github.com/locustio/locust/pull/1119).

Opening this so we can have a discussion (I'm particularly interested if someone has a big need for 2.7/3.5 support).

heyman:
I'd imagine that there are some old code bases out there that's still on python 2.7. I think we should drop the support at some point, but I'd prefer to keep it around for awhile longer. Since the extra work to support both versions is already in place, and it's usually not *that* much extra job to support both when implementing new features.

(Maybe I'm biased since I have quite a few website codebases (not locust related) in Python 2, and I'm dreading the day when I'll have to spend weeks migrating them)

heyman:
@cyberw Yeah, I think that's true in general. However, I do know of companies that have large old locust test codebases that would still take time to migrate. I'd say let's wait for now and revisit this after the 2.7 has actually reached end of life :).

lucrib:
if someone else agree with me that pytest is likely more used than locust inside python ecosystem, then consider the links below in order to drop support for py2.7 and py3.5.
- https://pytest.readthedocs.io/en/latest/py27-py34-deprecation.html
- https://python3statement.org/

mbeacom:
When I raised [this question](https://locustio.slack.com/archives/C3NUJ61DJ/p1569547661022600) in Slack, I was taking into consideration a codebase owner's ability to pin on the latest version to provide `2.7` support.

If an external repo is dependent on a version of `locust` able to be executed from `2.7`, it would be advisable for them to pin to a `0.12.x` series and we increment the version to a `0.13` or `0.2` minor to indicate the significance of the upgrade.  This would provide us with a great deal of benefits...

Additionally, I assume Locust is more so a stand alone CLI/service in support of other projects, rather than a library being directly used by other projects.

Some additional references in favor of this change:

- [Python Clock - 2 months and 5 days remaining as of this comment](https://pythonclock.org/)
- [Pandas drops Python 2 support - July 2019](https://pandas.pydata.org/#v0-25-0-final-july-18-2019)
[...]

PR #1243, opened 2020-01-24
Drop support for pre 3.6 Python versions (give error during installation)

cyberw:
Future releases will just not show support for these versions (so they wont be installed by pip), but this way we can show users what to do instead of just giving them an old package.

Solves #1120


cyberw:
I'm trying to figure out what is wrong with the docker build, but failing miserably...

Issue #1237, opened 2020-01-20
Retrieve failures.csv in --no-web mode


schorba:
### Describe the solution you'd like
When run locust in --no-web mode with --csv option, in addition to existed  requests_xxxx.csv and stats_history_xxxx.csv files, I'd like to get also failures_xxxx.csv file.

PR #1245, opened 2020-01-27
Save failures.csv in --no-web mode

ajt89:
#1237


cyberw:
Thanks! Is there a test that covers this? Would you consider adding one? Just checking the existence of the files is enough.

schorba:
Thank you a lot !

vstepanov-lohika-tix:
Very useful update, thanks

PR #1255, opened 2020-02-06
docs: remove the description of port 5558

orisano:
I remove the description of port 5558.
locust hadn't bind two sockets since from #927


heyman:
Thanks!

Issue #1264, opened 2020-02-21
Rename and restructure Locust/TaskSet


cyberw:
Some of locusts core concepts have a naming style that makes it unnecessarily confusing. Also, some of the code design doesn't make a lot of sense.

I'd like to:

* Merge the Locust and TaskSet classes into one. Separating the two never made sense to me. The user's Locust class almost never contains a significant number of things, and the things that ARE there are mostly things that should actually be set via command line parameters/config file "orthogonally" to the user behaviour (like wait times & target host). Another benefit of this is that it will make "the most basic locust script you could ever write" a lot shorter & easier to understand, so onboarding new users will be easier.
* Rename this Locust/TaskSet to User
* Change some command line parameters to use industry-standard names (avoid stuff like "hatch rate". Maybe this is also the time to drop the master/slave terminology)

Of course these things should be done carefully and, if possible, leaving the old functionality in place but marked deprecated for a while. I'm not sure if I'll ever have time to implement any of this, but I'd like to talk about it :P

heyman:
> Merge the Locust and TaskSet classes into one

I'm not in favor of merging them. I think they are conceptually different. In the beginning of Locust's development `TaskSet` didn't exist. Instead tasks were declared directly on the Locust class, and one could nest `SubLocust` classes. I think the introduction of the `TaskSet` class made it much better.

> The user's Locust class almost never contains a significant number of things

For most larger test codebases I've worked on the Locust classes has contained a bunch of stuff. Almost always wrapper methods for making HTTP requests, and often some state for the simulated users.

> Rename this Locust/TaskSet to User

I wouldn't mind renaming `Locust` to `User`.

> Change some command line parameters to use industry-standard names (avoid stuff like "hatch rate". Maybe this is also the time to drop the master/slave terminology)

Sounds like a good idea!

cyberw:
> > Merge the Locust and TaskSet classes into one
> 
> I'm not in favor of merging them. I think they are conceptually different. In the beginning of Locust's development `TaskSet` didn't exist. Instead tasks were declared directly on the Locust class, and one could nest `SubLocust` classes. I think the introduction of the `TaskSet` class made it much better.

To me, a TaskSet represents a user behaviour (or "scenario" if you will). What does a Locust represent?

> 
> > The user's Locust class almost never contains a significant number of things
> 
> For most larger test codebases I've worked on the Locust classes has contained a bunch of stuff. Almost always wrapper methods for making HTTP requests, and often some state for the simulated users.
[...]

PR #1274, opened 2020-03-02
Introduce SimpleTaskSet, a way to make super simple locust files without having to define a Locust class

cyberw:
This allows you to define a locust plan like this:

```
from locust import SimpleTaskSet

class MyTask(SimpleTaskSet):
    @task
    def task1(self):
        self.client.get("/1")

```
I havent updated documentation yet, but I think this can be the default in the documentation, because it is 100 times less fuss & unnecessary settings for a new user :)

You could argue that SimpleTaskSet should default to FastHttpLocust instead of HttpLocust, if that is "the future"

I spent a lot of time implementing this but in the end the change is so small, I almost feel bad :)

This relates to #1264


heyman:
Interesting! I like that this can be done with pretty few changes. Here are some initial thoughts:

* Couldn't this be done without changing `main.py` if one just imports the `SimpleHttpLocust` in the locustfile.py?

* Maybe the code could live in a separate module (something like `locust.contrib.simple`) (I think we should also move out `HttpLocust` from core.py, but that's another issue)?

* Here's an alternative API we should consider:

  ```python
  from locust.contrib.simple import SimpleHttpLocust
  
  @SimpleHttpLocust.task
  def my_task(t):
      pass
  ```
  that would also make this possible
    ```python
  from locust.contrib.simple import SimpleHttpLocust, constant
  
  SimpleHttpLocust.wait_time = constant(3)
  
  @SimpleHttpLocust.task
  def my_task(t):
      pass
  ```

> I think this can be the default in the documentation, because it is 100 times less fuss & unnecessary settings for a new user :)

I'm not sure if I think it should be the default in the docs. Locust's power does not come from very small Apache Bench-like load tests, but rather from the ability to create realistic load by simulating real user behaviour. I think that this simplified approach is worse at communicating that. Especially since there's an implicit wait_time of 0. The typical real Locust test isn't 10, 20 or even a 100 lines of code. Therefore I'd say that the current "overhead" is almost negligible, and I think it can be negative if the typical real use-case is very different from the ones presented in examples to new users.

cyberw:
> Interesting! I like that this can be done with pretty few changes. Here are some initial thoughts:
> 
> * Couldn't this be done without changing `main.py` if one just imports the `SimpleHttpLocust` in the locustfile.py?

Yes, that sounds like a very good plan, I will do that!

> * Maybe the code could live in a separate module (something like `locust.contrib.simple`) (I think we should also move out `HttpLocust` from core.py, but that's another issue)?
> * Here's an alternative API we should consider:
>   ```python
>   from locust.contrib.simple import SimpleHttpLocust
>   
>   @SimpleHttpLocust.task
>   def my_task(t):
>       pass
>   ```

Sure, but if I manage to convince you (see below) that it should be the first introduction in the documentation I think it should be part of locust module.

> > I think this can be the default in the documentation, because it is 100 times less fuss & unnecessary settings for a new user :)
> 
> I'm not sure if I think it should be the default in the docs. Locust's power does not come from very small Apache Bench-like load tests, but rather from the ability to create realistic load by simulating real user behaviour. I think that this simplified approach is worse at communicating that. Especially since there's an implicit wait_time of 0. The typical real Locust test isn't 10, 20 or even a 100 lines of code. Therefore I'd say that the current "overhead" is almost negligible, and I think it can be negative if the typical real use-case is very different from the ones presented in examples to new users.

I think most people start by making a very small tests and then expands from there (either because they thought the AB-like test would be enough and then change their minds, or just because that is a natural first step in developing a scenario). 

I think Locust needs to be documented primarily with the beginner in mind (because most of our users are not professional load testers), but allow for growth. I think the documentation can start out with simple_task but move on to the more complex cases later on.

In my experience test plans start out small but can grow pretty big (although I'd say I average around 100 loc, and I still have never used nested tasksets and only rarely more than one Locust class in the same test :) ). But of course, the benefit of simple_task is not decreasing the lines of code but decreasing the complexity and cognitive load on the new locust user.

heyman:
Hmm, another alternative (which I believe you've suggested before), would be to allow task declarations directly on the Locust user class:

```python
class User(Locust):
    @task
    def t1(self):
        pass
```

What I like about that approach is that it feels more similar to the default way of declaring tasks, compared to introducing an additional approach (with a `Locust.task` classmethod that can be used as a decorator to create tasks).

cyberw:
> Hmm, another alternative (which I believe you've suggested before), would be to allow task declarations directly on the Locust user class:
> 
> ```python
> class User(Locust):
>     @task
>     def t1(self):
>         pass
> ```
> 
> What I like about that approach is that it feels more similar to the default way of declaring tasks, compared to introducing an additional approach (with a `Locust.task` classmethod that can be used as a decorator to create tasks).

Hmm.. but I thought you didnt like that? :) I guess that could work, but I think it would require more changes under the hood. But I guess it could be done. It would have to be a different decorator than the usual `@task` though. This one would have to check if there is a TaskSet defined on the locust, create if it isnt there and then add itself there. 

But what would the self reference point to? The task set instance? That would be very weird (because you are defining it on the locust) The locust? That would be ok, but very different from the normal behaviour. And could be hard to implement.

>> Interesting! I like that this can be done with pretty few changes. Here are some initial thoughts:
>> Couldn't this be done without changing main.py if one just imports the SimpleHttpLocust in the locustfile.py?
> Yes, that sounds like a very good plan, I will do that!

Hmm... I tried that but couldnt get it to work. The tasks need to be defined before calling TaskSet.__new__ due to the magic meta-class stuff. I can of course do it using a different decorator (like `@simple_task`), but that is almost as magical as doing it on module level (also, people might get confused on when to call `@simple_task` and when to call `@task`).

cyberw:
> > > Interesting! I like that this can be done with pretty few changes. Here are some initial thoughts:
> > > Couldn't this be done without changing main.py if one just imports the SimpleHttpLocust in the locustfile.py?
> 
> > Yes, that sounds like a very good plan, I will do that!
> 
> Hmm... I tried that but couldnt get it to work. The tasks need to be defined before calling `TaskSet.__new__` due to the magic meta-class stuff. I can of course do it using a different decorator (like `@simple_task`), but that is almost as magical as doing it on module level (also, people might get confused on when to call `@simple_task` and when to call `@task`).

I guess I would be ok with this though, if you think it is better, and worthy of making it the "default" :)

heyman:
> Hmm.. but I thought you didnt like that?

I'm not sure that I do, though I still might prefer it to introducing an additional approach for declaring tasks. The more I think about it, the more I feel skeptical about having two different ways of declaring tasks, and allowing `@task` on the `Locust` class would make the hello world example simpler, while still diverging less from the default way of declaring tasks.

> But what would the self reference point to? The task set instance? That would be very weird (because you are defining it on the locust) The locust? That would be ok, but very different from the normal behaviour. And could be hard to implement.

Yes, that's all things that we would need to consider, and it's very much part of the reasons why I'm not sure if it's a good idea. 

I think that the only logical thing would be to have `self` point to the `Locust` instance. I'm not sure how much it would complicate the implementation.

> But I guess it could be done. It would have to be a different decorator than the usual @task though.

I think that it should be possible to do it with the same decorator.

cyberw:
hmm... that last commit broke stuff... will try to fix it...

cyberw:
So I'm quite happy with the change now. What could make it even easier for the user is if we just made this the default (so that it happens even when subclassing TaskSet, instead of SimpleTaskSet)

If we had just made constant(0) the default sleep time when it was introduced it would have been one less hacky thing to do in main.py :)

cyberw:
I think this would be a great simplification for new users of locust and I want it in as part of 1.0 (I'll update the docs too of course)

Maybe @max-rocket-internet or @timran1 want to weigh in?

cyberw:
@heyman Pretty please with sugar on top? :P I really think this change will do a lot of good for Locust's usability.

cyberw:
Closing this in favour of #1304

Issue #1158, opened 2019-11-18
Slave count stuck at 3 instead of decreasing to 1 due to "missing"


max-rocket-internet:
Master logs after slaves scaling down:

```
[2019-11-18 16:57:02,829] locust-opa-master-787476b668-74txj/INFO/locust.main: Starting web monitor at *:8089
[2019-11-18 16:57:02,830] locust-opa-master-787476b668-74txj/INFO/locust.main: Starting Locust 0.13.1
[2019-11-18 16:57:24,100] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-cjkwb_594d3fd33f864c34aaabbb987cd79e65' reported as ready. Currently 1 clients ready to swarm.
[2019-11-18 16:57:24,210] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-pljft_240da2aa2ada4a0993fad606fb5976e0' reported as ready. Currently 2 clients ready to swarm.
[2019-11-18 16:57:24,221] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-zckdc_b42adb57f2d34ef885b0bcc228068102' reported as ready. Currently 3 clients ready to swarm.
[2019-11-18 16:57:24,253] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-z2g29_7772633197e6471b9344951aba939c39' reported as ready. Currently 4 clients ready to swarm.
[2019-11-18 16:57:24,356] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-d95g2_a3a40cf055d749ea8b6c0e3246f22faa' reported as ready. Currently 5 clients ready to swarm.
[2019-11-18 16:57:24,381] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-mm7dt_ee79eead69c84802a096a12f6e6ef1e5' reported as ready. Currently 6 clients ready to swarm.
[2019-11-18 16:57:24,388] locust-opa-master-787476b668-74txj/INFO/locust.runners: Client 'locust-opa-slave-86cf6df9fb-b7c2k_a961e9574df048739d1057d70dff35bd' reported as ready. Currently 7 clients ready to swarm.
[...]

PR #1280, opened 2020-03-09
ensure the connection between master and slave in heartbeat

delulu:
With heartbeating enabled, I still noticed network issues (packet drop, invalid byte stream) in long-term running on k8s cluster (overlay network). 

And a straightforward fix is to reestablish the connection when any network issue is detected.

I've applied this fix in my locust tests for one week's running, and it works as expected.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -495,7 +495,8 @@ class WorkerLocustRunner(DistributedLocustRunner):
     def __init__(self, *args, master_host, master_port, **kwargs):
         super().__init__(*args, **kwargs)
         self.client_id = socket.gethostname() + "_" + uuid4().hex
-        
+        self.master_host = master_host
+        self.master_port = master_port
         self.client = rpc.Client(master_host, master_port, self.client_id)
         self.greenlet.spawn(self.heartbeat).link_exception(callback=self.noop)
         self.greenlet.spawn(self.worker).link_exception(callback=self.noop)
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -14,6 +14,8 @@ from gevent.pool import Group
 from .rpc import Message, rpc
 from .stats import RequestStats, setup_distributed_stats_event_listeners
 
+from .exception import RPCError
+
 logger = logging.getLogger(__name__)
 
 
@@ -426,15 +428,15 @@ class MasterLocustRunner(DistributedLocustRunner):
         try:
             self.server.close()
             self.server = rpc.Server(self.master_bind_host, self.master_bind_port)
-        except Exception as e:
-            logger.error("Exception found when resetting connection: %s" % ( e ) )
+        except RPCError as e:
+            logger.error("Temporay failure when resetting connection: %s, will retry later." % ( e ) )
 
     def client_listener(self):
         while True:
             try: 
                 client_id, msg = self.server.recv_from_client()
-            except Exception as e:
-                logger.error("Exception found when receiving from client: %s" % ( e ) )
+            except RPCError as e:
+                logger.error("RPCError found when receiving from client: %s" % ( e ) )
                 self.connection_broken = True
                 gevent.sleep(FALLBACK_INTERVAL)
                 continue
@@ -525,8 +527,8 @@ class WorkerLocustRunner(DistributedLocustRunner):
         while True:
             try:
                 self.client.send(Message('heartbeat', {'state': self.worker_state, 'current_cpu_usage': self.current_cpu_usage}, self.client_id))
-            except Exception as e:
-                logger.error("Exception found when sending heartbeat: %s" % ( e ) )
+            except RPCError as e:
+                logger.error("RPCError found when sending heartbeat: %s" % ( e ) )
                 self.reset_connection()
             gevent.sleep(HEARTBEAT_INTERVAL)
 
@@ -535,15 +537,16 @@ class WorkerLocustRunner(DistributedLocustRunner):
         try:
             self.client.close()
             self.client = rpc.Client(self.master_host, self.master_port, self.client_id)
-        except Exception as e:
-            logger.error("Exception found when resetting connection: %s" % ( e ) )
+        except RPCError as e:
+            logger.error("Temporary failure when resetting connection: %s, will retry later." % ( e ) )
 
     def worker(self):
         while True:
             try:
                 msg = self.client.recv()
-            except Exception as e:
-                logger.error("Exception found when receiving from master: %s" % ( e ) )
+            except RPCError as e:
+                logger.error("RPCError found when receiving from master: %s" % ( e ) )
+                continue
             if msg.type == "hatch":
                 self.worker_state = STATE_HATCHING
                 self.client.send(Message("hatching", None, self.client_id))
@@ -570,10 +573,8 @@ class WorkerLocustRunner(DistributedLocustRunner):
         while True:
             try:
                 self._send_stats()
-            except Exception as e:
-                logger.error("Connection lost to master server: %s. Aborting..." % (e))
-                break
-            
+            except RPCError as e:
+                logger.error("Temporary connection lost to master server: %s, will retry later." % (e))            
             gevent.sleep(WORKER_REPORT_INTERVAL)
 
     def _send_stats(self):
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -22,6 +22,7 @@ WORKER_REPORT_INTERVAL = 3.0
 CPU_MONITOR_INTERVAL = 5.0
 HEARTBEAT_INTERVAL = 1
 HEARTBEAT_LIVENESS = 3
+FALLBACK_INTERVAL = 5
 
 
 class LocustRunner(object):
@@ -50,6 +51,7 @@ class LocustRunner(object):
         
         self.environment.events.request_success.add_listener(on_request_success)
         self.environment.events.request_failure.add_listener(on_request_failure)
+        self.connection_broken = False
 
         # register listener that resets stats when hatching is complete
         def on_hatch_complete(user_count):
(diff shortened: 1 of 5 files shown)


locust/runners.py
```diff
@@ -521,12 +542,27 @@ def on_locust_error(locust_instance, exception, tb):
...
+                logger.error("Exception found when sending heartbeat: %s" % ( e ) )
+                self.reset_connection()
             gevent.sleep(self.heartbeat_interval)
 
+    def reset_connection(self):
+        logger.info("Reset connection to master")
+        try:
+            self.client.close()
+            self.client = rpc.Client(self.master_host, self.master_port, self.client_id)
+        except Exception as e:
```
cyberw: Would it be possible to change this to only catch specific exceptions?

delulu: It's intended to catch all the exceptions, to make it reliable with respect to all the possible failures. 

also it's not against the current logic, this reset_connection is newly introduced in function loop and it's not supposed to throw any exception out unless there's a strong reason.

I understand you concern, will add tests to cover this change.

cyberw: Catching all types of exceptions is generally considered bad practice as it may hide more serious issues or put the program in an unknown state, causing hard-to-debug problems later on. But in this case it may be warranted, I can't really tell :)

max-rocket-internet:
@delulu does this fix the occasional `missing` slave? I see this now and again and I'm also running on k8s.

https://github.com/locustio/locust/issues/1158

delulu:
> @delulu does this fix the occasional `missing` slave? I see this now and again and I'm also running on k8s.
> 
> #1158

Yes, it fixes it well.

locust/runners.py
delulu: I see, will update the catching based on the exceptions I observed during my tests.

cyberw:
Do we really need to wrap the exceptions i. Our own exception class? I dont really see what value that adds..

heyman:
> Do we really need to wrap the exceptions in our own exception class? I dont really see what value that adds..

Also, *if* we're wrapping the exceptions we should use `raise ... from e`.

delulu:
> > Do we really need to wrap the exceptions in our own exception class? I dont really see what value that adds..
> 
> Also, _if_ we're wrapping the exceptions we should use `raise ... from e`.

sure, will update it. 

The main purpose is to handle these exceptions in the same place rather than scattered in runners.py. Also it shows how to deal with RPCError, it reduces the maintenance effort.

delulu:
I would like to add test case to test reset_connection in test_runners.py, but I haven't figured out a good way, free to let me know if you have any ideas.

delulu:
I've added a test case to test_reset_connection:

1. check connection_broken will be true when there's a RPCError in recv_from_client. https://github.com/locustio/locust/blob/d45bac93e90ddd652ef9a3411c7e2ca2293381b9/locust/runners.py#L434

2. make sure RPCError is tolerated in reset_connection.

Please have a review.

cyberw:
Looks nice! But I still dont understand what wrapping the exception helps with? I would prefer catching zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError etc in the places where it is relevant instead of catching RPCError. Less code, less magic.

delulu:
I'm exactly "catching zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError etc in the places where it is relevant". 

I mean in runners.py, it deals with rpc and it has no idea or context about these errors (zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError) and how to handle them.

cyberw:
> I'm exactly "catching zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError etc in the places where it is relevant".
> 
> I mean in runners.py, it deals with rpc and it has no idea or context about these errors (zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError) and how to handle them.

I dont quite understand what you mean by "it has no idea or context about these errors and how to handle them"? 

Why can't runner do `except (zmq.error.ZMQError, <and whatever other exceptions you are wrappring>) as e:` instead of `except RPCError as e:`? Wrapping them adds no value imo, it just complicates the code and makes it less obvious what really happened.

delulu:
> > I'm exactly "catching zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError etc in the places where it is relevant".
> > I mean in runners.py, it deals with rpc and it has no idea or context about these errors (zmq.error.ZMQError, msgerr.ExtraData, UnicodeDecodeError) and how to handle them.
> 
> I dont quite understand what you mean by "it has no idea or context about these errors and how to handle them"?
> 
> Why can't runner do `except (zmq.error.ZMQError, <and whatever other exceptions we are wrappring>) as e:` instead of `except RPCError as e:` and handle it exactly the same way? Wrapping them adds no value imo, it just complicates the code and makes it less obvious what really happened.

rpc is dealing with zmq, msg decode and msgpack directly, so it has context of these exceptions and know the possible causes of them. so it can wrap them together and add the cause info. while runners don't have these context, and it have no knowledge in what scenarios these exceptions will be raised.

And I totally disagree that it complicates the code.

The handling of these three exceptions are the same, so wrap them in a unified exception with description and the callers only need to care this one exception, rather than handling them in different scenarios.

For example in some place one of these three should be caught, in another place two of these three or all the three should be caught. This is complicate and it require caller to understand the details of rpc.

cyberw:
Hi @delulu ! I'm sorry we were not able to agree on this. I would love to see more robust connection handling, but I won't merge with the (IMO) needless exception wrapping.

Perhaps I can at least convince you that the wrapping is not so important to you as to stop the fix?

If you make the requested changes & have a look at possibly speeding up the test case I'll be happy to merge.

heyman:
Unfortunately I don't have time to review the full PR at the moment. However I don't see any problem with with raising a common RPCError exception (from the specific one), as long as the proper way to handle them in runners.py is the same. I actually think it makes the abstraction less leaky.

max-rocket-internet:
I would love to see something merged to ensure better communication between slaves and masters 🙏

I still can't reproduce it reliably but we still see now and again missing or slaves that don't stop hatching.

cyberw:
If I'm the only one who thinks the exception wrapping is weird, and someone resolves the conflicts then I'm ok with merging.

delulu:
sorry for my late response.

@cyberw I have to say that the wrap is as important as the fix, because I think this is the right thing to do. if you see anything wrong you need to point it out and convince me, and I'll be happy to make the changes.

as for test case you mentioned above, it tests three scenarios:

1. RPCError is handled and the connection_broken will be set to "True".
2. When a normal msg is well received, the connection_broken will be set back to "False".
3. Other exceptions ( HeyAnException ) wont be handled, so no change on connection_broken.

for each scenario, it takes about 3 seconds to get the msg and the test case doesn't work as expected when I try to reduce the sleep time to 2 seconds.

I'll rebase it with latest master branch and test it. will commit the change when test looks good.

cyberw:
> @cyberw I have to say that the wrap is as important as the fix

I disagree, but as I said, if I'm the only one who considers it less clean than just catching the underlying exceptions I wont argue any more. Maybe I am missing something but let's move on. It's not that important.

As for the test cases, couldn't you just reduce the timeouts before the test (and reduce the sleeps)? Or is that not possible for some reason?

delulu:
> > @cyberw I have to say that the wrap is as important as the fix
> 
> I disagree, but as I said, if I'm the only one who considers it less clean than just catching the underlying exceptions I wont argue any more. Maybe I am missing something but let's move on. It's not that important.
> 
> As for the test cases, couldn't you just reduce the timeouts before the test (and reduce the sleeps)? Or is that not possible for some reason?

when a msg is sent, it seems take 3 seconds to go to the code line of updating connection_broken. because when I reduce it to 2 seconds, the status of connection_broken is not updated as expected.

so I can't reduce the sleep time or sleeps.

conflicts have been resolved and the latest branch has been tested, please take a further check.

cyberw:
Thanks for your contribution!

heyman:
> when a msg is sent, it seems take 3 seconds to go to the code line of updating connection_broken. because when I reduce it to 2 seconds, the status of connection_broken is not updated as expected

The reason for this was the FALLBACK_INTERVAL that was introduced by this pull request. The first sleep wasn't needed at all, but when both the first and second was 3, they surpassed the FALLBACK_INTERVAL that is 5 (that's why it failed when both sleeps were 2). I've now fixed this in 2ac0a84ae5656c240e7a051b3d779f0ba8f75392 by temporarily setting down the FALLBACK_INTERVAL.

What's the purpose of the `FALLBACK_INTERVAL` BTW, and how was the 5 seconds selected?

I also removed the case with an unhandled exception in `MasterLocustRunner.client_listener` because it was causing an error stack trace to be printed when the test was run, and if an unhandled exception happens in client_listener we're fucked anyways, so we don't need to check for it (unless we're planning to handle the exception in some way).

delulu:
@heyman you explains it well and your fix looks nice! I forgot the mocked rpc context in test.

as for `FALLBACK_INTERVAL` in my env `HEARTBEAT_INTERVAL` has been set to 3, and connection will be reset in heartbeat if connection broken is detected. so I simply choose the prime number to make sure `reset_connection` has been executed and the interval does not collide with the `HEARTBEAT_INTERVAL`.

and I'm fine that you removed the case. this case is to check that only RPCError will trigger the `reset_connection`.

heyman:
> as for FALLBACK_INTERVAL in my env HEARTBEAT_INTERVAL has been set to 3, and connection will be reset in heartbeat if connection broken is detected. so I simply choose the prime number to make sure reset_connection has been executed and the interval does not collide with the HEARTBEAT_INTERVAL.

Why do we need to call `reset_connection()` in `heartbeat_worker()`? Couldn't we just call `reset_connection()` in the exception handler (in `client_listener`) and skip the `sleep()` call?

delulu:
> Why do we need to call `reset_connection()` in `heartbeat_worker()`? Couldn't we just call `reset_connection()` in the exception handler (in `client_listener`) and skip the `sleep()` call?

Because I prefer to make it consistent with WorkerLocustRunner that the connection status check and connection reset are done in heartbeat.

Issue #896, opened 2018-10-01
Hatch rate in distributed mode spawns users in batches equal to number of slaves


tortila:
### Description of issue / feature request and actual behavior
It looks like hatch rate behavior highly depends on the number of slaves in Locust's distibuted mode. 

As an example:
I'm running Locust in distributed mode with master node and 10 slave nodes. I set the test execution to spawn 100 users with hatch rate of 1. It seems that instead of spawning 1 user per second, 10 users (1 on each slave) are being spawned at once in batches.

![screen shot 2018-10-01 at 14 25 10](https://user-images.githubusercontent.com/5598055/46288761-1fd35300-c587-11e8-94d1-0b4008943d99.png)

If I add 5 more slave (summing up to 15 slave nodes in total), and start new test with the same values: 100 users with hatch rate of 1, users are now spawned in batches of 15:

![screen shot 2018-10-01 at 14 38 49](https://user-images.githubusercontent.com/5598055/46288973-e2bb9080-c587-11e8-8ef8-e4a026dd22b9.png)

### Expected behavior
I would expect hatch rate to behave independent of the number of slaves. In the example above, I expect a smooth increase of 1 user every second.

### Environment settings (for bug reports)

- OS: Debian Stretch
- Python version: 3.6
- Locust version: 0.9.0

### Steps to reproduce (for bug reports)
As described above

heyman:
Yes, your description matches the current implementation. The slave nodes is unaware of each other and will get an instruction on launching X users with Y hatch rate.

This should only be a potential issue if you have a very low hatch rate (less than number of slave nodes) which I don't think is very common. 

Could be fixed but it would add quite a bit of extra complexity, which I currently don't think is justified.

tortila:
@heyman thank you for responding. 

> This should only be a potential issue if you have a very low hatch rate (less than number of slave nodes) which I don't think is very common.

When I filed this issue it was indeed the case - we used to run Locust in setups with 300 slaves. The reason behind it was that we aimed for a very large scale, and wanted to ramp-up slowly, ideally not changing the number of slaves on the fly as it was very problematic (but that's another story). So with this setup, the smallest possible number of users spawned at once was 300, and it was not small enough, as 300 users generated already a significant amount of load. So to sum up, this feature is important for a narrow use case, but I think it's still important to guarantee a smooth and gradual ramp-up. On top of that I also see it as a surprising and not intuitive behaviour - so maybe if it won't be fixed, at least it deserves a proper documentation.

[...]

PR #1281, opened 2020-03-10
Enable statistically meaningful constant RPS load generation distributions

timran1:
Locust is very userful for stress testing web services at constant RPS. Prior work has added support for `constant_pacing` based `wait_time` which ensures that RPS is independent of how long the web services takes to respond back. However, how the individual requests are spread out within a single second of say a 10 RPS load is not currently considered at all in Locust. This can impose undue bursty pressure of requests on the system under test. The situation worsens if a slave is added or removed during load generation.

To measure and understand the existing work, I developed a small web service which simply records and plots the request arrival time, inter-arrival time between consecutive requests and the frequency distribution of inter-arrival times. Refer to the screenshots below for particular examples.

The following commands are used to invoke locust with different `wait_time` schemes implemented in this PR.
```
locust -f <task_file> --host=<host_address> --csv=locust --no-web -c 20 -r 100 -t 1m
```
Similar commands are used for distributed master/slave configurations.

## Existing Support
Using `constant_pacing` on a single slave. Notice the bursts followed by periods of no requests. The exact pattern depends on when the locusts/users are hatched:
locust-constant_pacing-single

Using `constant_pacing` on a multiple slaves. Notice the inter-arrival time is affected when new slave joins after around 100 requests:
locust-constant_pacing-multiple

## Added Support
An open-loop load generation behavior has been added which considers the clock time, user ID and slave ID to decide when to trigger tasks, such that a statistical distribution of task/request arrival time is achieved.

### Constant Inter-Arrival Time
Using `constant_uniform` on a single slave:
locust-constant_uniform-single

Using `constant_uniform` on multiple slaves. Notice the system adjusts itself after a slave is added after around 100 requests.
locust-constant_uniform-multiple

### Poisson Distribution Inter-Arrival Time
Using `poisson` on a single slave:
locust-poisson-single

Using `poisson` on multiple slaves:
locust-poisson-multiple

## Implementation Details
Most of the edits in core locusts files are related to maintaining an ID for each locust such that IDs are consecutive at all times (even after random locust kills). Secondly, a similar scheme is used to track and communicate slave/client ID and hence the timeslots each client is supposed to issue requests at.

Currently the implementation does not account for difference in wall clock times of different nodes. However, if this support gets through, I can work on using a clock synchronization protocol to account for this affect as well.

Comments and suggestions are welcome.


max-rocket-internet:
> I can work on using a clock synchronization protocol to account for this affect as well.

Is that s a bit out of the scope of locust, no? Running and monitoring ntpd or chronyd should be the responsibility of the system, not locust.

heyman:
Hi! Thanks for a well described PR.

What happens when some users are killed or spawned during a test (e.g. if a new slave node connects during a test). Are Locust user instances assigned new IDs?

There is an existing issue where new Locust are spawned in bursts when when running Locust with a large number of slave nodes (#896). One proposed solution to this is to introduce a different delay for each slave node when spawning locust (to spread out the spawning). Would these changes be compatible with that behaviour?

cyberw:
> > I can work on using a clock synchronization protocol to account for this affect as well.
> 
> Is that s a bit out of the scope of locust, no? Running and monitoring ntpd or chronyd should be the responsibility of the system, not locust.

If this feature causes unexpected/weird behaviour if the clocks are out of sync then I think we need to at least fail the run or log a warning. I would not want to be the one debugging that :)

cyberw:
Oh, and speaking of time syncing, I think we should use time.monotonic() instead of time.time() for this timer and all others as well.

timran1:
> > I can work on using a clock synchronization protocol to account for this affect as well.
> 
> Is that s a bit out of the scope of locust, no? Running and monitoring ntpd or chronyd should be the responsibility of the system, not locust.

I agree.



> Hi! Thanks for a well described PR.
> 
> What happens when some users are killed or spawned during a test (e.g. if a new slave node connects during a test). Are Locust user instances assigned new IDs?
> 
Each slave maintains the IDs for its own locusts (starting from zero for every slave). Moreover, each slave gets a "timeslot" from the master (which is updated everytime slave connects/quits). This slave timeslot makes sure that requests from different slaves are interleaved nicely to get the required distribution. I have included a screenshot of what happens if a slave quits (slave connects around 150 requests and quits around 700 requests):
locust-constant_uniform-multiple-quits

> There is an existing issue where new Locust are spawned in bursts when when running Locust with a large number of slave nodes (#896). One proposed solution to this is to introduce a different delay for each slave node when spawning locust (to spread out the spawning). Would these changes be compatible with that behaviour?

This PR works to decouple the locust hatch time from the request generation pattern by using locust IDs instead. As long as locusts have consecutive IDs at all times, we should get expected results.

On the other hand, I think the `timeslot` information passed from master to clients in this PR may be used for spacing out hatch times as well.


> > > I can work on using a clock synchronization protocol to account for this affect as well.
> > 
> > 
> > Is that s a bit out of the scope of locust, no? Running and monitoring ntpd or chronyd should be the responsibility of the system, not locust.
> 
> If this feature causes unexpected/weird behaviour if the clocks are out of sync then I think we need to at least fail the run or log a warning. I would not want to be the one debugging that :)
> 
> (unless of course the behaviour would only impact the in-second distribution of requests in which case it doesnt really matter that much)

Out of sync clocks between slaves should affect the in-second distribution only. A task will be executed at the same interval, just the position in time for the task will be skewed.
 

> Oh, and speaking of time syncing, I think maybe we should use time.monotonic() instead of time.time() for this timer and all others as well.

Makes sense. I will update the new wait_time functions to use time.monotonic()

One comment I want to make is that these distributions can be disturbed if a task takes longer than the wait_time specified. I generally specify a timeout for requests which is less than the specified wait_time to avoid this.

timran1:
Additionally, I just pulled the lastest changes from master and looks like #1266 remove the global `runners.locust_runner`.

The information need in wait_time functions for this PR is:
1. Slave timeslot
2. Locust ID

What is the recommended way to bring in this information now? If I add arguments to wait_time function it will break existing wait_time functions.

heyman:
> Additionally, I just pulled the lastest changes from master and looks like #1266 removes the global runners.locust_runner.
>
> What is the recommended way to bring in this information now? If I add arguments to wait_time function it will break existing wait_time functions.

The  runner instance is now accessible through `locust_instance.environment.runner` (locust_instance is given as an argument to the wait_time functions).

cyberw:
@timran1 can you have a look at the conflicts? Having this in 1.0 would be nice!

heyman:
> If this feature causes unexpected/weird behaviour if the clocks are out of sync then I think we need to at least fail the run or log a warning. I would not want to be the one debugging that :)

That's already the case for Locust though, due to how response time stats aggregation works. 

There is some commented out code from back in 2012 that seemingly is supposed to check for this, but those lines seems to have almost been commited by mistake judging from the [commit message](https://github.com/locustio/locust/commit/dab68d28) from yours truly...

https://github.com/locustio/locust/blob/2ac0a84ae5656c240e7a051b3d779f0ba8f75392/locust/runners.py#L461-L463

timran1:
I have fixed the conflicts and updated test cases to account for additional messages sent by `broadcast_timeslot`.

cyberw:
LGTM. Ok to merge @heyman ?

locust/runners.py
```diff
@@ -128,19 +128,24 @@ def spawn_locusts(self, spawn_count, hatch_rate, wait=False):
...
             sleep_time = 1.0 / hatch_rate
+            hatch_count = 0
             while True:
                 if not bucket:
                     logger.info("All locusts hatched: %s (%i already running)" % (
                         ", ".join(["%s: %d" % (name, count) for name, count in occurrence_count.items()]), 
                         existing_count,
                     ))
+                    if not sorted([g.args[0].id for g in self.locusts]) == list(range(len(self.locusts))):
+                        logger.warning("Locust IDs are not consecutive.")
```
heyman, line 139: When could this happen?

locust/runners.py
```diff
@@ -166,7 +171,17 @@ def kill_locusts(self, kill_count):
...
+
+        to_kill_ids = sorted([user.id for user in to_kill])
+        remaining_count = len(self.locusts) - kill_count
+        for g in self.locusts:
+            if g.args[0].id >= remaining_count:
+                g.args[0].id = to_kill_ids.pop()
+
         self.kill_locust_instances(to_kill)
+        if not sorted([g.args[0].id for g in self.locusts]) == list(range(len(self.locusts))):
+            logger.warning("Locust IDs are not consecutive.")
```
heyman, line 183: When could this happen?

heyman:
Is it possible to write tests for the wait_time functions as well?

locust/runners.py
timran1: This is mostly paranoia. If everything goes to plan, this should never happen.

New locusts get increasing IDs and that should happen every time.

locust/runners.py
timran1: Same comment as above. Just making sure everything is good after some Locusts were killed.

Should I remove these checks?

timran1:
> Is it possible to write tests for the wait_time functions as well?

We may need to run the tests for a while as the wait functions actually waits but this should be doable. I will work on these tests in next couple of days.

locust/runners.py
heyman: I think the check is fine. Just wanted to know if there were some case where this could happen that I had missed :).

heyman:
> We may need to run the tests for a while as the wait functions actually waits but this should be doable. I will work on these tests in next couple of days.

That would be great! I can totally see that it might be far from trivial to write good tests for this, but I think it would be really good to have them since it might be hard to catch regressions in the future otherwise.

jheld:
Would freezegun (with tick) or mock of time/monotonic be helpful for the running of tests which require long runs? Or am I mistaken and the code would actually have to run for a certain amount of time for the test to be effective?

domik82:
@timran1 @heyman @cyberw - do you have plans to finish this PR?

cyberw:
@timran1 Will you have time to look at adding the tests & resolving the conflicts? Sorry for the slow response. Personally I'm not very invested in this change (so I havent really put the time in to determine if it is good nor not). If there are no further updates i will decline this PR in a week or so (we can always open a new one)

cyberw:
Closing due to inactivity. Feel free to reopen if someone (@timran1 ?) has the time to fix the conflicts & add tests.

PR #1287, opened 2020-03-11
Give a more descriptive error when the Locust or TaskSet has no tasks.

cyberw:
Give the user gets a better exception than `IndexError: Cannot choose from an empty sequence`

Should help with user problems like this one for instance https://stackoverflow.com/questions/60629518/error-while-executing-basic-code-in-locust


cyberw:
Merging this unless someone objects today :) @heyman

heyman:
Looks good to me!

PR #1289, opened 2020-03-13
Allow creating a Locust without specifying an environment 

cyberw:
Defautling to a basic one, with default options, including events.

if a user wants to spawn a basic Locust instance they should be able to do that without generating an environment with settings

This simplifies usage so that you can do (for example):

    HttpLocust().run()

Instead of the very awkward (and requiring knowledge of the internals of locust):

    from locust import env, events
    HttpLocust(env.Environment(events=events)).run()

I have not updated the test cases (there are a lot of them) that could use this approach, but I can do that later.

Tbh I'm only 90% sure this is a good idea. I'd love your opinions :)

### DIFF
--- a/locust/__init__.py
+++ b/locust/__init__.py
@@ -1,7 +1,7 @@
+from .event import Events
+events = Events()
 from .core import HttpLocust, Locust, TaskSet, TaskSequence, task, seq_task
 from .exception import InterruptTaskSet, ResponseError, RescheduleTaskImmediately
 from .wait_time import between, constant, constant_pacing
-from .event import Events
-events = Events()
 
 __version__ = "0.14.5"
--- a/locust/core.py
+++ b/locust/core.py
@@ -19,6 +19,8 @@ from .exception import (InterruptTaskSet, LocustError, RescheduleTask,
                         RescheduleTaskImmediately, StopLocust, MissingWaitTimeError)
 from .runners import STATE_CLEANUP, LOCUST_STATE_RUNNING, LOCUST_STATE_STOPPING, LOCUST_STATE_WAITING
 from .util import deprecation
+from .env import Environment
+from . import events
 
 
 logger = logging.getLogger(__name__)
@@ -155,9 +157,12 @@ class Locust(object):
         super(Locust, self).__init__()
         # check if deprecated wait API is used
         deprecation.check_for_deprecated_wait_api(self)
-        
-        self.environment = environment
-        
+
+        if environment:
+            self.environment = environment
+        else:
+            self.environment = Environment(events=events)
+
         with self._lock:
             if hasattr(self, "setup") and self._setup_has_run is False:
                 self._set_setup_flag()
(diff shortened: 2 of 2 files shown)


heyman:
Hmm, I don't think it's a good idea to let the Locust class create Environment instances. I'm fine with injecting a reference to an Environment instance because it's necessary (for clients to have access to events). 

This should still work:

```python
MyLocust(Environment()).run()
```

cyberw:
> Hmm, I don't think it's a good idea to let the Locust class create Environment instances. I'm fine with injecting a reference to an Environment instance because it's necessary (for clients to have access to events).
> 
> This should still work:
> 
> ```python
> MyLocust(Environment()).run()
> ```

That only works if you havent registered any listeners on the locust.events object. changing Environment() parameter to default to locust.events instead of creating a new Events() object might work though (env.py:53)

Although I guess I can deal with the above boilerplate, now that I've figured it out. It will be hard for others though... but I guess most people wont be spawning their own Locust instances :)

heyman:
Ah, yes, in that case:

```python
MyLocust(Environment(events=locust.events)).run()
```

Since it's not part of the official API (creating your own Locust instances) I think that's okay.

cyberw:
> Ah, yes, in that case:
> 
> ```python
> MyLocust(Environment(events=locust.events)).run()
> ```
> 
> Since it's not part of the official API (creating your own Locust instances) I think that's okay.

I've found running a single locust in the debugger is hugely useful, but to do that I need to create my own Locust instance (because otherwise gevent hangs during monkey patching). E.g. https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/debug.py

I can of course modify that example, but I'm afraid it will be confusing to the user, as they will be exposed to more of locust's internals. I guess that is ok, but I just want to show you why it may sometimes have to be exposed to users.

heyman:
Yeah, I guess this is a cost of using dependency injection compared to global variable.

The documentation on [using locust as a library](https://docs.locust.io/en/latest/use-as-lib.html) (which should definitely be extended with better explanations instead of just an example) should hopefully help mitigate confusion for end-users. Perhaps we could alsa have an example on how to use a debugger with Locust?

cyberw:
> Yeah, I guess this is a cost of using dependency injection compared to global variable.
> 
> The documentation on [using locust as a library](https://docs.locust.io/en/latest/use-as-lib.html) (which should definitely be extended with better explanations instead of just an example) should hopefully help mitigate confusion for end-users. Perhaps we could alsa have an example on how to use a debugger with Locust?

Good idea! I can add that to the documentation at some point. Should I try to move gevent_debugger_patch() into locust core somewhere (https://github.com/SvenskaSpel/locust-plugins/blob/master/locust_plugins/utils.py)

heyman:
> Should I try to move gevent_debugger_patch() into locust core somewhere

Hmm, maybe? It's specific to VSCode right? So in that case maybe place it somewhere like `locust.contrib.vscode`. I'm fine with it as long as we're prepared to keep maintaining it. I don't use VSCode, so in practice that probably means you :).

Issue #1250, opened 2020-01-30
Multi thread master node


JayCarr:
### Is your feature request related to a problem? Please describe.

The load I am putting on my master node is destroying it despite it being on a powerful box.  It looks like its only utilizing one core at the moment so...hard for me to scale that.  If it were multithreaded, I could throw more CPUs at it...

### Describe the solution you'd like

For the master node to be multithreaded.

### Describe alternatives you've considered

I need this enough that I'm happy to do it myself if someone will just point me at the place I shold start looking at the code.

### Additional context

Context:
I'm writing a load test for a content provider for a major travel company.  My target is to be able to run 1.5mil requests per minute (25k rps).

At the moment I have 4 amazon t2 ECSs (4processors, 16g of ram) running 4 slaves on each box.  I think I can probably get to 25k rps by just adding more boxes.

The issue I'm getting is that the master node is freezing up when I start to get past 4k rps.  It's pegging it's CPU pretty hard, so I keep hoping for multithreading....

JayCarr:
As I said, I'm happy to fork the project and figure this out if there's no real solution built into the code.  I'd really like this to work for us....

kurtfm:
I just ran into this issue as well.  The master is only using a single CPU from what I can tell... once that gets pegged, you're toast.

heyman:
Yes, Locust only utilises a single CPU core per node (master as well as slaves). However it does sound weird that the master would become a bottleneck at only 16 slave nodes. I've ran load tests with >100 slave nodes without the master node maxing out on CPU.

It should be possible to parallelize  some of the CPU work that is done in the master node (mainly aggregating stats data from the slave nodes), but I'd start with looking into why it's maxing out in the first place.  

Are you running any slave nodes on the machine where you're running the master (I would advice against that)? 

How many different named endpoints do you have in the test (e.g. endpoints that haven't been grouped using the HTTP client's `name` parameter)? Having *a lot* of different endpoints that haven't been grouped using the name parameters could probably increase CPU usage.

On a side note, I think you might need more than four 4-core machines to achieve 25k reqs/s, though it obviously depends a lot on the test scripts, and which HTTP client you're using. Though you should be able to see if that's the case by monitoring the CPU usage of the slave nodes.

kurtfm:
Running 32 slaves. Getting up to 12k requests per second with minimal wait time. Master is on an 8 core machine dedicated. There are 8 slave machines 4 cores each.  Testing APIs not UI. 
[...]

PR #1290, opened 2020-03-14
[WIP] Add multi core/cpu support when running in distributed mode

n89nanda:
### Issue:
Due Python's [GIL](https://docs.python.org/2/glossary.html#term-global-interpreter-lock), Locust runs only on a single CPU core and single Process. When running locust in a distributed mode, it essentially utilizes just a single core (one or more machines) which is not very resource efficient.

### Proposed solution:
Using Python's `subprocess` module, we can essentially side-step this limitation by spawning multiple processes for each `slave`. Each `slave` process will have its own _GIL_ and we can utilize all the cores in the CPU. 

### PR Change details:

1. Added additional `--fast-slave` & `--slave-count` arguments. By default `--slave-count` is set to the current number of CPU's available. (Physical and Logical). 
2. `--fast-slave` is just like `--slave`, except with the help of a small middleware, new _processes_ are spawned and the `args` are passed to these  _processes_ 

### Related to issues:
#1250 

**Appreciate any comments/feedbacks.**

### Todo
- [ ] Run github_changelog_generator to update CHANGELOG.md
github_changelog_generator -t $CHANGELOG_GITHUB_TOKEN locustio/locust
- [ ] Add highlights to changelog in docs: locust/docs/changelog.rst
- [ ] Update locust/__init__.py with new version number: __version__ = "VERSION"
- [ ] Tag master as "VERSION" in git
- [ ] Build VERSION package and upload to PyPI
- [ ] Update Automated Builds configuration in Docker Hub so that the newly created git tag is built as the "latest" docker tag

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -147,6 +147,17 @@ def setup_parser_arguments(parser):
         action='store_true',
         help="Set locust to run in distributed mode with this process as slave"
     )
+    parser.add_argument(
+        '--fast-slave',
+        action='store_true',
+        help="Set locust to run in distributed mode with multiple running processes as slaves"
+    )
+    parser.add_argument(
+        '--slave-count',
+        type=int,
+        default=multiprocessing.cpu_count(),
+        help="Number of running instances/processes of slaves. Only used when running with --fast-slave. Defaults to number of logical cores in the cpu."
+    )
     # master host options
     parser.add_argument(
         '--master-host',
--- a/locust/main.py
+++ b/locust/main.py
@@ -105,12 +105,10 @@ def create_environment(options, events=None):
         stop_timeout=options.stop_timeout,
     )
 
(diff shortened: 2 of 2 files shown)


heyman:
What happens when the `--logfile` command line option is specified?

So far we've chosen to not spawn and orchestrate multiple slave processes automatically. The reasoning has been that the extra effort of spawning >2 processes - when you're already starting at least one master and one slave process, often on multiple machines - is not large enough to warrant the complexity of managing multiple processes from within Locust.

cyberw:
I think launching a distributed run is better done completely outside of locust (like with https://github.com/SvenskaSpel/locust-swarm, or using a similar approach)

cyberw:
Closing this if there is no activity soon. I’m not sold on this, but could maybe be convinced...

PR #1291, opened 2020-03-16
Add reference to LocustRunner instance and WebUI instance on Environment

heyman:


cyberw:
LGTM. And I need it, so I'm merging :)

Issue #1292, opened 2020-03-17
More information in csv reports


alien11689:
### Problem
I am going to run tests with locust without web ui and it would be great to have much more detail information in locust csv report to create some charts on my own.

### What I need
I would like to have in `*_stats_history.csv` file information about:
- how many users are currently running (especially for step mode)
- statistics grouped by type and name (currently they are always set to `None,Aggregated`)

### Alternatives
As an alternative I have considered console output, but it is not very convienient for parsing.

KenClemson:
Hi, I also would agree with alien11689's request.
It would certainly but great to have how many users running per time stamp or something like that so it can be shown on a graph.

alien11689:
I have run some tests on current master with option '--csv=result --csv-full-history' then I receive error `StatsEntry.use_response_times_cache must be set to True if we should be able to calculate the _current_ response time percentile`

When I change property from the error:

```diff
-    def __init__(self, stats, name, method, use_response_times_cache=False):
+    def __init__(self, stats, name, method, use_response_times_cache=True):
```

Specific entries for tasks are collected, but after a few seconds percentiles are set to 0:

```csv
"Type","Name","Timestamp","# requests","# failures","Requests/s","Requests Failed/s","Median response time","Average response time","Min response time","Max response time","Average Content Size","50%","66%","75%","80%","90%","95%","98%","99%","99.9%","99.99%","99.999","100%"
"None","Aggregated","1585134213",0,0,0.00,0.00,0,0,0,0.00,0.00,"N/A","N/A","N/A","N/A","N/A","N/A","N/A","N/A","N/A","N/A","N/A","N/A"
"GET cached client","Get 1","1585134215",776,0,0.00,0.00,20,28,11,144.00,3154.00,0,0,0,0,0,0,0,0,0,0,0,0
"GET new client","Get 1","1585134215",714,0,0.00,0.00,35,44,25,144.00,3154.00,0,0,0,0,0,0,0,0,0,0,0,0
"None","Aggregated","1585134215",1490,0,0.00,0.00,31,36,11,144.00,3154.00,31,36,41,48,66,75,130,130,140,140,140,140
"GET cached client","Get 1","1585134217",886,0,0.00,0.00,21,35,11,402.00,3154.00,21,33,41,47,69,130,140,150,400,400,400,400
"GET new client","Get 1","1585134217",814,0,0.00,0.00,37,51,25,402.00,3154.00,37,52,64,68,83,130,140,150,400,400,400,400
[...]

PR #1297, opened 2020-03-25
[1292] Fix detailed results statistics in csv reports

alien11689:


cyberw:
Lgtm, but I dont see any tests? Did you maybe miss something in the tests-commit?

heyman:
This PR does not make sense to me. Why would be be storing a dict of user counts on every StatsEntry (which corresponds to a (URL, HTTP method) pair in the case of HttpLocust)?

heyman:
I assume (guessing since there is no PR description) that the purpose of this PR is to get a user_count column in the *history* CSV file? That functionality makes sense, but it would need a different implementation.

alien11689:
@heyman The purpose of this change is described in this issue https://github.com/locustio/locust/issues/1292

For me, the most important is to have detailed values for percentiles and additionally also information about current amount of spawned users in csv reports. 

With my PR I added support for counting users in aggregated csv line. If there is better place then it would be great.

I also added percentiles information, because locust doesn't report them as described in issue. The reported percentiles in PR are accumulated - it would be nice to have them calculated from current time window (configurable).

I need to publish the collected metrics to other place than csv file (e. g. to influx). Do you think that publishing a list of dicts of calculated metrics from stats.py as an event would be a good idea?

heyman:
> For me, the most important is to have detailed values for percentiles

I've opened an issue for this: https://github.com/locustio/locust/issues/1315

> also information about current amount of spawned users in csv reports.

This should be a separate issue/PR, and the way to retrieve the current user count is through `environment.runner.user_count`. One would have to change so that `locust.stats.stats_writer()` gets a reference to the `Environment` instance instead of the `RequestStats` instance.

> I need to publish the collected metrics to other place than csv file (e. g. to influx). Do you think that publishing a list of dicts of calculated metrics from stats.py as an event would be a good idea?

You could listen for the request_success and request_failure events and post info to some external DB there. 

If you'd rather post the aggregate stats periodically (could make sense in order to not overload the external DB) you could spawn a Greenlet that does this every X second. 

Something like this (note: haven't tested id, just wrote it here so I might have missed something, but hopefully you get the idea):

```python
import gevent
from locust import events

stats_greenlet = None

def on_test_start(environment, **kw):
    def stats_reporter():
        while True:
            # grab stats from environment.runner.stats here and push to Influx or whatever
            gevent.sleep(1)
    stats_greenlet = gevent.spawn(stats_reporter)

@events.test_stop.add_listener
def on_test_stop(environment, **kw):
    if stats_greenlet:
       stats_greenlet.kill()
```

Issue #817, opened 2018-06-07
Duplicate/confusing entry in pypi


gugu91:
### Description of issue / feature request

Is there any difference between: https://pypi.org/project/locust/ and https://pypi.org/project/locustio/ ?

Also, there is a this conda package: https://anaconda.org/conda-forge/locust and this: https://anaconda.org/auto/locustio

cgoldberg:
https://pypi.org/project/locustio/ is the official

cgoldberg:
hmm.. this *is* super confusing considering `pip install locust` (instead of `pip install locustio`) will get you an outdated version right now.

@heyman do you have the access to the `locust` namespace on PyPI?  If so, I think we should publish the next release there as well as `locustio`.  Could you add me as a maintainer to `locust`?

cyberw:
@heyman We should try to fix this. Ideally I think we should switch to using locust instead of locustio (but that would be a major change). IMO, the unused one should be removed or replaced with some sort of a deprecation notice.

heyman:
@cgoldberg Yes, I have access to the `locust` namespace. Currently what's uploaded there is just a setup.py with a `install_requires=["locustio"]` so one should get the latest locustio version when running `pip install locust`(though I guess you won't get the latest version when running `pip install locust --upgrade` if you already have an older version of the `locustio` package installed).

@cyberw I think that ideally we would switch to `locust`, keep all the already uploaded `locustio` packages so that we wouldn't break projects that have a version pinned, and then make so that the newest version of `locustio` was some kind of dummy package that would issue a warning telling people to user `locust` instead. I'm not sure if that is possible though?

heyman:
I guess that we could switch to `locust` and then keep uploading a new setup.py (to `locustio`) with `install_requires=["locust==x.y"]` for every version we release. That would be a maintenance burden though (but maybe we could automate it?).

cyberw:
There seems to be a way to output log messages (and maybe even throw an error) after package installation, maybe that could be useful.

https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

heyman:
I believe pip swallows all output, at least when the installation is successful (maybe it's possible to output a message while rasing an error).

cyberw:
Fixed! Release is now on locust, and trying to install locustio gives a helpful error message:

`**** Locust package has moved from 'locustio' to 'locust'. Please update your reference (or pin your version to 0.14.6 if you dont want to update to 1.0) ****`

PR #1302, opened 2020-03-27
Update package name to locustio

weitzman:
I'm pretty new to Python, so that makes me able to spot glaring docs bugs, or I'm a foolish newbie. Please close this if I'm mistaken. Thanks.


cyberw:
Your change is good, but as we're planning on renaming the package to locust (https://github.com/locustio/locust/issues/817) I think it is not needed.

Issue #1206, opened 2019-12-24
[Nested TaskSet] Child TaskSet's on_stop method is not called when GreenletExit


da-zhanglei:
### Describe the bug
I have a locustfile like this. The LightSet and HeavySet 's on_stop method is never called.
The background is, I want to do some cleanup works in LightSet and HeavySet respectively.
Is there anyway to do it?

```python

from locust import HttpLocust, task, TaskSet, constant


class LightSet(TaskSet):
    @task(1)
    def light_1(self):
        print('Light 1')

    def on_stop(self):
        print('Light stop')


class HeavySet(TaskSet):
    @task(1)
    def heavy_1(self):
        print('Heavy 1')

    def on_stop(self):
        print('Heavy stop')


class LightAndHeavySet(TaskSet):
    tasks = {LightSet: 2, HeavySet: 8}


class User(HttpLocust):
    task_set = LightAndHeavySet
    wait_time = constant(1)
    host = 'localhost'

```

### Expected behavior

The LightSet and HeavySet 's on_stop should be called.

Or there is a workaround to do the cleanup for child TaskSet.

### Actual behavior
The LightSet and HeavySet 's on_stop is not called.
I cannot find a way to do it(By looking at Docs and StackOverflow).

### Environment settings

- OS: Mac OS
- Python version: 3.7
- Locust version: 0.13.5

ertanden:
Is there some kind of a workaround for this?

da-zhanglei:
Finally I gave up using "Nested TaskSet".   Instead I use different "User" as a workaround.
It worked as expected.

Like this:

```python

from locust import HttpLocust, task, TaskSet, constant


class LightSet(TaskSet):

    @task(1)
    def light_1(self):
        print('Light 1')

    def on_stop(self):
        print('Light stop')


class HeavySet(TaskSet):
    @task(1)
    def heavy_1(self):
        print('Heavy 1')

    def on_stop(self):
        print('Heavy stop')


class HeavyUser(HttpLocust):
    task_set = HeavySet
    weight = 8
    wait_time = constant(1)
    host = 'localhost'


class LightUser(HttpLocust):
    task_set = LightSet
    weight = 2
    wait_time = constant(1)
    host = 'localhost'

```






```

heyman:
I can confirm and reproduce this bug.

The current code for calling `on_stop` does not take nested TaskSets into account:

[...]

PR #1306, opened 2020-04-02
Decouple Runner and Locust code by introducing Locust.start and Locust.stop methods

heyman:
Decouples the runner code from the Locust user code by introducing Locust.start and Locust.stop methods.

The start() and stop() methods takes a gevent.pool.Group instance, and is responsible for spawning and killing a greenlet for the locust user. With this change the runner does no longer need to keep track of the state of the individual Locust instances, and the Locust instance does no longer need to be given a reference to the runner.

(This PR is based on the tasks-directly-under-locust-class branch, and shouldn't be merged until that PR is merged).

### DIFF
--- a/locust/core.py
+++ b/locust/core.py
@@ -274,14 +274,9 @@ class TaskSet(object, metaclass=TaskSetMeta):
                     self.schedule_task(self.get_next_task())
                 
                 try:
-                    if self.locust._state == LOCUST_STATE_STOPPING:
-                        raise StopLocust()
+                    self._check_stop_condition()
                     self.execute_next_task()
-                    if self.locust._state == LOCUST_STATE_STOPPING:
-                        raise StopLocust()
                 except RescheduleTaskImmediately:
-                    if self.locust._state == LOCUST_STATE_STOPPING:
-                        raise StopLocust()
                     pass
                 except RescheduleTask:
                     self.wait()
@@ -363,12 +358,17 @@ class TaskSet(object, metaclass=TaskSetMeta):
             ))
     
     def wait(self):
+        self._check_stop_condition()
         self.locust._state = LOCUST_STATE_WAITING
         self._sleep(self.wait_time())
         self.locust._state = LOCUST_STATE_RUNNING
 
     def _sleep(self, seconds):
         gevent.sleep(seconds)
+    
+    def _check_stop_condition(self):
+        if self.locust._state == LOCUST_STATE_STOPPING:
+            raise StopLocust()
     
     def interrupt(self, reschedule=True):
         """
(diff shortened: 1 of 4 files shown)


locust/core.py
```diff
@@ -556,19 +560,59 @@ def _set_setup_flag(cls):
     def _set_teardown_flag(cls):
         cls._teardown_is_set = True
     
-    def run(self, runner=None):
+    def run(self):
+        self._state = LOCUST_STATE_RUNNING
         task_set_instance = self._task_set(self)
         try:
+            if hasattr(self, "on_start"):
```
cyberw, line 567: would it be possible to define an on_start/on_stop method on the TaskSet base class that does nothing?

locust/core.py
```diff
@@ -556,19 +560,59 @@ def _set_setup_flag(cls):
...
+        
+        *Arguments*:
+        
+        * gevent_group:  gevent.pool.Group instance where the greenlet will be spawned.
+        
+        Returns the spawned greenlet.
+        """
+        def run_locust(user):
+            """
+            Main function gor Locust user greenlet. It's important that this function takes the locust
```
cyberw, line 588: gor? :P

locust/runners.py
```diff
@@ -161,36 +154,32 @@ def kill_locusts(self, kill_count):
...
         if self.environment.stop_timeout:
             dying = Group()
-            for g in greenlets:
-                locust = g.args[0]
-                if locust._state == LOCUST_STATE_WAITING:
-                    self.locusts.killone(g)
-                else:
-                    locust._state = LOCUST_STATE_STOPPING
-                    dying.add(g)
+            for user in users:
```
cyberw, line 172: I guess you're using "user/s" instead of "locust/s" because we're renaming it? it is confusing  when the comments talk about locusts but the code about users :P

cyberw:
Big thumbs up, except for the minor questions/issues I mentioned.

locust/core.py
heyman: Yeah, that might be better. I guess that any performance difference should be negligible (haven't measure so I don't even know which one is fastest).

locust/core.py
heyman: Oops, will fix :)

locust/runners.py
heyman: Yeah, I figured we're renaming it, and even before we've done that I don't think `user` can be mistaken for anything else in that context.

heyman:
@cyberw:
Thanks! All you mentioned issues have now been addressed I think. This PR now also fixes #1206.

cyberw:
I'm ok to merge!

PR #1312, opened 2020-04-06
Httpbasicauth

anuj-ssharma:
Add the ability to enable basic authentication to the locust webui.


anuj-ssharma:
This is still WIP.

PR #1313, opened 2020-04-06
Add basic auth for webui

anuj-ssharma:
Added the ability to configure basic auth for the webui. This will be done via the command  line option --web-auth. If enabled, then the routes will need to authorised with the correct credentials before they can be accessed


cyberw:
LGTM, and I do see the use case. The only drawback is that the list of arguments for locust is getting pretty long. Could you find somewhere to add it in the documentation?

Actually we should probably just add the output of `locust --help` to the docs somewhere (to help people who are googling/searching the docs). If that could be automated it would be really nice.

cyberw:
Perhaps also add a test case where no credentials were provided?

locust/main.py
```diff
@@ -8,6 +8,7 @@
 import time
 
 import gevent
+from flask_basicauth import BasicAuth
```
heyman, line 11: This import doesn't seem to be used.

locust/main.py
```diff
@@ -229,6 +229,16 @@ def timelimit_stop():
         # spawn web greenlet
         logger.info("Starting web monitor at http://%s:%s" % (options.web_host or "*", options.web_port))
         web_ui = WebUI(environment=environment)
+        if options.web_auth is not None:
+            web_ui.app.config["BASIC_AUTH_ENABLED"] = True
```
heyman, line 233: I think it would be better to add an `auth` argument to `WebUI.__init__` (that defaults to None) and move most of this logic to the WebUI class.

locust/web.py
```diff
@@ -45,8 +47,23 @@ def __init__(self, environment):
         self.app = app
         app.debug = True
         app.root_path = os.path.dirname(os.path.abspath(__file__))
+        app.config["BASIC_AUTH_ENABLED"] = False
+        self.auth = BasicAuth()
+
+        def auth_required_if_enabled(view_func):
```
heyman, line 53: Looking at the [documentation](https://flask-basicauth.readthedocs.io/en/latest/) for Flask-BasicAuth, there seem to be an option called `BASIC_AUTH_FORCE`. Couldn't we use that and not have to decorate every endpoint?

locust/test/test_web.py
```diff
@@ -257,3 +258,37 @@ def my_task(self):
...
+        self.web_ui.app.config["BASIC_AUTH_PASSWORD"] = "doe"
+        auth = BasicAuth()
+        auth.init_app(self.web_ui.app)
+
+    def tearDown(self):
+        super(TestWebUIAuth, self).tearDown()
+        self.web_ui.stop()
+        self.runner.quit()
+
+    def test_index_with_basic_auth_enabled(self):
```
heyman, line 291: This should be two separate tests I think. We should also have a test where incorrect user/pass is supplied.

locust/test/test_web.py
```diff
@@ -257,3 +258,37 @@ def my_task(self):
...
+        self.stats = self.runner.stats
+
+        self.web_ui = WebUI(self.environment)
+        self.web_ui.app.view_functions["request_stats"].clear_cache()
+        self.setUpBasicAuth()
+        gevent.spawn(lambda: self.web_ui.start("127.0.0.1", 0))
+        gevent.sleep(0.01)
+        self.web_port = self.web_ui.server.server_port
+
+    def setUpBasicAuth(self):
```
heyman, line 279: If we move the logic of setting up basic auth to web.py as suggested in [this comment](https://github.com/locustio/locust/pull/1313/files#r403993789) we don't need a special method for setting up the BasicAuth.

anuj-ssharma: You're right, refactored and removed this method now.

locust/test/test_web.py
anuj-ssharma: Added multiple tests now (correct, incorrect, blank)

locust/main.py
anuj-ssharma: Made a change, let me know what you think.

locust/main.py
anuj-ssharma: > 
> 
> LGTM, and I do see the use case. The only drawback is that the list of arguments for locust is getting pretty long.
> 
> Could you find somewhere to add it in the documentation?

I agree but maybe a separate PR though, what do you think ?

locust/web.py
heyman: Hmm, but the `auth_required_if_enabled` isn't exposed in the API as far as I can tell?

locust/web.py
```diff
@@ -36,17 +39,45 @@
...
+            credentials = web_auth.split(':')
+            if len(credentials) == 2:
+                self.app.config["BASIC_AUTH_USERNAME"] = credentials[0]
+                self.app.config["BASIC_AUTH_PASSWORD"] = credentials[1]
+                self.app.config["BASIC_AUTH_ENABLED"] = True
+                self.auth = BasicAuth()
+                self.auth.init_app(self.app)
+            else:
+                sys.stderr.write("The credentials need to provided in the format username:password\n")
+                sys.exit(1)
```
heyman, line 64: Instead of writing to stderr and calling `sys.exit()` we should raise an exception here. For example `InvalidCredentialsError` that could inherit from `ValueError`. Then we can catch this exception in main where we can output a proper error message with `logger.error()` and call `sys.exit(1)`.

locust/web.py
```diff
@@ -36,17 +39,45 @@
 
 class WebUI:
     server = None
-    """Refernce to pyqsgi.WSGIServer once it's started"""
+    """Reference to pyqsgi.WSGIServer once it's started"""
     
-    def __init__(self, environment):
+    def __init__(self, environment, web_auth=None):
```
heyman, line 44: Maybe `auth_credentials` is a better variable name?

locust/argument_parser.py
```diff
@@ -357,7 +357,13 @@ def setup_parser_arguments(parser):
         metavar='LocustClass',
         help="Optionally specify which Locust classes that should be used (available Locust classes can be listed with -l or --list)",
     )
-
+    parser.add_argument(
+        '-B', '--web-auth',
```
heyman, line 361: I don't think we need a short hand `-B` for this option. `--web-auth` should be enough.

locust/argument_parser.py
```diff
@@ -357,7 +357,13 @@ def setup_parser_arguments(parser):
...
+        type=str,
+        dest='web_auth',
+        default=None,
+        help='Set the Basic Auth for the UI (fmt: username:password)'
+    )
```
heyman, line 366: You should pull the latest master branch and make sure that the argument is added to the Web UI argument group.

locust/web.py
anuj-ssharma: You're right !  I have made relevant changes now which exposes the decorator to be added to custom routes.

locust/web.py
heyman: We're still not raising an exception here. 

We shouldn't catch ValueError and IndexError (that could cause us to mistakingly swallow unrelated errors in BasicAuth). What I meant was that instead of calling `sys.exit` *here* we should raise our *own* exception (that can  inherit from ValueError) and then handle this exception in main.py where we can log the error and call `sys.exit(1)`. I see no problem with the `if len(credentials) == 2:` check.

anuj-ssharma: Okay I've had another attempt at this, let me know you're happy with it.

heyman: Looks great! Only some minor fixes that I could fix from the Github UI.

heyman:
Great job!

cyberw:
@anuj-ssharma I want to deprecate the support for basic auth. It relies on Flask-BasicAuth that hasnt been update for a REALLY long time. Do you have a problem with this? Or maybe you want to update it?

Issue #1283, opened 2020-03-10
Rename Locust to User/LocustUser


heyman:
Branched off from #1266. See #1264 for a longer discussion.

cyberw:
Definitely User, not LocustUser :)

cyberw:
I guess we should also rename the parameter (-c/--clients) to -u/--users?

heyman:
> I guess we should also rename the parameter (-c/--clients) to -u/--users?

Yes, that makes sense I think.

curtisgibby:
> I guess we should also rename the parameter (-c/--clients) to -u/--users?

This change is not indicated in the changelog. The docs correctly show that `-u` is the current parameter, but the changelog doesn't say that it's different than it once was.

cyberw:
> > I guess we should also rename the parameter (-c/--clients) to -u/--users?
> 
> This change is not indicated in the changelog. The docs correctly show that `-u` is the current parameter, but the changelog doesn't say that it's different than it once was.

Added to the chagelog now (6b94272943c7ac2d9d722cc7801739364fd66016)

PR #1314, opened 2020-04-06
Rename locust to user

anuj-ssharma:
#1283


cyberw:
Thanks! But there are lots of places where "User" are still called "Locust", so there is more work to be done.

Can you introduce a hidden parameter -c/--client that throws an error? Thanks!

We should also discuss whether things like LocustRunner should be renamed UserRunner? I'm in favour of that.

heyman:
> We should also discuss whether things like LocustRunner should be renamed UserRunner? I'm in favour of that.

Hmm, maybe just `Runner` ( and`LocalRunner`, `MasterRunner` and `WorkerRunner`)?

locust/core.py
```diff
@@ -447,7 +447,7 @@ def __new__(mcs, classname, bases, class_dict):
         return type.__new__(mcs, classname, bases, class_dict)
 
 
-class Locust(object, metaclass=LocustMeta):
```
heyman, line 450: `LocustMeta` should be renamed to `UserMeta`.

cyberw:
@anuj-ssharma Any chance you'll have time to look at the suggested changes & fix the conflicts soon? It would be great to be able to release 1.0 soon!

heyman:
:+1: 

All of the documentation stills needs updating as well

I currently get an exception when I run `locust --help` with this PR.

anuj-ssharma:
Still working through this sorry !

anuj-ssharma:
I think I've messed up with merge conflicts

anuj-ssharma:
Still need to update the documentation

anuj-ssharma:
@cyberw @heyman 
What are your opinions on renaming the option `locust_classes` to `user_classes` ?

cyberw:
> @cyberw @heyman
> What are your opinions on renaming the option `locust_classes` to `user_classes` ?

Go for it!

anuj-ssharma:
There are errors which have not been picked up by the existing tests.

cyberw:
The last commit broke the build (seems some tests need updating), but if you fix that then I then I am +1 to merge!

heyman:
Great work!

> The last commit broke the build (seems some tests need updating), but if you fix that then I then I am +1 to merge!

I've fixed the broken tests, but I'd like some time to go through the changes a bit more thoroughly before be merge. Hopefully I'll be able to do that later today/tonight.

anuj-ssharma:
My last commit (which is a merge) will break tests. Will analyse and fix them.

anuj-ssharma:
> 
> 
> My last commit (which is a merge) will break tests. Will analyse and fix them.

and fixed now !

heyman:
Merged!

PR #1316, opened 2020-04-07
Add User count to CSV history stats

### LINKED ISSUE #1292: More information in csv reports

heyman:
This PR does:

* Add a column "User count" to the history CSV stats (should fix #1292)
* Makes the "Timestamp" column in history CSV the first column
* Remove the "Download response time history CSV" from Web UI since it has never worked, and fixing it is non-trivial (since we would either have to always store the whole stats history in memory, or always create a disk file somewhere, and I don't think we want to do either).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,6 +27,8 @@ Breaking changes
       def on_test_start(**kw):
           print("test is stopping")
 * ``TaskSequence`` and ``@seq_task`` has been replaced with :ref:`SequentialTaskSet <sequential-taskset>`.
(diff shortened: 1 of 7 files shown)


cyberw:
LGTM!

PR #1324, opened 2020-04-13
Worker quitting then stopping via web UI bug fix

### LINKED ISSUE #1279: Web page is confused when I shutdown the only slave
### Describe the bug
I started locust as master-slave mode (just 1 slave). When locust is running, I shutdown the slave.
When I refresh the web page, it become confused, refresh is useless.
image
In console, master already set the slave missing
console

### Expected behavior
Web page is OK

### Actual behavior
See the image above

### Steps to reproduce
1. run locust master;
2. run 1 slave;
3. shutdown slave;
4. press "Stop" button
5. refresh the web page

### Environment

- OS: win 7 x64
- Python version: 3.7.0
- Locust version: 0.14.4
- Locust command line that you ran: locust -f boomer_master.py --master
- Locust file contents (anonymized if necessary):  Golang scripts (drived by boomer), I think it doesn't matter with this bug.

Trouv:
Fixes #1279 

The issue is that when all workers quit, the test is stopped AND the client_listener greenlet stops. So, when the user presses Stop after the worker quits, the state is set to `STATE_STOPPING`, but the client_listener will never set it from `STATE_STOPPING` to `STATE_STOPPED` again. This seems to be why the UI gets confused on refresh.

Ideally, I think all workers quitting would update the UI to its stopped form and the stop button would disappear. Even then, this if statement (or something similar) is probably still good to have for idempotency's sake.


Trouv:
Or would something like [this](https://github.com/locustio/locust/compare/master...Trouv:worker-quit-ui-fix-alt) be better?

cyberw:
I think we should do both fixes! The ui part for better clarity for the user, and the other part for more robustness.

Trouv:
Sounds good! I added the second fix I linked in. I also noticed that test_stop wasn't being fired when the workers quit, so now `stop()` is called when the last worker quits.

heyman:
Nice!

Could you write tests for this as well?

> Or would something like [this](https://github.com/locustio/locust/compare/master...Trouv:worker-quit-ui-fix-alt) be better?

I think it's fine if clicking the stop button when in a stopped/stopping state results in a no-op. Also, `message` is actually never displayed to the user AFAIK, so currently those extra lines doesn't really do anything. I believe `message` was added some time in Locust's infancy and was never actually used (so we should probably remove it in another PR).

locust/runners.py
```diff
@@ -490,6 +491,8 @@ def client_listener(self):
                 if msg.node_id in self.clients:
                     del self.clients[msg.node_id]
                     logger.info("Client %r quit. Currently %i clients connected." % (msg.node_id, len(self.clients.ready)))
+                    if self.worker_count == 0:
+                        self.stop()
```
heyman, line 495: We could log a message here. `Last worker quit. Stopping test` or something.

cyberw: Isnt that implied by the last ”client x quit. Currently 0 clients connected” though?

cyberw: Side note: I dont like the word ”clients” for load gen workers, as there is a chance of confusion with test clients/users. That is a different issue..

Trouv: I think this log message would be helpful.

Also, I was thinking I should add a comment here to clarify that calling stop here isn't responsible for stopping the client_listener greenlet, and mention where it is being stopped/killed. Buuut I'm not totally sure what IS stopping it when the last worker quits?

heyman: > Isnt that implied by the last ”client x quit. Currently 0 clients connected” though

Maybe, but I don't think it hurts to be extra clear about it. One could perhaps expect that the master would re-balance the simulated users and the test would continue to run if a new worker connects (just as would happen if all workers except one quits, and then a new worker connects).

> Side note: I dont like the word ”clients” for load gen workers

Definitely agree! It should say workers.

> Also, I was thinking I should add a comment here to clarify that calling stop here isn't responsible for stopping the client_listener greenlet

The client_listener is not supposed to end when workers disconnect, so I don't think that's necessary.

Trouv: > The client_listener is not supposed to end when workers disconnect, so I don't think that's necessary.

Oh okay, I see what's happening now. It's just hanging when it tries to recv_from_client() since there are no more workers to receive from.

Trouv:
I believe I've addressed the issue in a more complete way now. I noticed that making the worker go missing instead of quitting would still cause the bug. So, I moved things around a little bit.

I think the real core of the issue was that setting it to STATE_STOPPED was handled in client_listener. When client_listener has nothing to listen to, the state will never be updated to STOPPED. Also, setting it to STATE_STOPPED isn't directly related to dealing with the messages from the workers, it has more to do with the state of the self.clients.

I moved setting the STATE_STOPPED to the heartbeat_worker function instead, since it doesn't hang when there are no messages coming in. I think it fits here since this function is used to check on the state of self.clients already. I also moved checking the # of workers remaining to here, and made this compatible with missing workers.

This broke one test case that was using sleep() without sending a mock heartbeat, so I've updated it to do so.

I still plan to write tests for this soon, and maybe look into why refreshing when the state is TEST_STOPPING makes it look so weird in the first place.

locust/runners.py
```diff
@@ -410,11 +414,15 @@ def start(self, locust_count, hatch_rate):
         
         self.state = STATE_HATCHING
 
-    def stop(self):
-        self.state = STATE_STOPPING
-        for client in self.clients.all:
-            self.server.send_to_client(Message("stop", None, client.id))
-        self.environment.events.test_stop.fire(environment=self.environment)
+    def stop(self, info=None):
```
heyman, line 417: I don't think we should add the info parameter to the `stop` method. Instead we should log the message where we make the call to `stop()`.

locust/runners.py
```diff
@@ -440,6 +449,12 @@ def heartbeat_worker(self):
                 else:
                     client.heartbeat -= 1
 
+            if self.worker_count - len(self.clients.missing) <= 0:
+                self.stop(info='No workers remaining, stopping test.')
```
heyman, line 453: Unless I'm mistaken, this check and call to `stop()` should be able to go into the if-block within the for loop above (immediately after `client.user_count = 0` at line 448):

```python
if client.heartbeat < 0 and client.state != STATE_MISSING:
```

We could then first log something like `The last worker went missing, stopping test.` and we wouldn't need to pass in `info` to stop.

locust/runners.py
```diff
@@ -499,8 +514,6 @@ def client_listener(self):
             elif msg.type == "exception":
                 self.log_exception(msg.node_id, msg.data["msg"], msg.data["traceback"])
 
-            if not self.state == STATE_INIT and all(map(lambda x: x.state != STATE_RUNNING and x.state != STATE_HATCHING, self.clients.all)):
-                self.state = STATE_STOPPED
```
heyman, line 503: I think we still want this check here so that we'll change the state to STOPPED immediately when we've received the stop message from the last worker (under normal operation where workers don't go missing).

locust/test/test_runners.py
```diff
@@ -734,11 +743,15 @@ def test_spawn_locusts_in_stepload_mode(self):
             for i in range(5):
                 server.mocked_send(Message("client_ready", None, "fake_client%i" % i))
 
+            def heartbeat():
+                for i in range(5):
+                    server.mocked_send(Message("heartbeat", {'state': STATE_RUNNING, 'current_cpu_usage': 50}, "fake_client%i" % i))
+
```
heyman, line 749: I'm trying to understand why this test broke. Was it because the workers went missing before the test ended? 

If so, maybe we could instead just decorate the test with:

```python
@mock.patch("locust.runners.HEARTBEAT_LIVENESS", new=30)
```
instead of sending the fake heartbeat messages?

Trouv: I originally thought that was why it was failing but now I'm not sure. That patch didn't work either. I'll look into this more when I get off work

Trouv: So, what's actually happening is not that fake workers go missing, instead their state is never changed from ready since they never send a heartbeat.  If all the workers are set to ready, then the state will be set to stopped on the next check.

This wasn't a problem before since there was never a check to stop after the test was started. The client_listener would just hang after it received all the fake client_ready messages. Having the heartbeat_worker check to stop was responsible for breaking the test.

However, it's no longer a problem after I implement your suggestion to move the check inside the if statement, so I'll revert the changes to this test, and write some new tests

locust/runners.py
Trouv: I made a new function for this and call it both here and in heartbeat_worker

locust/runners.py
Trouv: This change fixed the broken test case!

locust/test/test_runners.py
```diff
@@ -452,6 +452,47 @@ def test_master_marks_downed_workers_as_missing(self):
...
+            server.mocked_send(Message("hatching", None, "fake_client1"))
+            server.mocked_send(Message("hatching", None, "fake_client2"))
+
+            server.mocked_send(Message("quit", None, "fake_client1"))
+            sleep(1)
+            self.assertEqual(1, len(master.clients.all))
+            self.assertNotEqual(STATE_STOPPED, master.state, "Not all workers quit but test stopped anyway.")
+
+            server.mocked_send(Message("quit", None, "fake_client2"))
+            sleep(1)
```
heyman, line 471: If I'm not mistaken, these two could be `sleep(0)` to speed up the test.

locust/test/test_runners.py
```diff
@@ -452,6 +452,47 @@ def test_master_marks_downed_workers_as_missing(self):
...
+            server.mocked_send(Message("hatching", None, "fake_client2"))
+
+            sleep(3)
+            server.mocked_send(Message("heartbeat", {'state': STATE_RUNNING, 'current_cpu_usage': 50}, "fake_client1"))
+
+            sleep(3)
+            self.assertEqual(1, len(master.clients.missing))
+            self.assertNotEqual(STATE_STOPPED, master.state, "Not all workers went missing but test stopped anyway.")
+
+            sleep(3)
```
heyman, line 492: We should be able to speed up this test by overriding the heartbeat constants using mock. Smething like this

```python
@mock.patch("locust.runner.HEARTBEAT_INTERVAL", new=0.1)
def test_last_worker_missing_stops_test(self):
    ...
    sleep(0.31)
    ...
    sleep(0.31)
```

locust/test/test_runners.py
Trouv: Yup! done

locust/test/test_runners.py
Trouv: Done

Trouv:
Cool, thanks for your help!

heyman:
Thanks for the contribution!

Issue #1285, opened 2020-03-10
Change how logging is handled


heyman:
Currently we're wrapping sys.stdout and sys.stderr which seems wrong.

PR #1325, opened 2020-04-13
Improve how we do logging

heyman:
PR for #1285. Does the following:

* We no longer wrap `sys.stdout` and `sys.stderr` (yay!). `print()` statements in locust scripts will no longer automatically go into the log. If someone wants to add entries to the log they can use the root logger instead (`logging.info(…)`), or any custom logger named `"locust.*"`.
* The "console_logger" logger has been removed. For printing stats to the console we now use a logger called `"locust.stats_logger"`. For printing output when using flags such as `--show-task-ratio` or `-l` we now use `print()` statements instead.
* We now set up an exception handler that logs any unhandled exception, for every greenlet we spawn (using Greenlet.link_exception) unless they already have their own exception handler (e.g. Locust User greenlets).
* Only allow valid `--loglevel` values.
* Better tests.

### DIFF
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -4,7 +4,25 @@
 Logging
 =======
 
-Locust comes with a basic logging configuration that optionally takes ``--loglevel`` and/or ``--logfile`` to modify the configuration. If you want to control the logging configuration you can supply the ``--skip-log-setup`` flag, which ignores the other parameters.
+Locust uses Python's `built in logging framework <https://docs.python.org/3/library/logging.html>`_ for 
+handling logging.
+
+The default logging configuration that Locust applies, writes log messages directly to stderr. ``--loglevel`` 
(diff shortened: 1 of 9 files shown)


cyberw:
Looks good! But is there some reason plain print() statements cant also work?

heyman:
> But is there some reason plain print() statements cant also work?

Do you mean why `print()` statements won't automatically go into the log file? That was an effect of us wrapping `sys.stdout` which seems like a strange and invasive behaviour. Print statements still work, it's just that they now does what one would expect them to do in python (just print to the console).

cyberw:
> > But is there some reason plain print() statements cant also work?
> 
> Do you mean why `print()` statements won't automatically go into the log file? That was an effect of us wrapping `sys.stdout` which seems like a strange and invasive behaviour. Print statements still work, it's just that they now does what one would expect them to do in python (just print to the console).

Ah, I misunderstood you and thought you meant that print statements no longer work :)

+1

PR #1326, opened 2020-04-14
Update main.py about gevent.signal()

YUFEIZED:
For update in gevent 1.5.0(https://github.com/gevent/gevent/issues/1529), the usage of gevent.signal() should also be updated.


heyman:
Thanks!

Issue #1328, opened 2020-04-14
clean up core.py & __init__.py


cyberw:
core.py is big and messy. I think we should:

* Remove NoClientWarningRaiser because I dont think it is needed any more? (at least the comment would suggest that)
* Move Locust & HttpLocust into separate file (maybe locusts.py?)
* Move TaskSet* into separate file (maybe tasks.py? maybe also move the contents of sequential_locust.py into this file?)

We can still export them in `__init__.py`, to keep the external interface unchanged. 

Some things should probably be removed from `__init__.py` though: 
`from .exception import InterruptTaskSet, ResponseError, RescheduleTaskImmediately` 
and
`from .wait_time import between, constant, constant_pacing` (the documentation instructs the user to get them from locust.wait_time module anyway...)

If it lgty @heyman I can make a PR for this.

heyman:
I think it's a good idea!

> Remove NoClientWarningRaiser because I dont think it is needed any more? (at least the comment would suggest that)

It's still used by `Locust.client`, but could be removed. A long time ago the `Locust` class had an HTTP client, and when that was moved to `HttpLocust`, we created the warning class to be able to give a better error message if people inherited from `Locust` and expected there to be HTTP client on the `client` attribute.

> * Move Locust & HttpLocust into separate file (maybe locusts.py?)
> * Move TaskSet* into separate file (maybe tasks.py? maybe also move the contents of sequential_locust.py into this file?)

Yes, I think that's a good idea. I think we should still have those modules under a common `core` package though (since they are still closely related). I think this structure would make sense:
```
core/
  - __init__.py
  - task.py
    - get_tasks_from_base_classes  (needed by both LocustMeta and TaskSetMeta)
    - @task
    - TaskSetMeta
    - TaskSet
    - SequentialTaskSet
    - DefaultTaskSet
  - user.py
    - LocustMeta
    - Locust
    - HttpLocust
```

> Some things should probably be removed from __init__.py though:
> from .exception import InterruptTaskSet, ResponseError, RescheduleTaskImmediately

:+1:

> and
> from .wait_time import between, constant, constant_pacing (the documentation instructs the user to get them from locust.wait_time module anyway...)

I think we should keep at least `constant` and `between` in the root namespace as a convenience shortcut. But I'm fine with having to import other wait time functions from other modules.

cyberw:
cool, I'll get started!

[...]

PR #1330, opened 2020-04-15
Import wait functions from locust instead of locust.wait_time

cyberw:
In documentation as well as test cases. Relates to #1328


heyman:
Looks good to me!

PR #1331, opened 2020-04-16
Make all LocustRunners aware of their target_user_count, not just MasterLocustRunner

cyberw:
I need this for my global rps target to work (this change was part of the rejected #1241)


heyman:
Looks good!

PR #1332, opened 2020-04-16
Removed double consideration of same timestamp

Oribow:
When searching for an acceptable timestamp dictionary `t-CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW` is considered twice, because `t-CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW-0 == t-CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW+0`


heyman:
Thanks!

Issue #1335, opened 2020-04-17
resp.failure() shouldnt immediately log a failed request, just mark it a such


cyberw:
Because resp.failure() immediately logs a failure ~~trows an exception~~, it is kind of hard to work with.

edit: I was confused, this first thing is entirely possible today

~~For instance, if you want to tag the request as a failure and then break out of the task entirely, you cant really do that in a straight forward way.~~

~~I would like to be able to do this:~~

        with self.client.post("/", catch_response=True) as resp:
            if resp.status_code != something:
                resp.failure()
                return

or even this (particularly useful when you have wrapped the request method and it does some kind of error checking internally, which you may want to override sometimes):

        with self.client.post("/", catch_response=True) as resp:
            if resp.status_code != something:
                resp.failure()
            if known_error:
                resp.success()

Probably we should also change it from being a method call to being a property.

Because this is potentially a breaking change for some testplans I would like to do it for 1.0

heyman:
Hmm, `catch_response=True` + `resp.failure()` doesn't raise an exception? Maybe I'm misunderstanding your post?

The following code works perfectly fine for me (with either HttpLocust or FastHttpLocust):

```python
from locust import HttpLocust, task, constant
from locust.contrib.fasthttp import FastHttpLocust


class TestUser(HttpLocust):
    host = "http://127.0.0.1:8089"
    wait_time = constant(1)
    
    @task
    def test_task(self):
        with self.client.post("/non-existing", catch_response=True) as resp:
            resp.failure("Noooo, fail!")
            print("request failed")
            return
        
        print("end of task (should never run)")
```

Command line output:
```
~/projects/locust master ❯ locust -f examples/issue_1335.py -c 1 -r 1 -t 3 --headless
[2020-04-17 16:57:44,237] MacBook-Air.localdomain/INFO/locust.main: Run time limit set to 3 seconds
[2020-04-17 16:57:44,237] MacBook-Air.localdomain/INFO/locust.main: Starting Locust 1.0.0
[2020-04-17 16:57:44,238] MacBook-Air.localdomain/INFO/locust.runners: Hatching and swarming 1 users at the rate 1 users/s (0 users already running)...
[2020-04-17 16:57:44,238] MacBook-Air.localdomain/INFO/locust.runners: All locusts hatched: TestUser: 1 (0 already running)
 Name                                                          # reqs      # fails     Avg     Min     Max  |  Median   req/s failures/s
[...]

PR #1336, opened 2020-04-17
Allow multiple calls to response.failure() or response.success() within the same with block

heyman:
PR for #1335

### DIFF
--- a/locust/test/test_client.py
+++ b/locust/test/test_client.py
@@ -4,6 +4,7 @@ from requests.exceptions import (InvalidSchema, InvalidURL, MissingSchema,
 
 from locust.clients import HttpSession
 from locust.env import Environment
+from locust.exception import ResponseError
 from .testcases import WebserverTestCase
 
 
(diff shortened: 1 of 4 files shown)


locust/test/test_client.py
```diff
@@ -122,3 +123,69 @@ def test_get_with_params(self):
...
+        self.assertEqual(1, self.environment.stats.total.num_requests)
+        self.assertEqual(0, self.environment.stats.total.num_failures)
+    
+    def test_catch_response_response_error(self):
+        s = self.get_client()
+        try:
+            with s.get("/fail", catch_response=True) as r:
+                raise ResponseError("response error")
+        except ResponseError as e:
+            self.fail("ResponseError should have be raised")
```
cyberw, line 174: do you mean "should have been caught"?

cyberw:
LGTM apart from that error message I didnt understand :)

locust/test/test_client.py
heyman: Oops, copy & paste error. Fixed!

PR #1339, opened 2020-04-19
Populate UI fields with -c, -r, --step-clients, and --step-time options

### LINKED ISSUE #1186: command line arguments for clients and spawn rate should populate ui in the same way as url
### Describe the bug

The command parameters should consistently populate the UI boxes.

### Expected behavior

Command line arguments for clients and spawn rate should populate ui in the same way as url

### Actual behavior

While the command line arguments for url is populated into the web ui, those are clients and spawn rate are ignored and only used if the web ui is not started.

### Steps to reproduce

```
python3.6 locust.py -c 1000 -s 100 
```

### Environment settings

- OS: RHEL7.5
- Python version: 3.6
- Locust version: latest

Trouv:
Closes #1186

I had trouble thinking of test cases to write for this (I did write one), so if you have any ideas please let me know!


Trouv:
I also noticed that web_port and web_host are not used in master right now. Should I go ahead and fix this in this PR or should there be a separate one?

heyman:
I'm not sure about adding four extra parameters to the `create_web_ui` method whose only purpose is to prefill a text box. I think that I'd prefer to add an *optional* reference to the parsed argument options object, to the `Environment` instance, which could then be used to prefill the text inputs (if it's set). 

> I also noticed that web_port and web_host are not used in master right now. Should I go ahead and fix this in this PR or should there be a separate one?

Oh, nice catch. A separate PR for that would be great!

locust/main.py
```diff
@@ -241,12 +241,20 @@ def timelimit_stop():
...
                              len(runner.clients.ready), options.expect_workers)
                 time.sleep(1)
         if not options.worker:
+            # apply headless mode defaults
+            if options.num_clients is None:
+                options.num_clients = 1
+            if options.hatch_rate is None:
+                options.hatch_rate = 1
+            if options.step_clients is None:
+                options.step_clients = 1
```
heyman, line 256: I think we should consider changing so that the `-c` and `-r` arguments are mandatory in `--headless` mode.

Trouv:
> I think that I'd prefer to add an _optional_ reference to the parsed argument options object, to the `Environment` instance, which could then be used to prefill the text inputs (if it's set).

That's a great idea! I'll get on it

locust/main.py
Trouv: Should I change this now or would you prefer to discuss it more?

heyman: Hmm, let's keep the current behaviour for this PR. We can do a separate issue/PR for that later :).

locust/test/test_web.py
```diff
@@ -52,6 +52,19 @@ def test_web_ui_no_runner(self):
     
     def test_index(self):
         self.assertEqual(200, requests.get("http://127.0.0.1:%i/" % self.web_port).status_code)
+
+    def test_index_with_options(self):
+        self.environment.parsed_options = parse_options(["-r", "10"])
+        self.assertEqual(200, requests.get("http://127.0.0.1:%i/" % self.web_port).status_code)
```
heyman, line 58: We should check that the values has been set in the response HTML as well. Either just check for the whole `<input ...>` string, or with a simple regex, or perhaps even using [pyquery](https://pyquery.readthedocs.io/en/latest/) (which we would have to add to the `tests_require` dependencies in setup.py).

cyberw:
I'm +1 to merge. Maybe a mention of this feature somewhere in the documentation would be good though.

heyman:
Merged! Good job!

> Maybe a mention of this feature somewhere in the documentation would be good though.

Added an entry to the changelog in 2d0cc6769c1a8527fbf559e0dc892f28bf647e91

PR #1340, opened 2020-04-20
Fix web options

Trouv:
As mentioned in #1339, web_host and web_port aren't actually being used. Fixed this and added a test for it.

Still not sure about saying "*" is the default web_host when it no longer is. I changed the docstring for this anyway but idk if "" still means "all interfaces"?

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -159,7 +159,7 @@ def setup_parser_arguments(parser):
     web_ui_group.add_argument(
         '--web-host',
         default="",
-        help="Host to bind the web interface to. Defaults to '' (all interfaces)"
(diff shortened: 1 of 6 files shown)


heyman:
Looks good!

I ran into an issue that was caused by MacOS (Catalina) only setting up the loopback interface for `127.0.0.1`. I suspect you're not running MacOS, and therefore I went ahead and made a work-around for it.

While at it, I also made a few other changes:

* The test now picks an available TCP port dynamically.
* I changed so that the default argument for `--web-host` is now `'*'` which will be replaced with `''` in main.py. The reason for this is to be consistent with how `--master-bind-host` is specified.

Issue #1238, opened 2020-01-21
Add `allow_redirects` option to FastHttpSession.request


kevinwainczak:
### Is your feature request related to a problem? Please describe.
I'm trying to convert my use of HttpLocust to FastHttpLocust, but I am unable to do so because my code critically relies on being able to set `allow_redirects` to False.

### Describe the solution you'd like
I want an `allow_redirects` option supported in FastHttpSession.request such that passing the value True (or not passing a value at all) will automatically perform redirects from a 3xx response and passing the value False will not perform any redirect resulting from a 3xx response.

### Describe alternatives you've considered
I have considered looking at the Referer/Location that comes back, however I'm unable to inspect the headers after the redirect has been performed. I'm also unable to find the current location after the response has been completed.

### Additional context
I need to parse the location URL where the response is saying to redirect.

cyberw:
It is super annoying, but I dont think geventhttpclient supports this, so I'm not sure we can build it (without patching geventhttpclient, which is of course possible, but pretty dirty...)

cyberw:
Actually, there might be a way. I was looking for a way to control timeouts and found a setting for redirects as well. I'll get back to you...

cyberw:
I found a (dirty) workaround. Before your request (preferably in the TaskSet.on_start method or something) do this:
`self.client.client.redirect_resonse_codes = []`

heyman:
Since we’re already wrapping the geventhttpclient, I think we could implement support for an `allow_redirects` parameter pretty easily (by setting the `redirect_response_codes`).

masthijohna:
@heyman any update on this ?

heyman:
@masthijohna Nope, not that I know of.

cyberw:
Once that PR is merged, you'll be able to set `redirect_resonse_codes` directly on the Locust class. Set it to [] if you want no redirects to happen.

cyberw:
Fixed!

PR #1341, opened 2020-04-20
Allow overriding redirect response codes for FastHttpLocust

cyberw:
Adresses #1238

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -84,6 +84,9 @@ class FastHttpLocust(Locust):
     insecure: bool = True
     """Parameter passed to FastHttpSession. Default True, meaning no SSL verification."""
 
+    redirect_resonse_codes = None
+    """Override for redirect_response_codes for FastHttpSession"""
+
     abstract = True 
     """Dont register this as a locust that can be run by itself"""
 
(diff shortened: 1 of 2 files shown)


heyman:
I wonder if it wouldn't be better to add an `allow_redirects` argument to the `FastHttpSession.request()` method (and `get()`, `post()`, etc.), and  dynamically set the `redirect_resonse_codes` to `[]` if `allow_redirects` is set to `False`?

That way we we would get the same API as the Requests client, and one wouldn't have to choose between turning off/on redirects for *all* requests.

cyberw:
> I wonder if it wouldn't be better to add an `allow_redirects` argument to the `FastHttpSession.request()` method (and `get()`, `post()`, etc.), and dynamically set the `redirect_resonse_codes` to `[]` if `allow_redirects` is set to `False`?
> 
> That way we we would get the same API as the Requests client, and one wouldn't have to choose between turning off/on redirects for _all_ requests.

How’s about I do both?

heyman:
> How’s about I do both?

If we have `allow_redirects` I actually don't really see the need for being able to set `redirect_resonse_codes` for the session instance. Seems like such an uncommon edge case?

cyberw:
> > How’s about I do both?
> 
> If we have `allow_redirects` I actually don't really see the need for being able to set `redirect_resonse_codes` for the session instance. Seems like such an uncommon edge case?

Damn, you are right. Not a lot of people will be used to geventhttpclient properties anyway. Ok, will fix.

PR #1342, opened 2020-04-20
Give better error message when kubernetes env vars collide with locust's

cyberw:


cyberw:
doh, didnt base this on master...

PR #1344, opened 2020-04-20
Add allow_redirects parameter to FastHttpLocust.client.request(). Fixes #1238

### LINKED ISSUE #1238: Add `allow_redirects` option to FastHttpSession.request

cyberw:


heyman:
Nice!

I suspect that one issue with this solution is that it's not thread safe if one were to use the same HTTP client session in multiple greenlets (imagine one greenlet starting a HTTP redirect chain with allow_redirects=True, and another greenlet doing an HTTP requests using allow_redirect=False before the first redirect chain has finished). I can live with that though, since it seems like somewhat of an edge-case. I can't see any way we could fix it without re-implementing `geventhttpclient.useragent.UserAgent.urlopen()`

cyberw:
Yea, I was thinking that as well. But as you said, it would have to be a very strange test plan to encounter that error...

PR #1346, opened 2020-04-21
Fix the package name in the sample pip3 command

LawnGnome:
The link to `locustio` is correct, but the copy/paste command uses `locust` as the package name.


cyberw:
We want to fix this by pushing to `locust` on pypi instead, see #817

@heyman that is on your todo, right? :)

PR #1354, opened 2020-04-25
Add CLI params for TLS cert and key - serves over HTTPS

mattdodge:
This adds support for serving the web UI over HTTPS. This shouldn't be the default behavior probably but if you have a cert I think you should be able to use it. Especially if basic auth is enabled.

This was first attempted and closed several years ago in locustio#282. I'm hoping that it can be readdressed today though. Most of the web is moving to HTTPS and I believe anything that ships a web server should also come with the ability to run it securely - especially with a package as widely used as Locust is.

Comments/feedback/suggestions welcome, happy to iterate and make this work however we can! I love the Locust project and am happy to help improve it.

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -124,6 +124,10 @@ class Environment:
                      which means all interfaces
         :param port: Port that the web server should listen to
         :param auth_credentials: If provided (in format "username:password") basic auth will be enabled
+        :param tls_cert: An optional path (str) to a TLS cert. If this is provided the web UI will be
+                         served over HTTPS
+        :param tls_key: An optional path (str) to a TLS private key. If this is provided the web UI will be
+                        served over HTTPS
         """
         self.web_ui = WebUI(self, host, port, auth_credentials=auth_credentials, tls_cert=tls_cert, tls_key=tls_key)
         return self.web_ui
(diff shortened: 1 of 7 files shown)


locust/env.py
```diff
@@ -125,5 +125,5 @@ def create_web_ui(self, host="", port=8089, auth_credentials=None):
         :param port: Port that the web server should listen to
         :param auth_credentials: If provided (in format "username:password") basic auth will be enabled
```
heyman, line 126: Could you add docstrings for the `tls_cert` and `tls_cert` params to the `create_web_ui()` method? (Since this method is part of the public API, and included in the API docs)

heyman:
This looks good!

Do we need to say something about the expected format of the cert and key, or is that standardised nowadays?

locust/env.py
mattdodge: Whoops yeah sure thing, missed this one.

mattdodge:
I'm certainly no expert but I think most cert types are handled fine out of the box. I tested with an existing LetsEncrypt cert I had, a new cert using `mkcert`, and also the self-signed certs from the unit test. It seems like gevent is using [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket) under the hood which should be pretty flexible.

heyman:
Ok, great!

PR #1355, opened 2020-04-28
Environment variable configuration changes

heyman:
* Automatically generate documentation for all available environment variables when documentation is built.
* Explicitly set environment variable names for the command line options.
* Rename env variables `LOCUST_MASTER`, `LOCUST_MASTER_PORT` in order to make it less likely that they collide with env vars automatically added by Kubernetes. Also change the name of `LOCUST_WORKER` and `LOCUST_MASTER_HOST` for consistency.
* Rename env variable `CSVFILEBASE` to `LOCUST_CSV`.
* Remove `--csv-base-name` command line option.¨
* Removed some redundant comments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ locust.egg-info/**
 locustio.egg-info/**
 docs/_build/**
 docs/cli-help-output.txt
+docs/env-options.rst
 mock.*.egg
 dist/**
 .idea/**
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -45,6 +45,20 @@ attribute instead:
         tasks = [MyTaskSet]
 
 
+Environment variables changed
+-----------------------------
+
+The following changes has been made to the configuration environment variables
+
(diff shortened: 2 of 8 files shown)


heyman:
>  I also slightly prefer the csv option to be stored as csv and not csv_prefix, but no biggie

The argument is still ``--csv``. The only reason I changed the `dest` param to be `csv_prefix` was that I think the following help text:

```
  --csv CSV_PREFIX  Store current request stats to files in CSV format. Setting this option
                    will generate three files: [CSV_PREFIX]_stats.csv,
                    [CSV_PREFIX]_stats_history.csv and [CSV_PREFIX]_failures.csv
```

is slightly better than:

```
  --csv CSV      Store current request stats to files in CSV format. Setting this option
                 will generate three files: [CSV]_stats.csv,
                 [CSV]_stats_history.csv and [CSV]_failures.csv
```

(in the second help text there's a risk of confusion due to the "files in CSV format" string in the first sentence)

cyberw:
Ok!

PR #1361, opened 2020-04-30
Split core.py into two files in separate python package

heyman:
I took a stab at the second part of #1328 and split `locust.core` into a separate `locust.user` package containing two moduels; `locust.user.task` and `locust.user.users`, as described here: https://github.com/locustio/locust/issues/1328#issuecomment-613930788.

I also moved the `wait_time`, `sequential_taskset` and `inspectlocust` modules (which I renamed to `inspectuser`) into the `locust.user` package. 

When moving/renaming files I took extra care to have separate commits that would *only* move/rename a file without any changes, so that we will get to keep as much git history as possible.

Here is the current structure of the `locust.user` package:

```
user/
  - __init__.py
  - task.py
    - get_tasks_from_base_classes  (needed by both LocustMeta and TaskSetMeta)
    - @task
    - TaskSetMeta
    - TaskSet
    - SequentialTaskSet
    - DefaultTaskSet
  - users.py
    - LocustMeta
    - Locust
    - HttpLocust
  - wait_time.py
  - sequential_taskset.py
  - inspectuser.py
```

Overall I think it makes a lot of sense to have a `user` package namespace, though I'm a bit split when it comes to having the two separate `user.task` and `user.users` modules, since their code is very related to each other. I guess an alternative would be to recreate this PR, but only moving `core.py` into the user package, instead of splitting it into `task.py` and `users.py`.


cyberw:
Lgtm! Im ok with splitting user/task into separate files, it doesnt matter that much. Git is actually quite good at recognizing a rename+changes unless the changes are too big.

Ok to merge.

heyman:
> Git is actually quite good at recognizing a rename+changes unless the changes are too big.

Oh, I thought it only worked if the file renames/moves were made without *any* other changes to the file that was moved. That info might be outdated though, and maybe that has improved?

cyberw:
I dont know how small the changes need to be - Ive experienced it working and not working. Dont know if it has been improved :)

cyberw:
You can see it before committing though

heyman:
> I dont know how small the changes need to be - Ive experienced it working and not working. Dont know if it has been improved :)

Ah! I guess doing a move/rename commit with no other changes should be super safe then :). Looking at the code for this PR, git does realise that `user/task.py` was previously `core.py`, but it hasn't kept the history for the lines of code that I moved out into `user/users.py`.

cyberw:
Yea, I dont think git recognizes splitting a file in two.

heyman:
Forgot to update documentation and docstrings for this PR. Fixed in 265a3bfce94ec73d48a2eb79d75d701753074907.

PR #1366, opened 2020-05-08
fix the doc

MarcSteven:
because when you executed the command line order, it installed empty package


cyberw:
Installing locust *should* work (I tried it just now). But the proper fix will be to push the correct package (locust), which @heyman is going to do _sooon_ (maybe next week, right? :)

Closing. Also this is a duplicate of https://github.com/locustio/locust/pull/1346

PR #1370, opened 2020-05-12
Remove task arguments

heyman:
This PR removes `args` and `kwargs` arguments from the internal `TaskSet.schedule_task` and `TaskSet.execute_task` methods.

Passing arguments to tasks was an old legacy feature that was never documented, and never used internally (except in tests). I think removing this feature in the name of KISS makes sense, since the feature aren't used internally, even though the methods are internal.

This will however be a breaking change for people who have looked at the Locust source and chosen to use this feature.


cyberw:
LGTM. That code always looked strange to me :)

PR #1373, opened 2020-05-14
Fixing docker-compose example

apardo04:
Had to change --worker to --slave and was missing 'locust' in command


heyman:
The `--slave` parameter has been renamed to `--worker` in the upcoming 1.0 version. To use 1.0 you can change the docker image to `locustio/locust:1.0` or `locustio/locust:master` for the latest master version.

PR #1375, opened 2020-05-15
Try to increase open files limit and warn if it is still too low afterwards

### LINKED ISSUE #1368: We should check limits (ulimit) and warn if they are too low
Occasionally people file tickets/SO issues that seem to relate to this:

https://github.com/locustio/locust/issues/496

https://stackoverflow.com/questions/61633170/locust-error-nodename-nor-servname-provided-or-not-known/61651507?noredirect=1#comment109058148_61651507

We could add something like this:
```
if resource.getrlimit(resource.RLIMIT_NOFILE)[0] < 10000:
    logger.warning("System open file limit setting is not high enough for load testing. See https://docs.locust.io/en/stable/installation.html#increasing-maximum-number-of-open-files-limit for more info.")
```

cyberw:
Fixes #1368

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -6,6 +6,7 @@ import signal
 import socket
 import sys
 import time
+import resource 
 
 import gevent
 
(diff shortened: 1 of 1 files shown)


cyberw:
Note: the warning is the most important thing, trying to set it (which may only work on MacOS) is more of a convenience

locust/main.py
```diff
@@ -153,6 +154,14 @@ def main():
         # list() call is needed to consume the dict_view object in Python 3
         user_classes = list(user_classes.values())
     
+    try:
+        if resource.getrlimit(resource.RLIMIT_NOFILE)[0] < 10000:
+            # Increasing the limit to 10000 within a running process should work on at least MacOS.
+            # It does not work on all OS:es, but we should be no worse off for trying.
+            resource.setrlimit(resource.RLIMIT_NOFILE, [10000, resource.RLIM_INFINITY])
+    except:
```
cgoldberg, line 162: bare except also catches `KeyboardInterrupt` and `SystemExit`....  change this to `except Exception`

PR #1384, opened 2020-05-19
Rework quickstart documentation and update some documentation for 1.0

cyberw:
Also update some UI text.


docs/quickstart.rst
```diff
@@ -2,54 +2,17 @@
...
-simulate any number of those users while gathering request statistic. The entrypoint for defining the 
-user behaviour is the `locustfile.py`.
-
-.. note::
-
-    The ``locustfile.py`` is a normal Python file that will get imported by Locust. Within it you 
-    can import modules just as you would in any python code.
-    
-    The file can be named something else and specified with the `-f` flag to the ``locust`` command.
-
```
heyman, line 18: I think the note about the locustfile.py being a normal Python file, that gets imported by Locust, and where one can import other python modules, is important. I've seen many examples where people seem to haven't realised that they can split their load test suite across multiple modules.

docs/quickstart.rst
```diff
@@ -2,54 +2,17 @@
 Quick start
 =============
 
-Example locustfile.py
```
heyman, line 5: Why remove "example" from this headline?

docs/quickstart.rst
```diff
@@ -2,54 +2,17 @@
...
-        def view_post(self):
-            post_id = random.randint(1, 10000)
-            self.client.get("/post?id=%i" % post_id, name="/post?id=[post-id]")
-        
-        def on_start(self):
-            """ on_start is called when a User starts before any task is scheduled """
-            self.login()
+locustfile.py
+=============
+.. literalinclude:: ../examples/locustfile.py
```
heyman, line 9: I think it's better to keep the example inline. Since we're "breaking it down" immediately below it'll make it slitghtly easier to maintain this documentation file (and not risking that we accidentally do some changes to `../examples/locustfile.py` without updating quickstart.rst at some point in the future).

docs/quickstart.rst
```diff
@@ -60,38 +23,27 @@ users will start running within their own green gevent thread.
 
 .. code-block:: python
 
-    wait_time = between(5, 9)
-
-Our class defines a ``wait_time`` function that will make the simulated users wait between 5 and 9 seconds after each task 
-is executed. 
-
```
heyman, line 67: Why move this to the bottom of the class?

docs/quickstart.rst
```diff
@@ -2,54 +2,17 @@
...
-
-Let's break it down:
---------------------
+.. rubric:: Let's break it down
 
 .. code-block:: python
 
-    class WebsiteUser(HttpUser):
+    class MyUser(HttpUser):
```
heyman, line 15: I don't see why `MyUser` would be a better name than `WebsiteUser`? I think it's better to use examples that more closely resembles some real world scenario, than to use very generic names.

docs/quickstart.rst
```diff
@@ -60,38 +23,27 @@ users will start running within their own green gevent thread.
...
-.. code-block:: python
-
-    @task(2)
-    def index(self):
-        self.client.get("/")
-        self.client.get("/ajax-notifications/")
+    @task
+    def hello(self):
+        self.client.get("/hello")
+        self.client.get("/world")
```
heyman, line 29: Same here. I much prefer potential real world examples than very generic names.

docs/quickstart.rst
```diff
@@ -60,38 +23,27 @@ users will start running within their own green gevent thread.
...
     
-    @task(1)
-    def view_post(self):
+    @task(3)
+    def view_item(self):
+        item_id = random.randint(1, 10000)
+        self.client.get(f"/item?id={item_id}", name="/item")
```
heyman, line 34: Same here. I much prefer potential real world examples than very generic names.

docs/quickstart.rst
```diff
@@ -60,38 +23,27 @@ users will start running within their own green gevent thread.
...
         ...
 
-We've also declared two tasks by decorating two methods with ``@task`` and given them 
-different weights (2 and 1). When a simulated user of this type runs it'll pick one of either ``index`` 
-or ``view_post`` - with twice the chance of picking ``index`` - call that method and then pick a duration 
+We've also declared two tasks by decorating two methods with ``@task``, one of which has been given a higher weight. When a simulated user of this type runs it'll pick one of either ``hello`` 
+or ``view_item`` - with three times the chance of picking ``view_item`` - call that method and then pick a duration
```
heyman, line 38: I think it's worth mentioning the `weight` values here just to be extra clear.

docs/quickstart.rst
```diff
@@ -60,38 +23,27 @@ users will start running within their own green gevent thread.
...
+        item_id = random.randint(1, 10000)
+        self.client.get(f"/item?id={item_id}", name="/item")
         ...
 
-We've also declared two tasks by decorating two methods with ``@task`` and given them 
-different weights (2 and 1). When a simulated user of this type runs it'll pick one of either ``index`` 
-or ``view_post`` - with twice the chance of picking ``index`` - call that method and then pick a duration 
+We've also declared two tasks by decorating two methods with ``@task``, one of which has been given a higher weight. When a simulated user of this type runs it'll pick one of either ``hello`` 
+or ``view_item`` - with three times the chance of picking ``view_item`` - call that method and then pick a duration 
 uniformly between 5 and 9 and just sleep for that duration. After it's wait time it'll pick a new task
```
heyman, line 39: The mentioning of 5 and 9 here makes more sense with the `wait_time` at the top of the class.

docs/quickstart.rst
heyman: I just realised that in this PR the `locusfile.py` and the "break down" is already out of sync. The first task is called `index` in the file and `hello` in the break down.

docs/quickstart.rst
```diff
@@ -100,69 +52,54 @@ the :ref:`name parameter <name-parameter>` to group all those requests under an
 Additionally we've declared a `on_start` method. A method with this name will be called for each simulated 
 user when they start. For more info see :ref:`on-start-on-stop`.
 
+.. code-block:: python
+
+    wait_time = between(5, 9)
+
+Our class defines a ``wait_time`` function that will make the simulated users wait between 5 and 9 seconds after each task 
+is executed. For more info see :ref:`wait-time`.
```
heyman, line 60: Good thing to add reference to the wait-time docs :+1:

docs/quickstart.rst
```diff
@@ -100,69 +52,54 @@ the :ref:`name parameter <name-parameter>` to group all those requests under an
...
 
 Start Locust
 ============
 
-To run Locust with the above Locust file, if it was named *locustfile.py* and located in the current working
-directory, we could run:
+Put the above code in a file named *locustfile.py* in your current directory and run:
```
heyman, line 65: :+1:

docs/quickstart.rst
```diff
@@ -100,69 +52,54 @@ the :ref:`name parameter <name-parameter>` to group all those requests under an
...
 
 .. code-block:: console
 
     $ locust
 
 
-If the Locust file is located under a subdirectory and/or named different than *locustfile.py*, specify
-it using ``-f``:
+If your Locust file is located somewhere else, you can specify it using ``-f``
```
heyman, line 72: :+1:

docs/quickstart.rst
```diff
@@ -100,69 +52,54 @@ the :ref:`name parameter <name-parameter>` to group all those requests under an
...
     $ locust -f locust_files/my_locust_file.py
 
 
-To run Locust distributed across multiple processes we would start a master process by specifying
-``--master``:
-
-.. code-block:: console
-
-    $ locust -f locust_files/my_locust_file.py --master
-
```
heyman, line 129: I do think we should mention how to run Locust distributed in the Quickstart docs, since it's such an important USP.

docs/writing-a-locustfile.rst
```diff
@@ -5,17 +5,19 @@ Writing a locustfile
...
 
 A user class represents one user (or a swarming locust if you will). Locust will spawn (hatch) one 
 instance of the User class for each user that is being simulated. There are a few attributes that 
 a User class should typically define. 
 
-The *wait_time* attribute
--------------------------
+.. _wait-time:
 
-In addition to the *task_set* attribute, one should also declare a
```
heyman, line 18: Good catch!

docs/writing-a-locustfile.rst
```diff
@@ -93,12 +94,6 @@ Usually, this is specified in Locust's web UI or on the command line, using the
 If one declares a host attribute in the user class, it will be used in the case when no :code:`--host` 
 is specified on the command line or in the web request.
 
-The *tasks* attribute
----------------------
-
-A User class can have tasks declared as methods under it using the :py:func:`@task <locust.task>` decorator, but one can also
-specify tasks using the *tasks* attribute which is described in more details :ref:`below <tasks-attribute>`.
-
```
heyman, line 101: Why remove this?

examples/nested_inline_tasksets.py
```diff
@@ -23,3 +23,4 @@ def stop(self):
         def stats(self):
             self.client.get("/stats/requests")
 
+    tasks = [TopLevelTaskSet]
```
heyman, line 26: We could also decorate the `TopLevelTaskSet` with `@task` instead of setting the `tasks` attribute.

locust/main.py
```diff
@@ -226,7 +226,7 @@ def timelimit_stop():
     if not options.headless and not options.worker:
         # spawn web greenlet
         protocol = "https" if options.tls_cert and options.tls_key else "http"
-        logger.info("Starting web monitor at %s://%s:%s" % (protocol, options.web_host, options.web_port))
+        logger.info("Starting web interface at %s://%s:%s" % (protocol, options.web_host, options.web_port))
```
heyman, line 229: :+1:

examples/add_command_line_argument.py
```diff
@@ -20,7 +20,6 @@ class WebsiteUser(HttpUser):
     """
     host = "http://127.0.0.1:8089"
     wait_time = between(2, 5)
-    class task_set(TaskSet):
-        @task
-        def my_task(self):
-            pass
+    @task
+    def my_task(self):
+        pass
```
heyman, line 25: :+1:

locust/test/test_main.py
```diff
@@ -149,7 +149,7 @@ def my_task():
             stdout, stderr = proc.communicate()
             self.assertEqual(0, proc.returncode)
             stderr = stderr.decode("utf-8")
-            self.assertIn("Starting web monitor at", stderr)
+            self.assertIn("Starting web interface at", stderr)
```
heyman, line 152: :+1:

locust/user/task.py
```diff
@@ -384,7 +384,7 @@ def interrupt(self, reschedule=True):
         and execute, a new task.
         
         This method should not be called by the root TaskSet (the one that is immediately, 
-        attached to the User class's *task_set* attribute), but rather in nested TaskSet
+        attached to the User class's *tasks* attribute), but rather in nested TaskSet
```
heyman, line 387: Actually we should be able to remove this whole sentence now, since the root TaskSet is an internal one (since the tasks-under-user change).

locust/user/wait_time.py
```diff
@@ -34,14 +34,13 @@ def constant_pacing(wait_time):
     execution equal to the time specified by the wait_time argument. 
     
     In the following example the task will always be executed once every second, no matter 
-    the task execution time::
+    the task execution time:
```
heyman, line 37: The double colon is actually not a typo (though I can definitely see why one would think so 😄). It's ReStructuredText syntax.

Screenshot of API docs without double `:`:

image

locust/user/wait_time.py
```diff
@@ -34,14 +34,13 @@ def constant_pacing(wait_time):
...
-        cclass MyUser(User):
+        class MyUser(User):
             wait_time = constant_pacing(1)
-            class task_set(TaskSet):
-                @task
-                def my_task(self):
-                    time.sleep(random.random())
+            @task
+            def my_task(self):
+                time.sleep(random.random())
```
heyman, line 43: :+1:

docs/quickstart.rst
cyberw: I think that should be covered by "In Locust you define your user behaviour in plain Python code." at the top, but if you dont agree, you can add it back in where you think is appropriate.

docs/quickstart.rst
cyberw: Because it is more clear that the file name is important. But you are right, it was better before. I'll add it back in.

docs/quickstart.rst
cyberw: I split it out to be able to test run it exactly as it is. I almost commited multiple out of sync things before moving it to a separate file, so it doesnt really help :)

What *would* help (and I think improve readability) is if we just had a locustfile with inline comments instead. How do you feel about that?

docs/quickstart.rst
cyberw: Because it is less important then everything else. According to 6/8 voters it shouldnt even be mandatory ;) https://github.com/locustio/locust/issues/1308

docs/quickstart.rst
heyman: True, it should, but I think it can be worth making it super clear. That was the reason I added it in IIRC (I think the text about locustfiles being normal python files were already there when the note was added).

docs/quickstart.rst
cyberw: A good name would be FooUser if the site were named Foo, but we dont know what the name of the site is... Maybe QuickstartUser?

WebsiteUser(HttpUser) is just cookie-on-cookie ;) I just changed it to make it clear it is not important.

docs/quickstart.rst
heyman: > I split it out to be able to test run it exactly as it is. I almost commited multiple out of sync things before moving it to a separate file, so it only helps marginally :)

Ah, I see! What I did when I wrote it initially was that that I just used a temporary file to test it, and then copy it back into the quickstart.rst.

> What would help (and I think improve readability) is if we just had a locustfile with inline comments instead. How do you feel about that?

Hmm, I'm not sure. I think the current format allow us to easier highlight what part of the code the comment is referring to E.g:
image

It would definitely be easier to maintain a file with comments, but I do think it's worth the effort to use the current format (since it's not something that should change super frequently).

docs/quickstart.rst
heyman: I disagree. Even though `wait_time` can be declared as a method, in this example it's effectively a mandatory configuration paramater to the user class, so I think it should go at the top.

docs/quickstart.rst
heyman: Ah, I was thinking of `WebsiteUser` as opposed to `ApiUser`, but I see what you mean :). `QuickstartUser` sound ok, or maybe `DesktopUser` or `MobileUser`?

docs/quickstart.rst
cyberw: Ok, I'll add it back inline, no biggie.

docs/quickstart.rst
cyberw: In general, I agree that parameters should be first, but in this case (a quickstart) I believe we should introduce things in order of importance and wait_time is not as important as task definitions.

docs/quickstart.rst
cyberw: I'll go with QuickstartUser if that is ok.

docs/quickstart.rst
cyberw: it is, just further down

docs/quickstart.rst
heyman: With the current version (#1308 is a different issue I think), I'd argue that `wait_time` is just as important as tasks since they are both mandatory. And it's a parameter. Also the fact that we further down reference the wait time before having introduced it (see my other comment).

docs/quickstart.rst
heyman: Ah, sorry, I didn't realize that!

docs/quickstart.rst
cyberw: I added the (3)

docs/writing-a-locustfile.rst
cyberw: Mainly because the index on the left hand side looks way better, but mainly because repetition/duplication bad :)

docs/quickstart.rst
cyberw: re-added it further down

docs/quickstart.rst
cyberw: Hmm... I prefer using obviously made up names instead of names that could be interpreted as meaningful but aren't (such as /ajax-notifications/)

docs/writing-a-locustfile.rst
heyman: I think the "Writing a locustfile" should be a full reference of all important aspects on writing test scripts, so I think we should include it.

docs/quickstart.rst
cyberw: I'm sure it could be improved further, but I think the old examples are worse (for the reasons stated above)

docs/quickstart.rst
cyberw: good catch. Fine. I'll change it back.

docs/quickstart.rst
heyman: Hm, in this case I don't think there's a big risk of people thinking those names have special meaning for locust.

heyman: > I'm sure it could be improved further, but I think the old examples are worse (for the reasons stated above)

I'm sorry to say I don't agree :/

heyman: Perhaps we could come up with different names that has no chance of misleading people into believing they might have special meaning, while still resembling some potential real world scenario?

docs/writing-a-locustfile.rst
cyberw: Uh. my previous comment was actually about why I had removed "the" and "property" from lots of headings. Dont know why it ended up here...

cyberw: The reason for removing was that the tasks attribute is already mentioned under Tasks, no point in repeating it under User.

heyman: I see. When I wrote it, the idea was to list all the available public attributes, even though a more detailed description exists under Tasks (which the text refers to).

I'm not sure that I prefer this structure:

image

to this:

image

We could definitely drop the "The", but I actually think the second version is more clear.

cyberw: Ok, I'll add back the "attribute". 

TBH I'm a bit frustrated by the amount of time I've had to spend on this PR. I know you're just trying to make things as nice as possible, but maybe next time file your own PR with adjustments, if the gist of my PR is good? Otherwise I dont think I'll bother fixing the docs any more...

heyman: I'm sorry you feel that way :/, but there were quite a few of the changes to the Quickstart docs that I thought didn't make it better. 

I spent quite a few hours re-writing those docs *very recently* for the 1.0 changes (mostly in these commits: e40edfed6e08966c722e03c053214097009ccbde, e95983f0760e1ce53e1b36cf2f973f95fd4ed498, 51547a0e58fb8a8d3784a6dac5333de95ab61f63, 1bdbcdbb6592c7dfacea7196b5d58804151b1d61), which is why I really didn't want to merge changes that I thought made it worse. I might be wrong about on how to best structure the quickstart docs, but I definitely just want to make the docs as good as possible, and not trying to be a pain in the ass or anything :/.

cyberw: Oh yea, I wasnt thinking about the fact that you had been working on this recently. Then it makes more sense to be picky. Maybe next time when a documentation update is needed we can do it together or something.

docs/quickstart.rst
cyberw: Maybe, but I dont have the energy to do it atm. Feel free to update as you see fit.

heyman: Ok, sounds good! 👍

docs/writing-a-locustfile.rst
heyman: That sounds good! I do realize that all my comments did make me a pain in the ass, especially if you didn’t know I recently did a rewrite of that documentation page.

Issue #1116, opened 2019-10-22
Installing 0.12.1 requires "pipenv lock --pre"


schinckel:
The latest release of `locustio` cannot really be managed using pipenv. It includes a dependency of `geventhttpclient-wheels`, which has only pre-release versions available.

As such, it is not possible to generate a `Pipfile.lock` file without using `pipenv lock --pre`.

Doing this will install pre-release versions of other packages, which is problematic. (In my case, it installed a 5.0dev version of coverage, which uses a different coverage file format).

schinckel:
(It's possible that this was also present in earlier versions: my last lock file that worked correctly indicated that it worked fine with 0.11.0).

schinckel:
I’m not sure: I think maybe Pipenv would still complain.

heyman:
I just tried installing geventhttpclient-wheels using pipenv:
```
pipenv install geventhttpclient-wheels
```
Which failed because I didn't use the `--pre` flag. However the installation succeeded when I ran:
```
pipenv install geventhttpclient-wheels==1.3.1.dev2
```
Wouldn't that indicate that it would work if we pinned the version locust's setup.py? I haven't used  pipenv much, and not sure how to test it.

heyman:
I pinned the version in master, and now `pipenv install -e git+https://github.com/locustio/locust.git#egg=locustio` succeeds, so I'm assuming it fixed it.

martinpeck:
I'm not sure this has fixed things. The version of gevent being specified for locust is 1.5a2. This is a pre-release version. When I `pipenv install locustio` I get the same errors as the original poster.

easherma:
Same, I get the same errors. @heyman Can you re-open this?

heyman:
That's unfortunate. I could issue a non pre-release version of the geventhttpclient-wheels package, though I'd prefer not to, since I think it'd be better to follow the same version number as the official geventhttpclient package.

Gevent 1.5 fixes fixes a bug which would cause Locust's Web UI to crash on Python 3.8 (#1154), so I don't think we'd want to downgrade to gevent 1.4 (the latest non pre-release version). 

Therefore, I don't see a good fix for this at the moment, except from maybe documenting it.

cyberw:
@heyman You reopened this, but say there is no good solution :) I guess we are stuck waiting for a new release on geventhttpclient? Or should we "solve" the ticket by just documenting it?

heyman:
[...]

PR #1388, opened 2020-05-22
Bump geventhttpclient and switch back to use its original repo 

cyberw:
... instead of its fork, now that there is a new release (I can trigger new releases later on as well). Relates to #1116

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -6,7 +6,6 @@ import signal
 import socket
 import sys
 import time
-import resource 
 
 import gevent
 
@@ -155,6 +154,7 @@ def main():
         user_classes = list(user_classes.values())
     
     try:
+        import resource 
         if resource.getrlimit(resource.RLIMIT_NOFILE)[0] < 10000:
             # Increasing the limit to 10000 within a running process should work on at least MacOS.
             # It does not work on all OS:es, but we should be no worse off for trying.
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
         "requests>=2.9.1", 
         "msgpack>=0.6.2", 
         "pyzmq>=16.0.2", 
-        "geventhttpclient-wheels==1.3.1.dev3",
+        "geventhttpclient>=1.4.0",
         "ConfigArgParse>=1.0",
         "psutil>=5.6.7",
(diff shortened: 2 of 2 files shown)


heyman:
Before we can merge this we need to update the geventhttpclient repo and setup building + uploading of wheels for both Linux, Mac and Windows. For Windows I think we need to use Appveyor. This has already been done for the forked repo (https://github.com/locustio/geventhttpclient), so we should be able to take the changes from there. If we merge this without binary wheels available for common platforms it would cause installation of Locust to fail for people who don't have a build environment set up with the necessary lib dependencies installed.

cyberw:
I’ll have a look

cyberw:
Ok, super confusing. I can see we've uploaded the windows build in the log https://ci.appveyor.com/project/cyberw/geventhttpclient/builds/33063424 but in pypi there is no windows build :-/

heyman:
> My guess is that APPVEYOR_REPO_TAG is not true, but I cant understand why. It may be related to the fact that builds are not triggered automatically (I have to do it manually)

That sounds plausible. Are you lacking permissions to set up the triggers? If so I guess @gwik has to set it up or give us admin access. I guess an alternative could be to see if he'd be okay with moving the official geventhttpclient repo over to the locustio org?

gwik:
As I said before I have no issue moving the repo wherever there are people willing to maintain it.

gwik:
Just let me know what you need.

cyberw:
I think it it least disruptive to keep it where it is (but moving it is also fine I guess). I guess you have given us all the permissions you can with the repo itself @gwik ? Maybe appveyor needs to be added to the project level?

cyberw:
I changed appveyor to always try uploading, even when it doesnt detect a tag, so we now should have a good 1.4.2 package! (I still need to push a button on appveyor to get the windows wheel out, but that is acceptable for now)

cyberw:
Tested on windows and it works (found a bug while I was at it). Merging this, but we can continue discussing here if you want :)

heyman:
> Tested on windows and it works (found a bug while I was at it). Merging this, but we can continue discussing here if you want :)

Great work!

We should still try to get automatic triggering to work I think so that windows releases aren't forgotten a few years down the line :).

PR #1395, opened 2020-05-25
Update flask requirement. Fixes #1394

### LINKED ISSUE #1394: Update flask version
Our minimum required flask version is too old (saw at least one person having an issue https://stackoverflow.com/questions/61969924/typeerror-when-i-run-a-locustfile-py)

https://flask.palletsprojects.com/en/1.1.x/changelog/#version-0-12-5 is a minimum, but we should probably go to 1.x right away.

I can do the PR

cyberw:


cyberw:
Once this is merged, I'd like to make a new locust release (primarily because I think the previous one is borked on windows)

heyman:
Looks good!

> Once this is merged, I'd like to make a new locust release (primarily because I think the previous one is borked on windows)

Sounds like a good idea! 1.0.2 I guess?

heyman:
I'll make the new release!

cyberw:
Github still lists 0.14.6 as the latest release :-/

cyberw:
I guess it has its own idea of releases, completely separate from git tags?

heyman:
Oh, yeah.. Will fix a release for the latest version.

heyman:
Done! They are tied to tags, but you have to create them manually (though I guess it might be possible to automate it somehow).

Issue #587, opened 2017-05-15
Custom Request/sec exit code


odannyc:
Hello,

I'm running Locust on my CI server (jenkins) and would like to know if there is any way to specify when my test is fail or pass.
For example if my test does less than 100req/sec I want it to fail.

Is this possible?

cgoldberg:
>  is any way to specify when my test is fail or pass

Locust returns an exit code of 1 when any failed requests were reported.

> For example if my test does less than 100req/sec I want it to fail.

Exit status code base on throughput (rps) has not been implemented.

cyberw:
I'm going to reopen this. Not saying I'm going to build it, but I think it would be neat to have, and I think it is a very common use case in CI. Throughput is the most important metric, but response times could be very useful as well.

heyman:
I think this makes sense, and I think we should provide a hook for letting end-users set the exit code from within their test scripts.

Maybe we could add a `process_exit_code` attribute to `Environment`. Something like this maybe:

```python
# locustfile.py
from locust import events

@events.init.add_event_listener
def _(self, environment, **kw):
    environment.process_exit_code = lambda env: env.stats.total.total_rps < 100 and 1 or 0
```

Or we could change so that if a listener for the `quitting` event returns True or False, it'll set the exit code to 0/1 (and ignore the default behaviour):

```python
# locustfile.py
from locust.import events

@events.quitting.add_event_listener
def _(self, environment, **kw):
    return environment.stats.total.total_rps >= 100
```

Though I'm not sure how we would handle multiple event listeners returning different bools, and it would also require us to rework the event system to support events that return values.

cyberw:
The process exit code attribute on environment seems more flexible. This, in combination with a good example should be enough (so no need for a specific parameter or anything). 

Maybe (just putting this out there) we could even remove the `—exit-code-on-failure` or whatever it is called.

heyman:
> The process_exit_code attribute on environment seems more flexible. This, in combination with a good example should be enough (so no need for a specific parameter or anything).

Yeah. I guess we could even let it be either a callable (in which case it could be used like my first example in the previous post) or just an integer in which case one could also set it in a listener for the `quitting` event like this:

```python
from locust import events

@events.quitting.add_listener
def _(environment, **kw):
    environment.process_exit_code = environment.stats.total.total_rps < 100 and 1 or 0
```

cyberw:
[...]

PR #1396, opened 2020-05-25
Ability to control the Locust process' exit code

heyman:
PR for #587

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,12 @@ Changelog Highlights
 
 For full details of the Locust changelog, please see https://github.com/locustio/locust/blob/master/CHANGELOG.md
 
+In development
+==============
+
+* Ability to control the exit code of the Locust process by setting :py:attr:`Environment.process_exit_code <locust.env.Environment.process_exit_code>`
+
+
 1.0.2
 =====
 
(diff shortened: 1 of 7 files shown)


cyberw:
Looks good! Could you add an example on how to set the return code based on rps or avg response time as well?

heyman:
I've extended the example to also check for average response time as well as response time percentile. 

I'm hesitant on adding an RPS check to the example because I think it's a slightly weird condition to use (since it's very dependent on both the number of simulated users as well the actual code of the User classes). It shouldn't be hard to figure out for anyone who still wants to do it, but I don't think it should be encouraged by an example in the official docs.

> And maybe do a log entry in the example as well.

Good idea!

cyberw:
Ok to merge! I still think an RPS check is a good idea, especially for CI purposes. Yes, RPS depends on your test plan, but that is constant over CI runs. There are cases where RPS might drop even though response times do not rise (e.g. delays reading test data, load gen overload etc) which you would definitely want to catch. 

But if you are -1 on that, I can just comment it in the ticket (which was about RPS, just like the slack question that made me reopen it :)

cyberw:
Lets worry about (possibly) adding rps thresholds to the documentation later. Merging.

PR #1400, opened 2020-05-28
Rename 3 remaining instances of slave to worker

ibrahima:
Just noticed this when I was browsing the documentation.

### DIFF
--- a/docs/running-locust-docker.rst
+++ b/docs/running-locust-docker.rst
@@ -14,7 +14,7 @@ The docker image can be used like this (assuming that the ``locustifile.py`` exi
 Docker Compose
 ==============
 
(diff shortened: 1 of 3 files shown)


ibrahima:
Actually, there's probably a couple more that should be changed:

https://github.com/locustio/locust/search?q=slave&unscoped_q=slave

https://github.com/locustio/locust/blob/bd198acf2694272707a23429a475255e8754ab42/locust/runners.py#L512

https://github.com/locustio/locust/blob/cd08d0e206769bb2ed03b2c6530c8d4ee07cd529/locust/test/test_runners.py#L359

I think the other instances are places where the old options need to be handled with a deprecation warning.

cyberw:
Nice, thanks!

PR #1409, opened 2020-06-03
update __init__ file

manifiko:
just add some commas 
this will avoid warnings like " 'events' is not declared in \_\_all\_\_ "


heyman:
Thanks!

Issue #1408, opened 2020-06-02
Ability to run test_start on workers.


Zooce:
### Is your feature request related to a problem? Please describe.
I have a load test where some setup is needed, but I really only need to do it once to set some global variables, that all spawned users to use.

Previously this was done with the `HttpLocust.setup` method. Unfortunately, now there is the `test_start` functionality which only runs on the master.

### Describe the solution you'd like
Allow the `test_start` functionality to run on workers.

### Describe alternatives you've considered
I can use the `HttpUser.on_start` method, but it results in ever single user that is spawned making a request (part of the work that I have to do in the setup), and it's not ideal.

Zooce:
I seem to have found a workaround for this -- bypassing the logic in `HttpUser.on_start` if the global variables I'm trying to set have already been set by another spawned user.

```python
class MyUser(HttpUser):
    def on_start(self):
        global SOME_GLOBAL
        if not SOME_GLOBAL:
            # do the setup work
```

cyberw:
Yes, the setup method was removed (which may have been a bad decision, but it had its reasons, because it was running in a bit of a strange place - people ended up getting confused because it was only being run on the first user, so if you set an instance property it would only be available on the FIRST instance of the User)

Maybe we should reintroduce it in some form.

Another way to worka around it is like this:

```
class MyUser(User):
    _first_instance = True

    def __init__(self, parent):
        super().__init__(parent)
        if MyUser._first_instance:
            MyUser._first_instance = False
            # do the setup work
```

cyberw:
Also, if what you want it just to do some general one-off initialization, you can do it using the init event:

```
from locust import events
...
@events.init.add_listener
def on_locust_init(environment, **_kwargs):
    if not isinstance(environment.runner, MasterRunner):
```

heyman:
We could consider adding an additional event that is triggered on the worker nodes when they start spawning users. We would have to consider if it should trigger for new worker nodes that connects while a test is running (I guess it should?). Though I'm not sure it's necessary, since there are work-arounds like the one by @Zooce above.

The setup method was removed for a good reason, and I don't think it makes sense to re-add it.

Zooce:
Removing the `setup` method was probably a good idea, I agree.

[...]

PR #1415, opened 2020-06-06
Add simple documentation with use case for init event

Zooce:
This PR adds some simple documentation to the "Writing a locustfile" page regarding the `init` event.

#1408


cyberw:
thx!

Issue #1417, opened 2020-06-08
locust doesn't escape the double quotes in the csv output 


tobeczm:
### Describe the bug
```
    @task(1)
    def get(self, size):
        self.client.get("/cash/{}".format(size), name='{"scenario":"get cashes","path":"/cash/[amount]","arguments":[{"size":1}]}'))

```

### Expected behavior
the quotes should be escaped as below:
```
"Type","Name",...
"GET","{""scenario"": ""get cashes"", ""path"": ""/cash/[amount]"", ""arguments"": [{""size"": 1}]}", ...
```

### Actual behavior
in the csv output is looks as below, the quotes are not escaped.

```
"Type","Name",...
"GET","{"scenario": "get cashes", "path": "/cash/[amount]", "arguments": [{"size": 1}]}", ...
```

### Environment

- OS:
- Python version:3.8
- Locust version: 1.0.2
- Locust command line that you ran:
- Locust file contents (anonymized if necessary):

heyman:
Thanks for reporting!

We should consider using the python builtin `csv` module for generating the CSV data.

cyberw:
fixed by #1428!

PR #1428, opened 2020-06-15
Use csv module to generate csv data

ajt89:
#1417

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -172,7 +172,11 @@ class WebUI:
         @app.route("/stats/requests/csv")
         @self.auth_required_if_enabled
         def request_stats_csv():
-            response = make_response(requests_csv(self.environment.runner.stats))
+            data = StringIO()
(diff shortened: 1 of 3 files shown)


locust/web.py
```diff
@@ -182,7 +186,11 @@ def request_stats_csv():
         @app.route("/stats/failures/csv")
         @self.auth_required_if_enabled
         def failures_stats_csv():
-            response = make_response(failures_csv(self.environment.runner.stats))
+            data = StringIO()
+            writer = csv.writer(data)
+            failures_csv(self.environment.runner.stats, writer)
+            data.seek(0)
```
cyberw, line 192: I think you can use data.getvalue() instead of .seek(0) and then .read()

cyberw:
Cool stuff! I'll merge this if you fix that minor nit-pick :)

cyberw:
Thanks!

PR #1434, opened 2020-06-17
Quickstart documentation clarification.

JamesB41:
Quickstart documentation has one bit that's slightly confusing. Just a small edit.


heyman:
Thanks!

PR #1435, opened 2020-06-17
Base Locust Docker image on non-alpine python image

heyman:
This PR makes Locust use the `python:3.8` docker image instead of `python:3.8-alpine` as the base image.

The reason for this is that it'll make it possible to `pip install ...` most PyPI packages out of the box, which should be a very common thing to do in project specific docker images. 

It comes with a ~900 MB cost in total image size, though I think the `python:3.8` image should be a very common base image to already have pulled, in which case there will effectively be no higher disk usage, and I think it makes sense to prioritize developer convenience and happiness over disk usage.

This is what the description at the [official docker python repo](https://hub.docker.com/_/python) says about the `python:3.8` image:

> This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.

We could consider maintaining two branches of docker images (`locustio/locust` and `locustio/locust-alpine`), but I don't think it's worth the extra maintenance work.

We would have to bump the Locust version to 1.1 if we release this, since it might break end-users custom docker images (the fixes should be very easy though, and mostly consist of removing a bunch of lines no longer needed from their Dockerfiles).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ build/
 .coverage
 .tox/
 docs/
+.git/
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@ dist/
 build/
 .coverage
 .tox/
+docs/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,13 @@
 FROM python:3.8
 
-COPY . /src
-WORKDIR /src
-RUN pip install .
-RUN rm -rf /src
+COPY . /build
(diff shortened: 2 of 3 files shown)


cyberw:
Nice! 

Can we drop the use of a builder image while we’re at it? I dont think it saves much in image size, and it simplifies things a little when users make derived images.

heyman:
Great! I'll merge this, and then we'll bump the locust version to 1.1 with the next release.

Issue #1461, opened 2020-07-03
--run-time flag is not respected if there is an exception in a test_stop listener


camilojimenez:
### Describe the bug
On a headless session, if there is an exception thrown from a `test_stop` listener, the run-time flag is not respected and the 
session keeps running indefinitely.

### Expected behavior
The session should stop even if the listeners throw exceptions.

### Actual behavior
The traceback is printed, and the session keeps running.

### Steps to reproduce
Take, for example, the following `locustfile.py`:
```
from locust import HttpUser, events, task
import time

@events.test_stop.add_listener
def on_test_stop(**kwargs):
  1/0

class NOOP(HttpUser):
  @task
  def index_page(self):
    time.sleep(10)
```

If locust is run with the following command line:

```locust -t1 -r1 -u10 --headless  --host http://www.google.com```

you see the exception being raised and the session continues:
```
[2020-07-03 03:37:18,086] instance-1/INFO/locust.main: Time limit reached. Stopping Locust.
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 854, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/camilojimenez/locust/venv/lib/python3.7/site-packages/locust/main.py", line 231, in timelimit_stop
    runner.quit()
  File "/home/camilojimenez/locust/venv/lib/python3.7/site-packages/locust/runners.py", line 294, in quit
    self.stop()
  File "/home/camilojimenez/locust/venv/lib/python3.7/site-packages/locust/runners.py", line 340, in stop
    self.environment.events.test_stop.fire(environment=self.environment)
  File "/home/camilojimenez/locust/venv/lib/python3.7/site-packages/locust/event.py", line 33, in fire
    handler(**kwargs)
  File "/home/camilojimenez/locust/locustfile.py", line 6, in on_test_stop
    1/0
ZeroDivisionError: division by zero
2020-07-03T03:37:18Z <Greenlet at 0x7fcad134b9d8: timelimit_stop> failed with ZeroDivisionError

[2020-07-03 03:37:18,091] instance-1/CRITICAL/locust.main: Unhandled exception in greenlet: <Greenlet at 0x7fcad134b9d8: timelimit_stop>
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 854, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/camilojimenez/locust/venv/lib/python3.7/site-packages/locust/main.py", line 231, in timelimit_stop
[...]

PR #1462, opened 2020-07-03
Gracefully handle exceptions in event listeners

camilojimenez:
An exception thrown during the handling of an event will stop the execution of other listeners.

Solves #1461


camilojimenez:
I would prefer adding some logging, but looking at the file I didn't see how that could be accomplished

PR #1477, opened 2020-07-10
Refactoring stats to handle custom percentiles

vstepanov-lohika-tix:
**Goal**
Currently, if we set a custom list of percentiles, we'll get a broken csv report & console stats for percentiles.
This PR is intended to:

- Handling percentiles list in a single place
- Get proper csv report with custom percentiles set 
- Get proper console stats with custom percentiles set
- Decrease of duplication of percentile values in code

**How-to check:**
In the locust file, set the list of percentiles you need to get:

```
import locust.stats
locust.stats.PERCENTILES_TO_REPORT = [0.50, 0.90, 0.95, 0.99]
```

Both csv report & console stats should be correct with a proper columns number and data.


vstepanov-lohika-tix:
Updated tests -> Passed. Moving to Open.

vstepanov-lohika-tix:
Found issue with rounding:
image
Will fix in the next commit

vstepanov-lohika-tix:
Fixed:
image

vstepanov-lohika-tix:
Fixed console stats separator to any number of percentiles:
image

cyberw:
Cool stuff! I think this may be better implemented as a command line switch, but I’m not wholly against it as is. It needs a proper test case (that actually modifies the setting and ensures it changes the output), and documentation.

cyberw:
@heyman what do you think? Should this be a command line switch? I think we need to give some thought to how we manage settings in general (as the list of command line settings keeps growing)

vstepanov-lohika-tix:
@cyberw Added a doc for customization of stats settings. Not sure about command-line options, but probably it's worth to add it to configuration file.

vstepanov-lohika-tix:
@cyberw @heyman  Please let me know if something else should be added/updated

locust/stats.py
```diff
@@ -711,33 +699,22 @@ def print_stats(stats, current=True):
...
-        '90%',
-        '95%',
-        '98%',
-        '99%',
-        '99.9%',
-        '99.99%',
-        '100%',
-    ))
-    console_logger.info("-" * (90 + STATS_NAME_WIDTH))
+    headers = ('Type', 'Name', '# reqs') + tuple([f"{round(percentile*100, 4)}%" for percentile in PERCENTILES_TO_REPORT])
```
lhupfeldt, line 702: Is a max number of decimal points assumed? Rounding like this can cause different percentiles to round to the same number. If a man number of decimal points is assumed, then there should be validation of the PERCENTILES_TO_REPORT.

lhupfeldt:
stats_history_csv_header still has hardcoded list of percentiles

lhupfeldt:
web.py has hardcoded percentiles in different places:

`                    "ninetieth_response_time": s.get_response_time_percentile(0.9),`
```
                report["current_response_time_percentile_95"] = environment.runner.stats.total.get_current_response_time_percentile(0.95)
                report["current_response_time_percentile_50"] = environment.runner.stats.total.get_current_response_time_percentile(0.5)

```
Not sure how median plays into this?
Maybe it should be required that the percentiles referenced from web.py must be in the percentiles list.

locust/stats.py
vstepanov-lohika-tix: At the moment maximum percentile in default list is 0.99999, that matching to 3 digits after point (99.999%)
I've updated to 6 digits and believe that no-one ever will try to use percentiles with higher precision, so I think to do the additional calculation from percentile list will be redundant.

lhupfeldt: That will probably do, but you should still to verify the PERCENTILES_TO_REPORT so that the web interface does not break.

vstepanov-lohika-tix: These changes assume that a user will manually set a list of percentiles based on the default list, and this assumes that a user understands what does he do and why. Otherwise, we should handle all possible wrong or not relevant parameters that could be input ( >1, <0, not a number etc). There's no reason to set a percentile precision over 99.999999% in a performance testing world, and even if it will happen - it's just will cause a wrong header name in results stats.
 So, imho it's a very little sense to add additional calculation in code to cover a non-breaking issue that unlikely could occur.

lhupfeldt: In `web.py` there are assumptions about a few specific percentiles. I think it should be validated that those are present in the list. I agree that it is otherwise fine to assume that users will know to put relevant numbers in the list.

cyberw:
This looks pretty good to me.

Can you drop the _SEC from the variable names? It is enough that the documentation says that it is in seconds.

cyberw:
This looks good to me, if nobody objects, I will merge it this weekend.

lhupfeldt:
`stats_history_csv_header` still has hardcoded list of percentiles

cyberw:
> stats_history_csv_header still has hardcoded list of percentiles

Good point. I will hold off.

locust/stats.py
vstepanov-lohika-tix: That's a good point, I'll check web.py additionally

vstepanov-lohika-tix:
Will check stats_history_csv_header and web.py additionally

vstepanov-lohika-tix:
@cyberw @lhupfeldt Updated `stats_history_csv_header` as well and extracted repetative action to a function.
Regarding web.py - checked additionally, percentile values in this file are not linked to PERCENTILES_TO_REPORT parameter and calculated "on a fly", so looks there's no need to check those values according to this list.

cyberw:
Nice. Let's give @heyman one last chance to comment before I merge :)

cyberw:
Thanks!

Issue #1463, opened 2020-07-03
test_start event blocks the UI


denniseffing:
### Describe the bug
When a user starts a load test and the `test_start` event is fired, the UI is basically doing nothing until the `test_start` event hooks have finished executing.

### Expected behavior
UI should open up the statistics dashboard and change the status from `READY` to a new state called `INITIALIZING`.

### Actual behavior
UI hangs on "Start new Locust run screen" until `test_start` execution has finished.

### Steps to reproduce
1. Add a `test_start` listener that pauses the thread for 5 seconds. 

```python
@events.test_start.add_listener
def on_test_start(**kwargs):
    time.sleep(5)
```

2. Start a test

### Environment

- OS: macOS 10.15.3
- Python version: 3.7.7
- Locust version: 1.0.3
- Locust command line that you ran: `locust`
- Locust file contents (anonymized if necessary):

```python
from locust import events, HttpUser, task, between
import time


@events.test_start.add_listener
def on_test_start(**kwargs):
    time.sleep(5)


class TestUser(HttpUser):
    wait_time = between(5, 9)

    @task
    def index_page(self):
        self.client.get(url=self.host)
```

camilojimenez:
I looked into this and I found that there is a relatively easy way to fix it (without changing too much the current ways). I put a PR for this (#1479). However, the displayed state is not 'fully' consistent (it says ready while in reality is preparing).

A better (but more involved) solution would be to add an extra state signaling that the test_start events are running (something like PREPARING).

Thoughts?

denniseffing:
Well that was indeed an easy fix. However, I think that adding an additional state would be the most optimal solution.

cyberw:
the solution feels good enough for me. Marking as low prio.

PR #1479, opened 2020-07-11
fade into the running screen before getting a response from the server

camilojimenez:
Attempts to solve #1463, by fading into the dashboard screen.


cyberw:
LGTM! Thanks!

PR #1491, opened 2020-07-23
Remove "Loadgen" from CPU warning log messages

max-rocket-internet:
I don't know what `Loadgen` means but AFAIK it's not a term used either in Locust or in the broader load testing field. I think it makes more sense without it? When I first saw this message I thought it was the name of a class or something 😅


cyberw:
Loadgen in this case means worker or standalone. I dont want to remove it but maybe there is a better term?

cyberw:
I was trying to be specific to worker/standalone but as it is actuallt checked on all Runners (even masters), that is not necessary.

Issue #1487, opened 2020-07-22
Negative min_response_time shown in stats


max-rocket-internet:
### Describe the bug

Negative value shown for `min_response_time`:

![Screen Shot 2020-07-22 at 17 58 54](https://user-images.githubusercontent.com/8859277/88199373-124bf900-cc45-11ea-9186-6632cd1a0eb4.png)

### Expected behavior

0 or positive value.

### Actual behavior

Negative value shown for `min_response_time`

### Steps to reproduce

I have a branch local docker-compose setup to reproduce it easily. But it's just 1 worker calling nginx in another docker container.

### Environment

- OS: Docker
- Python version: -
- Locust version: current master

max-rocket-internet:
This line is returning a negative value: https://github.com/max-rocket-internet/locust/blob/b18cf1e96646a4d43dc16f9d80e435c8f1596a33/locust/clients.py#L117

Perhaps a docker clock source issue 🤔

Docker itself just using the host clock source but docker on Mac (what I'm using) uses HyperKit VM as Mac kernel doesn't support native docker. Perhaps this VM clock source is not reliable?

max-rocket-internet:
There's lots of docker clock issues but as far as I can see they are related to sleep: https://github.com/moby/hyperkit/issues/110

If it was a docker issue, it would be easy to reproduce outside of locust with a file like this:

```python
import time

while True:
    t = time.time()
    if time.time() < t:
        print('ERROR: {0} is less than {1}'.format(time.time(), t))
        break
```

And then run it like `docker run -v $PWD:/mnt python:latest python /mnt/time_test.py`

max-rocket-internet:
Ah, after leaving this for a while I get an error 😅

```
ERROR: 1595435189.4236984 is less than 1595435189.4583442
```

OK, good to know!

max-rocket-internet:
Should we make a change to the locust code though? Like to record the response time as `0` in this case?

cyberw:
Can you try replacing the two calls to time.time() with time.monotonic()?

I thought we did that a long time ago, but maybe it was only done for FastHttpUser or something.

heyman:
Interesting.

> Should we make a change to the locust code though? Like to record the response time as 0 in this case?

I don't think we should "hide" the clock issue by setting it to zero if it's negative.

max-rocket-internet:
> I don't think we should "hide" the clock issue by setting it to zero if it's negative.

Makes sense

[...]

PR #1492, opened 2020-07-23
Replace time.time() with time.monotonic()

max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/1487

I didn't replace it in parts related to stats as I'm not sure of the effect there. So I've replaced in these 3 parts where time is measured by subtracting a previous time from `time.time()`.

`FastHttpUser` is not affected by this problem in my testing. It uses a timer from the `timeit` package.


cyberw:
I’m merging this because it is an improvement, but maybe we should look in to using timeit (like FastHttpUser is doing)

PR #1496, opened 2020-07-28
Print valid URL when --web-host is not specified

dmitrytokarev:
Currently if `--web-host` option is not specified the message is:
`Starting web interface at http://:8089` which is an invalid URL and may confuse new users.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -235,13 +235,13 @@ def main():
     if not options.headless and not options.worker:
         # spawn web greenlet
         protocol = "https" if options.tls_cert and options.tls_key else "http"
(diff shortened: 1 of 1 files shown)


heyman:
Saying `http://127.0.0.1:8089` might mislead people to believe that the server is only accepting connections from localhost. 

In the case when web_host is `*`/`''`, maybe it should say something like this instead:

```
Starting web interface at http://127.0.0.1:8089 (accepting connections from all network interfaces)
```

dmitrytokarev:
@heyman I replaced 127.0.0.1 to 0.0.0.0 since . Added the note you proposed.

heyman:
Nice!

PR #1494, opened 2020-07-25
Autoformat using Black. Also fix a few style complaints (trailing whitespaces)

cyberw:
Fixes #1489 (kinda)

* Black is very popular, so many will be familiar with it.
* It is good at minimizing diffs
* It is very opinionated (the only config I've added is the line length), so there should be minimal discussion about how to configure it
* IMO it strikes a good balance between keeping the code nicely formatted (and PEP8 compliant) without making it unnecessarily long. 

Here's a comparison of various formatters https://medium.com/3yourmind/auto-formatters-for-python-8925065f9505

As the next step, we could (should) increase the number of rules covered by flake8 (probably switching to ignoring the few rules that we dont like instead of enabling a few favourite ones. something like --ignore E501,W503,E701,E731,E741,E712,E402,F401).

PR #1498, opened 2020-08-03
Add 2 things to .gitignore

max-rocket-internet:
Since it looks like we're gonna [keep the messy formatting](https://github.com/locustio/locust/pull/1494) then I think we need to ignore `.editorconfig` 🙂

Also adding `__pycache__` as it shows these directories in a colour as if they would be committed by git (at least in Atom) and this is a little confusing:

Screen Shot 2020-08-03 at 17 17 46

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ build/
 .coverage
 .tox/
 docs/env-options.rst
+.editorconfig
+__pycache__


heyman:
`__pycache__` sounds good, though I can recommend that you add it to your global .gitignore.

Maybe it would be better to add the `.editorconfig` with the project specific settings to the project instead?

max-rocket-internet:
> Maybe it would be better to add the .editorconfig with the project specific settings to the project instead?

I never used it until now TBH. Do people put custom stuff in these files? I just added:

```
root = true

[*]
trim_trailing_whitespace = false
insert_final_newline = false
```

Issue #1488, opened 2020-07-23
Allow negative hatch rate for ramping down


max-rocket-internet:
### Describe the solution you'd like

Currently if you edit a running test with a lower amount of users it jumps straight down to this number of users. It would be cool if we could ramp down to this new value with a negative hatch rate 🙂

cyberw:
Maybe this can be solved as part of #1432?

max-rocket-internet:
I'll try 😃

PR #1502, opened 2020-08-04
Allow ramping down of users

max-rocket-internet:
Resolves: https://github.com/locustio/locust/issues/1488

- Merge `stop_user_instances()` into `stop_users()`
- Allow `stop_rate` to be passed to `stop_users()`
- Negative or positive value for `stop_rate` does the same thing
- Copy same ramping behaviour from `spawn_users()`, it seems robust
- If no `stop_rate` passed to `stop_users()` then don't "ramp down", just stop all without sleeping
- Add one extra debug log message when running test is edited with new `user_count`/`hatch_rate` values
- Add test runner to ensure ramping down works

This is my first proper PR for this project so any and all feedback welcome 🙂

I have still yet to update any documentation...

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -168,19 +168,21 @@ class Runner(object):
                     logger.debug("%i users hatched" % len(self.user_greenlets))
                 if bucket:
                     gevent.sleep(sleep_time)
-        
+
         hatch()
         if wait:
             self.user_greenlets.join()
             logger.info("All users stopped\n")
 
-    def stop_users(self, user_count):
+    def stop_users(self, user_count, stop_rate=None):
         """
-        Stop a stop_count of weighted users from the Group() object in self.users
+        Stop `user_count` weighted users at a rate of `stop_rate`
         """
+        if user_count == 0 or stop_rate == 0:
+            return
+
         bucket = self.weight_users(user_count)
         user_count = len(bucket)
-        logger.info("Stopping %i users" % user_count)
         to_stop = []
         for g in self.user_greenlets:
             for l in bucket:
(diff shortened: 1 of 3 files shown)


locust/runners.py
```diff
@@ -189,25 +191,38 @@ def stop_users(self, user_count):
...
-                    stopping.add(user._greenlet)
-            if not stopping.join(timeout=self.environment.stop_timeout):
-                logger.info("Not all users finished their tasks & terminated in %s seconds. Stopping them..." % self.environment.stop_timeout)
-            stopping.kill(block=True)
-        else:
-            for user in users:
-                user.stop(self.user_greenlets, force=True)
-        
+                    stop_group.add(user_to_stop._greenlet)
+                if not stop_group.join(timeout=self.environment.stop_timeout):
```
cyberw, line 214: This timeout will prolong the iteration time for this while loop. 

So if you had for example stop_rate 1 and stop_timeout 2, you would end up (if no users stopping within the timeout) taking 3 seconds to complete an iteration (giving you a stop rate of 1/3 users per second). As a user, that is not what I would expect. Ideally these things should be done async (I guess you'll have to spawn a "killer greenlet" or something)

max-rocket-internet: OK I see what you mean.

What is we just do `stop_group = Group()` outside the loop and then `stop_group.kill(block=True)` after loop is done?

cyberw: You mean *inside* the loop, right? Sounds good! Can you add a test case for it?

cyberw: Sounds good! Can you add a test for it? I think it is one of those things that could trip us up :) (now or in the future)

max-rocket-internet: I can but what exactly is the test? Ramping down with `stop_timeout` set? How does the test fail? If ramping down doesn't happen? Or it happens too slowly?

If I understand correctly, if `stop_timeout` is set, we can't really guarantee any ramp down rate accurately, but could do best effort.

cyberw:
> If I understand correctly, if stop_timeout is set, we can't really guarantee any ramp down rate accurately, but could do best effort.

Well, in my worst-case example, we should still be stopping roughly 1 user per second, over time (except at the very last step where we last users will take around 1s extra)

For example, if we have 4 users and 2s timeout, it should take 5-6s to ramp down to zero if no users finish their iteration in time and ramp down is 1/s (and the iterations should be killed after 2 seconds and not sooner)

max-rocket-internet:
@cyberw OK I've added a second test under `TestStopTimeout` that basically repeats the other test but with `stop_timeout` set. Please have a look and see what you think.

cyberw:
> @cyberw OK I've added a second test under `TestStopTimeout` that basically repeats the other test but with `stop_timeout` set. Please have a look and see what you think.

This test is too lenient:

`self.assertTrue(user_count < 10, "User count has not decreased at all: %i" % user_count)`

After 2.1 seconds of ramp down at 2/s and with one second delay I would expect to see 6 users running, any more than that is an error.

You can run higher ramp down rate to shorten the test.

max-rocket-internet:
> This test is too lenient

OK fair enough.

I did some testing and did see some variation, like +/- 1 when using whole numbers for sleeping. So I increased the hatch rate to `4` and tightened the sleep times also. This makes the tests run faster and it also very tight with higher hatch rate. If I make it tighter than this I see some failures just from variation.

max-rocket-internet:
> Looks great now

🎉

> Only thing left is some documentation + an example.

Can you give me a hint here? I've looked through the docs and there's nothing really specifically mentioning hatch rate when decreasing users, decreasing users at all or really anything specific about hatch rate. And what kind of example would I show?

cyberw:
I think for the web gui it is kind of supposed to be self explanatory, so maybe a note there about the ability to set a lower number of users (tbh I'm not much of a web user myself). Actually just trying to reduce the number of Users using the web gui (running on master, without your changes) just kind of worked!? Maybe this feature already exists in some form!?

```
[2020-08-12 16:44:34,507] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.main: Starting web interface at http://127.0.0.1:8089
[2020-08-12 16:44:35,036] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.main: Starting Locust 1.1.1
[2020-08-12 16:44:44,361] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.runners: Hatching and swarming 2 users at the rate 1 users/s (0 users already running)...
[2020-08-12 16:44:46,377] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.runners: All users hatched: WebUser: 2 (0 already running)
[2020-08-12 16:44:52,926] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.runners: Stopping 1 users
^CKeyboardInterrupt
2020-08-12T14:46:07Z
[2020-08-12 16:46:07,024] lafp-mac-JG5J.int.svenskaspel.se/INFO/locust.main: Running teardowns...
...
```

max-rocket-internet:
> I think for the web gui it is kind of supposed to be self explanatory, so maybe a note there about the hatch rate affecting both up and down ramping (tbh I'm not much of a web user myself).

OK true. I made some small changes to the wording around starting and editing a test.

> Maybe there should be a an example similar to the unit test in /examples as well?

I don't think this is possible? The examples are all various types of locust files and you can't edit a running test in them. Or really do anything to show "ramping down" I don't think. I think this will be addressed in my other PR which we are also close to merging 😉

cyberw:
Nice!

Issue #1501, opened 2020-08-04
Excessive precision of metrics in losust csv stats 


vstepanov-lohika-tix:
### Describe the bug
After upgrade from 0.14.4 to 1.1.1 the most of metrics in csv report became with excessive precision (> 10 digits after point) 
It doesn't make sense and makes report hard readable. 
Not reproduced in console stats

### Expected behavior
Precision for time metrics & content size should be integer number.
Precision for requests rate - float with maximum 2-3 digits after point.

### Actual behavior
Over 10 digits after point:
![image](https://user-images.githubusercontent.com/48280294/89289212-803dea80-d65f-11ea-9a6a-55fef9f3c1d3.png)

### Steps to reproduce
Run any test with HTTP requests & --headless & --csv=report option and check report-stats.csv result file.

### Environment

- OS:
- Python version: 3.7.2
- Locust version: 1.1.1
- Locust command line that you ran: 
locust -f {locust_file_path} --headless -u {users_number} -r {hatch_rate} --run-time={run_time} --csv=results --exit-code-on-error=0
- Locust file contents (anonymized if necessary): - complicated.

heyman:
It seems likely that this was caused by #1428. Previously - when we generated the CSV data manually - we formatted the values using string formatting (`%.2f`). Now when we're using the built in csv module, we should probably wrap the values in `round()` calls.

vstepanov-lohika-tix:
@heyman Please take a look a draft above - will it enough?

vstepanov-lohika-tix:
@heyman please take a look draft PR above, will it be ok?

heyman:
:+1::+1:

PR #1503, opened 2020-08-05
Fix csv stats precision

vstepanov-lohika-tix:
Fix for #1501 issue
image


locust/stats.py
```diff
@@ -822,13 +822,13 @@ def requests_csv(stats, csv_writer):
...
             s.num_requests,
             s.num_failures,
-            s.median_response_time,
-            s.avg_response_time,
-            s.min_response_time or 0,
-            s.max_response_time,
-            s.avg_content_length,
-            s.total_rps,
-            s.total_fail_per_sec,
+            round(s.median_response_time) if s.median_response_time else s.median_response_time,
```
rafaelrds, line 825: I think you can define a function above such as `_adjust_precision(stat)` instead of fixing the precision at this scope. Moreover, having a separate function allows customization using different precision, if needed.

vstepanov-lohika-tix: These changes are not intended to bring some new feature, but just to return an old view of csv report after regression.
Previously, data format was done also at this scope and was broken with an injection of csv_writer in 1.x.x version:
https://github.com/locustio/locust/pull/1428/files#diff-5d5f310549d6d596beaa43a1282ec49eL822
If we'll create a new function, it will be a complicated because of different data types in stats entries (time metrics, content metric, rate metrics). So it should be a some reason to add a new tangled function that probably will be used only once for this specific case.

Issue #1432, opened 2020-06-16
Support for completely custom load pattern / shape


max-rocket-internet:
### Describe the solution you'd like

To be able to programmatically control the user and hatch rate over a period of time. This would allow us to scale up and down and generate spikes which allows more accurate representation of real world load.

We can generate this easily:

![Screen Shot 2020-06-15 at 13 05 52](https://user-images.githubusercontent.com/8859277/84650671-0021c100-af09-11ea-974f-9061c09a7468.png)

But not this:

![Screen Shot 2020-06-15 at 13 07 49](https://user-images.githubusercontent.com/8859277/84650838-4545f300-af09-11ea-8ec8-bcb27a47a08e.png)

And not this either:

![Screen Shot 2020-06-16 at 14 10 03](https://user-images.githubusercontent.com/8859277/84772362-1f891e80-afdb-11ea-82e4-b306b57c46dc.png)

### Describe alternatives you've considered

I tried something quite hacky but it's not good enough:

```python

def traffic_spike(spike_after, spike_length, normal_wait_time):
    def wait_time_func(self):
        if not hasattr(self,"_traffic_spike_start"):
            self._traffic_spike_start = time()
            return normal_wait_time
        else:
            run_time = time() - self._traffic_spike_start
            if run_time > (spike_after + spike_length) or run_time < spike_after:
                return normal_wait_time
            else:
                return 0

    return wait_time_func

class QuickstartUser(HttpUser):
    wait_time = traffic_spike(600, 300, 1)

    @task
    def index_page(self):
        self.client.get("/hello")
        self.client.get("/world")
```

Also I know we could implement this outside of locust and call the locust API but it's not ideal.

### Additional context

k6 has something called [stages](https://k6.io/docs/getting-started/running-k6#stages-ramping-up-down-vus) that looks great:

```javascript
export let options = {
  stages: [
    { duration: '30s', target: 20 },
    { duration: '1m30s', target: 10 },
    { duration: '20s', target: 0 },
  ],
};
```

It would be great to see something like this in locust. Perhaps it could look like this:

```python
from locust import User, TaskSet, between, LoadShape


class LoadWithSpike(LoadShape):
    stages = [
        {'duration': 3600, 'users' 1000, 'hatch': 5},
        {'duration': 300, 'users' 5000, 'hatch': 50},
        {'duration': 1800, 'users' 1000, 'hatch': -50}
    ]


class MyUser(HttpUser):
    wait_time = between(1, 3)

    @task(2)
[...]

PR #1505, opened 2020-08-06
Adding ability to generate any custom load shape with LoadTestShape class

max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/1432

Adding a base `LoadTestShape` class for users to inherit in order to generate any `user_count`/`hatch` at 1 second resolution.

- The the locustfile will be searched for a `LoadTestShape` class and used it if present
- If present, this will run as a `greenlet` on the master and will continuously edit/update and optionally stop a running load test
- Added 3 examples to show how it could be used, including step-load test like Visual Studio, stages like k6 and a completely customised double wave.
- User and hatch rate in the UI is effectively ignored if a `LoadTestShape` class is present. This is stated in a log message.
- Locust command will exit with error if a `LoadTestShape` is present and a conflicting option is also specified as argument (users, hatch-rate or step-load)
- Create documentation

Tests:
- Update 4 tests to ensure the `LoadTestShape` class is not in `user_classes`
- A full distributed test with master + 3 workers scaling up, down and stop
- Scale up and scale down tests for `TestMasterRunner`

Initial testing looks good 🚀

Screen Shot 2020-07-31 at 13 40 13

Comments from @cyberw

> I think the shape should be possible to have in a separate file

> We'll need a couple really of good unit tests for this

Please give me suggestions if we need more tests as I'm not sure. This addition is simply editing a running test so much of this functionality should be covered already.

> make sure there is no error for parameters (-u, -t etc) not being specified when there is a shape.

It will exit with an error.

> correctly applying stop_timeout when ramping down
> where new slaves are added during the test

The current behaviour here won't change?

Comments from @heyman

> Unfortunately the stepload implementation isn't very good and I think it should be removed

We can replace the stepload code with the more general approach in this PR but also this PR is quite similar to the stepload implementation. Please clarify.


locust/runners.py
```diff
@@ -270,12 +276,37 @@ def stepload_worker(self, hatch_rate, step_users_growth, step_duration):
...
+                logger.info("Step Load is finished")
                 break
             self.start(current_num_users, hatch_rate)
-            logger.info('Step loading: start hatch job of %d user.' % (current_num_users))
+            logger.info("Step loading: start hatch job of %d user" % (current_num_users))
             gevent.sleep(step_duration)
 
+    def start_shape(self):
+        logger.info("Shape test starting")
+        if self.shape_greenlet:
```
cyberw, line 287: Can this really happen?

locust/test/test_runners.py
```diff
@@ -784,7 +784,56 @@ def test_spawn_fewer_locusts_than_workers(self):
...
+                    num_users += msg.data["num_users"]
+            self.assertEqual(1, num_users, "Total number of users in first stage of shape test is not 1: %i" % num_users)
     
+            # Wait for shape_worker to update user_count again
+            sleep(2)
+            num_users = 0
+            for _, msg in server.outbox:
+                if msg.data:
+                    num_users += msg.data["num_users"]
+            self.assertEqual(3, num_users, "Total number of users in second stage of shape test is not 3: %i" % num_users)
```
cyberw, line 831: Maybe also test ramp down?

locust/runners.py
```diff
@@ -71,10 +73,13 @@ def on_request_failure(request_type, name, response_time, response_length, excep
         def on_hatch_complete(user_count):
             self.state = STATE_RUNNING
             if environment.reset_stats:
-                logger.info("Resetting stats\n")
+                logger.info("Resetting stats")
```
cyberw, line 76: is this change intentional?

locust/runners.py
```diff
@@ -71,10 +73,13 @@ def on_request_failure(request_type, name, response_time, response_length, excep
...
                 self.stats.reset_all()
         self.environment.events.hatch_complete.add_listener(on_hatch_complete)
-    
+
+        if self.environment.shape_class:
+            logger.debug("Starting with shape class")
```
cyberw, line 81: Looks like you are kind of logging this twice? (here and on line 286)

locust/runners.py
max-rocket-internet: Yes, I thought the `\n` was a mistake. It's intentional?

locust/runners.py
max-rocket-internet: True. I'll remove this one.

locust/test/test_runners.py
max-rocket-internet: Good idea. I'll add another test. I had a problem that the test only runs for 5 seconds and is then stopped, regardless of what I set. I couldn't see why. Is that correct?

max-rocket-internet: OK now there are 2 tests, one for scaling up and one for scaling down. Ideally they would be in the same test but that 5s issue makes it too tight.

locust/runners.py
max-rocket-internet: I'll revert it anyway

cyberw: Tbh I dont know :)

locust/test/test_runners.py
cyberw: Hmm... I didnt know there was a 5 second limit and cant really understand where it comes from :)

examples/custom_shape/stages.py
```diff
@@ -0,0 +1,59 @@
...
+    A simply load test shape class that has different user and hatch_rate at
+    different stages.
+
+    Keyword arguments:
+
+        stages -- A list of dicts, each representing a stage with the following keys:
+            duration -- When this many seconds pass the test is advanced to the next stage
+            users -- Total user count
+            hatch_rate -- Hatch rate
+            stop -- A boolean that can stop that test at a specific stage
```
cyberw, line 27: Maybe a None entry could be used to signal the end of the test? Because you'll never need to do (x, y, True) for any other values of x or y than zero, right?

locust/shape.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Resets start time back to 0
+        """
+        self.start_time = time.monotonic()
+
+    def get_run_time(self):
+        """
+        Calculates run time in seconds of the load test
+        """
+        return round(time.monotonic() - self.start_time)
```
cyberw, line 22: I'm not sure a rounded time is a good thing to expose. This can give people the illusion that tick is called every second (which cannot be guaranteed because of gevent)

examples/custom_shape/double_wave.py
```diff
@@ -0,0 +1,49 @@
...
+
+class DoubleWave(LoadTestShape):
+    """
+    A shape to immitate some specific user behaviour. In this example, midday
+    and evening meal times.
+
+    Keyword arguments:
+
+        min_users -- minimum users
+        peak_one -- first peak size
```
cyberw, line 25: peak_one sounds to me like "the number of users at the first peak"

Not sure what would be a better name though.

examples/custom_shape/double_wave.py
```diff
@@ -0,0 +1,49 @@
...
+        peak_two -- second peak size
+        amplitude -- size of the test
+        time_limit -- total length of test
+    """
+    def __init__(self, min_users=20, peak_one=1, peak_two=1.5, amplitude=150, time_limit=600):
```
cyberw, line 30: This looks a little weird to me. It's not as if this constructor can be called from somewhere where these parameters are ever set?

Probably all these variables should be class level and then you can get rid of the parameters for the constructor

examples/custom_shape/stages.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    def __init__(self,
+            stages=[
+                {'duration': 60, 'users': 10, 'hatch_rate': 10},
+                {'duration': 80, 'users': 30, 'hatch_rate': 10},
+                {'duration': 100, 'users': 50, 'hatch_rate': 10},
+                {'duration': 120, 'users': 70, 'hatch_rate': 10},
+                {'duration': 180, 'users': 100, 'hatch_rate': 10},
+                {'duration': 220, 'users': 10, 'hatch_rate': 10},
+                {'duration': 230, 'users': 5, 'hatch_rate': 10},
```
cyberw, line 39: Include a stage with ramp down just to show that it can be done :)

examples/custom_shape/stages.py
max-rocket-internet: Well we need to return at least `(Int, Int)` to edit the running test so you are saying to stop the test return `(None, None)`? Or `(Int, Int, None)`? Or something else?

examples/custom_shape/double_wave.py
max-rocket-internet: I just threw this together to show how you could generate something totally custom and not related to "stages" or "step-load". I'll rework the wording here and maybe improve the maths.

examples/custom_shape/double_wave.py
max-rocket-internet: True. I'll fix.

examples/custom_shape/stages.py
max-rocket-internet: This does ramp down, see `users` key, it goes up and then down 🙂

But OK, I'll reduce number of stages here.

locust/shape.py
max-rocket-internet: OK

max-rocket-internet:
@cyberw I've addressed all your comments

examples/custom_shape/stages.py
cyberw: Ah, I kind of conflated two issues: 

I meant that tick() should do `return None` instead of `return (0, 0, True)` (which of course requires modifications in runners.py on line 297 as well.

But I also think that stop_at_end is an unnecessary complication. If you want to run your test forever, cant you just set the last step to use `'duration': 999999` ?

examples/custom_shape/stages.py
cyberw: Ah, my comment was about having so many stages, not about the very last stage :)

examples/custom_shape/stages.py
max-rocket-internet: > I meant that tick() should do return None instead of return (0, 0, True)

Is that consistent? Wouldn't something like `(Int, Int)`, `(None, None)` or `()` be better?

> But, I also think that stop_at_end is an unnecessary complication

This is just an example of what I imagine someone might want to use or see before they write their own. But I can remove `stop_at_end` it to make it simpler, no worries.

> cant you just set the last step to use 'duration': 9999999

I would say no, that's not very precise. What if you forget it's running and it gets to 10000000 seconds😃

locust/shape.py
```diff
@@ -0,0 +1,35 @@
...
+        Returns a tuple with 3 elements to control the running load test.
+
+            user_count -- Total user count
+            hatch_rate -- Hatch rate to use when changing total user count
+            stop_test -- A boolean to stop the test
+
+        """
+        run_time = self.get_run_time()
+
+        return (0, 0, True)
```
max-rocket-internet, line 35: @cyberw comments about what to return should go here

max-rocket-internet: > I meant that tick() should do return None instead of return (0, 0, True)

Is that consistent? Wouldn't something like `(Int, Int)`, `(None, None)` or `()` be better?

examples/custom_shape/stages.py
max-rocket-internet: Just to reiterate, these are examples. I imagine they would be in the docs or just here for people to find. They aren't to be copy/pasted and used as they are. I just imagined them to help a beginner to see how this class could be used 🙂

cyberw: Sure. Lets not spend too much time on them. On the other hand, if they are just examples then it makes sense to keep them very simplified (and let our users worry about edge cases like how to make it run forever etc).

locust/shape.py
cyberw: I dont know what you mean by (Int, Int) - how would that be used indicate the end of the test?

I think None is the most consistent of the options, also it requires the least amount of code and thought from the user. 

It is not a ramp down to null users using ramping down at a rate of null per second, so your second option doesnt make sense to me.

I guess we could do "anything that is "falsy" (to allow `()`, `None` and `False`), if that sounds better to you.

max-rocket-internet: I meant trying to keep the type that is returned consistent in all cases. It returns a tuple (of 2 integers normally `(Int, Int)`) so if you we want to be consistent then it should always return a tuple, perhaps sometimes an empty one, `()`.

But I know this is a [contentious topic in Python](https://www.reddit.com/r/Python/comments/30yb5t/return_none_or_not_to_return_none/), especially for people coming from other languages who find it a bit too loose.

Ultimately I don't care enough to write more comments about this so I'll just change it to return `(Int, Int)` or `None` 😃

max-rocket-internet:
@cyberw please re-review and when you're happy I'll work on docs

locust/runners.py
```diff
@@ -293,13 +293,13 @@ def shape_worker(self):
         logger.info("Shape worker starting")
         while self.state == STATE_INIT or self.state == STATE_HATCHING or self.state == STATE_RUNNING:
             new_state = self.environment.shape_class.tick()
-            user_count, hatch_rate, stop_test = new_state
-            if stop_test:
+            if new_state == None:
```
cyberw, line 296: use `is None` (not that it matters, but it is just nicer :)

max-rocket-internet: As you wish ✅

cyberw:
Looks great now. One question. How does it work when running web UI? 

@heyman I'll probably merge this soon. Last chance to give your view. Step two could be to remove the step-load thingy if you agree this is better.

max-rocket-internet:
I still need to write docs and this is not a quick task IMO.

> How does it work when running web UI?

Like I said in my opening comment, basically if a load test is started via the UI or API, the hatch rate and users is simply ignored. I think that's fair, if you're using a custom shape you will be a pretty advanced user anyway. I was thinking we could log a warning? What do you think?

cyberw:
I think logging a warning is fine (we should change that for -u/-r/-t parameters to make them behave that way as well tbh, but that may be a later thing)

max-rocket-internet:
OK I've included a few words in the log message to indicate the user_count/hatch_rate are ignored. I've also disabled editing a running shape test as this doesn't make sense.

I'll work on docs now.

max-rocket-internet:
OK I wrote a nice doc. Personally I think this feature deserves a mention in `docs/what-is-locust.rst`? But I'm obviously biased 😅

Anything else @cyberw?

max-rocket-internet:
Do I have to generate the API docs myself? Or is that done in CI or something?

```
$ make build_docs
sphinx-build -b html docs/ docs/_build/
Running Sphinx v2.4.3
Running `locust --help` command and storing output in /Users/max.williams/git/locust/docs/cli-help-output.txt
Generating RST table for Locust environment variables and storing in /Users/max.williams/git/locust/docs/config-options.rst

Extension error:
Could not import extension sphinx_search.extension (exception: No module named 'sphinx_search')
make: *** [build_docs] Error 2
```

cyberw:
I think it fits best in `running-locust-without-web-ui.rst` (but maybe it deserves a mention somewhere else too)

While you are adding the warning about ignoring -u/... when running with web ui, can you do it for non-custom-load-shapes as well? I've been annoyed by that many times + it is good to be consistent. Bonus points if you prefill the web ui form with the values from parameters instead of warning :)

Api docs are generated in CI (but if you install the readthedocs-sphinx-search package then you should be able to build them locally as well)

cyberw:
Regarding the docs, I think you can add it the same way as we did step load (a mention in the menu and a link in a suitable place, but keep it in a separate rst file). And then after a while, we remove step load :P Or maybe re-implement it using this method.

max-rocket-internet:
> While you are adding the warning about ignoring -u/... when running with web ui, can you do it for non-custom-load-shapes as well?

There's not really a place for me to log warning if user/hatch is passed because these are ALWAYS passed. In master, locust will throw an exception or error if you don't set user/hatch when starting a test.

> can you do it for non-custom-load-shapes as well? I've been annoyed by that many times + it is good to be consistent

I'm happy to add something but I'm not 100% sure what you mean. You mean a combination of command arguments and the UI? Or?

Here's what I have now in this PR, when running locust from command line:

```
    if shape_class and (options.num_users or options.hatch_rate or options.step_load):
        logger.error("The specified locustfile contains a shape class but a conflicting argument was specified: users, hatch-rate or step-load")
        sys.exit(1)
```

If you start a load test from the UI or API and there is a shape class in use then user/hatch is ignored and this is logged:

```
logger.info("Shape test starting. User count and hatch rate are ignored for this type of load test")
```

Editing a running load test with a shape class nothing but will log:

```
logger.info("There is an ongoing shape test running. Editing is disabled")
```

max-rocket-internet:
>  I've been annoyed by that many times + it is good to be consistent. Bonus points if you prefill the web ui form with the values from parameters instead of warning :)

If you create an new issue (it sounds separate) then I'll happily address it in a new PR 🙂

cyberw:
Looking really good now. Is it worth adding a test case for master/worker setup as well?

max-rocket-internet:
>  Is it worth adding a test case for master/worker setup as well?

OK @cyberw, done, with master + 3 workers scaling up, down and stop.

Are we ready? 😃 🙏

cyberw:
Looks amazing. One last question: what will happen if additional workers connect during the test? Will new users launch on the newly connected workers?

Tbh I'm not sure that is a use case we need to provide full support for, but some people argued differently (discussing global rps wait time I think). If it *should* work then maybe add it to the test? (adding a worker and then ramping up, making sure the users remain balanced after ramp up)

max-rocket-internet:
> what will happen if additional workers connect during the test? Will new users launch on the newly connected workers?

This behaviour is unchanged by this PR. Users are rebalanced across all workers when new workers connect. Does a test of that functionality have a relation to this PR? There is already the `test_rebalance_locust_users_on_worker_connect` test

cyberw:
I was just thinking those two features might be worth testing together. Maybe not. Merging now! Thanks a bunch!

max-rocket-internet:
@cyberw when is a new release likely to be made?

cyberw:
Tonight if you’re lucky, tomorrow if you are not :)

cyberw:
You weren't lucky :(

Also, I think I'll want to hold of for some more merges (1468 in particular seems to be very close)

cyberw:
Also, can you make a PR that greys out the numbers that cant be edited in the UI when running with a shape?

Also, I noticed something weird. What is going on here: https://github.com/locustio/locust/blob/62fa37bb7e6b71ac31cb013d22a310b50daa9693/locust/shape.py#L34

max-rocket-internet:
> Also, can you make a PR that greys out the numbers that cant be edited in the UI when running with a shape?

I can.

> the variable run_time is never used...

That is intentional. The user has to override `tick()` anyway and this was to show that they most likely want to calculate the run time and that this is how you do it. All 3 examples have the same line (but it's used). I can remove it? Or add a comment?

cyberw:
👍 

I think it makes sense to remove it, feels like it adds about as much confusion as it removes :) Also, pylance complains about it.

max-rocket-internet:
OK: https://github.com/locustio/locust/pull/1524

PR #1509, opened 2020-08-10
Fix typos

phil-davis:
in comments, ordinary text output, changelog and some test method names.

None of these should have any impact on "real" run-time functionality.


cyberw:
changelog.md is auto generated from PRs & commits made to github, so correcting spelling there is bad because any corrections would be overwritten on next release. 

Also, having an uncorrupted git-blame history for the lines in changelog.rst is probably more important than having the correct spelling.

If you file a new PR that leaves those two files untouched is then this is really nice! 

I'm so embarrased over all the typos :)

phil-davis:
Yeh, I suspected that. Once a commit is merged with a typo then the changelog is forever stuck that way.
I will make a reduced commit.

cyberw:
awesome, thanks!

PR #1510, opened 2020-08-10
Fix typo of redirect_response_codes

phil-davis:
This looks like it could be a bug, but I don't see where `redirect_resonse_codes` or `redirect_response_codes` is mentioned anywhere else.

Update: it comes from https://github.com/locustio/geventhttpclient

PR https://github.com/locustio/geventhttpclient/pull/3 would be needed first to make this pass. And I guess that would be a backward-compatibility break for https://github.com/locustio/geventhttpclient

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -221,14 +221,14 @@ class FastHttpSession(object):
                 headers['Accept'] = "application/json"
 
         if not allow_redirects:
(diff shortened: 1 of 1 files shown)


cyberw:
I dont think I can be bothered to make a geventhttpclient release for just this. Maybe when some other changes have been added.

phil-davis:
yes, I agree!

cyberw:
Closing this for now. I will try to remember it if we ever make a release of geventhttpclient.

PR #1520, opened 2020-08-14
Allow setting pass/fail... result - Add status url

lhupfeldt:


lhupfeldt:
This makes it possible to include locust in a CO/CD pipeline and have the pipeline break when performance test 'fails'.
Documentation is still missing but I would like some feedback on this.

cyberw:
Hi! I'm not sure this is a core use case. Interpreting the run as pass/fail is useful in command line/CI, but not so much in the UI I think (as the user can just as easily do that in their head). For CI there is some support for this functionality in locust-plugins https://github.com/SvenskaSpel/locust-plugins#command-line-options

lhupfeldt:
Hi

We are using the 'UI' as an api to monitor the run with the added 'status'
url. We hav locust deployed in containers and having a 'rest' api seems
like the simplest solution to keep track of things. I added the UI element
as a nice to have - if you have automatically determined that the test has
failed, then why not show it in the UI? Note that it only shows if a result
is actually set by the user.
I really disagree that determining whether the test is failed or passed can
easily be done in the head. The test will run in many different
environments with different requirements, and it will not always be the
same people there to look at it - automating the comparison of tests
statistics against your requirements/baseline seems to me as the only way
to go in a CI/CD pipeline.

Best regards / Venlig Hilsen
Lars Hupfeldt Nielsen
Hupfeldt IT

Phone: +45 2060 3546


Den lør. 15. aug. 2020 kl. 17.28 skrev Lars Holmberg <
notifications@github.com>:

cyberw:
I didnt mean to say it isnt useful to *anyone* (obviously it is, otherwise we would not be having this conversation :)

I'm just saying that in my opinion it is not useful to _enough_ people to motivate the added complexity and future maintenance costs to include this in core. I think it is entirely possible to do everything in this PR using the init integration point (https://docs.locust.io/en/stable/extending-locust.html#adding-web-routes), except maybe the UI element. Perhaps it would be a nice "advanced" example to include in the documentation somewhere.

lhupfeldt:
I will reimplement as as plugin.

It would be nice if the UI was pluggable as well then.
Since you are intent on keeping core very small it kind of limits what you can do when you cannot extend the UI.

I also think that keeping a list of plugins (as a very prominent link) would be of great benefit to your users.

cyberw:
Thanks! If you make a PR for https://github.com/SvenskaSpel/locust-plugins I will consider it!

Also, I should make the link to plugins more prominent I guess...

PR #1504, opened 2020-08-06
Download full history - see issue 1468

lhupfeldt:
Adds another download link to the UI when --csv-full-history was specified

See: https://github.com/locustio/locust/issues/1468

Does not implement the suggested 'download-all'

PR #1522, opened 2020-08-16
Keep csv files open

lhupfeldt:
The intent of is to reduce os calls and disk IO.
This code goes on top of my other PR https://github.com/locustio/locust/pull/1504, so submitting as draft.

### DIFF
--- a/docs/retrieving-stats.rst
+++ b/docs/retrieving-stats.rst
@@ -26,4 +26,5 @@ You can also customize how frequently this is written if you desire faster (or s
 .. code-block:: python
 
     import locust.stats
-    locust.stats.CSV_STATS_INTERVAL_SEC = 5 # default is 2 seconds
+    locust.stats.CSV_STATS_INTERVAL_SEC = 5 # default is 1 second
+    locust.stats.CSV_STATS_FLUSH_INTERVAL_SEC = 60 # Determines how often the data is flushed to disk, default is 10 seconds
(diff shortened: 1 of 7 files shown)


cyberw:
Does this also split stats into stats and stats_csv? It might make sense, but as I dont really know the web interface code very well, I can't really judge if that is a good idea.

lhupfeldt:
> Does this also split stats into stats and stats_csv? It might make sense, but as I dont really know the web interface code very well, I can't really judge if that is a good idea.

Yes, it splits out the csv generation. I think it gives better readability, but I can merge it back into stats if you prefer.

cyberw:
Is this ok to merge now?

lhupfeldt:
Let me rebase on master first.

cyberw:
Good to go?

lhupfeldt:
> ~Good to go?~ oh right, still waiting to un-split stats/stats_csv (preferably by rewriting the history so that git-blame is not distorted)

I think I will be to blame for the stats-csv stuff regardless how it is un-split, since all lines will be indented under the class.
So the easiest will be to simply copy stats_csv content back into stats.
Any objection to that?

lhupfeldt:
Created a new single commit and force pushed, should keep the blame changes to a minimum, although not sure there is much difference.

cyberw:
LGTM. Thanks!

PR #1524, opened 2020-08-18
Grey out UI input fields when LoadTestShape is in use

max-rocket-internet:
- Grey out input fields for starting and editing users and hatch rate when `LoadTestShape` is in use
- Add CSS style for above
- Remove unused `run_time = self.get_run_time()` from `shape.py`

### DIFF
--- a/locust/shape.py
+++ b/locust/shape.py
@@ -31,6 +31,5 @@ class LoadTestShape(object):
         If `None` is returned then the running load test will be stopped.
 
         """
-        run_time = self.get_run_time()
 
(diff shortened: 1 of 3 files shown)


max-rocket-internet:
Starting a load test looks like this:

Screen Shot 2020-08-18 at 10 31 49 (2)

And editing looks like this and is completely disabled:

Screen Shot 2020-08-18 at 10 33 35 (2)

cyberw:
LGTM. Thanks!

PR #1526, opened 2020-08-18
Make hatch-rate parameter deprecated instead of killing it right away.

cyberw:
The renamed events are still incompatible, but that should impact much fewer users.

Should we still call it 2.0? I guess we should.

ping @lhupfeldt

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,7 @@ For full details of the Locust changelog, please see https://github.com/locustio
 1.2
 ===
 
-* Rename hatch rate to spawn rate (this may be a breaking change in some cases!)
+* Rename hatch rate to spawn rate (the --hatch-rate parameter is only deprecated, but the hatch_complete event has been renamed spawning_complete)
 * Ability to generate any custom load shape with LoadTestShape class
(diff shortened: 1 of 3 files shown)


lhupfeldt:
How does the renamed events affect users? I'm too new to Locust to determine whether it is necessary to call it 2.0.
If you follow strict semver then any breaking change should get a new major version - but in my opinion that can also be confusing for users that will generally expect major new functionality in a new major version.

cyberw:
I think there is enough new functionality to call it 2.0. The breaking change is that if a user listens to the hatch_complete event, they will have to change it to use spawning_complete instead (example: https://github.com/SvenskaSpel/locust-plugins/blob/8a84100bff1ecffe6164be2a9bb631c631ef11b4/locust_plugins/listeners.py#L109)

Whether we call it 1.2 or 2.0 is mostly a matter of taste I guess. Parts of the general OS community (and browser developers etc) seem to have gone the way of releasing lots of major versions (chrome is now 84.x and firefox something like 78.x :)

lhupfeldt:
2.0 seems fine to me.

PR #1532, opened 2020-08-20
Update repo README with new wording, locust example, screenshots

max-rocket-internet:
- Copy latest wording from docs
- Add very simple example that shows clearly how a locust file is written
- Add some screen shots from the UI
- Format features correctly using `####` instead of bold bullet points
- Use consistent bullet point syntax
- Remove unnecessary HTML
- Add github links for authors
- Move `Links` to the end as these are boring
- Remove `Documentation` section as this is already in the `Links` section


cyberw:
Thx!

cyberw:
Huh. It broke the pypi page: https://pypi.org/project/locust/

Perhaps we need to use absolute URLs to github (unless there is some other solution)

max-rocket-internet:
OK I can make the image links use `raw` links to this repo

PR #1539, opened 2020-08-21
Fix load shape worker in headless.

cyberw:


max-rocket-internet:
> Cool, then you'll get all the UI bugs from now ;P

😅

> Btw, have you checked out the Timescale+Grafana stuff in locust-plugins?

I had a quick look just now. We don't use the locust UI for evaluating the test results or looking at the performance of the application under test. If we are running a non-automated test then we just use it to see if the test is running, if it's healthy (exceptions etc) and check worker count.

We are big Prometheus users and have custom dashboards for our applications (and for our locust k8s infrastructure). We use these for evaluating the test results as they are also what we use for live environments so the results are comparable.

For automated load tests we use Spinnaker pipeline to scale up locust and application infrastructure, start the load test via locust API, monitor exceptions and error rates via the locust API for a set duration, then stop the test afterwards (or if exceptions and error rates are too high).

cyberw:
> > Btw, have you checked out the Timescale+Grafana stuff in locust-plugins?
> 
> I had a quick look just now. We don't use the locust UI for evaluating the test results or looking at the performance of the application under test. If we are running a non-automated test then we just use it to see if the test is running, if it's healthy (exceptions etc) and check worker count.
> 
> We are big Prometheus users and have custom dashboards for our applications (and for our locust k8s infrastructure). We use these for evaluating the test results as they are also what we use for live environments so the results are comparable.

Nice! Havent really used Prometheus that much myself, but it looks really nice.

> For automated load tests we use Spinnaker pipeline to scale up locust and application infrastructure, start the load test via locust API, monitor exceptions and error rates via the locust API for a set duration, then stop the test afterwards (or if exceptions and error rates are too high).

Cool stuff. I wanted Klarna to use Spinnaker when I worked there but I dont think I ever convinced anyone (when I was there I think they were building something similar in house :-/ )

One of the nice things about timescale/the logging plugin is I get exact metrics for every request, logging for the entire test run (and trends for aggregated results as well).

But I guess if you have good monitoring on the system side then you may not need that (as I'm more of a consultant performance tester, I dont always have access/knowledge of the underlying system's metrics)

PR #1565, opened 2020-09-16
Update documentation for schedule_task parameters in TaskSet (task.py)

kmels:
The parameters "args" and "kwargs" were removed in commit 82c2d6a68f946bdd303642dff6a89e0d5c2bf298. However they still show in the latest documentation https://docs.locust.io/en/latest/api.html?highlight=schedule_task#locust.TaskSet.schedule_task


cyberw:
thx

PR #1566, opened 2020-09-17
Add friendlier message about expected limit

parberge:
Will now print something like this to standard out:

```
[2020-09-17 09:39:33,411] 7131dd94fc3b/WARNING/locust.main: System open file limit '2048' is below recommended setting '10000'. It's not high enough for load testing, and the OS didn't allow locust to increase it by itself. See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-number-of-open-files-limit for more info.
```

Since the expected setting isn't documented without this message it's
hard to know what the recommended value is.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -230,7 +230,8 @@ def main():
 
     if options.master:
         runner = environment.create_master_runner(
-            master_bind_host=options.master_bind_host, master_bind_port=options.master_bind_port,
+            master_bind_host=options.master_bind_host,
+            master_bind_port=options.master_bind_port,
         )
     elif options.worker:
         try:
--- a/locust/main.py
+++ b/locust/main.py
@@ -172,17 +172,17 @@ def main():
         try:
             import resource
 
-            recommended_open_file_limit = 10000
+            minimum_open_file_limit = 10000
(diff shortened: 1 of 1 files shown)


cyberw:
The recommended limit is actually "as high as possible". The reason I chose 10000 is because that is the highest we can automatically increase it to on macOS. Not sure what would be the best way to describe it...

cyberw:
Overall I prefer a short log entry but maybe some more details in the wiki entry.

parberge:
@cyberw Ok, so what action should I take? I'm not against any of the suggestions, but I'm not sure what you want me to do with this PR.

cyberw:
I guess if you just change the word ”recommended” to ”minimum” that is good enough for me to merge.

cyberw:
Dont forget to format using Black :) (if you can configure your editor to automatically do that, then feel free to add that config to the repo, like it is for VS Code at the moment)

parberge:
@cyberw I do have black with vscode. So my changes are formatted with black.

However I chose not to include the formatting changes that had nothing to do with my changes to make it clear.
I can include those if you really want

parberge:
Oh you have that as a step in CI. I see.
I'll include those changes as well then.

cyberw:
thanks!

Issue #1535, opened 2020-08-20
After starting test from UI, state shows as STOPPED even though it's running


max-rocket-internet:
### Describe the bug

After starting test from UI, state shows as `STOPPED` even though it's running.

If I click start again, then the status is updated to `SPAWNING` correctly

![Screen Shot 2020-08-20 at 15 32 25](https://user-images.githubusercontent.com/8859277/90776440-63114900-e2fa-11ea-83e1-62293bc18a0c.png)

### Expected behavior

Status should show as "SPAWNING" I believe.

### Steps to reproduce

Just start test from UI.

### Environment

- OS: k8s
- Locust version: 1.2.1

max-rocket-internet:
I can't really reproduce this reliably though 🤔

max-rocket-internet:
```
bash-5.0$ curl http://locust-crs:8089/swarm --data-raw 'user_count=20000&spawn_rate=20'
{
  "host": "https://crs-01.ldt.restaurant-partners.com",
  "message": "Swarming started",
  "success": true
}

# Wait 10 seconds

bash-5.0$ curl -s http://locust-crs:8089/stats/requests | grep -m 1 state
  "state": "stopped",
```

cyberw:
Are you using load shapes? (just guessing here, I dont know this part of the code base so you're on your own :) )

max-rocket-internet:
> Are you using load shapes?

Nope

max-rocket-internet:
I'm in a good position to debug it myself but thought I'd post an issue in case others had the same problem.

mboutet:
I also experienced this a few times and came to the conclusion that it was happening when the workers were overloaded (high CPU usage). In my case, the workers were performing some blocking tasks on start (random texts generation). To prevent this, I limited the amount of users per worker and I also sprinkle some `gevent.sleep(0)` in the blocking code so that the event loop is not completely blocked.

cyberw:
@max-rocket-internet did you manage to figure it out?

max-rocket-internet:
> did you manage to figure it out?

Not yet! But we are seeing this every day. We might need to role back to a previous version. I'm still debugging it.

cyberw:
Perhaps it only happens with a large number of workers? Were you using an on_stop method? (in that case, try running latest master with the above mentioned fix)

With no more details I think I'll have to close this.

max-rocket-internet:
> Were you using an on_stop method? 

[...]

PR #1570, opened 2020-09-28
Adding more debug logging for runners.py

max-rocket-internet:
It's almost impossible to investigate issues like https://github.com/locustio/locust/issues/1535 and some of my previous issues without making custom local changes to add more logging, which is annoying and time consuming.

So here I'm adding some more debug log messages and add a function for updating the `state`, with a debug log message.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -84,7 +84,7 @@ class Runner(object):
 
         # register listener that resets stats when spawning is complete
         def on_spawning_complete(user_count):
-            self.state = STATE_RUNNING
+            self.update_state(STATE_RUNNING)
             if environment.reset_stats:
                 logger.info("Resetting stats\n")
                 self.stats.reset_all()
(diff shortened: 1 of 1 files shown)


cyberw:
lgtm.

locust/runners.py
```diff
@@ -578,27 +586,33 @@ def start(self, user_count, spawn_rate):
...
+            logger.debug("Sending spawn message to client %s" % (client.id))
             self.server.send_to_client(Message("spawn", data, client.id))
 
-        self.state = STATE_SPAWNING
+        self.update_state(STATE_SPAWNING)
 
     def stop(self):
         if self.state not in [STATE_INIT, STATE_STOPPED, STATE_STOPPING]:
-            self.state = STATE_STOPPING
+            logger.debug("Stopping...")
```
cyberw, line 596: oh, but I think this line is no longer needed. you can remove it as part of the actual fix later on :)

max-rocket-internet: Why is it not needed?

cyberw: Because we’re also going to be logging ”Updating state to STATE_CLEANUP, old state was X” right after that.

Issue #1530, opened 2020-08-19
Add ability to easily extend Locust web UI


solowalker27:
### Is your feature request related to a problem? Please describe.
It would be very nice to be able to extend the Locust web UI to make custom data visible natively in the same UI as the basic Locust stats. A separate route for each data set I want to view that's just a dict/JSON or some other generic raw data view that I have to manually refresh is a very poor experience. They're also completely undiscoverable to other people who may be using my Locust set up to run tests. Being able to extend the UI would solve all these problems and open up a lot of other interesting possibilities.

### Describe the solution you'd like
We already can add new routes to Flask relatively easily. And data can be passed from workers to the master so the master can do things with data from the workers, including serve it via added routes, and I have done so. What's missing is a UI component.

This would first require adding some extension `blocks` to `index.html`. That would allow users to create their own HTML file that `extends 'index.html'`. That's the easy part. I've attached an example of this that I'm currently using.
[index.html.txt](https://github.com/locustio/locust/files/5098400/index.html.txt)

The problems come from actually trying to go any further. I was ultimately able to accomplish the end result of serving up an extended UI page but ran into several issues that made it more difficult than I think it could be. To make this a viable feature, I feel like the following would need to be addressed:

1. Before I can serve additional files such as my extended HTML page or a Javascript file to go with it, I have to create my own jinja instance to look for the path I want to store my files in. There may be an easier way to do this, but this is what I found to make it work:
```
        templateLoader = jinja2.FileSystemLoader(searchpath="./")
        templateEnv = jinja2.Environment(loader=templateLoader)
        template_file = "file.html"
        template = templateEnv.get_template(template_file)
        …
        template.render(
            # all the stuff index.html needs
        )
```
2. I had to copy almost all the code for the `/` route in `web.py` to make sure I got everything the `index.html` page needs to render my extended page properly because of the number of variables `index.html` relies on.
3. Since there's only a single CSS style sheet allowed per page, I do have to do my styling in Javascript.

If there's no better way to do number 1, it'd have to be accepted and just included as part of the documentation.

[...]

PR #1574, opened 2020-09-28
More easily extend web UI

solowalker27:
Addresses #1530. Includes the following additions and changes:

* `index.html` is extendable with blocks for tabs, panes, and scripts
* Args required to properly render `index.html` are retrievable via `web_ui.update_template_args()`
* Event for web UI's Reset Stats button to aid in custom data resetting
* Make some UI classes and functions more generic and more easily reusable for extended UI

### DIFF
--- a/locust/static/locust.js
+++ b/locust/static/locust.js
@@ -1,5 +1,5 @@
-$(window).ready(function() {
-    if($("#user_count").length > 0) {
+$(window).ready(function () {
+    if ($("#user_count").length > 0) {
         $("#user_count").focus().select();
     }
 });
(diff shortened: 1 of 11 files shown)


cyberw:
Cool stuff! 

Can you add some documentation and an example for how to extend the UI?

As I dont use the web ui and dont know the code behind very well, I'll take your word for it being sensible & working :) If something does break in the UI related to these changes, will you have some time to help fix it? Maybe you'd even be interested in being "the" UI maintainer for Locust :)

locust/static/locust.js
```diff
@@ -40,15 +40,19 @@ $(".close_link").click(function(event) {
...
-        responseTimeChart.resize();
-        usersChart.resize();
-    }
+$("ul.tabs").tabs("div.panes > div").on("onClick", function (event) {
+    // trigger resizing of charts
+    resizeCharts();
 });
 
+var charts = []
+function resizeCharts() {
```
max-rocket-internet, line 49: Will this resize the charts when the window is resized?

solowalker27: Yes, the same as it did before—it's triggered only if you click on a different tab then click back on the tab with the charts.

solowalker27:
Added example of extending the web UI.

I can help with the UI code where I can. I'm not exactly an expert, just hacking away at it. I needed to do this for my work and decided to pass the improvements upstream so I didn't have to maintain them in a fork. I don't know if I can sign on to be a maintainer of the web UI but I can provide feedback and review PRs if I can.

cyberw:
Lgtm. Can you just add a reference to the example in the documentation (extending-locust.rst)

solowalker27:
How's that?

Issue #1577, opened 2020-09-29
After starting and then stopping a load test, master is updating state in a loop


max-rocket-internet:
After this PR: https://github.com/locustio/locust/pull/1570

Then master is logging this repeatedly in a loop after stopping a load test:

```
[2020-09-29 15:15:40,859] locust-vps-master-746cf87dc7-rmws5/DEBUG/locust.runners: Updating state to stopped, old state was stopped
```

I will make a PR to fix.

cyberw:
Wow, that is weird. Is it completely stuck, or just logging it a bunch of times before actually finishing?

max-rocket-internet:
Sorry, it's not stuck, everything works, but it's continually updating the state. Only visible with debug log level.

PR #1580, opened 2020-09-30
Don't log state change if it's the same

max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/1577

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -618,8 +618,10 @@ class MasterRunner(DistributedRunner):
         self.greenlet.kill(block=True)
 
     def check_stopped(self):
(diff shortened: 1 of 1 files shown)


cyberw:
Are you sure there is not an underlying issue? I dont see why we should make tons of state changes to the same state...

max-rocket-internet:
True. I can find out why something is continually updating the state to the same value.

max-rocket-internet:
It's from here: https://github.com/locustio/locust/blob/master/locust/runners.py#L624

I will test a different fix then 🙂

cyberw:
Nice! maybe just add `and not self.state == STATE_STOPPING` to that if statement...

max-rocket-internet:
OK done that now some test is failing. I will look into it.

Issue #1557, opened 2020-09-10
LoadTestShape.get_run_time is not relative to start of test


Bubba74:
Admittedly, I'd say this particular bug is open to interpretation and I can see benefit to the current behavior.
However, I believe the current behavior is the opposite of what most implementations would require.

### Describe the bug

The get_run_time() method in LoadTestShape returns the time delta relative to the start of Locust, not the start of the test.
This is exacerbated when there is a significant delay between Locust startup and test start (for example through the WebUI)

### Expected behavior

LoadTestShape.get_run_time() should return roughly 0 seconds the first time LoadTestShape.tick() is called.
If the test resets (say, through the WebUI), the LoadTestShape timer should reset too.

### Actual behavior

When Locust calls LoadTestShape.tick() for the first time, a call to self.get_run_time() will return a time delay that incorporates Locust startup and all time spent in the WebUI.

Restarting the example below is basically impossible as the timer has already exceeded 10 seconds if the LoadTest finished normally.

### Workaround
A subclass of LoadTestShape could keep a `self.first_loop = True` variable then do:
``` .py
if self.first_loop:
    self.reset_time()
    self.first_loop = False
```

### Steps to reproduce
 
``` locustfile.py
# locustfile.py
from locust import User, task, LoadTestShape, constant

class MyUser(User):
    wait_time = constant(10)

    @task
    def noop(self):
        pass

class Shape(LoadTestShape):
    def tick(self):
        time = self.get_run_time()
        if time > 10:
            return None
        print("Time:", time)
        return (1, 1)
```
The above runs for ~10s when run with: ` locust --headless `
But might never run if started manually (10s later) via: `locust`

### Environment
- OS:
- Python version: PyPy 7.3.1 with Python 3.6.9 (but this problem is evident in the master branch source)
- Locust version: 1.2.3
- Locust command line that you ran:
- Locust file contents (anonymized if necessary):

max-rocket-internet:
You're right @Bubba74. I like your solution, please open a PR 🙂

PR #1581, opened 2020-09-30
Run time relative to start when using LoadTestShape

DennisKrone:
Adresses #1557

The start time of a test with a shape should now be reset every time a new test is being run. Not from the time of locust init.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -193,10 +193,9 @@ def main():
     environment = create_environment(user_classes, options, events=locust.events, shape_class=shape_class)
 
     if shape_class and (options.num_users or options.spawn_rate or options.step_load):
-        logger.error(
-            "The specified locustfile contains a shape class but a conflicting argument was specified: users, spawn-rate or step-load"
+        logger.warning(
+            "The specified locustfile contains a shape class but a conflicting argument was specified: users, spawn-rate or step-load. Ignoring arguments"
         )
-        sys.exit(1)
 
     if options.show_task_ratio:
         print("\n Task ratio per User class")
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -6,7 +6,7 @@ import sys
 import traceback
 import warnings
 from uuid import uuid4
-from time import time, monotonic
+from time import time
 
 import gevent
 import greenlet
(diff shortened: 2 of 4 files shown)


locust/shape.py
```diff
@@ -7,7 +7,8 @@ class LoadTestShape(object):
     during a load test.
     """
 
-    start_time = time.monotonic()
+    def __init__(self):
+        self.start_time = time.monotonic()
```
max-rocket-internet, line 11: I don't think this is ever called? Because the class is never instantiated, it's just declared [eg](https://github.com/locustio/locust/blob/master/examples/custom_shape/step_load.py#L17), so `__init__ ` is never run.

locust/main.py
```diff
@@ -193,10 +193,9 @@ def main():
     environment = create_environment(user_classes, options, events=locust.events, shape_class=shape_class)
 
     if shape_class and (options.num_users or options.spawn_rate or options.step_load):
-        logger.error(
-            "The specified locustfile contains a shape class but a conflicting argument was specified: users, spawn-rate or step-load"
+        logger.warning(
```
max-rocket-internet, line 196: Like it 👍

locust/runners.py
```diff
@@ -352,6 +352,7 @@ def start_shape(self):
         self.update_state(STATE_INIT)
         self.shape_greenlet = self.greenlet.spawn(self.shape_worker)
         self.shape_greenlet.link_exception(greenlet_exception_handler)
+        self.environment.shape_class.start_time = monotonic()
```
max-rocket-internet, line 355: Could you not just call `self.environment.shape_class.reset_time()` here and that would be enough?

locust/shape.py
cyberw: Isn't it instantiated here: https://github.com/locustio/locust/blob/08d5173c2551e2f8c16404345816482f5af25394/locust/main.py#L93 (making the `shape_class` variable name a bit of a misnomer...)

max-rocket-internet: Ah yes, you are right. I wrote that, I should know!

cyberw:
nice!

PR #1603, opened 2020-10-22
Replace '<' and '>' for swarm 'host' field. Fix for XSS attack.

lhupfeldt:


cyberw:
Hmm... I think this might be the wrong place to escape this. 

Right now `host` doesnt take a complete url (http:/foo.com/THISSTUFF/) to be used for relative names, but we want to change that, and in this case we might even want to allow http:/foo.com/?a=1&b=2, which would be broken by this change. Can we instead change it to escape it before showing it back to the user?

XSS is not really a major thing for us, as the web ui should never be exposed to external users (just spawning a test with 10000000 users could at least crash the machine it is running on, so that will always be a bad idea, and I'm sure there are other vulnerabilities as well)

lhupfeldt:
I think the place is correct. It would be very hard to ensure the security by attempting to replace at every point of use. The actual error found was from injecting a script and then viewing the exception log!
Maybe html.escape is too aggressive, but I was reluctant to pull in a big input validation framework like WTForms for this one field.

```
 html.escape("http:/foo.com/?a=1&b=2")
'http:/foo.com/?a=1&amp;b=2'
```
Maybe just replace `&amp;` back to `&` would be a quick solution. And then it can be revisited when you allow for more complex URLs.

I agree that most people will not expose Locust to the internet, which is also why I mentioned as not being very critical, regardless of that I still think it should be fixed (I personally will not be allowed to roll out the use of Locust in our organisation until a fix is in place).

lhupfeldt:
This was the only input validation bug found, after all there is a very limited number of input fields.

cyberw:
It seems html.escape only escapes &, < and > anyways. If you just do a string replace for < and > instead, then I'm fine with it.

lhupfeldt:
Done

cyberw:
Thx!

lhupfeldt:
Thank you for the quick response

lhupfeldt:
:( the black check failed I see, but it only says "would be refomattted". I did not see that because I had not raised the number of open files, so never got the black test.

lhupfeldt:
Shouldn't all the tests be run on the PRs?

cyberw:
They would have been if I hadnt been so sloppy as to merge it before they had completed :)

Fixed in master now.

lhupfeldt:
😃

lhupfeldt:
@cyberw I'm hoping for this to get released soon. The security team keeps asking for this to be deployed.

cyberw:
I can do it next week

lhupfeldt:
Thank you

PR #1611, opened 2020-11-01
Added --html option to generate HTML report

vsapronov:
This is very useful when running locust in headless mode.


cyberw:
Cool stuff! Can you add a test case?

vsapronov:
> Cool stuff! Can you add a test case?

Hi, I think the functionality that I have extracted is under tests that are hitting `/stats/report` here: https://github.com/locustio/locust/blob/master/locust/test/test_web.py#L300-L345.

cyberw:
Can you add a "real" test in test_main.py ? I agree that some of the underlying features are tested, but not the writing to file etc.

cyberw:
@vsapronov ?

cyberw:
replaced by #1637

PR #1612, opened 2020-11-02
Add key commands for increasing and stopping users 

### LINKED ISSUE #1600: Control user count from terminal
It would be nice if we could control the number of users with keyboard presses from terminal, particularly when running headless.

Something like Alt+Up/Down to start/stop one user and Alt+Shift+Up/Down to start/stop 10 users. We should also log what happened (new target user count). I would have preferred Ctrl, but macOS hijacks Ctrl+Up/Down by default.

DennisKrone:
w = increase by 1
W = increase by 10
s = stop 1
S = stop 10

fixes #1600

### DIFF
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -39,7 +39,7 @@ class KeyPoller:
         return self
 
     def __exit__(self, type, value, traceback):
-        if not os.name == "nt" and getattr(self, "tattr"):
+        if not os.name == "nt" and hasattr(self, "tattr"):
             termios.tcsetattr(self.stdin, termios.TCSANOW, self.tattr)
 
     def poll(self):
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -18,72 +18,88 @@ else:
     import tty
 
 
-class KeyPoller:
-    def __enter__(self):
-        if os.name == "nt":
-            self.read_handle = GetStdHandle(STD_INPUT_HANDLE)
-            self.read_handle.SetConsoleMode(ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT | ENABLE_PROCESSED_INPUT)
+class UnixKeyPoller:
+    def setup(self):
+        try:
+            self.stdin = sys.stdin.fileno()
+            self.tattr = termios.tcgetattr(self.stdin)
+            tty.setcbreak(self.stdin, termios.TCSANOW)
+        except termios.error:
+            pass
+        return self
 
-            self.cur_event_length = 0
-            self.cur_keys_length = 0
+    def cleanup(self):
+        termios.tcsetattr(self.stdin, termios.TCSANOW, self.tattr)
 
-            self.captured_chars = []
-        else:
-            try:
-                self.stdin = sys.stdin.fileno()
-                self.tattr = termios.tcgetattr(self.stdin)
-                tty.setcbreak(self.stdin, termios.TCSANOW)
-            except termios.error:
-                pass
+    def poll(_self):
+        dr, dw, de = select.select([sys.stdin], [], [], 0)
+        if not dr == []:
+            return sys.stdin.read(1)
+        return None
+
+
+class WindowsKeyPoller:
+    def setup(self):
+        self.read_handle = GetStdHandle(STD_INPUT_HANDLE)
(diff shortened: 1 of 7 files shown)


locust/main.py
```diff
@@ -344,11 +345,24 @@ def timelimit_stop():
...
+    input_listener_greenlet = None
     if not options.only_summary and (options.print_stats or (options.headless and not options.worker)):
         # spawn stats printing greenlet
         stats_printer_greenlet = gevent.spawn(stats_printer(runner.stats))
         stats_printer_greenlet.link_exception(greenlet_exception_handler)
 
+        input_listener_greenlet = gevent.spawn(
+            input_listener(
+                {
+                    "w": lambda: runner.spawn_users(1, runner.spawn_rate),
```
cyberw, line 357: no need to use runner.spawn_rate, just do it immediately (set it to maybe 100). If people want to spawn slowly then let them wait between keypresses :) (when spawning only one user it will not matter anyway)

locust/main.py
```diff
@@ -344,11 +345,24 @@ def timelimit_stop():
         spawn_run_time_limit_greenlet()
 
     stats_printer_greenlet = None
+    input_listener_greenlet = None
     if not options.only_summary and (options.print_stats or (options.headless and not options.worker)):
         # spawn stats printing greenlet
         stats_printer_greenlet = gevent.spawn(stats_printer(runner.stats))
         stats_printer_greenlet.link_exception(greenlet_exception_handler)
 
+        input_listener_greenlet = gevent.spawn(
```
cyberw, line 354: Is this really the right place to register it? Why not register it all the time (before this if statement, maybe skipping options.worker)

The current way, running locust with --only-summary, would silently disable key commands (which is weird)

And I dont think there is any down side to having it running even in web mode (although you might need to verify that locust is actually running before someone tries to ramp up/down)

locust/input_events.py
```diff
@@ -0,0 +1,91 @@
+import time
+
+isWindows = False
```
cyberw, line 3: This is a nicer way to know if we're on Windows (no need to try/catch & no need to store our own variable)

```
import os

if os.name == 'nt':
     ...
```

locust/input_events.py
```diff
@@ -0,0 +1,91 @@
...
+            self.captured_chars = []
+        else:
+            self.stdin = sys.stdin.fileno()
+            self.tattr = termios.tcgetattr(self.stdin)
+            tty.setcbreak(self.stdin, termios.TCSANOW)
+
+        return self
+
+    def __exit__(self, type, value, traceback):
+        if isWindows:
```
cyberw, line 40: use 

```
if not something:
     # do something 
```

instead of 

```
if something:
    pass
else:
    # do something
```

locust/input_events.py
```diff
@@ -0,0 +1,91 @@
...
+                    if cur_event.EventType == KEY_EVENT:
+                        if ord(cur_event.Char) == 0 or not cur_event.KeyDown:
+                            pass
+                        else:
+                            cur_char = str(cur_event.Char)
+                            self.captured_chars.append(cur_char)
+
+                self.cur_event_length = len(events_peek)
+
+            if not len(self.captured_chars) == 0:
```
cyberw, line 66: I'm 90% sure this can be replaced with just
     if self.captured_chars:

locust/input_events.py
```diff
@@ -0,0 +1,91 @@
...
+
+            events_peek = self.read_handle.PeekConsoleInput(10000)
+
+            if len(events_peek) == 0:
+                return None
+
+            if not len(events_peek) == self.cur_event_length:
+                for cur_event in events_peek[self.cur_event_length :]:
+                    if cur_event.EventType == KEY_EVENT:
+                        if ord(cur_event.Char) == 0 or not cur_event.KeyDown:
```
cyberw, line 58: Most likely, this:

                        if ord(cur_event.Char) == 0 or not cur_event.KeyDown:
                            pass
                        else:
                            cur_char = str(cur_event.Char)
                            self.captured_chars.append(cur_char)

can be replaced by 

                        if ord(cur_event.Char) and cur_event.KeyDown:
                            cur_char = str(cur_event.Char)
                            self.captured_chars.append(cur_char)

PR #1617, opened 2020-11-06
issue-1571 : Added a logging line when workers connect

zdannar:


cyberw:
Make it debug instead of info and you have a deal :) People might have a lot of workers and we shouldnt spam their logs by default.

zdannar:
Thanks @cyberw.  Done.

PR #1623, opened 2020-11-11
bugfix fail_ratio calculation error

solxnp:


cyberw:
image

num_requests contains both failures and successes, so the current behaviour is correct (except in the case you are experiencing :) applying this change would count failures twice.

cyberw:
It is really weird that neither the current version nor the suggested change fails any unit tests though. We should add such a test.

cyberw:
Actually, your suggested change fails the tests, but for some reason no builds were executed. Weird.

solxnp:
I think I know what the problem is
thanks

PR #1630, opened 2020-11-16
Fix bug causing 100% CPU usage when starting Locust without interactive shell

### LINKED ISSUE #1629: Locust docker version 1.4.0 using 100% CPU on idle
Test environment:
Ubuntu 18.04.1
Docker version 19.03.6

Environment stays the same only the docker image version is changed.
Running the docker container with the same locustifle.py which is an example from https://hub.docker.com/r/locustio/locust

```python

    def on_start(self):
        self.client.post
("/login", json={"username":"foo", "password":"bar"})

    @task
    def hello_world(self):
        self.client.get("/hello")
        self.client.get("/world")

    @task(3)
    def view_item(self):
        for item_id in range(10):
            self.client.get(f"/item?id={item_id}", name="/item")
```

When running locustio/locust version 1.3.2 the idle CPU load is around 0.0% and stable. 

Running locustio/locust version 1.4.0 the idle CPU load goes up to 100.0% at once when the container is started and stays between 96.7% and 100.0%.

heyman:
Should fix #1629 .

### DIFF
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -94,6 +94,7 @@ def input_listener(key_to_func_map):
                     gevent.sleep(0.2)
         except Exception as e:
             logging.warning(f"Exception in keyboard input poller: {e}")
+            raise
         finally:
             poller.cleanup()
 
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -94,7 +94,6 @@ def input_listener(key_to_func_map):
                     gevent.sleep(0.2)
         except Exception as e:
             logging.warning(f"Exception in keyboard input poller: {e}")
-            raise
         finally:
             poller.cleanup()
 
--- a/locust/main.py
+++ b/locust/main.py
@@ -342,7 +342,7 @@ def main():
         pass  # dont log anything - not having a time limit is normal when not running headless
 
     input_listener_greenlet = None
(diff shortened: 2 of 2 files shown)


locust/input_events.py
```diff
@@ -94,6 +94,7 @@ def input_listener_func():
                     gevent.sleep(0.2)
         except Exception as e:
             logging.warning(f"Exception in keyboard input poller: {e}")
+            raise
```
heyman, line 97: I've added this for a more robust code, but I'm not sure if there are any valid exceptions that could happen during a normal session. If so, we should probably handle them seprately.

cyberw:
Cool stuff! Me and @DennisKrone have a fix on the way, but will integrate your check for isatty() into it (instead of relying on catching an exception, which was our first attempt)

heyman:
> Cool stuff! Me and @DennisKrone have a fix on the way, but will integrate your check for isatty() into it (instead of relying on catching an exception, which was our first attempt)

Great!

cyberw:
Superseded by #1631

kurtmckee:
Looks like the `fix-1629` branch can be deleted.

cyberw:
Thx!

PR #1631, opened 2020-11-16
Fix 100% cpu usage when running in docker/non-tty terminal

DennisKrone:
Raise and catch an exception when failing to get terminal attributes.

### DIFF
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -18,14 +18,18 @@ else:
     import tty
 
 
+class InitError(Exception):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


cyberw:
@heyman LTGY? If so I can merge and make a new release asap.

heyman:
Looks good as far as I can tell! I also tried running it in a Docker container without tty without being able to reproduce the issue.

heyman:
I took the liberty to fix test that had broke in this PR.

PR #1636, opened 2020-11-23
WIP: added example for influxdb listener

pjcalvo:
Hi,

This is a tiny PR to add a working locust example connected to influxDB for persistent metrics storage. I think many people will benefit from an example like this. I spent the weekend looking for a library that works with locust latests and I couldn't find it.

WIP: The mentioned repository is not ready and has also WIP, but first I want to make sure I can add this example section to the locust documents.

### DIFF
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -101,11 +101,11 @@ InfluxDB Listener
 =================
 
 As an alternative to the provided locust reports, you can use the :py:obj:`locust.events` to expose the locust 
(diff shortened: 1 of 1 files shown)


cyberw:
hi @pjcalvo ! Would you consider making a PR to add the listener (and example documentation) to https://github.com/SvenskaSpel/locust-plugins ?

We could probably make a more clear list of what is included in plugins and include it in locust core documentation as well, but having the actual documentation for stuff that isn't really in locust itself is a bit overkill.

PR #1643, opened 2020-12-02
URLs with non default (param=value) query strings doesn't work with FastHttpUser

heyman:
Currently it's not possible to make a request - using the FastHttpUser client - to URLs such as:

`http://example.com/get-file?path/to/file`

The fix needs to be implemented in the geventhttpclient lib, and I've made a PR for that here:
https://github.com/gwik/geventhttpclient/pull/134

This PR has a (currently failing) test for this. Once the above PR is merged, and a new geventhttpclient version released, we should bump the required geventhttpclient version, and merge this PR.

### DIFF
--- a/locust/test/test_client.py
+++ b/locust/test/test_client.py
@@ -196,3 +196,8 @@ class TestHttpSession(WebserverTestCase):
             pass
         self.assertEqual(1, self.environment.stats.total.num_requests)
         self.assertEqual(1, self.environment.stats.total.num_failures)
+
+    def test_querystring_different_format(self):
+        s = self.get_client()
+        response = s.get("/querystring?this/is/a/non-standard/querystring")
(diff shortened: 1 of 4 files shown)


cyberw:
I'm fighting that build error for macos on geventhttpclient..

cyberw:
looks nice now

cyberw:
(ignore that force push, I accidentally commited some stuff that was actually intended for master)

cyberw:
Hmm... I think the update to geventhttpclient had a breaking change. Let me confirm/fix that before we merge this..

cyberw:
See comments in https://github.com/gwik/geventhttpclient/pull/134

cyberw:
@heyman Should we fix the geventhttpclient to be backwards compatible, or bump its version (at least minor version) to indicate a breaking change?

 I would prefer not to keep this open for much longer...

PR #1645, opened 2020-12-03
Add example that manually adds stats entries

heyman:
This PR adds an example with a `manual_report()` decorator/context manager that can be used to easily measure the time of functions/code blocks and manually report them as stats entries to the Locust statistics.

I wrote up the `manual_report()` functionality as an example for a user in the Locust Slack, and I think it turned out pretty.

We could also consider actually making `manual_report` part of Locust's API. For example under`locust.contrib.report.manual_report`. Though I'm not sure it's useful enough to warrant including it with locust itself?

**Usage example as a context manager:**
```python
with manual_report("stats entry name"):
    # Run time of this block will be reported under a stats entry names "stats entry name" (of type "manual")
    # If an Exception is raised, it'll be reported as a failure
```

**Usage as a decorator:**
```python
@task
@manual_report
def my_task(self):
   # The run time of this task will be reported under a stats entry named "my_task" (of type "manual")
   # If an Exception is raised, it'll be reported as a failure
```

Even if we don't want to add it udner `locust.contrib.report.manual_report` I think it can be a useful example.

### DIFF
--- a/examples/manual_stats_reporting.py
+++ b/examples/manual_stats_reporting.py
@@ -1,5 +1,5 @@
 """
-Example of a manual_report() function that can be used either as a context manager 
+Example of a manual_report() function that can be used either as a context manager
 (with statement), or a decorator, to manually add entries to Locust's statistics.
 
 Usage as a context manager:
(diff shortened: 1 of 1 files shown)


cyberw:
LGTM! When you make PRs and I approve them, do you want to click the merge button yourself or should I do it? :)

heyman:
Feel free to merge them unless they are drafts or I've explicitly said that they aren't ready to be merged for some reason :).

PR #1655, opened 2020-12-10
Bump Echarts version to show charts legends

### LINKED ISSUE #1651: Show legends on charts

Sometimes when we need to share screenshots of the Requests per seconds we need to explain explicitly what does green and red colored lines mean, 

It will be handy to have legends on the charts so that is easy for the reader
### Additional context

I've tried passing legend option `this.chart.setOption` in [chart.js](http://github.com/talhajunaidd/locust/blob/afea8c0cfb842a8a28d3d18ab56326b4db9898b2/locust/static/chart.js#L28-L28)
but it did not show the legend.

aek:
The current Echarts library is [3.2.2](https://github.com/apache/incubator-echarts/releases/tag/3.2.2) which is 4y old
This PR update it to the latest version 5.0.0 to allow charts to have legends like the following:
image
image
image
image

This PR close #1651


cyberw:
thx!

Issue #613, opened 2017-06-26
Question: debugging in pycharm (or other arbitrary IDE)


paul-nelson-baker:
Hello, so quick question. I have locust installed in a venv on python 3 (locustio==0.8a2). I've been building and testing a locust python script, but I'm trying to traverse some complicated json responses. In order to run a locust script, you have to run `locust loadtest-file.py`

Is it possible to attach a debugger while running a locust script?

cgoldberg:
not really an answer, but a few comments:

> Is it possible to attach a debugger while running a locust script?
> I've been building and testing a locust python script, but 

a `locustfile` is a normal python file, so you can debug it directly without the rest of the framework.  It's easier to debugging a single `Locust` than a swarm. 

> I'm trying to traverse some complicated json responses.

add code so your `Locust` class gets created when `locustfile` is run as a script (not imported) ... you can usually do most develop against a single instance of `Locust`.

MattF-NSIDC:
Took me a minute to get there so I figured I'd share:

I create a subclass of HttpLocust called WebsiteUser. To execute a single instance from the CLI, I added:

```
if __name__ == '__main__':
    x = WebsiteUser()
    x.run()
```

If I have a `pdb` breakpoint in one of my tasks, this triggers the breakpoint.

GabLeRoux:
Here I managed to get Intellij Idea (Similar to PyCharm) to break while `locust` is running.

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/1264761/35758775-9fd6013a-0843-11e8-8c2e-d744428cfa2e.png">

nelsonjchen:
1. Turn on the gevent compatible option in the debugger
2. Run the locust executable as the script, (`which locust`)

This seems to work for me. Also, note that the remote debug suggestion above requires a paid version of PyCharm or its cousin of a plugin in IntelliJ. I think the steps I have above will work for PyCharm CE as well.

ad34:
is there a similar tip for visual studio code ?

dalegaspi:
just to add to what @nelsonjchen has stated, for IntelliJ set the "Script path" to what `which locust` returns from the command line then set the "Parameters" to `-f script_you_want_to_debug.py --host http://whatever`.  you should be able hit breakpoints where you placed them in  `script_you_want_to_debug.py`

r3dlin3:
@ad34 
> is there a similar tip for visual studio code ?

This is what I put in my `.vscode\launch.json`

```
{
[...]

PR #1656, opened 2020-12-10
Add feature locust launcher useful for run/debug on IDEs like PyCharm

aek:
Add feature locust launcher useful for run/debug on IDEs like PyCharm
Borrowed from the virtualenv installed package used for execute locust in the command-line

This will help to setup locust in an IDE like the following image
image

Using this script to run locust in the IDE will fix import errors related to the usage of others scripts like `main.py`, `__init__.py` and `__main__.py`

It's only an utility file to easy the develop of locust. An entry in the documentation will be needed if this PR become merged. I could include it as part of this PR. 
This PR is related to the issue #613

### DIFF
--- a/locust.py
+++ b/locust.py
@@ -1,0 +1,9 @@
+# -*- coding: utf-8 -*-
+import re
+import sys
+
+from locust.main import main
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+    sys.exit(main())


heyman:
How about the suggestion in this comment: https://github.com/locustio/locust/issues/613#issuecomment-465319635 ?

Doesn't that work? If so, I think the fix should be to just write proper documentation for that. I'm not keen on adding a `locust.py` file to the root.

aek:
@heyman I have used that way until I needed to start to develop and debug locust itself. It will not use your locust project files, it will be using the installed locust package instead so any modification to locust itself wouldn't be effective to be run without an issue of the command `python setup.py install` needing to packaging and reinstall locust for every modification in order to be able to see your changes

heyman:
What if you install locust using `pip install -e .` (https://pip.pypa.io/en/stable/reference/pip_install/#install-editable)?

aek:
I guess that could be an option but for some users but in my case I will prefer to have a way to directly run/debug locust without `pip install` it. 
Not need to merge it then but I found myself asking why locust doesn't have this launcher script that it's generated after `pip install` or `python setup.py install`?

heyman:
If you don't want to use pip you could use `python setup.py develop`

aek:
It's the same, and don't answer my previous question `why locust doesn't have this launcher script that it's generated after pip install or python setup.py install/develop?`. I'm curious about it. Maybe to have it generated as a command-line program to be executed? Just because of that or there is anything else?

heyman:
Sorry, I'm not sure I understand your question. Do you know of other python libraries that expose console script entrypoints that does this?

> I guess that could be an option but for some users but in my case I will prefer to have a way to directly run/debug locust without pip install it.

If you use `python setup.py develop` doesn't that provide you a way of directly running locust without pip installing it?

aek:
Yes, I'm aware that others do this. It's ok. I was asking about why locust doesn't provide the launcher script physically and have the console script entrypoint to call/use it. 
Nevermind. I confirm that the use of `pip install -e path/to/locust/source` works. I contains an implicit call to `python setup.py develop` so both ways are valid. 
I close the PR then

heyman:
I think we could maybe add an entrypoint file called `__main__.py` in the `locust/` package directory (not directly in the git root). That seems to be the way both `Flask` and `Django` does it, so it might at least be some informal standard. I'd be willing to merge that if you update the PR (ideally we would also add a test to `test.test_main.LocustProcessIntegrationTest` as well).

aek:
The `__main__.py` file already exist and cannot be used directly because there is no defined parent package. 
The exact error is:
```
ImportError: attempted relative import with no known parent package
```
That's the main reason to have an script importing locust package at the root level having the locust package imported and available to all the scripts that issue imports like:
```python
from .main import main
```
```python
from . import log
from .argument_parser import parse_locustfile_option, parse_options
from .env import Environment
from .log import setup_logging, greenlet_exception_logger
from . import stats
from .stats import print_error_report, print_percentile_stats, print_stats, stats_printer, stats_history
from .stats import StatsCSV, StatsCSVFileWriter
from .user import User
from .user.inspectuser import get_task_ratio_dict, print_task_ratio
from .util.timespan import parse_timespan
from .exception import AuthCredentialsError
from .shape import LoadTestShape
from .input_events import input_listener
from .html import get_html_report
```

Issue #1469, opened 2020-07-04
Mouse over in charts view should show number of users for all graphs


lhupfeldt:
Currently when looking at a specific graph you need to move your mouse to the users graph to see the number of users (and you need to place it at the same time point). It would be nice to extend the mouse over to always show number of users.
Possible other information should also be shown across all graphs. There is still room for more info without making it feel cluttered.

PR #1658, opened 2020-12-14
Feature chart tooltip custom values - show user count

aek:
This PR allow to show the number of users that are running when a metric was recorded like in the image bellow
image

This solve #1469


heyman:
Nice!

aek:
👍

PR #1659, opened 2020-12-14
add default values and validation for Web UI swarm form values

### LINKED ISSUE #237: ValueError when missing hatchrate
Not a big deal, but if you forget to specify hatch_rate in web interface you'll get ugly exception in logs, flask app could check for blank values in form and return some friendly error message to client

```
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: Traceback (most recent call last):
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: File "/usr/local/lib/python2.7/dist-packages/gevent/pywsgi.py", line 508, in handle_one_response
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: self.run_application()
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: File "/usr/local/lib/python2.7/dist-packages/gevent/pywsgi.py", line 494, in run_application
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: self.result = self.application(self.environ, self.start_response)
[2015-01-27 15:16:56,474] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1903, in __call__
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: return self.wsgi_app(environ, start_response)
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1887, in wsgi_app
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: response = self.make_response(self.handle_exception(e))
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1465, in handle_exception
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: reraise(exc_type, exc_value, tb)
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1884, in wsgi_app
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: response = self.full_dispatch_request()
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1539, in full_dispatch_request
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: rv = self.handle_user_exception(e)
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1443, in handle_user_exception
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: reraise(exc_type, exc_value, tb)
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1537, in full_dispatch_request
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: rv = self.dispatch_request()
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/home/pawel/py/flask/flask/app.py", line 1523, in dispatch_request
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: return self.view_functions[rule.endpoint](**req.view_args)
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: File "/usr/local/lib/python2.7/dist-packages/locust/web.py", line 50, in swarm
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: hatch_rate = float(request.form["hatch_rate"])
[2015-01-27 15:16:56,555] pawel-All-Series/ERROR/stderr: ValueError: could not convert string to float:
[2015-01-27 15:16:56,559] pawel-All-Series/ERROR/stderr: {'CONTENT_LENGTH': '27',
```

Overall locust is great though.

aek:
Minor fix for form default values to 1 and validations to avoid type conversion errors
Fixes #237

### DIFF
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -78,6 +78,8 @@
                         {% endif %}
                     </label>
                     <input type="text" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
+                    <br/>
+                    <div id="swarm_errors" align="center" style="color: #e68686;"></div>
                     <button type="submit">Start swarming</button>
                 </form>
                 <div style="clear:right;"></div>
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -67,9 +67,9 @@
                         <input type="hidden" name="spawn_rate" id="spawn_rate" value="0"/><br>
                     {% else %}
                         <label for="user_count">Number of total users to simulate</label>
-                        <input placeholder="Please enter a number" type="number" required="required" name="user_count" id="user_count" class="val" value="{{ num_users or "" }}"/><br>
+                        <input placeholder="1" type="number" required="required" name="user_count" id="user_count" class="val" value="{{ num_users or "" }}"/><br>
                         <label for="spawn_rate">Spawn rate <span style="color:#8a8a8a;">(users spawned/second)</span></label>
-                        <input placeholder="Please enter a number" type="number" required="required" name="spawn_rate" id="spawn_rate" class="val" value="{{ spawn_rate or "" }}"/><br>
+                        <input placeholder="1" type="number" required="required" name="spawn_rate" id="spawn_rate" class="val" value="{{ spawn_rate or "" }}"/><br>
                     {% endif %}
                     <label for="host">
                         Host <span style="color:#8a8a8a;">(e.g. http://www.example.com)</span>
@@ -77,7 +77,7 @@
                             <span style="color:#8a8a8a; font-size:12px;">(setting this will override the host on all User classes)</span></label>
                         {% endif %}
                     </label>
-                    <input placeholder="Please enter a Host" type="text" required="required" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
+                    <input type="text" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
                     <button type="submit">Start swarming</button>
                 </form>
                 <div style="clear:right;"></div>
@@ -98,9 +98,9 @@
                         <input type="text" name="spawn_rate" id="new_spawn_rate" class="val_disabled"  value="-" disabled="disabled" title="Disabled for tests using LoadTestShape class"/><br>
                     {% else %}
                         <label for="new_user_count">Number of users to simulate</label>
-                        <input placeholder="Please enter a number" type="number" required="required" name="user_count" id="new_user_count" class="val"  value="{{ num_users or "" }}"/><br>
+                        <input placeholder="1" type="number" required="required" name="user_count" id="new_user_count" class="val"  value="{{ num_users or "" }}"/><br>
                         <label for="spawn_rate">Spawn rate <span style="color:#8a8a8a;">(users to start per second (or stop))</span></label>
-                        <input placeholder="Please enter a number" type="number" required="required" name="spawn_rate" id="new_spawn_rate" class="val"  value="{{ spawn_rate or "" }}"/><br>
+                        <input placeholder="1" type="number" required="required" name="spawn_rate" id="new_spawn_rate" class="val"  value="{{ spawn_rate or "" }}"/><br>
                     {% endif %}
                     {% if is_shape %}
                         <button type="submit" disabled>Start swarming</button>
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -79,7 +79,7 @@
                     </label>
                     <input type="text" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
                     <br/>
-                    <div id="swarm_errors" align="center" style="color: #e68686;"></div>
+                    <div id="start_swarm_errors" align="center" style="color: #e68686;"></div>
                     <button type="submit">Start swarming</button>
                 </form>
                 <div style="clear:right;"></div>
@@ -103,6 +103,8 @@
                         <input placeholder="1" type="number" required="required" name="user_count" id="new_user_count" class="val"  value="{{ num_users or "" }}"/><br>
                         <label for="spawn_rate">Spawn rate <span style="color:#8a8a8a;">(users to start per second (or stop))</span></label>
                         <input placeholder="1" type="number" required="required" name="spawn_rate" id="new_spawn_rate" class="val"  value="{{ spawn_rate or "" }}"/><br>
+                        <br/>
+                        <div id="edit_swarm_errors" align="center" style="color: #e68686;"></div>
                     {% endif %}
                     {% if is_shape %}
                         <button type="submit" disabled>Start swarming</button>
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -67,9 +67,9 @@
                         <input type="hidden" name="spawn_rate" id="spawn_rate" value="0"/><br>
                     {% else %}
                         <label for="user_count">Number of total users to simulate</label>
-                        <input type="text" name="user_count" id="user_count" class="val" value="{{ num_users or "" }}"/><br>
+                        <input placeholder="Please enter a number" type="number" required="required" name="user_count" id="user_count" class="val" value="{{ num_users or "" }}"/><br>
                         <label for="spawn_rate">Spawn rate <span style="color:#8a8a8a;">(users spawned/second)</span></label>
-                        <input type="text" name="spawn_rate" id="spawn_rate" class="val" value="{{ spawn_rate or "" }}"/><br>
+                        <input placeholder="Please enter a number" type="number" required="required" name="spawn_rate" id="spawn_rate" class="val" value="{{ spawn_rate or "" }}"/><br>
                     {% endif %}
                     <label for="host">
                         Host <span style="color:#8a8a8a;">(e.g. http://www.example.com)</span>
@@ -98,9 +98,9 @@
                         <input type="text" name="spawn_rate" id="new_spawn_rate" class="val_disabled"  value="-" disabled="disabled" title="Disabled for tests using LoadTestShape class"/><br>
                     {% else %}
                         <label for="new_user_count">Number of users to simulate</label>
-                        <input type="text" name="user_count" id="new_user_count" class="val"  value="{{ num_users or "" }}"/><br>
+                        <input placeholder="Please enter a number" type="number" required="required" name="user_count" id="new_user_count" class="val"  value="{{ num_users or "" }}"/><br>
                         <label for="spawn_rate">Spawn rate <span style="color:#8a8a8a;">(users to start per second (or stop))</span></label>
-                        <input type="text" name="spawn_rate" id="new_spawn_rate" class="val"  value="{{ spawn_rate or "" }}"/><br>
+                        <input placeholder="Please enter a number" type="number" required="required" name="spawn_rate" id="new_spawn_rate" class="val"  value="{{ spawn_rate or "" }}"/><br>
                     {% endif %}
                     {% if is_shape %}
                         <button type="submit" disabled>Start swarming</button>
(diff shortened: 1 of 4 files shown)


heyman:
If we are to allow starting a test with no values entered I think we should set placeholder values for the input boxes. But perhaps it would be better to do a client side check and complain (e.g. make the input boxes redish) if the values are missing?

aek:
@heyman I have add client side validations and placeholder, now it shows like this image
image

locust/templates/index.html
```diff
@@ -67,17 +67,17 @@ <h2>Start new load test</h2>
...
+                        <input placeholder="Please enter a number" type="number" required="required" name="spawn_rate" id="spawn_rate" class="val" value="{{ spawn_rate or "" }}"/><br>
                     {% endif %}
                     <label for="host">
                         Host <span style="color:#8a8a8a;">(e.g. http://www.example.com)</span>
                         {% if override_host_warning %}
                             <span style="color:#8a8a8a; font-size:12px;">(setting this will override the host on all User classes)</span></label>
                         {% endif %}
                     </label>
-                    <input type="text" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
+                    <input placeholder="Please enter a Host" type="text" required="required" name="host" id="host" class="val" autocapitalize="off" autocorrect="off" value="{{ host  or "" }}"/><br>
```
heyman, line 80: The host shouldn't be mandatory.

locust/web.py
```diff
@@ -134,8 +134,11 @@ def index():
         @self.auth_required_if_enabled
         def swarm():
             assert request.method == "POST"
-            user_count = int(request.form["user_count"])
-            spawn_rate = float(request.form["spawn_rate"])
+            user_count = spawn_rate = 1
+            if request.form.get("user_count"):
+                user_count = int(request.form["user_count"])
+            if request.form.get("spawn_rate"):
+                spawn_rate = float(request.form["spawn_rate"])
```
heyman, line 141: Instead of defaulting `user_count` and `spawn_rate` to `1` we should just return something like this (and ideally set the HTTP response code to `400`):
```javascript
{"success": False, "message": "You must specify user_count and spawn_rate"}
```

It would also be good to have a test for this (in `test_web.py`)

heyman (changes requested):
Thanks! I think making the values mandatory by adding the required attributes is great.

However the placeholder text is a little redundant and should be removed. (What I  meant with the placeholders in my previous comment was that *if* we were to allow submitting the form without having entered values, we should set the placeholders to `1` (or whatever default that we chose), but I realize now that that might not have been obvious, sorry!)

locust/templates/index.html
aek: If the host is not mandatory means that all the test client (HTTP) requests need to be using absolute URLs, not relative ones. I'm ok with that but there are some checks inplace to enforce the value for host. 
Here is the log that throws FastHttpUser and HttpUser:

https://github.com/locustio/locust/blob/db2f7649d32191beb359d0a6f9c86342e63d545e/locust/contrib/fasthttp.py#L310-L313
```log
  File "/Users/axel/git-workspace/locust/locust/contrib/fasthttp.py", line 312, in __init__
    "You must specify the base host. Either in the host attribute in the User class, or on the command line using the --host option."
locust.exception.LocustError: You must specify the base host. Either in the host attribute in the User class, or on the command line using the --host option.
2020-12-16T15:29:17Z <Greenlet at 0x7fe7effda290: <lambda>> failed with LocustError
```
https://github.com/locustio/locust/blob/db2f7649d32191beb359d0a6f9c86342e63d545e/locust/user/users.py#L212-L215
```log
  File "/Users/axel/git-workspace/locust/locust/user/users.py", line 214, in __init__
    "You must specify the base host. Either in the host attribute in the User class, or on the command line using the --host option."
locust.exception.LocustError: You must specify the base host. Either in the host attribute in the User class, or on the command line using the --host option.
```

aek: I think that host it's not exactly needed in order to make client requests working. It's better to check if the URL is relative and there is no base_url(host value) in the session to build the full URL
Let me know and I could make the adjustments and remove the host validations Locust errors

locust/web.py
aek: Have commit the change of placeholders. Looks like this now
image

aek:
I have commited the changes of the placeholders
image

locust/templates/index.html
heyman: People might be using custom `User` classes that doesn't use the HTTP client. For `HttpUser` and `FastHttpUser` those checks makes sense, but the Web UI will be the same even if you're using a custom `User` class that doesn't use the host.

aek: That's not completely True, host it's just needed if the requests doesn't use an absolute URL. Otherwise the host value wouldn't been used at all. I think it should concern of the Http User Test Class to provide the value of the host for base URL or not. Also it could be the case that 1 test class use a host value and others Test Classes use a different host value. I don't see the point of asking for a host value in the form before swarm begin when maybe it wasn't been used by any User Test Class. 
I think it's more consistent to have a base host value in the argparse options that all the Test Classes could pick if they don't provide one. Because in the form you couldn't pick which User Test Classes will get the value you supply for their host field or provide a value for each of the User Test Classes that will be running when the swarm begins

heyman: Yes, one could argue for removing the host input from the UI altogether. However, people *are* using that functionality in their Locust work flows, and removing it for those user is in my opinion much worse than the negative effect of showing an empty input for people not using it.

In any case, I think it's better to leave that out of the scope of this PR.

locust/web.py
aek: @heyman I have added this response and removed the defaults values. Tests for this validations have been add to test_web.py
Also I have reorder the JS form submit code in order to check for response success in order to show the Locust Interface when tests begin to swarm, otherwise show the error message from the response data like:
image

The error message in the image is shown because I have locally remove the required attributes from the form fields to test and show the error messages, but I keep it in the PR
Also I have added some UI improvements to prevent some clunky behavior of the dialogs that lead to this kind of weird visualization
image
That happens when you edit the test and without closing the edit form click on the stop button and then click on the new tests link that shows the start swarm form. So now when a dialog is about to open it close the other dialog if it's open

locust/web.py
```diff
@@ -134,6 +134,9 @@ def index():
         @self.auth_required_if_enabled
         def swarm():
             assert request.method == "POST"
+            if not request.form.get("user_count") or not request.form.get("spawn_rate"):
+                return jsonify({"success": False, "message": "You must specify user_count and spawn_rate"})
```
stanislawskwark, line 138: this should not be required if shape class is used, I would move it to line 152

lhupfeldt:
Does a default of **1** really make sense? Maybe it would be better to leave it empty to avoid running with 1 user which is probably not what you normally want to do.

aek:
Conflict resolved, 
Updated with master
@lhupfeldt The default value of 1 was removed several commits ago, what you see it's just a placeholder of the inputs

cyberw:
@heyman is this ready to merge, or are there still things you want changed?

cyberw:
Hmm... I realize this may just be conflicting opinions but I really dont like there being a number in the field which isnt "real".

This makes zero sense to me:

image

I would very much prefer the form being prefilled with 1:s (real), but once you select the field the value should become selected so that typing overwrites it.

I run tests with 1 user all the time (particularly while debugging a script, is a significant portion of launches, perhaps even a majority) and find it a completely sensible default. In almost all cases, accidentally launching a test with just a single user has no real harm anyway.

lhupfeldt:
I would prefer that there was a help text instead of a number. If you want it to be pre-filled with an actual/real number you can specify it on the commandline. I still think having any built-in default (like 1) is a bad idea. We run the UI in an automated fashion and I prefer that at misconfiguration result in an error rather than running with 1 by mistake. E.g. a successful run (stats within defined limits) with 1 instead of 1000 would cause a false approval to be registered.

cyberw:
If you are talking about the endpoint itself I agree: the API should not use default values. 

But the UI (the web form) could/should be pre-filled with reasonable defaults if it makes it more user friendly. 

Also, it is reasonable to assume the number of people starting runs using the web UI outnumber the people using the UI for automation by something like 100 to 1, so even if there was some conflict, it should be clear which one is the most prioritized.

aek:
Let's get a consensus about what needs to be done here and I push the changes, because this PR also solve several UI bugs that cause Locust UI to see it broken

lhupfeldt:
Sorry for my slow response.
I agree that it should be clear which one is prioritized, but I don't think we agree on which one that is :)
Saving 4 keystrokes by having a default value of '1' does not make sense to me. Preventing something which does not perform as required from going to production should be much more important. If you prefer a default of 1 when invoking manually, then you can just make a wrapper script and name i 'l', then you don't even have to type locust. 
Is there support for personal config files? That would make it even simpler to have whatever defaults you prefer.
I think we agree that the 'dummy/help text' value of 1 is not a good idea. A single greyed out character is not easily distinguishable from a proper value. A good help text would be better.
Since this PR contains other fixes, we should probably not delay it much longer, so whatever is chosen will be fine with me.

Regarding the use of UI for automation, that is caused by the lack of a proper rest API. There is no simple way to control locust running in a container. But I see no reason that the UI should be disabled for automatic runs - why prevent people from seeing the progress? The enabling/disabling of the UI should really not be coupled to whether Locust is run automatically.

cyberw:
> Since this PR contains other fixes, we should probably not delay it much longer, so whatever is chosen will be fine with me.

Agreed. Lets just make it empty for now. In the future I'd like it to default to 1, because I think UI automation is not a use case we should care about (maybe we need to make some adjustements to make that it completely unnecessary though), and for actual users 4 keystrokes (or 2 clicks and 2 keystrokes) just to start the run __is__ a real issue when debugging your scripts and you do 10-20 launches in a row. And except for automation there is little risk of mistakes or confusion (unless you completely fail to read the form).

> Regarding the use of UI for automation, that is caused by the lack of a proper rest API. There is no simple way to control locust running in a container. But I see no reason that the UI should be disabled for automatic runs - why prevent people from seeing the progress? The enabling/disabling of the UI should really not be coupled to whether Locust is run automatically.

Agreed, the connection between "auto-start" and headless mode has never made sense to me either. Lets fix that some time. Are you using selenium to start the test? You could just POST to /swarm
with form variables for user_count: 1, spawn_rate: 1, host: https://whatever (of course, this is 2021 and it should be a REST endpoint :), but does that make it hard to do without a simulated browser?

lhupfeldt:
I'm just using POST. Currently it is as simple as using a REST API. So if the default is simply a UI thing, maybe I will have nothing to worry about :)

cyberw:
> I'm just using POST. Currently it is as simple as using a REST API. So if the default is simply a UI thing, maybe I will have nothing to worry about :)

Thats exactly what I meant when I said

> If you are talking about the HTTP endpoint itself I agree: the API should not use default values.
> But the UI (the web form) could/should be pre-filled with reasonable defaults if it makes it more user friendly.

:)

lhupfeldt:
I still don't think have a default in the UI is a good idea though. 1 is really an edge case, if you want to make a trial run, you are probably better off with 2 :). But it is not a big deal for me, better to get it merged than having such long lived PRs.
Maybe later there will be a user config file where people can set your own defaults if they like.

cyberw:
I think the defaults could/should be controlled by the regular settings (-u / -r)

cyberw:
While we are at it, I think we should change the wording to ”total _concurrent_ users”, because people often misunderstand that.

cyberw:
@aek If you can change it so that the defaults are set by the regular settings (-u / -r) I will merge this.

PR #1668, opened 2020-12-26
Show legends on charts #1651

talhajunaidd:
It will be handy to have legends on the charts so that is easy for the reader to read the chart. #1651 

Here is a desktop view

image

Downloaded image:

total_requests_per_second_1609013311


talhajunaidd:
Closing it, It seems like someone already did that. 
Thank you!

PR #1671, opened 2021-01-04
/swarm adjusted for tests with shape class

### LINKED ISSUE #1670: /swarm web endpoint should not require user_count and spawn_rate when shape_class is used
/swarm endpoint requires two int values for user_count and spawn_rate, which makes sense only for non-shaped locustfiles

in web.py get user_count and spawn_rate could be used only if necessary

stanislawskwark:
for web /swarm endpoint moved extracting user_count and spawn_rate after checking if shape_class is used. if shape class is used these values are ignored
Fixes #1670

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -134,8 +134,6 @@ class WebUI:
         @self.auth_required_if_enabled
         def swarm():
             assert request.method == "POST"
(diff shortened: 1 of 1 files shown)


cyberw:
ty

Issue #1673, opened 2021-01-08
currently locust is supporting for stas, failures, stas history in csv format while running without web


nags28:
### Is your feature request related to a problem? Please describe.

If locust supports to get exception csv file in without web UI it will be great .
and even if it supports report file which is available in web UI It might help for detail investigation.

PR #1674, opened 2021-01-11
Feature stats exceptions csv

aek:
Solve #1673

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -827,6 +827,13 @@ class StatsCSV:
             "Occurrences",
         ]
 
+        self.exceptions_columns = [
+            "Count",
+            "Message",
+            "Traceback",
+            "Nodes",
+        ]
+
     def _percentile_fields(self, stats_entry):
         return (
             [int(stats_entry.get_response_time_percentile(x) or 0) for x in self.percentiles_to_report]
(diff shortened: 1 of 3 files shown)


cyberw:
Lgtm. I dont use the built in csv writing though, so it would be great if someone else can weigh in as well.

lhupfeldt:
Looks good.
I think that as a followup change it should be consideret to start using lists and loops to reduce the number of places needing change if new files are added.

aek:
@lhupfeldt I don’t understand what you mean 
Could you elaborate please?

lhupfeldt:
I mean your PR looks fine, but after it is merged, maybe there should another PR to use be a list of filesnames such that e.g.:

`    def setUp(self):
        super().setUp()
        self.remove_file_if_exists(self.STATS_FILENAME)
        self.remove_file_if_exists(self.STATS_HISTORY_FILENAME)
        self.remove_file_if_exists(self.STATS_FAILURES_FILENAME)
        self.remove_file_if_exists(self.STATS_EXCEPTIONS_FILENAME)
`
Could be written as:

`    def setUp(self):
        super().setUp()
        for fn in  self.stats_files:
            self.remove_file_if_exists(fn)
`

aek:
I see it now, I agree with that. I could include it in this one if it will be considered to be merged too

lhupfeldt:
I would personally prefer is as a separate PR. Makes it easier to see what this PR is about, but i'm not a maintainer.

aek:
Let's wait to see this one merged then
Thanks

PR #1675, opened 2021-01-15
Add scenario concept

mathcantin:
Bt default, in locust each User Class is loaded. When we have different applications or use cases we need to have many locustfiles and we have to redeploy the application each time we want to change of locustfile.

For that I add the concept of scenario in Locust. Instead of load every User instance, we have the ability to add a scenario and specify which User instance to run for each scenario.

The UI is modified only when they have scenario instance : 
image

This change is full compatible for each client without scenario.


cyberw:
This is really cool. 

I think it would be even more useful if you could specify user weights in the scenario as well, like this:
`users = {User1: 3, User2: 5}`

Can you also add a unit test that runs a scenario using the Web UI?

mathcantin:
@cyberw I add UTs on WebUI.

Yes, I'm agree with you and it be also useful to add shape on scenario object. I think that is a good first iteration.

The weight could by fake like that:

```
        class MyUser(User):
            @task(1)
            def my_task(self):
                pass

        class MyUserForScenario1(MyUser):
            weight = 3

        class MyUserForScenario2(MyUser):
            weight = 1

        class Scenario1(Scenario):
            name = "My Scenario 1"
            users = [MyUserForScenario1]

        class Scenario2(Scenario):
            name = "My Scenario 2"
            users = [MyUserForScenario2]
```

cyberw:
Lgtm. I will merge soon, unless someone has any objections.

cyberw:
I guess the one objection one might have is that this does the same thing as the `@tags` functionality.

oh, and one more thing: we need to add command line support as well.

mathcantin:
I think the command line support isn't full useful, because there are already possible to have multiple locustfile with different scenari. But if you want, I could add them to the PR.

cyberw:
hmm... yea, I guess this is mainly useful for UI runs. 

@heyman , do you have an opinion on whether to merge this? I think we talked about a while back but decided on adding the tags feature instead of this. I see the need for it, but I'm kind of +/- 0.

cyberw:
@mathcantin , can you merge the latest master into your branch (the automated build stuff has been updated, and it would be nice to see that it works on your branch)

heyman:
I wonder if we really need to introduce a whole new concept (`Scenario`) for this. Wouldn't we achieve the same thing if we instead implemented the ability to specify which User classes to run, from the Web UI? We could even add the ability to override the User classes' `weight` from the UI.

mathcantin:
I merge the master into my branch. I still think the scenario object is more flexible in the future than a mix between User, Weigth, Shape on UI. But it's your call.

cyberw:
I'm +/- 0 (so I wont merge right now, but I wont object if somebody else does).

I see the point and the code seems well written. I think @heyman 's suggestion would be more flexible, but if nobody builds that (within reasonable time) then that point is moot :)

aek:
I indeed see the point for the scenarios usages but I don’t like either to continue modifying the current codebase with new python class based structures
I will vote/go for a redesign of the current codebase to make it more based on data rather than custom python classes construction to define more complex designs of load tests concurrent scenarios 
Maybe it sounds crazy and don’t get me wrong... I find locust very interesting and awesome project but I think it needs a redesign to make it more lean and robust than today. Also with this redesign I will go also for a reactive ui using maybe reactjs or vuejs that reflect the status of the data tests options definitions

cyberw:
I kind of agree with @aek

Now that we have switched to ConfigArgparse, so we already support config files, perhaps the best solution would be to expand on that?

Scenarios, Load Shapes and possibly even User weights could fit in there (havent tried it or anything, but imo it would be worth exploring)

heyman:
I see the use of this PR, but I'd rather not introduce a whole new concept just for specifying a group of User classes to use in a test. 

What I suggested in my [comment above](https://github.com/locustio/locust/pull/1675#issuecomment-766887687) should be possible to implement with less or equal amount of code, and should also have the added benefit of making the Web UI more powerful in general, allowing end-users to run different set of User classes "on the fly" (without having to create a new Scenario).

Regarding @aek's comment, that sounds a bit more like separate issue(s) to me. There's definitely parts of Locust's code base that could benefit from a rewrite (rewriting the Web UI in Vue.js is one such thing that I've thought about). However, I also think one should be wary of second system syndrome :).

PR #1678, opened 2021-01-19
Feature chart sync

### LINKED ISSUE #1677: Report charts plot data points after the test has ended
imagen
imagen

### Describe the bug
After stopping a test on the UI, when I click on "Download report" the plots show a date range between _test-start_ and _now_, instead of _test-start_ and _test-end_. 

### Expected behavior
All the plots should only contain data between _test-start_ and _test-end_ dates I guess.

### Actual behavior
All three plots in the report continue plotting either 0 (Response Times and Number of Users) or RPS (Total Requests per Second) after the test has ended up until the moment you open or refresh the Download Report page.

### Steps to reproduce

- Start the UI with any test file (the Quick Start one for example: https://docs.locust.io/en/stable/quickstart.html): `locust -f example.py`
- Parameters don't matter as far as I know, just run the test
- Let it run for a few seconds
- Stop it
- Wait for 10 or 20 seconds so it is easier to see the bug
- Click on **Download Data -> Download Report**
- Notice the flatlines on the plots and the date ranges, extending way after the test was stopped. If you refresh this page (the report) the plots just keep growing.
- Interestingly enough, the start and end dates at the top of the report are correct

### Environment

- OS: Ubuntu 20.04
- Browser: Firefox 84.0.2 and Chromium 87.0.4280.141
- Python version: 3.8.5
- Locust version: 1.4.1
- Locust command line that you ran: `locust -f example.py`
- Locust file contents (anonymized if necessary): 

```
import time

    @task
    def hello_world(self):
        self.client.get("/hello")
        self.client.get("/world")

    @task(3)
    def view_item(self):
        for item_id in range(10):
            self.client.get(f"/item?id={item_id}", name="/item")
            time.sleep(1)

    def on_start(self):
        self.client.post("/login", json={"username":"foo", "password":"bar"})
```

BTW I think this is a great tool, I have just discovered it and I am loving it! :heart:

aek:
This PR improves the charts of the index and the report page with the following changes:
- Prevent to continue reporting stats to the history of stats when the runner is stopped
- Usage of shared template to build stats_history data to reload to charts data in the index.html and report.html templates
- Fix Report Charts tooltips user count values

Fixes #1677

### DIFF
--- a/locust/static/chart.js
+++ b/locust/static/chart.js
@@ -47,7 +47,9 @@
                                 var param = params[i];
                                 str += '<br><span style="color:' + param.color + ';">' + param.seriesName + ': ' + param.data.value + '</span>';
                             }
-                            str += '<br><span style="color:#a3b3ac;">Users: ' + param.data.users + '</span>';
+                            if(param.data.users != undefined){
+                                str += '<br><span style="color:#a3b3ac;">Users: ' + param.data.users + '</span>';
+                            }
                             return str;
                         } else {
                             return "No data";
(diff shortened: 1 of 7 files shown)


cyberw:
Awesome!

aek:
@cyberw Sorry, I need to add a new way to pass the data from python to the js without need to generate the instructions. I will make another PR

PR #1679, opened 2021-01-19
Improve stats data sharing from python to js

aek:
related to #1678

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -4,13 +4,7 @@ import pathlib
 import datetime
 from itertools import chain
 from .stats import sort_stats
-
-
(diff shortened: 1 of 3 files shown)


cyberw:
I’m going to assume you know what you are doing :)

wonder why the builds are not running...

aek:
@cyberw Let me add the jinja2.ext.do extension to the tests to fix that build issue. Also I'm trying to fix the chart persistence integration with a running test, because it's getting cleaned by new metrics

cyberw:
unfortunately travis itself is borked at the moment. I have reached out to @heyman to see if he can fix it.

PR #1680, opened 2021-01-19
Fix issues with render_template

aek:
Restore the custom html.py render_template function to fix tests


aek:
@cyberw this should fix the Travis too because I have remove it in order to use the Flask version but there are some tests that runs without Flask

cyberw:
Cool! Good night! (In my time zone at least :)

aek:
Good night to you too,
It's 5pm here, so my day continue

PR #1686, opened 2021-01-22
Use Github Actions for CI

heyman:
This PR also fixes a couple of broken tests.

Releases to PyPI are *not* handled yet, but that can be a different PR I think.


cyberw:
Lgtm!

Issue #1690, opened 2021-01-27
Charts not working well in version 1.4.2


mthaak:
Hi Locust developers,

I discovered your tool today and it looks like exactly what I need. However, I found that in version 1.4.2 (release 21 hours ago) the charts are not working well.

### Describe the bug
The _Reponse Times (ms)_ and _Number of Users_ charts don't update at all. The _Total Requests per Second_ chart updates but the tooltip values are `undefined`. This only happens in Locust 1.4.2. In Locust 1.4.1 the charts work perfectly.

### Actual behavior
Locust 1.4.2:
![image](https://user-images.githubusercontent.com/6129322/105986912-d1beb380-609d-11eb-9b01-10f88cfa9607.png)

This is what I see in my browser's dev console:
![image](https://user-images.githubusercontent.com/6129322/105987011-f31f9f80-609d-11eb-83bc-dd02a983583c.png)
(many many times the same error)

### Expected behavior
Exact same set up but then using Locust 1.4.1:
![image](https://user-images.githubusercontent.com/6129322/105985266-873c3780-609b-11eb-8fbd-10e6dc0d32ac.png)

### Steps to reproduce
_See environment below_

### Environment

- OS: Windows 10
- Python version: 3.8.5
- Locust version: 1.4.2
- Locust command line that you ran: `locust --host=...`
- Locust file contents (anonymized if necessary):
```
from locust import HttpUser, task, between

class QuickstartUser(HttpUser):
    wait_time = between(1, 2.5)

    @task
    def get_1(self):
        self.client.get(...)

    @task
    def get_2(self):
        self.client.get(...)

    @task
    def get_3(self):
        self.client.get(...)

    def on_start(self):
        self.client.headers['Authorization'] = f'Bearer {AUTH_TOKEN}'
```

50 total users spawned at 10/s

cyberw:
@aek Can you take a look? I think your changes are the only ones relating to graphs in 1.4.2.

aek:
@cyberw Yeah, let me take a look about it and I will get back to this issue
@mthaak Thanks for reporting this one and glad you like the project too

heyman:
This should now be fixed in 1.4.3, which should be up on PyPI in a couple of minutes once the build script has finished. Thanks for reporting @mthaak, and thanks for fixing @aek!

PR #1691, opened 2021-01-27
fix stats values for chart tooltips

aek:
This PR solves the regression issue #1690 
Combine the dynamic stats values with the new persistence chart stats


cyberw:
Thanks! I will try it out and make a release tomorrow if it works out.

heyman:
I've tested the changes, and can confirm that it fixes the issue!

heyman:
New release tag pushed, so it should be up on PyPI in a few minutes.

PR #1705, opened 2021-02-15
fix type hinting on Events

mrijken:
When I add an event listener, my IDE complains about the type of the decorated function. This will fix that by making sure that the annotation of the attributes of Events are an instance of EventHook (and not the class EventHook).

### DIFF
--- a/locust/event.py
+++ b/locust/event.py
@@ -48,7 +48,7 @@ class EventHook:
 
 
 class Events:
-    request_success = EventHook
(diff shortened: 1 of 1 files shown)


cyberw:
Great stuff. I tried fixing this a while back, but it impacted documentation generation so I backed out (this seems to look fine though). There are some other places too that need this kind of fix (in the User class for instance) but we can do it some other time...

PR #1736, opened 2021-03-23
fix doc error in the running-locust-docker.rst file

CaixiangFan:
The section "Use docker image as a base image" is wrong by directly adding RUN pip3 install XX after the base image. It will cause module import error in docker run because packages imported before installed.
The correct way is to add the installation command with other python packages installation in the locust Dockerfile or any place before the ENTRYPOINT.

    FROM python:3.8
    COPY . /build
    RUN cd /build && pip install . && pip install some-python-package && rm -rf /build
    ...

### DIFF
--- a/docs/running-locust-docker.rst
+++ b/docs/running-locust-docker.rst
@@ -28,7 +28,7 @@ Use docker image as a base image
 ================================
 
 It's very common to have test scripts that rely on third party python packages. In those cases you can add the package installation command in the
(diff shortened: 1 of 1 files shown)


cyberw:
Thanks for the PR, but it looks more like a workaround than a proper fix. It used to work the way it is described  in the docs, and we should preferably fix that (dont have time to look into it myself though)

CaixiangFan:
@cyberw Thanks. I will close this PR.

Issue #1718, opened 2021-03-01
LoadTestShape run_time broken when using test_start and test_stop decorators


laakejs:
### Describe the bug

When using the LoadTestShape class in combination with @events.test_start.add_listener and @events.test_stop.add_listener, the run_time does not work as expected. The tick function is called once BEFORE the  @events.test_start.add_listener function runs. Next, the  content of the @events.test_start.add_listener function is executed which in our case takes quite a while as we are faking data. When it is finished the actual CustomShape loadtest starts, however the time was running in the background all the time, so the get_run_time() function returns pretty high numbers. If there are stages that would still run at that time locust now seems to start them and we get some load on our Api. But even more confusing the @events.test_start.add_listener seems to run again, this time in parallel wit the actual loadtest.

### Expected behavior

If there is a class extending the  LoadTestShape class, it should be instantiated (and the time should start to tick) AFTER the @events.test_start.add_listener function finished. Otherwise it is dependent on the runtime of preparation actions. Also it should not cause @events.test_start.add_listener to run again.

### Actual behavior

The tick function is called once BEFORE the @events.test_start.add_listener function runs and the time keeps running in the background.  Additionally it seems to start events.test_start.add_listener again in parallel to the loadtest.
![Locust_bug_proof_long](https://user-images.githubusercontent.com/33700709/109479634-df8b9e00-7a7a-11eb-8315-bbb6bb4f27a5.JPG)

### Steps to reproduce
 

import logging
import time 

from locust import HttpUser, LoadTestShape, between, events, task
from utils import get_api_credentials, get_host

@events.test_start.add_listener
def on_test_start(**kwargs):
    logging.info("on_test_start started running")
    time.sleep(317)
    logging.info("on_test_start finished")

@events.test_stop.add_listener
def on_test_stop(**kwargs):
    logging.info("on_test_stop started running")
    time.sleep(317)
    logging.info("on_test_stop finished")

class Api1(HttpUser):
    wait_time = between(1, 10)
    host = get_host()
    credentials = get_api_credentials()

    @task
    def task1(self):
        self.client.get("/foo", auth=self.credentials, timeout=0.2)

    @task
    def task2(self):
        self.client.get("/bar", auth=self.credentials, timeout=0.2)

    @task
    def task3(self):
        self.client.get("/misc", auth=self.credentials, timeout=0.2)

class StagesShape(LoadTestShape):

    stages = [
        {"duration": 120, "users": 100, "spawn_rate": 1},
[...]

PR #1738, opened 2021-03-23
Call shape_class.reset_time() after test_start event so that tick time is correct

max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/1718


cyberw:
I'm looking at why the build is failing, most likely unrelated to your change...

PR #1746, opened 2021-04-09
Bump minimum versions of various dependencies

cyberw:
Bump versions of requests, msgpack, pyzmq and configargparse. Using pretty much mid 2020-versions of everything.

The purpose is to reduce the time pip takes to backtrack to find appropriate versions in packages that depend on locust, and to ensure people dont run into some rare issue because they use a really old version of a dependency.


cyberw:
Huh, it didnt seem to help very much with resolve times...

PR #1749, opened 2021-04-21
[SGP-16076] update ReadMe to avoid accidentally messing up locustio/locust

ecedmondson:
This is the first PR to the specific `bevy/locust` fork. It isn't any code changes, but, instead, a ReadMe change suggesting patterns for handling the fork.

The reason for this is twofold:
1) We hope to move `bevy/load-testing` to use this fork instead of `locustio/locust` in order to add in Bevy specific server-side features
2) We do not want to accidentally mess up an open-source project. That would be bad. 

This ReadMe is fairly straightforward. I am haappy to change anything that doesn't make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,74 @@
-# Locust
+# Bevy's Locust Fork
+
+## Manging the Bevy Locust Fork
+
+Bevy maintains a custom fork of locust to allow us to have custom features server-side that are not in the base locustio/locust package.
+Out of consideration for the maintainers of this open-source tool, it is recommended that if you intend to contribute to the Bevy 
+Locust fork that you follow the git recommended patterns for working with upstream remotes. An overview of those upstream patterns 
+can be found [here](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes) and a how-to guide for managing remotes can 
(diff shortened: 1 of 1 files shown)


cyberw:
Dont worry, you can't push to pypi locust project without credentials, so it can never happen accidentally :)

What kind of features are you thinking about adding? For most things you shouldnt need to fork locust.

ecedmondson:
Hey Lars, 

This was actually a test to see if my ReadMe had any holes in it! Since it showed up on your end, it turns out that my instructions did have some holes in it so I decided to make it clearer never to push to locustio and to only open PRs to our fork.

Due to roles and permissions with user accounts in our application, we have some pretty complex scenarios where we need to be able to more precisely map the number of each type of user class. I’ve looked into the locust shaping in order to try and get this. It didn’t end up solving the exact problem I had, but it did solve another problem so it was a nice feature!

In addition to this specification, I wrote a little bit of vanilla JS and some tweaks to the Flask code that allow someone to pick which UserClass they wish to use from a dropdown in the locust main page dialog. This feature was intended to try and make load testing more accessible to our developers who don’t spend the majority of their time writing test code. This is mostly only a need for folks who are running using the WebUI, which is a portion of the folks who I intended for the feature. A lot of the time I will run tests in headless mode.

Hope that helps!

Issue #1676, opened 2021-01-18
GRPC compatibility : Locust load test throws greenlet.GreenletExit exception on reaching test time limit


pranavgupta1234:
### Describe the bug

Locust load test throws greenlet.GreenletExit exception on reaching test time limit while testing GRPC based services. To monkey patch GRPC with gevent I am using grpc's:
```
import grpc.experimental.gevent as grpc_gevent
grpc_gevent.init_gevent()
```
but no luck.

### Expected behavior

Should gracefully stop all greenlets and generate the necessary outputs.

### Actual behavior

Test keeps on generating load, does not stop, has to be stopped via CTRL+C Keyboard Interrupt.
Thrown exception:
```
[2021-01-18 13:40:47,715] C02CD0BNMD6N/INFO/locust.main: Run time limit set to 30 seconds
[2021-01-18 13:40:47,715] C02CD0BNMD6N/INFO/locust.main: Starting Locust 1.4.1
[2021-01-18 13:40:47,716] C02CD0BNMD6N/DEBUG/locust.runners: Updating state to 'spawning', old state was 'ready'
[2021-01-18 13:40:47,716] C02CD0BNMD6N/INFO/locust.runners: Spawning 5 users at the rate 3 users/s (0 users already running)...

 a bunch of related stats generated .......

[2021-01-18 13:41:17,717] C02CD0BNMD6N/INFO/locust.main: Time limit reached. Stopping Locust.
[2021-01-18 13:41:17,717] C02CD0BNMD6N/DEBUG/locust.runners: Stopping all users
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Updating state to 'cleanup', old state was 'running'
[2021-01-18 13:41:17,718] C02CD0BNMD6N/INFO/locust.runners: Stopping 5 users
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Stopping Greenlet-0
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Stopping Greenlet-1
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Stopping Greenlet-2
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Stopping Greenlet-3
[2021-01-18 13:41:17,718] C02CD0BNMD6N/DEBUG/locust.runners: Stopping Greenlet-4

[...]

PR #1755, opened 2021-05-04
Add gRPC load test example

beandrad:
This PR has two commits: 

- The first one changes the call to the `stop()` instance method in the runner so that it can be overridden.
- The second one (which needs the first commit) adds an example on how to implement a gRPC load test to avoid #1676; please note that I created a PR (https://github.com/grpc/grpc/pull/26058) to solve the root cause in `grpcio`, but I'm not sure if/when will be merged.


cyberw:
Looks good! Just auto-format the files using Black and I can merge.

cyberw:
thanks!

Liuyanzhi:
@beandrad Please ask weather support grpc BidirectStrem interface? I used Generator to grpc stream method, but the test case was stuck

PR #1756, opened 2021-05-05
allow passage of arguments to main if used externally

pranavgupta1234:
In case if main is imported while using locust as a library then allowing taking args can be helpful

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -116,16 +116,16 @@ def create_environment(user_classes, options, events=None, shape_class=None):
     )
 
 
-def main():
+def main(args=None):
     # find specified locustfile and make sure it exists, using a very simplified
     # command line parser that is only used to parse the -f option
-    locustfile = parse_locustfile_option()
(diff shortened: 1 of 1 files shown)


cyberw:
can you add a test case and an example (in /examples) of this?

cyberw:
I'm not sure about this change actually. Isnt it better to use the standard use as library functionality? https://docs.locust.io/en/latest/use-as-lib.html

pranavgupta1234:
It might be useful is someone is using whole core functionality of main() and wants to do some wrapping around it. Our parse_args() functions already accept args as required in test cases so it wouldn't hurt to connect the chain to so they are receivable from main() itself. Let me know if this makes sense.

cyberw:
Sorry, I didnt see that you had responded! I guess it might make sense. Can you show me your use case and add a test case for it?

cyberw:
closing due to inactivity

PR #1758, opened 2021-05-07
Measure elapsed time using time.perf_counter() 

cyberw:
... instead of a mix of time.monotonic(), which sometimes has less precision, and timeit (used in FastHttpUser), which has undesirable effects like disabling GC.

Also, in FastHttpUser, dont round response times to millisecond for no reason.

Note: On Linux/mac .perf_counter() calls .monotonic() in the end, so there is no difference there


cyberw:
Force push for better commit messages :)

Issue #1762, opened 2021-05-19
status "stopped" instead of "spawning", tick() method of LoadShape called only once


stanislawskwark:
### Describe the bug
This happens with high number of workers, for me it sometimes happens when I use ~80 workers, it does not happen each time though, I didn't find the root cause so far. It never happened for me when running standalone locust, with low number of workers it happened just once as far as I know. 
Sometimes build starts with "stopped" status instead of "spawning". Number of users grows until it reaches expected value, then status changes to "running". The most problematic thing for me is that when it happens, locust stops calling tick() method of load shape, so shape doesn't really work and number of users get stuck on value returned in first call of tick() method.
### Expected behavior
Status shown in GUI represents actual status and tick method is called every second until it returns None or the build is stopped manually

### Actual behavior
GUI hows status "stopped" instead of "spawning", tick() method of LoadShape called only once

### Steps to reproduce
set up master and around 80 workers (maybe it would happen more often with higher number of workers, I didn't have access to more machines to check it)
### Environment

- OS: centos 7
- Python version: 3.8.2
- Locust version: 1.4.3, 1.4.4, 1.5.2
- Locust command line that you ran: locust --master --skip-log-setup --only-summary
- Locust file contents (anonymized if necessary): this happens regardless of locust file, even with dummy locust file which doesn't do any requests, just task which creates random list

cyberw:
If you can, then have a look at fixing it your self @stanislawskwark . Or maybe @max-rocket-internet can have a look.

stanislawskwark:
@cyberw I would be happy to fix it, I would need some guidance though. from what I checked until now - as I suspected shape failure is direct consequence of changes state - tick() is called only if state is STATE_INIT, STATE_SPAWNING or STATE_RUNNING.

from what I see first two calls are within milliseconds, and state is set to STOPPED, so my guess is that `check_stopped` is misbehaving on higher load and 
`        if (
            not self.state == STATE_INIT
            and not self.state == STATE_STOPPED
            and all(map(lambda x: x.state != STATE_RUNNING and x.state != STATE_SPAWNING, self.clients.all))
        )
`  evaluates to True because correct status was not set yet, could somebody confirm or give me some hint?

cyberw:
That code was written by @max-rocket-internet . Max, can you help out?

stanislawskwark:
I was not able to reproduce it after applying  the change which I added in pull request - it seems to be solving the issue

PR #1769, opened 2021-05-27
modified check_stopped condition

stanislawskwark:
added STATE_INIT to condition in check_stopped condition for clients - it looks like there's possibility that master is not in STATE_INIT anymore, but some client didn't have enough time to change the state

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -608,7 +608,7 @@ class MasterRunner(DistributedRunner):
         if (
             not self.state == STATE_INIT
             and not self.state == STATE_STOPPED
-            and all(map(lambda x: x.state != STATE_RUNNING and x.state != STATE_SPAWNING, self.clients.all))
+            and all(map(lambda x: x.state not in (STATE_RUNNING, STATE_SPAWNING, STATE_INIT), self.clients.all))
         ):
             self.update_state(STATE_STOPPED)
 


stanislawskwark:
From what I tested - this seems to be fixing https://github.com/locustio/locust/issues/1762

cyberw:
thx!

mboutet:
@stanislawskwark @cyberw, when I merged this change into https://github.com/locustio/locust/pull/1621, one of the [tests](https://github.com/mboutet/locust/blob/users-distribution/locust/test/test_runners.py#L657) began to fail because the assertion checking if the master's state is `stopped` fails. (the test is not in the locust's master branch, it is new from https://github.com/locustio/locust/pull/1621). 

I also observed the behavior when a distributed load test is stopped (i.e. `self.stop()` is called in the master), the state of the master stays at `stopping` indefinitely.

I investigated and found that the master stays `stopping` because when a worker receives the stop message, it also [sends back](https://github.com/mboutet/locust/blob/users-distribution/locust/runners.py#L997) a `client_ready` message to the master. Thus, the `check_stopped` never evaluates to `True`. In summary this is what happens:

1. Master sends the `stop` message to all workers.
2. Upon reception by the worker, the worker [sends](https://github.com/mboutet/locust/blob/487873b3f55a57763e2301b243f9375dc85d36f6/locust/runners.py#L996) a `client_stopped` to the master which result in the master [removing](https://github.com/mboutet/locust/blob/487873b3f55a57763e2301b243f9375dc85d36f6/locust/runners.py#L786) the client from its list of workers.
3. Immediately after sending the `client_stopped`, the worker also [sends](https://github.com/mboutet/locust/blob/487873b3f55a57763e2301b243f9375dc85d36f6/locust/runners.py#L997) a `client_ready` message which result in the master [adding back](https://github.com/mboutet/locust/blob/487873b3f55a57763e2301b243f9375dc85d36f6/locust/runners.py#L774) the worker to its list of workers.
4. Because the list of worker is now populated with client being in the `STATE_INIT` state, the `check_stopped` never evaluates to `True` and the master never transitions to the `STATE_STOPPED` state.

Is this intentional?

mboutet:
How about the master keeps track of when a test was started and use this in the `check_stopped` method? That way, the condition would be inhibited for the first `t` seconds of starting a load test thus giving time for all workers to transition state.

Something like:

```python
    def check_stopped(self):
        if (
            time.time() - self.started_at > int(os.getenv("STARTED_AT_THRESHOLD", 10))
            and self.state not in [STATE_INIT, STATE_STOPPED]
            and all(map(lambda x: x.state not in (STATE_RUNNING, STATE_SPAWNING), self.clients.all))
        ):
            self.update_state(STATE_STOPPED)
```

By making it configurable via an environment variable, it would be possible to accommodate different load test scale such as yours with lots of workers.

jbarz1:
curious if this will be released soon

cyberw:
> curious if this will be released soon

Yes, 1.6.0 should be out in a minute or two!

Issue #1506, opened 2020-08-07
Supplying data centrally. Coordinating in cluster mode.


mingp:
I'm unsure if this is a request for a new feature or an inquiry about how to use existing features. I would be happy with either outcome.

Thanks in advance.

### Is your feature request related to a problem? Please describe.

I need to either supply data centrally to individual locusts or coordinate between locusts in cluster mode.

The canonical use-case is as follows. Suppose that I have a limited set of test user accounts, that I need to distribute between my locusts in cluster mode. I have more test user accounts than locusts, but not necessarily by that large of a margin. I would ideally like to hand them out sequentially, round-robin.

Random assignment, where each locust randomly selects from the whole list, by and large works, but has significant drawbacks. In particular, if the set of available test accounts is not much larger than the set of locusts, then we start to get collisions, where multiple locusts pull the same test account, at a rate determined by the Birthday Paradox. This is bad if, for example, running multiple concurrent test scripts against the same test account would cause test script failure or potentially even leave the test account wedged in an invalid state.

### Describe the solution you'd like

There should ideally be a way from the central coordinator, ideally via some sort of hook that the test script can add, to hand out data to individual locusts in cluster mode. Given that the central coordinator already needs to send start and stop requests to locusts, it seems like this data can piggy-back onto existing calls.

Notable prior art comes from [Gatling](https://gatling.io/docs/current/), a competing load-testing framework, which has a feature that it calls [feeders](https://gatling.io/docs/current/session/feeder), by which data can be centrally injected into individual sessions. IMO, something similar could benefit Locust as well.

### Describe alternatives you've considered

1. Random assignment, as detailed above. It works, but has caveats, and the failure mode can be situationally catastrophic, as detailed above.
2. An external API endpoint handles this logic. The load test script calls the API endpoint to fetch parameters for the current run. It works, but is quite clunky, to have to spin up additional seemingly unrelated infrastructure.

### Additional context

N/A.

cyberw:
Hi! I think this would be nice but need a lot of work. Personally, I keep this sort of data in a MongoDb, see locust-plugins, specifically MongoReader https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/mongoreader_ex.py

max-rocket-internet:
[...]

PR #1782, opened 2021-06-07
Allow cross process communication using custom messages

### LINKED ISSUE #1780: Allow master node to supply data to worker nodes directly

Pretty much a duplicate of #1506, but it would be extremely beneficial for locust to be able to communicate data between nodes easily. An example that you're probably familiar with is needing to retrieve user accounts from a database for login, and trying to coordinate that between several worker nodes so that the same users are not logged in across multiple workers. 

Adding hooks to allow communication between nodes would be ideal. I'm not sure exactly what the best solution would be but allowing the user to define custom messages and handlers for the greenlets seems like it'd be easy enough to implement and massively improve locust's ability to coordinate itself in distributed mode. 

The currently accepted workarounds (e.g. mongodb or a REST api) are much less than ideal, especially for something like a docker environment where each additional component must be installed and set up regularly.

### Additional context

I'm available to give this a shot, but I wanted to make sure A) it has value and B) the best implementation from those who are more experienced on this project.

nathan-beam:
This pull fixes #1780 and #1506. 

Added two methods, `register_message` and `send_message` to the two `DistributedRunner` classes.

`register_message` takes a custom message type and a callback function to call when the specified message is received. The callback's parameters are `environment` (obviously the runner's environment) and `msg`, which is the message itself. 

`send_message` takes a custom message type and optional data to send with the message, allowing a user to easily send their own custom messages.

Updated tests/documentation where necessary, added an example implementation to the examples directory (`examples\custom_messages.py`).

### DIFF
--- a/docs/running-locust-distributed.rst
+++ b/docs/running-locust-distributed.rst
@@ -90,8 +90,8 @@ nodes has connected before the test is started.
 Communicating across nodes
 =============================================
 
-When running Locust in distributed mode, you may want to communicate between nodes in order to coordinate 
-data. This can be easily accomplished with custom messages using the built in messaging hooks:
+When running Locust in distributed mode, you may want to communicate between master and worker nodes in 
+order to coordinate data. This can be easily accomplished with custom messages using the built in messaging hooks:
 
 .. code-block:: python
 
@@ -110,20 +110,26 @@ data. This can be easily accomplished with custom messages using the built in me
 
     @events.init.add_listener
     def on_locust_init(environment, **_kwargs):
-        if isinstance(environment.runner, WorkerRunner):
+        if not isinstance(environment.runner, MasterRunner):
             environment.runner.register_message('test_users', setup_test_users)
-        elif isinstance(environment.runner, MasterRunner):
+        if not isinstance(environment.runner, WorkerRunner):
             environment.runner.register_message('acknowledge_users', on_acknowledge)
 
     @events.test_start.add_listener
     def on_test_start(environment, **_kwargs):
-        if isinstance(environment.runner, MasterRunner):
+        if not isinstance(environment.runner, MasterRunner):
             users = [
                 {"name": "User1"},
                 {"name": "User2"},
                 {"name": "User3"},
             ]
             environment.runner.send_message('test_users', users)  
+
+Note that when running locally (i.e. non-distributed), this functionality will be preserved; 
+the messages will simply be handled by the same runner that sends them.  
+
+A more complete example can be found in the `examples directory <https://github.com/locustio/locust/tree/master/examples>`_ of the Locust 
+source code.
 
 
 Running distributed with Docker
--- a/docs/running-locust-distributed.rst
+++ b/docs/running-locust-distributed.rst
@@ -87,6 +87,44 @@ listen to. Defaults to 5557.
 Used when starting the master node with ``--headless``. The master node will then wait until X worker
 nodes has connected before the test is started.
 
+Communicating across nodes
+=============================================
+
+When running Locust in distributed mode, you may want to communicate between nodes in order to coordinate 
+data. This can be easily accomplished with custom messages using the built in messaging hooks:
+
+.. code-block:: python
+
+    from locust import events
+    from locust.runners import MasterRunner, WorkerRunner
+
+    # Fired when the worker recieves a message of type 'test_users'
+    def setup_test_users(environment, msg, **kwargs):
+        for user in msg.data:
+            print(f"User {user['name']} recieved")
+        environment.runner.send_message('acknowledge_users', f"Thanks for the {len(msg.data)} users!")
+
+    # Fired when the master recieves a message of type 'acknowledge_users'
+    def on_acknowledge(msg, **kwargs):
+        print(msg.data)
+
+    @events.init.add_listener
+    def on_locust_init(environment, **_kwargs):
+        if isinstance(environment.runner, WorkerRunner):
+            environment.runner.register_message('test_users', setup_test_users)
+        elif isinstance(environment.runner, MasterRunner):
+            environment.runner.register_message('acknowledge_users', on_acknowledge)
+
+    @events.test_start.add_listener
+    def on_test_start(environment, **_kwargs):
+        if isinstance(environment.runner, MasterRunner):
+            users = [
+                {"name": "User1"},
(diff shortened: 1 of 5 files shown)


locust/runners.py
```diff
@@ -709,13 +718,27 @@ def client_listener(self):
...
 
+    def send_message(self, msg_type, data=None):
+        """
+        Sends a message to all attached worker nodes
+
+        :param msg_type: The type of the message to send
+        :param data: Optional data to send
+        """
+        for client in self.clients.all:
+            logger.debug("Sending {msg_type} message to client {client_id}")
```
cyberw, line 738: Are you trying to do an f-string? If so I think you are missing the `f` :)

examples/custom_messages.py
```diff
@@ -0,0 +1,37 @@
+from locust import HttpUser, task, events, between
```
cyberw, line 1: Can we do a more "real world" example? Like the worker asking for some test data (a customer ID or something) and the master replying with it? I think it is one of the most common use cases for this.

examples/custom_messages.py
```diff
@@ -0,0 +1,37 @@
...
+
+@events.test_start.add_listener
+def on_test_start(environment, **_kwargs):
+    if isinstance(environment.runner, MasterRunner):
+        users = [
+            {"name": "User1"},
+            {"name": "User2"},
+            {"name": "User3"},
+        ]
+        environment.runner.send_message('test_users', users)
```
cyberw, line 29: Does it make sense for the master to be able to send a message to a specific worker? Sending to all workers is also useful, but sending to a specific one is probably even more so.

cyberw:
Cool stuff. You saw the linting/formatting issues right? I'd like to hold off on this until we have merged and released the new worker distribution style though. So 2.1 at the earliest.

examples/custom_messages.py
nathan-beam: Sure, let me add in something a little less theoretical

examples/custom_messages.py
nathan-beam: Can you give an example of a use-case where the targeting would be beneficial? In my opinion, since the Master node has so little information with which to target the nodes anyways, it'd be easier to either only register the events you'd want to target on the nodes you're interested in, or include the targeting criteria in the message data somehow, rather than trying to figure out from the master which node you'd want to receive the message.

nathan-beam:
Any ETA on 2.1? I have some functionality that would depend on this, so just want to have a timeline so I can plan things out as best I can. Thanks!

examples/custom_messages.py
nathan-beam: Actually after updating the `custom_messages` example I see the usefulness. I'll add that in as well. 👍

cyberw:
It mostly depends on that other PR, and I’m not writing it myself. I’d say maybe a month?

benesadia2000 (approved):

cyberw:
A small note: please write commit messages in a way that makes sense even if the reader doesnt know what branch/PR they were part of.

cyberw:
Maybe some test cases for when a message is sent but no receiving method is registered? (should log a warning)

pappacena:
+1 for this PR.

I'm facing a similar situation, and I had to mostly copy the `WorkerRunner.worker` method in a subclass. 

I was [preparing a refactoring](https://github.com/locustio/locust/compare/master...pappacena:refactoring-worker-message?expand=1) to make it easier to subclass `WorkerRunner` and add new message types, but this proposal is way better! 

Good job!

cyberw:
@nathan-beam If you add the test cases I mentioned I will merge this and it will become part of the next minor version. No point in waiting for that other PR (and this is not a breaking change anyway).

nathan-beam:
@cyberw Added. Let me know if you think it's missing anything else

cyberw:
Looks pretty good now. 

Just one thing: as most plans are testes in a stand alone environment before moving to distributed, could this be made to work on stand alone (not distributed) runs? (or does it already work that way? If so then just add a test case for it)

nathan-beam:
> Just one thing: as most plans are testes in a stand alone environment before moving to distributed, could this be made to work on stand alone (LocalRunner) runs?

Not sure what "working" would look like on a local runner, since there isn't any messaging infrastructure. Are you alright with just logging a warning? The alternative I can think of is to covertly register a new event and have the `send_message()` method fire that event to emulate sending/receiving a message.

cyberw:
It would be great if we could emulate sending/receiving, so that the same test plan can work standalone and distributed without changes. I havent thought about it too much though, but see if you can't solve it without resorting to any weird hacks (if it is impossible/very hard then lets not do it)

nathan-beam:
Was actually even easier than I was thinking. Just moved the custom messages dictionary and registration the base Runner class and simply called the registered callback from the LocalRunner's `send_message()` function. Should be 100% functional between Master/Worker and Local modes now 👍

docs/running-locust-distributed.rst
```diff
@@ -87,6 +87,44 @@ listen to. Defaults to 5557.
 Used when starting the master node with ``--headless``. The master node will then wait until X worker
 nodes has connected before the test is started.
 
+Communicating across nodes
+=============================================
+
+When running Locust in distributed mode, you may want to communicate between nodes in order to coordinate
```
cyberw, line 93: between master and worker nodes

docs/running-locust-distributed.rst
```diff
@@ -87,6 +87,44 @@ listen to. Defaults to 5557.
...
+    @events.test_start.add_listener
+    def on_test_start(environment, **_kwargs):
+        if isinstance(environment.runner, MasterRunner):
+            users = [
+                {"name": "User1"},
+                {"name": "User2"},
+                {"name": "User3"},
+            ]
+            environment.runner.send_message('test_users', users)  
+
```
cyberw, line 127: Maybe add an actual User class, so that the example can run for real?

nathan-beam: Most inline code blocks in the documentation seem to only include the relevant code snippets, I would think to avoid cluttering up the page with familiar concepts. What about adding just a link to the full example in the examples directory?

cyberw: Sounds good!

cyberw:
Nice work!

PR #1784, opened 2021-06-09
Ensure that the exception dictionaries are not mutated when generating a html report

mboutet:
This PR fixes the following issue:

```
[2021-06-09 20:36:13,702] master1/CRITICAL/locust.runners: Unhandled exception in greenlet: <Greenlet at 0x7ff41a2846a0: <bound method MasterRunner.client_listener of <locust.runners.MasterRunner object at 0x7ff3ea9d4b80>>>
  File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/adminuser/.venv/lib/python3.8/site-packages/locust/runners.py", line 842, in client_listener
    self.log_exception(msg.node_id, msg.data["msg"], msg.data["traceback"])
  File "/home/adminuser/.venv/lib/python3.8/site-packages/locust/runners.py", line 422, in log_exception
    row["nodes"].add(node_id)
AttributeError: 'str' object has no attribute 'add'
```

If a user generated a html report during a load test run, then the `nodes` field in every `exception` dict was mutated to a string.

This also caused the following when hovering on one of the exception in the web UI:

Screen Shot 2021-06-09 at 18 08 27

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -1,5 +1,3 @@
-from copy import deepcopy
-
 from jinja2 import Environment, FileSystemLoader
 import os
 import pathlib
@@ -37,11 +35,9 @@ def get_html_report(environment, show_download_link=True):
 
     requests_statistics = list(chain(sort_stats(stats.entries), [stats.total]))
     failures_statistics = sort_stats(stats.errors)
-    exceptions_statistics = []
-    exceptions = deepcopy(environment.runner.exceptions)
-    for exc in exceptions.values():
-        exc["nodes"] = ", ".join(exc["nodes"])
-        exceptions_statistics.append(exc)
+    exceptions_statistics = [
+        {**exc, "nodes": ", ".join(exc["nodes"])} for exc in environment.runner.exceptions.values()
+    ]
 
     history = stats.history
 
--- a/locust/html.py
+++ b/locust/html.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 from jinja2 import Environment, FileSystemLoader
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


mboutet:
PS: I will incorporate this fix in the distribution branch because it is causing my load tests to crash (I use the html report feature quite a lot).

locust/html.py
```diff
@@ -36,7 +38,8 @@ def get_html_report(environment, show_download_link=True):
     requests_statistics = list(chain(sort_stats(stats.entries), [stats.total]))
     failures_statistics = sort_stats(stats.errors)
     exceptions_statistics = []
-    for exc in environment.runner.exceptions.values():
+    exceptions = deepcopy(environment.runner.exceptions)
+    for exc in exceptions.values():
         exc["nodes"] = ", ".join(exc["nodes"])
         exceptions_statistics.append(exc)
```
cyberw, line 44: Why not just do something like `exceptions_statistics.append(", ".join(exc["nodes"]))` instead? No need to copy runner.exceptions if `exceptions` is never used again, right?

mboutet: The `exceptions_statistics` must also contain the other fields of the exception dictionary as the html template also needs the other properties:

https://github.com/locustio/locust/blob/60cde73cee8c08bd5c6efb88eb0d511c6c025620/locust/html.py#L66-L80

If you prefer, it could be something like this instead of the for-loop:

```python
exceptions_statistics = [
    {**exc, "nodes": ", ".join(exc["nodes"])}
    for exc in environment.runner.exceptions.values()
]
```

mboutet: I changed it. Looks better now in my opinion.

cyberw:
Thx!

PR #1795, opened 2021-06-21
Use `None` instead of mutable `[]` default argument

mboutet:
This is a draft.

See https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -66,7 +66,7 @@ class Environment:
     def __init__(
         self,
         *,
-        user_classes=[],
+        user_classes=None,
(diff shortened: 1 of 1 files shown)


cyberw:
I was going to say that this is unnecessary, but when I think about it, any code in the class, not just in this function, could accidentally modify the default value. So I'm +1 now :)

cyberw:
I cherry picked this commit instead of merging (dont want too many PR:s in the changelog). Sorry for the extra work :)

PR #1799, opened 2021-06-28
Added last request timestamp validation to join the dots connection i…

EshwarMuthu:
**Before** _(in master)_
Screenshot 2021-06-28 at 8 50 50 PM

**After** _(in Fix-1702)_
Screenshot 2021-06-28 at 9 15 11 PM


cyberw:
Hmm. This does not look like an improvement to me. A missing data point can’t be represented as zero ms.

PR #1805, opened 2021-07-05
Check version of workers when they connect. Warn if there is a mismatch, refuse 1.x workers to connect

cyberw:


cyberw:
Force pushing a ton of times, like a boss

PR #1807, opened 2021-07-06
Introduce `LOCUST_SPAWN_AT_LEAST_ONE_USER_OF_EACH_USER_CLASS`

mboutet:
This boolean environment variable allows to change the behaviour of the users distribution. If `true`, it behaves as before by attempting to spawn at least one user of each class when the user count >= to the number of user classes.

If `false`, the real distribution is calculated, meaning that users with small weights won't always be spawned when a lower user count is set.

The test helper `patch_env` has been relocated in `test/util.py`.

Based on the conversion with "voyager" on Slack.

### DIFF
--- a/locust/distribution.py
+++ b/locust/distribution.py
@@ -1,4 +1,5 @@
 import math
+import os
 from itertools import combinations_with_replacement
 from operator import attrgetter
 from typing import Dict, List, Type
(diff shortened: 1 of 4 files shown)


mboutet:
Closing because no longer applicable as of https://github.com/locustio/locust/pull/1809  because the distribution is handled by the `roundrobin` library.

PR #1809, opened 2021-07-07
Refactoring of the dispatch logic to improve performance

mboutet:
NOTE: Still a work in progress, tests are not yet all updated so they fail.

This PR addresses the performance issues introduced by [this other PR](https://github.com/locustio/locust/pull/1621) when one or more of the following is true:
* Lots of workers (> 200)
* Lots of user classes (> 25)
* High spawn rate (> 100)

I completely refactored the `UsersDispatcher`. In fact, it is almost completely different from before and a lot simpler. I also ditched the `distribution.py` logic as I'm now using [this great little library](https://github.com/linnik/roundrobin) which allows for a nginx-like weighted round-robin dispatch of the users. Ramp-down is also now supported (i.e. stopping users at a given rate).

I've implemented most of the tests to validate this new implementation in `test_dispatch.py`, but I've not yet updated the `runners` module.

I'm still missing the logic to ensure that all workers run the expected users prior to beginning a ramp-up/down. I'll implement it in the next few days.

I implemented a small benchmark with the following config:
* Ramp-up from 0 to 100 000 users
* 1000 workers
* 50 user classes with varying weights
* 5000 spawn rate

Each dispatch iteration takes around 130ms to compute which is very good. It's orders of magnitude faster than before. The performance is similar for the inverse scenario from 100 000 to 0 users.


locust/dispatch.py
```diff
@@ -49,548 +48,134 @@ class UsersDispatcher(Iterator):
...
-        return self._user_count_left_to_assigned == 0
+    def _dispatcher(self) -> Generator[Dict[str, Dict[str, int]], None, None]:
+        # TODO: Ensure initial users on workers are ok. If not, send a single dispatch
+        #       to set the workers with the adequate users. After having done that,
+        #       we need to update `self._users_on_workers` and `self._current_user_count`.
 
-    @property
-    def _user_count_left_to_assigned(self) -> int:
-        return sum(map(sum, map(dict.values, self._users_left_to_assigned.values())))
+        if self._current_user_count < self._target_user_count:
```
cyberw, line 117: this if and elif statements seem unnecessary?

mboutet:
```python
        if self._current_user_count == self._target_user_count:
            yield self._initial_users_on_workers
            return

        while self._current_user_count < self._target_user_count:
            with self._wait_between_dispatch_iteration_context():
                yield self._ramp_up()

        while self._current_user_count > self._target_user_count:
            with self._wait_between_dispatch_iteration_context():
                yield self._ramp_down()
```

better?

locust/dispatch.py
```diff
@@ -49,548 +48,134 @@ class UsersDispatcher(Iterator):
...
+            if self._current_user_count >= min(
+                initial_user_count + self._user_count_per_dispatch_iteration, self._target_user_count
+            ):
+                return deepcopy(self._users_on_workers)
+
+    def _ramp_down(self) -> Dict[str, Dict[str, int]]:
+        initial_user_count = self._current_user_count
+        for user in self._user_gen_generator:
+            while True:
+                worker_name = next(self._worker_node_id_generator)
```
cyberw, line 166: does this ensure that if you ramp from for example 10->20->10 users you get back the distribution you originally had? (that was what I was getting at with my stopped_users list)

locust/dispatch.py
```diff
@@ -49,548 +48,133 @@ class UsersDispatcher(Iterator):
...
-        # `self._try_next_user_class_in_order_to_stay_balanced_during_ramp_up` is invoked, we need
-        # `self._dispatched_user_classes_count` to be recomputed.
-        self._dispatched_user_classes_count.cache_clear()
-
-        if all(user_count > 0 for user_count in self._dispatched_user_classes_count().values()):
-            # We're here because each user class have at least one user running. Thus,
-            # we need to ensure that the distribution of users corresponds to the weights.
-            if not self._adding_this_user_class_respects_distribution_better_than_adding_any_other_user_class(
-                user_class_to_add
+    def _ramp_up(self) -> Dict[str, Dict[str, int]]:
```
cyberw, line 150: the way I'm understanding this method is that it adds on user? Maybe rename it to indicate that?

locust/dispatch.py
```diff
@@ -49,548 +48,133 @@ class UsersDispatcher(Iterator):
...
-
-                number_of_users_left_to_assign = user_count - self._number_of_assigned_users_across_workers(
-                    assigned_users
-                )
-
-                if (
-                    self._number_of_assigned_users_for_worker(assigned_users, worker_node) == users_per_worker
-                    and number_of_users_left_to_assign > remainder
-                ):
+                return deepcopy(self._users_on_workers)
```
cyberw, line 159: I think doing a deepcopy of _users_on_workers makes this method O(number_of_users) instead of O(number_of_workers) like in my PoC. But maybe there is no way around that?

locust/dispatch.py
cyberw: and does it also preserve the distribution if a worker disconnects and users have to be spawned across the other workers?

locust/dispatch.py
cyberw: > Each dispatch iteration takes around 130ms to compute which is very good. It's orders of magnitude faster than before.

Is that the average time? If so, the last few iterations are probably painfully slow. If I understand it correctly, the last dispatch iteration (going from 95.000 to 100.000) has to copy almost 100.000 entries in _users_on_workers 5000 times.

cyberw: @mboutet ?

benchmarks/dispatch.py
```diff
@@ -0,0 +1,632 @@
+import itertools
```
cyberw, line 1: Add a comment to this file to make sure people understand it is for benchmarking locust itself

locust/runners.py
```diff
@@ -678,12 +692,12 @@ def start(self, user_count: int, spawn_rate: float, **kwargs) -> None:
```
cyberw, line 692: Maybe remove the high spawn rate warning on line 680, it shouldnt be a problem any more, right?

locust/runners.py
```diff
@@ -760,7 +778,7 @@ def _wait_for_workers_report_after_ramp_up(self) -> float:
 
         assert False, "not supposed to reach that"
 
-    def stop(self):
```
cyberw, line 763: "not supposed to reach that"-assert on 779 is unnecessary, change  the if statements to if-else and remove this.

locust/runners.py
```diff
@@ -760,7 +778,7 @@ def _wait_for_workers_report_after_ramp_up(self) -> float:
...
+    def stop(self, send_stop_to_client: bool = True):
```
cyberw, line 781: client is synonymous with worker, right? We dont have to change that in this PR, but I think we should clean this up in the future and user "worker" everywhere.

locust/runners.py
```diff
@@ -816,6 +837,9 @@ def heartbeat_worker(self):
                     logger.info("Worker %s failed to send heartbeat, setting state to missing." % str(client.id))
                     client.state = STATE_MISSING
                     client.user_classes_count = {}
+                    if self._users_dispatcher is not None:
```
cyberw, line 840: it seems we're checking whether _user_dispatcher is not None in a lot of places. Is there a better way to handle this? Spawn the dispatcher earlier so that it always exists maybe? Idk, maybe a bad idea, but it looks a little ugly atm.

locust/dispatch.py
```diff
@@ -49,548 +49,241 @@ class UsersDispatcher(Iterator):
...
+    def add_worker(self, worker_node: "WorkerNode") -> None:
+        self._worker_nodes.append(worker_node)
+        self._worker_nodes = sorted(self._worker_nodes, key=lambda w: w.id)
+        self._prepare_rebalance()
 
-        return False
+    def remove_worker(self, worker_node_id: str) -> None:
+        self._worker_nodes = [w for w in self._worker_nodes if w.id != worker_node_id]
+        if len(self._worker_nodes) == 0:
+            # TODO: Test this
```
cyberw, line 176: There is a test for this now, right? Remove the todo :)

locust/dispatch.py
```diff
@@ -49,548 +49,241 @@ class UsersDispatcher(Iterator):
...
 
-    @property
-    def _workers_user_count(self) -> Dict[str, int]:
-        """User count currently running on each of the workers"""
-        return {
-            worker_node_id: sum(dispatched_users_on_worker.values())
-            for worker_node_id, dispatched_users_on_worker in self._dispatched_users.items()
-        }
+    def _prepare_rebalance(self) -> None:
+        users_on_workers, user_gen, worker_gen, active_users = self._distribute_users(self._current_user_count)
```
cyberw, line 181: Is the todo on line 188 still relevant?

locust/dispatch.py
```diff
@@ -49,548 +49,241 @@ class UsersDispatcher(Iterator):
...
-    def _user_class_cannot_be_assigned_to_any_worker(self, user_class_to_add: str) -> bool:
-        """No worker has enough place to accept this user class"""
-        effective_user_count_of_that_user_class_that_can_be_added = sum(
-            self._desired_users_assigned_to_workers[worker_node.id][user_class_to_add]
-            - self._dispatched_users[worker_node.id][user_class_to_add]
-            for worker_node in self._worker_nodes
-        )
-        if effective_user_count_of_that_user_class_that_can_be_added <= 0:
-            return True
+        # TODO: What to do with self._initial_users_on_workers?
```
cyberw, line 188: is this still relevant?

locust/dispatch.py
```diff
@@ -49,548 +49,241 @@ class UsersDispatcher(Iterator):
...
+    @contextlib.contextmanager
+    def _wait_between_dispatch_iteration_context(self) -> None:
+        t0_rel = time.perf_counter()
 
-    def _worker_is_full(self, worker_node_id: str) -> bool:
-        """The worker cannot accept more users without exceeding the maximum user count it can run"""
-        return self._workers_user_count[worker_node_id] >= self._workers_desired_user_count[worker_node_id]
+        # We don't use `try: ... finally: ...` because we don't want to sleep
+        # if there's an exception within the context.
+        yield
```
cyberw, line 198: Why use a context manager at all if we're not catching exceptions?

cyberw: Ah, it is just for checking the timing?

locust/runners.py
```diff
@@ -678,12 +692,12 @@ def start(self, user_count: int, spawn_rate: float, **kwargs) -> None:
...
 
+        self._users_dispatcher.new_dispatch(target_user_count=user_count, spawn_rate=spawn_rate)
+
         try:
-            for dispatched_users in UsersDispatcher(
-                worker_nodes=self.clients.ready + self.clients.running + self.clients.spawning,
-                user_classes_count=self.target_user_classes_count,
-                spawn_rate=spawn_rate,
-            ):
+            dispatched_users = None
```
cyberw, line 698: I think it is best to remove this

locust/runners.py
```diff
@@ -711,7 +725,11 @@ def start(self, user_count: int, spawn_rate: float, **kwargs) -> None:
                     "Currently spawned users: %s" % _format_user_classes_count_for_log(self.reported_user_classes_count)
                 )
 
+            assert dispatched_users is not None
```
cyberw, line 728: remove this

locust/runners.py
```diff
@@ -816,6 +837,9 @@ def heartbeat_worker(self):
                     logger.info("Worker %s failed to send heartbeat, setting state to missing." % str(client.id))
                     client.state = STATE_MISSING
                     client.user_classes_count = {}
+                    if self._users_dispatcher is not None:
+                        self._users_dispatcher.remove_worker(client.id)
+                        # TODO: If status is `STATE_RUNNING`, call self.start()
```
cyberw, line 842: double back to this later

locust/runners.py
```diff
@@ -299,25 +305,23 @@ def start(self, user_count: int, spawn_rate: float, wait: bool = False):
...
-
         if self.state != STATE_INIT and self.state != STATE_STOPPED:
             self.update_state(STATE_SPAWNING)
 
+        if self._users_dispatcher is None:
+            self._users_dispatcher = UsersDispatcher(
+                worker_nodes=[self._local_worker_node], user_classes=self.user_classes
+            )
+
         logger.info("Ramping to %d users using a %.2f spawn rate" % (user_count, spawn_rate))
```
cyberw, line 316: Maybe rephrase this to "Ramping to %d users at a rate of %.1f per second"? (particularly the "per second" makes the sentence make more sense)

cyberw:
I found two issues: (kind of minor, but I'll hold off on the merge for now) 

Connecting a single worker and then shutting it down prints a message that locust will shut down, but it never happens:

```
locust -f demo.py -u 120 --master --expect-workers 2
[2021-07-12 13:33:05,825] lars-mbp.local/INFO/root: Waiting for workers to be ready, 0 of 2 connected
[2021-07-12 13:33:06,054] lars-mbp.local/INFO/locust.runners: Client 'lars-mbp.local_640ed5421ae94f4889c952b12cc63c8e' reported as ready. Currently 1 clients ready to swarm.
[2021-07-12 13:33:06,828] lars-mbp.local/INFO/root: Waiting for workers to be ready, 1 of 2 connected
...
[2021-07-12 13:33:19,871] lars-mbp.local/INFO/root: Waiting for workers to be ready, 1 of 2 connected
[2021-07-12 13:33:20,670] lars-mbp.local/INFO/locust.runners: Client 'lars-mbp.local_640ed5421ae94f4889c952b12cc63c8e' quit. Currently 0 clients connected.
[2021-07-12 13:33:20,671] lars-mbp.local/INFO/locust.runners: The last worker quit, stopping test.
[2021-07-12 13:33:20,872] lars-mbp.local/INFO/root: Waiting for workers to be ready, 0 of 2 connected

```
Eventually I shut down the master, and got a stacktrace:
```
...
[2021-07-12 13:34:42,129] lars-mbp.local/INFO/root: Waiting for workers to be ready, 0 of 2 connected
^CKeyboardInterrupt
2021-07-12T11:34:42Z
  File "/usr/local/bin/locust", line 33, in <module>
    sys.exit(load_entry_point('locust', 'console_scripts', 'locust')())
  File "/Users/lafp/git/locust/locust/main.py", line 323, in main
    time.sleep(1)
  File "/usr/local/Cellar/python@3.7/3.7.10_3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gevent/hub.py", line 166, in sleep
    hub.wait(t)
  File "src/gevent/_hub_primitives.py", line 46, in gevent._gevent_c_hub_primitives.WaitOperationsGreenlet.wait
  File "src/gevent/_hub_primitives.py", line 55, in gevent._gevent_c_hub_primitives.WaitOperationsGreenlet.wait
  File "src/gevent/_waiter.py", line 154, in gevent._gevent_c_waiter.Waiter.get
  File "src/gevent/_greenlet_primitives.py", line 61, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_greenlet_primitives.py", line 61, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_greenlet_primitives.py", line 65, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_gevent_c_greenlet_primitives.pxd", line 35, in gevent._gevent_c_greenlet_primitives._greenlet_switch
KeyboardInterrupt
```

cyberw:
Lets discuss my proposed changes after merging.

locust/dispatch.py
mboutet: > does this ensure that if you ramp from for example 20->10->20

Yes, I added the `active_users` list in commits made afterward and there's a test specifically to ensure this use case works.

> and does the algorithm also preserve the distribution if a worker disconnects and users have to be spawned across the remaining workers?

Absolutely. Again, the actual logic was added in commits made after your comment.

locust/dispatch.py
mboutet: In commits made after and after profiling using `line_profiler`, I removed `deepcopy` and implemented `_fast_users_on_workers_copy` which is a lot more faster. If we're sure that the `users_on_workers` returned by `locust.dispatch.UsersDispatcher.__next__` is never mutated outside the `UsersDispatcher`, then we could even skip the `_fast_users_on_workers_copy` call in `locust.dispatch.UsersDispatcher.__next__`.

> Is that the average time? If so, the last few iterations are probably painfully slow.

I'm not sure if you wrote this after I shared the benchmark results on Slack. Anyway, as you can see in these benchmarks, there's the min, max and average, so you can see for yourself 😛 Spoiler: no it's not painfully slow. As I mentioned on Slack, there's some "glitches" in this benchmark which I suspect might be due to GC. It took a fairly long time to run all these combinations. For instance, this case:

```
+---------+---------+--------------+------------+----------------------------+------------------------------+
| Workers | Users   | User Classes | Spawn Rate | Ramp-Up (avg/min/max) (ms) | Ramp-Down (avg/min/max) (ms) |
+---------+---------+--------------+------------+----------------------------+------------------------------+
| 20,000  | 500,000 | 100          | 25,000     |   91.736/70.764/196.392    |     74.724/58.504/83.601     |
+---------+---------+--------------+------------+----------------------------+------------------------------+
``` 
recorded `56586.944ms` max for the ramp-up, but when I run it again, it's fast as shown above. Honestly, I don't know if we'd need more performance than that.

locust/dispatch.py
mboutet: Something like `_add_users_on_workers` and `_remove_users_on_workers`?

locust/runners.py
mboutet: Done (in my other PR)

locust/runners.py
mboutet: Yes, I did those modifications quickly before my short vacation.

The reason this is like this is to handle cases where locust is stopped or is asked to quit when there's no load test running. When a load test is not running, `self._users_dispatcher` is `None`. Perhaps it's ugly, but with the manual testing I performed, it worked well.

locust/runners.py
mboutet: Done in the other PR

locust/runners.py
mboutet: Done in the other PR

locust/runners.py
mboutet: Yes, I tried to use `client` in the `Runner` classes because it was the nomenclature already used, but workers would make more sense.

locust/runners.py
mboutet: Done in the other PR

benchmarks/dispatch.py
mboutet: Done in the other PR

locust/runners.py
mboutet: Yes, of course! Thanks. It's done in the other PR.

locust/dispatch.py
mboutet: Let me check tomorrow haha.

locust/dispatch.py
mboutet: I don't think there's a test for the zero worker case yet. I'll add one.

locust/dispatch.py
cyberw: yes that sounds better!

locust/dispatch.py
mboutet: I removed it in the other PR.

Issue #1822, opened 2021-07-23
locust should add extending blocks to index.html to change the start test and edit test options


TWoeste:
### Is your feature request related to a problem? Please describe.

I want to be able to configure different options when starting a test with another field/toggle other than just the spawn rate and number of users. 

### Describe the solution you'd like

I would like to have `{% extended_on_start %}` and `{% extended_edit_test %}` to `index.html` so I can add in my own code new input fields to specify what kind of test I want. I am writing changes to my loadtester and based off of this toggle I want the code to set a flag that will tell what object to create. 

### Describe alternatives you've considered

I have considered forking the respository and changing the locust UI myself but I think is a little overkill for the small change I am thinking of. I have also considered adding a screen before you get to the start menu with the toggles but I am not sure how to do this currently and I think it will not look the best.

TWoeste:
This could also be a solution to [this](https://github.com/locustio/locust/issues/1817#issue-945838819) problem.
You could use this UI change if you already had the different tasksets with your locust files

cyberw:
I dont think that problem (launching a completely different locust file) can be solved this way, but other than that I understand that being able to add UI fields might be useful. I’d be happy to look at a PR.

TWoeste:
@cyberw I am trying to push a PR and it is telling me I do not have access to push a PR? Any help? I get this error, 
```remote: Permission to locustio/locust.git denied to TWoeste.
fatal: unable to access 'https://github.com/locustio/locust.git/': The requested URL returned error: 403
```

cyberw:
You need to push to your fork and then open a PR from your repo to locust master branch.

PR #1823, opened 2021-07-23
Added blocks to start and edit

TWoeste:
This will allow people to add different options to start and edit a test with.

### DIFF
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -121,6 +121,10 @@ A working example of extending the web UI, complete with HTML and Javascript exa
 in the `examples directory <https://github.com/locustio/locust/tree/master/examples>`_ of the Locust 
 source code.
 
+Another working example of how to change the web UI to customize the start of your load tests can be found in the
+`custom_test_start directory <https://github.com/locustio/locust/tree/master/examples/custom_test_start>`_ of the
+Locust source code.
+
 
 
 Run a background greenlet
--- a/examples/custom_test_start/README.md
+++ b/examples/custom_test_start/README.md
@@ -1,0 +1,11 @@
+# Custom Start UI
+
+This Example shows how to use flasks Blueprints and templates to add custom items to the start of a loadtest.
+
+## Files:
+- `extend_test_start.py:` This file contains a sample locustfile with the necessary routes to get the data from the forms
+and share them to the requests. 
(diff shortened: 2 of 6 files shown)


TWoeste:
PR for https://github.com/locustio/locust/issues/1822#issue-951558060
@cyberw

cyberw:
I dont understand how this would work and it needs some documentation and ideally a test case.

TWoeste:
`{% block extended_start %}{% endblock extended_start %}` & `{% block extended_edit %}{% endblock extended_edit %}`

The two lines added are block statements that work the same way as the three already added block statements do. It will allow the user to add html code inside these blocks to extend the way their code starts and stops load tests. You can couple this with the scripts block at the bottom of `index.html` to handle the requests and in turn start the load tests. I am not sure of a good way to write tests for this addition. If you point me towards a place you may want more documentation I will. The current place that has the documentation for the currently added blocks is just a example project that uses them.

cyberw:
I guess you can add the example in a new folder under examples, like the other example. Explain what it does and link it from the same place (extending-locust.rst /https://docs.locust.io/en/stable/extending-locust.html#extending-web-ui)

I guess a test may be overly ambitious. You can skip that.

TWoeste:
@cyberw Just added an example to the `examples` file.
I had to change where the blocks statements were in the code because with how the extending framework is set up I need the user to click a button to submit the changes of the new forms to a new route in the backend. 

image

That is what the example UI looks like. Let me know if you have any other suggestions :smile:

cyberw:
Hmm. This is really a nice feature and I understand why you'd want it. But I think we need to make it nicer & more integrated into Locust.

Locust already has a (poorly documented) way to add custom arguments https://docs.locust.io/en/latest/api.html#locust.event.Events.init_command_line_parser (example usage: https://github.com/locustio/locust/blob/master/examples/add_command_line_argument.py)

If we just showed any custom arguments as input fields (all arguments minus "normal" arguments) it would be very flexible and much easier for the user to use (no need to know jinja to be able to add arguments)

It would also be nice to make an "advanced" tab that lists *all* the arguments (but that is just nice to have, and some arguments are not possible to change at test start time)

TWoeste:
Hmm ya that would probably be a more sustainable way to add custom starts params. I do not have the time now to make all of that so I will probably just proceed with my private fork. Thanks for you help. 
You can close this branch if you want to.

cyberw:
Ok!

PR #1825, opened 2021-07-25
Remove `user_classes_count` from heartbeat payload

mboutet:
`user_classes_count` is already part of the stats payload so having it in the heartbeat payload is redundant and makes the payload larger.

TODO:
* [ ] Add/update test to validate the content of the heartbeat payload


cyberw:
You should probably also remove these lines from runner.py (910-912):

                        user_classes_count = msg.data.get("user_classes_count")
                        if user_classes_count:
                            c.user_classes_count = user_classes_count

cyberw:
ok to merge? (you still have it marked as draft)

mboutet:
Yes

Issue #1827, opened 2021-07-28
How can we stablize the client-server protocol?


myzhan:
### Is your feature request related to a problem? Please describe.
I'm maintaining boomer and locust4j, which are written in other languages. After version check is added to the protocol, it's difficult for me to keep compatible with locust across all versions.

Related code: https://github.com/locustio/locust/blob/master/locust/runners.py#L871
Related issue: https://github.com/myzhan/boomer/issues/126

### Describe the solution you'd like
In locust 2.0.0, I'd like to use grpc instead of zeromq + msgpack. Then the protocol may be stablized in the protobuf file.

cyberw:
As long as you send *some* version number then there is only a warning in the master log (refusing pre 2.0 clients to connect is intentional, because they arent going to work anyway, because of the changes in spawning logic) 

I would be ok with changing the behaviour so that if clients send `-1` as their version number the master would just skip the version check. 

I might be ok with switching to grpc if it has similar performance, but that would be another breaking change..

myzhan:
I'm planning to deprecate the support for 1.x versions and move forward to 2.x, in both boomer and locust4j.

I'm suggesting gRPC because it's easier to use more languages. For example, the data in message may be treated as string, or dict. It's easy to do that in Python, but not in Go or Java.

cyberw:
gRPC makes sense (for the reasons you mentioned as well as more popularity), but it would probably mean another major release (3.0), and I'd want to give 2.0 lots of time "in the wild" before making another big change. At least 3-4 months.

myzhan:
Since we make a big change to message in 2.0, it's not easy to keep compatible with 1.x. I may add some workaround to work with both and wait for the change in 3.0.

cyberw:
Ok! In case I wasnt being clear, I'm not going to write the code to switch to grpc, someone actually needing that (like you) would have to build it, and it would have to be tested thoroughly.

And again to clarify, the client-server protocol shouldnt change that often, and I'll definitely strive to keep it backwards-compatible when it does.

myzhan:
@cyberw Hi, I'm still considering. I noticed that some users were using boomer for GRPC load test. If we introduce GRPC in locust, it may result in dependency conflict (e.g. , GRPC version)

PR #1830, opened 2021-07-29
Allow workers to bypass version check by sending -1 as version

cyberw:
Also add test case for version checking. Helps a little with #1827.


cyberw:
Should be a "safe" change, and I'd like all versions to have some possible workaround for non-python worker implementations (whose versions will never follow Locusts 1:1) to not give errors

Issue #1832, opened 2021-07-30
Display current testscript used


tyge68:
### Is your feature request related to a problem? Please describe.
Currently when we start locust ui , there are no hints on the page to find out what will be executed exactly.
### Describe the solution you'd like
It would be nice to at least display the current testscript filename that was used to start the command a.k.a -f option value.
If not at least the list of tasks that are going to be executed would be also nice to have.
### Describe alternatives you've considered
I have no idea of any possible alternative.

### Additional context
Distributed installation with many worker machines and a dedicated master running with UI.

cyberw:
Good idea! Shouldnt be too difficult to do either. Maybe have a go att doing it yourself?

tyge68:
@cyberw not sure how that could be done by myself unfortunately. I guess more experienced person on this project could add this quicker.

cyberw:
I’m sure it would be quicker, but it is unlikely to happen. Particularly the UI is ”you want it, you build it”, as I dont use it myself :)

tyge68:
is there information how to build locust locally ? or can we simply do pip install . maybe ? not sure, I am not really expert in this area :(

cyberw:
Clone the repo from github, cd into the folder and run 
pip3 install -e .

tyge68:
@cyberw  thanks a lot, I managed to do something, you can check my draft PR on #1833

PR #1833, opened 2021-07-30
#1832 Displaying locustfile and tasks ratio information on index.html

tyge68:
As discribed on #1832 , here is the suggested changes to allow displaying a bit more information about the current locustfile / tasks that will be executed in the Web UI.

About the look n feel any suggestions to make it looks nicer would be welcome.

### DIFF
--- a/docs/use-as-lib.rst
+++ b/docs/use-as-lib.rst
@@ -26,6 +26,17 @@ The :py:class:`Environment <locust.env.Environment>` instance's
     env.runner.start(5000, spawn_rate=20)
     env.runner.greenlet.join()
 
+It is also possible to bypass the dispatch and distribution logic, and manually control the spawned users:
+
+.. code-block:: python
+
+    new_users = env.runner.spawn_users({MyUserClass.__name__: 2})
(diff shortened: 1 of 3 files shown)


tyge68:
Here is how it would looks like
Screenshot 2021-07-30 at 18 22 27

cyberw:
Cool stuff. I'll give it a look some time next week (after we've released 2.0 & followed up on any issues that arise from it)

cyberw:
I see the need for this, but would like to hide it behind an advanced tab or something. Needs some tests too.

cyberw:
Github seems to have issues with the tests. Try merging master.

cyberw:
Ignore the docker fail in the build, that is probably my bad.

tyge68:
@cyberw  do you know if tests can be retriggered (without any commits on my side) ?

cyberw:
> @cyberw do you know if tests can be retriggered (without any commits on my side) ?

Done.

cyberw:
@tyge68 If you fix this soon it might be in time for 2.2 release :)

PR #1841, opened 2021-08-05
Show custom arguments in web ui and forward them to worker

### LINKED ISSUE #1822: locust should add extending blocks to index.html to change the start test and edit test options

cyberw:
This is WIP. I'm not sure showing all custom arguments in the UI is the best solution (and it would also be nice to be able to show some of the default ones). What do you think @TWoeste

I tried to make that "advanced" tab thingy but it turned out to be too much work :)

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -7,7 +7,7 @@ Configuration
 Command Line Options
 -----------------------------
 
-The most straight forward way to Configure how Locust is run is through command line arguments. 
+Locust is configured mainly through command line arguments.
 
 .. code-block:: console
 
@@ -76,6 +76,11 @@ All available configuration options
 Here's a table of all the available configuration options, and their corresponding Environment and config file keys:
 
 .. include:: config-options.rst
+
+Custom arguments
+----------------
+
+See :ref:`parametrizing-locustfiles`
 
 Customization of statistics settings
 ------------------------------------
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -152,6 +152,26 @@ For example, you can monitor the fail ratio of your test and stop the run if it 
             gevent.spawn(checker, environment)
 
 
+Parametrizing locustfiles
+=========================
+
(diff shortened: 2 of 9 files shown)


TWoeste:
I do think this is a cleaner approach to adding custom start UI. I do think that the whole point to add these arguments would be to access them somehow in your tasks to change how they act (so on the worker runner for distributed tests). For instance if you wanted to specify the username and password of the requests it would be best if you could access these in the worker to set them in the header of the requests. Another downside to this approach is all of your custom fields will be text boxes so its hard to predict what users will put there. In the end I think its a good approach to solve my issue. In my local branch I solved the problem with not being able to access the variables in the worker branch by sending a message from the master runner to the worker runner as suggested [here](https://stackoverflow.com/a/68657455/16236463). @cyberw

TWoeste:
Would the parsed_options be populated with the correct data on the worker instance of env?

cyberw:
> I do think that the whole point to add these arguments would be to access them somehow in your tasks to change how they act (so on the worker runner for distributed tests). For instance if you wanted to specify the username and password of the requests it would be best if you could access these in the worker to set them in the header of the requests.

This approach makes them accessible in Users by `self.environment.parsed_options.my_argument`

> Would the parsed_options be populated with the correct data on the worker instance of env?

Not automatically no, that would also have to be addressed (hence the draft status :)

TWoeste:
This all looks amazing to me :). I think it will become one of the more uses locust feature when implements. Thanks!

cyberw:
@TWoeste Can you try this out before I merge it?

TWoeste (approved):
LGTM, Tested it with the example file!

Issue #1838, opened 2021-08-04
Add flag to run with uniform weights


shekar-stripe:
Hi! I work at [Stripe](https://stripe.com/). Our reliability infrastructure team uses locust to load test. Oftentimes there are tasks that occur with extremely low probability, so it is infeasible to simply rely on chance for a task to be run when testing a locustfile locally. Instead, engineers switch all task weights to 1, then switch them back once they're confident their locustfile is correct. This is proving untenable for large locustfiles.

### Is your feature request related to a problem? Please describe.
Switching weights often takes place across multiple files (under many layers of indirection). As a result, engineers have to remember past weights, sift through multiple files, etc. simply to test their implementation. 

### Describe the solution you'd like
This process would be made much easier if there was a flag that simply overrode task weights to an even distribution, or even guaranteed that each task was run sequentially. Somewhat like a "test" mode. I'm happy to open a PR for this.

### Describe alternatives you've considered
We've considered two alternatives:
- write a syntax tree parser that changes weights in the python files to 1 
- modify locust internals in our Docker container

Of these options, the team thinks that adding a flag would be the cleanest solution.

mboutet:
Have you considered using an environment variable to set all weights to 1?

e.g.
```python
class MyUser(FastHttpUser):
    @task(1 if os.getenv("SET_ALL_TASKS_WEIGHT_TO_1", "false") == "true" else 3)
    def my_task(self):
        ...
```

shekar-stripe:
While this could work, my main concern would be that we assign task weights in a few different ways depending on the load test being performed. Some are like your example, some are assigned in a dictionary, some are assigned in other files through indirection and downstream computation. As a result, I think that using an env var in a few different ways across hundreds of tasks is non ideal. What do you think?

mboutet:
I see.

Would using [tags](https://docs.locust.io/en/stable/writing-a-locustfile.html#tag-decorator) be an option? When only a few tasks have to be tested to validate that they work as expected, simply add tags on them and then run locust with the `--tags` flag to select them.

Would that work?

Otherwise, the logic you want probably needs to be implemented in the `locust.user.task.get_tasks_from_base_classes` function. I'm not a maintainer, so you'll have to wait for @cyberw's opinion.

cyberw:
[...]

PR #1842, opened 2021-08-06
Add equal weights flag

shekar-stripe:
Issue: https://github.com/locustio/locust/issues/1838

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -418,10 +418,10 @@ def setup_parser_arguments(parser):
         env_var="LOCUST_STOP_TIMEOUT",
     )
     other_group.add_argument(
-        "--same-task-weights",
+        "--same-weights",
         action="store_true",
         default=False,
-        dest="same_task_weights",
+        dest="same_weights",
         help="Use uniformly distributed task weights, overriding the weights specified in the locustfile.",
     )
 
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -417,6 +417,13 @@ def setup_parser_arguments(parser):
         help="Number of seconds to wait for a simulated user to complete any executing task before exiting. Default is to terminate immediately. This parameter only needs to be specified for the master process when running Locust distributed.",
         env_var="LOCUST_STOP_TIMEOUT",
     )
+    other_group.add_argument(
+        "--use-uniform-task-weights",
+        action="store_true",
+        default=False,
+        dest="use_uniform_task_weights",
+        help="Use uniformly distributed task weights, overriding the weights specified in the locustfile.",
+    )
 
     user_classes_group = parser.add_argument_group("User classes")
     user_classes_group.add_argument(
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -418,11 +418,11 @@ def setup_parser_arguments(parser):
         env_var="LOCUST_STOP_TIMEOUT",
     )
     other_group.add_argument(
-        "--same-weights",
+        "--equal-weights",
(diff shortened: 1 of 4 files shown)


cyberw:
👍👍 as for wording, I kind of prefer ”uniform” to ”same” but. 

And as you are (correctly) changing weights for Users and not just tasks, I think the parameter name is kind of confusing. 

Does this also work for TaskSets?

shekar-stripe:
I also prefer uniform, the reason I had to change it to same was because argparse thought a flag with --uniform-* was too close to --user (and could get confusing with the --u shorthand).

I can see why you think the parameter name is confusing, I can definitely change it. Something like `same-weights`? (Is there some way around the argparse error? Should I want to get around it?) 

Yes, I believe I tested TaskSets as well. Please let me know if there's something else I should test further. I'm new to locust (and Stripe, just joined less than a month ago) so it's definitely a possibility I missed something.

cyberw:
> I also prefer uniform, the reason I had to change it to same was because argparse thought a flag with --uniform-* was too close to --user (and could get confusing with the --u shorthand).
> 
> I can see why you think the parameter name is confusing, I can definitely change it. Something like `same-weights`? (Is there some way around the argparse error? Should I want to get around it?)

Not sure whether there is a way, but go ahead and try. The risk of confusion should be small. —uniform-weights would be ideal. 

> Yes, I believe I tested TaskSets as well. Please let me know if there's something else I should test further. I'm new to locust (and Stripe, just joined less than a month ago) so it's definitely a possibility I missed something.

I think the one thing you maybe havent tested is tasks specified as `tasks = {my_task: 3, another_task: 1}`?

https://docs.locust.io/en/stable/writing-a-locustfile.html#id2

or does it work for those too?

shekar-stripe:
Unfortunately, I don't think it's possible without enabling a flag that disables abbreviations which would introduce breaking changes: https://realpython.com/command-line-interfaces-python-argparse/
I've kept the internal calls `uniform` and just the flag `same`. 

I didn't test user tasks added as a dict, that's updated now. LMK if this is good to merge.

cyberw:
Cool! I think you can ignore that failing test in the latest run. I’m confused by the argument parsing issue though. How come —stop-timeout doesnt collide with --show-task-ratio for example?

cyberw:
Maybe —equal-weights works?

shekar-stripe:
I'm honestly not sure why the argument parsing issue is happening. I tried switching the flags to mirror the `--stop-timeout` and `--show-task-ratio` example, but to no avail. I also tried moving the new argument into the same arg group as users, but that also threw an error. If you have any other ideas, please let me know! 

I changed the flag to --equal-weights.

cyberw:
Does the name —equal-weights work then? If so, update the method name etc as well.

shekar-stripe:
Done!

cyberw:
Cool!

shekar-stripe:
Hey @cyberw, when do you think this change will be released as part of an official version? Per my company's security policy, we are only allowed to use official releases rather than the "bleeding edge" as mentioned in the docs...

cyberw:
You can use prereleases, right? I’m working on adding a git workflow that builds and pushes a prerelease version (something like 2.1.0devX) on every PR merge.

cyberw:
Ok, that took me longer than I had hoped :)

Anyways, you should now be able to do
`pip install -U --pre locust`
or (to get the latest one just now)
`pip install locust==2.1.1.dev53`

shekar-stripe:
Awesome, thank you! Really appreciate it.

Issue #1837, opened 2021-08-03
Stats History CSV is always growing


SoonerLR:
When running in distributed mode and in containers, one master and X workers, even when the test is stopped, the stats_history.csv file is always being written to if the container is up and running. It seems like there should also be at check to see if it is not a MasterRunner. MasterRunner stats probably should not be written to the history (the entry is all 0s so it does not make sense anyway). There might be a better way to prevent the stats history from always growing. 
 
See [Stats History writer here](https://github.com/locustio/locust/blob/6cb81973275f41875fbfdc20ab6612a8ab6ea869/locust/stats.py#L792)

cyberw:
I dont quite follow. I see the issue with master writing all 0s even before a test has started/after it has finished, and agree it should be fixed (by someone who is not me, because I do too much already :)

But writing stats for master *when it is running* is not a bug, is it?

SoonerLR:
No, the master writing stats when a test is running is fine. The bug is that nothing should be writing to the CSV when there is no test running.

SoonerLR:
I think I have a fix for this issue. How do I get the code to you?

Thanks,
Lance
GitHub user: SoonerLR

mboutet:
1. Fork the repo
2. Make your changes in your fork
3. Open a pull request from your fork to locust master branch

You can read this [doc](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork) for more info.

JonathanB-96:
I also have this issue and it's not just _stats_histroy.csv_ that keeps updating after the test is stopped while master is still running but also _result_exceptions.csv_, _results_failures.csv_ and also _results_stats.csv_. Once the test is stopped from the web UI after a minute or so the _results_stats.csv_ will return 0 values.

PR #1844, opened 2021-08-06
Fix for issue #1837 - Do not write rows to stats_history.csv when a …

SoonerLR:
…test is not running

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -789,15 +789,17 @@ def stats_history(runner):
         stats = runner.stats
         if not stats.total.use_response_times_cache:
             break
-        r = {
-            "time": datetime.datetime.now().strftime("%H:%M:%S"),
-            "current_rps": stats.total.current_rps or 0,
-            "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
-            "response_time_percentile_95": stats.total.get_current_response_time_percentile(0.95) or 0,
-            "response_time_percentile_50": stats.total.get_current_response_time_percentile(0.5) or 0,
-            "user_count": runner.user_count or 0,
-        }
-        stats.history.append(r)
+        # Do not try to write rows with all 0s in stats_history.csv when a test is not running
+        if (runner.state != "stopped" and runner.state != "ready"):
+            r = {
+                "time": datetime.datetime.now().strftime("%H:%M:%S"),
+                "current_rps": stats.total.current_rps or 0,
+                "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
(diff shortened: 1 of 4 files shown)


locust/stats.py
```diff
@@ -993,6 +992,10 @@ def _stats_history_data_rows(self, csv_writer, now):
         Note that this method differs from the other methods as it appends time-stamped data to the file, whereas the other methods overwrites the data.
         """
 
+        """ Do not try to write rows with all 0s in stats_history.csv when a test is not running """
```
mboutet, line 995:
```suggestion
        # Do not try to write rows with all 0s in stats_history.csv when a test is not running
```

cyberw:
Github seems to have issues with the tests. Try merging master to this branch.

locust/stats.py
```diff
@@ -995,6 +995,10 @@ def _stats_history_data_rows(self, csv_writer, now):
         Note that this method differs from the other methods as it appends time-stamped data to the file, whereas the other methods overwrites the data.
         """
 
+        # Do not try to write rows with all 0s in stats_history.csv when a test is not running
+        if (self.environment.runner.state == "stopped" or self.environment.runner.state == "ready"):
+            return
```
cyberw, line 1000: I wonder. What if for some reason the runner state is stopped, but for some reason there is still data coming in? Maybe it is safer to not write when there are only zeros?

cyberw: Like if there is a timining issue...

SoonerLR: I am also beginning to think that the tests are failing because they expect lines to be written to the CSV and that is exactly what this PR is addressing.

cyberw: I dont think so. Looking at for example TestWebUIFullHistory.test_request_stats_full_history_csv() it expects *valid* data in some of those rows that are now missing, so it seems like this really broke something...

cyberw: 👍 Tbh, I dont understand that test case. What does it even do, and does it still make sense?

cyberw: Ah, I understand the test now. Can you modify it so that it actually fires a request event? Testing that there are zero entries in the master stats isnt the same as testing that there is one :)

cyberw:
@SoonerLR If you fix it soon it might be in time for 2.2 release :)

SoonerLR:
I know. I'm working on it, but can't seem to get it to write an entry. LOL! :)

cyberw:
@SoonerLR Any luck fixing the fix? :)

cyberw:
Not off the top of my head... I think it is a real issue, but it could just be the unit test, sorry, dont have much time to spend on this...

cyberw:
No progress I guess? I'll probably have to close this soon...

PR #1851, opened 2021-08-16
Use UTC time for server stats history, localize times on the client

### LINKED ISSUE #1835: time display in live charts switches to the local time upon refresh
### Describe the bug

There is a bug in the display of the time in live graphs (that thankfully doesnt seem to appear when you go to “Download Report”), where the hour (time zone?) switches. It will show the time in the server's own timezone... and then at some point in the graph (depending on when you refreshed your browser to load the web UI), the time display will switch to the local timezone of your browser system.

### Expected behavior

The time display should always be consistent. Either take the cue from the server (this is the default, and is fantastic)... or from the browser system.

### Actual behavior

See "Describe the bug" section.

Here is a screenshot of the problem, showing how the hour switches from 11 (the server is GMT) to 19 (my system is GMT+8):
Screenshot 2021-07-30 at 19 14 22

### Steps to reproduce

Set up a server in a different time zone from your local system. Load the web UI and start a run. Refresh your browser, and click "Charts". You will notice that the last time as shown in the charts will *always* be in your local timezone, and as more points get added to the graphs, all of those will be in your local time.

This behaviour is repeatable. Refresh your browser again, click on "Charts" again... and you will see again that all of the time values will reset to the server's time zone save for the last one.

### Environment

- OS: Ubuntu 20.04
- Python version: 3.8.10
- Locust version: 1.6.0

obradovichv:
Server-side stats history time sourcing changed from server time(time zone dependent) to UTC, aligning it with other server-side time values sourced from `time.time()`. This enables the same frame of reference for all clients by knowing it is UTC. Localization is now performed client-side on server-rendered dates and times.  

Localization of the `report.html` start and end datetimes intentionally formats the value according to the client's default locale to keep with the localization theme of this change.

The date strings are manually parsed instead of using `Date.parse()` as recommended by MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parsed
`"It is not recommended to use Date.parse as until ES5, parsing of strings was entirely implementation dependent. There are still many differences in how different hosts parse date strings, therefore date strings should be manually parsed (a library can help if many different formats are to be accommodated)."`

Fixes #1835

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -791,7 +791,7 @@ def stats_history(runner):
             break
         if runner.state != "stopped":
             r = {
-                "time": datetime.datetime.now().strftime("%H:%M:%S"),
+                "time": datetime.datetime.utcnow().strftime("%H:%M:%S"),
                 "current_rps": stats.total.current_rps or 0,
                 "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
                 "response_time_percentile_95": stats.total.get_current_response_time_percentile(0.95) or 0,
--- a/locust/templates/report.html
+++ b/locust/templates/report.html
@@ -44,7 +44,7 @@
             {% if show_download_link %}
                 <p class="download"><a href="?download=1">Download the Report</a></p>
             {% endif %}
-            <p>During: <span>{{ start_time }} - {{ end_time }}</span></p>
+            <p>During: <span class="l10n datetime">{{ start_time }}</span> - <span class="l10n datetime">{{ end_time }}</span></p>
             <p>Target Host: <span>{{ host }}</span></p>
         </div>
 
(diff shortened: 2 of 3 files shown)


cyberw:
I think supporting pre ES5 browsers doesnt really make it to our priority list (10 year old browsers are likely to have all kinds of other issues anyway), and I would prioritize keeping the code simple. Other than that it looks good!

obradovichv:
Agreed, the spread operator would likely invalidate any 10 year old browser anyway :laughing: 
Updated PR accordingly to use `Date.parse()` to parsing the datetime instead of manually parsing the value.

I have left `stats_data.html` manual parsing unchanged because it uses only the time component and would need an ISO8601-like date format to work with `Date.parse()`, perhaps the {current date} or {unix epoch} would be appropriate "fake" date components to pass into `Date.parse()` but it does not feel like a simplification of `"HH:MM".split(":")` by comparison. I can update to this if preferred.

If a future feature stores stat history with the full datetime for the time series, then the client-side localization can be updated to `Date.parse()` without any complicated date injections.

cyberw:
Thanks!

Do you by any chance have time to look into another thing related to graphing? I'd like to not have a line for response times at all until requests have actually been reported (the first part of the line in this screen shot)

image

obradovichv:
Sure, I'll take a look.
Created new issue to track it: #1852

cyberw:
Awesome, thanks!

PR #1858, opened 2021-08-19
Alternative grouping

Serhiy1:
- Add an alternative way to group requests together, do deal with situations where passing in a name parameter is not possible
- Add examples using the alternative grouping approaches.
- Added an example of session patching when working with SDK's

### DIFF
--- a/docs/testing-other-systems.rst
+++ b/docs/testing-other-systems.rst
@@ -1,7 +1,7 @@
 .. _testing-other-systems:
 
 ========================
-Testing other systems 
+Testing non-HTTP systems
 ========================
 
 Locust only comes with built-in support for HTTP/HTTPS but it can be extended to load test almost any system. You do this by writing a custom client that triggers :py:attr:`request <locust.event.Events.request>`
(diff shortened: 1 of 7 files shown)


docs/testing-other-systems.rst
```diff
@@ -1,7 +1,7 @@
 .. _testing-other-systems:
 
 ========================
-Testing non-HTTP systems
```
cyberw, line 4: Maybe split this into two sections? ”Testing non-HTTP systems” and ”Testing requests-based SDKs”?

cyberw:
Cool stuff!

* it needs a unit test
* it needs formatting fixed (just run Black)
* I think the field name should be ”request_name” and not ”group_name” Grouping is just one use case for naming - you might do it to clarify or differentiate requests with the same URL. For the context version maybe rename() instead of group()?

Serhiy1:
> Cool stuff!
> 
> * it needs a unit test
> * it needs formatting fixed (just run Black)
> * I think the field name should be ”request_name” and not ”group_name” Grouping is just one use case for naming - you might do it to clarify or differentiate requests with the same URL. For the context version maybe rename() instead of group()? Or maybe name()? Not as sure about thay one.

cheers for the feedback, I will do all the items listed out.

Serhiy1:
@cyberw All the suggested changes are done though I'm not 100% happy with the function name for the context manager.

I think maybe instead of `name_request` changing it to `request_name_context` will keep it's intention clear and avoid any potential mix-ups with the field name.

cyberw:
Code looks great now! I think you've duplicated the documentation? Also, the new page needs to be added to index.rst to become visible in the left hand menu in the docs.

Maybe **re**name_request is more clear?

Serhiy1:
- Rats, you're right about the duplicated documentation. 
- I think I've already added the new page to index.rst? Unless I mis-understood something.  
- And yeah I like `rename_request` more

cyberw:
Looks great!

cyberw:
Two small issues, now that I see the documentation first hand (https://docs.locust.io/en/latest/)

The title is weird, and there is an extra heading that you should probably remove.

image

Serhiy1:
@cyberw agreed, that doesn't look great. 

I've fixed that, and slightly changed the wording, shall I make another PR?

image

cyberw:
Please do! Looks good! I think SDKs is better then SDK's, but that is a detail :)

PR #1860, opened 2021-08-19
Handle user classes with weight = 0

mboutet:
WIP

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -106,6 +106,8 @@ class Environment:
 
         self._filter_tasks_by_tags()
 
+        self._remove_user_classes_with_weight_zero()
+
         # Validate there's no class with the same name but in different modules
         if len(set(user_class.__name__ for user_class in self.user_classes)) != len(self.user_classes):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


locust/env.py
```diff
@@ -216,6 +218,19 @@ def _filter_tasks_by_tags(self):
         for user_class in self.user_classes:
             filter_tasks_by_tags(user_class, self.tags, self.exclude_tags)
 
+    def _remove_user_classes_with_weight_zero(self):
+        """
+        Remove user classes having a weight of zero.
+        """
+        if len(self.user_classes) == 0:
+            # Preserve previous behaviour that allowed no user classes to be specified.
```
cyberw, line 226: Was it ok to specify no user classes before? I dont know when that should not be considered an error?

mboutet: I agree that this should make locust fail fast because it is invalid. The reason I put it there is that otherwise more than 20 tests are failing. If I remove this condition, I'll have to update all those tests so that they contain fake user class.

cyberw: Ok! Yea I guess we still fail in main.py on line 165 if you actually launch locust proper without any classes. Ready to merge, or is there something else that is WIP?

mboutet: Ready

PR #1861, opened 2021-08-21
Improve some of the doc issues that were missed from the previous PR

Serhiy1:


cyberw:
Ignore the build error, that is unrelated...

PR #1866, opened 2021-08-31
Locust Automation

Aliya2817:
This will generate the locust script automatically only needed to provide the curl using that script it will do performance testing and provide result in csv .


cyberw:
I think you selected the wrong target branch (1.6.x instead of master) :)

PR #1872, opened 2021-09-03
Ability to inject custom html elements in the `head` element

fabito:
# Problem

There is no (SSR) way to inject custom elements in the `head` when extending the `index.html` jinja2 template

# Solution

Add a new jinja2 block `extend_head` to the `index.html` template to improve customizability. For example:

```jinja2

{% extends "index.html" %}

{% block extended_head %}
<link rel="stylesheet" type="text/css" href="./extend/static/extend.css?v={{ version }}" media="screen">
{% endblock extended_head %}

```

# Test

This is how the `extend_web_ui` example looks like with the new table styling:

Screenshot from 2021-09-03 18-44-03

### DIFF
--- a/examples/extend_web_ui/static/custom-stats-table.css
+++ b/examples/extend_web_ui/static/custom-stats-table.css
@@ -1,3 +1,7 @@
+/*
+ Customises stats tables color, alignment. Adds row highlighting.
+ Just a simple extension example. The sky's the limit here :-)
+**/
 
 table.stats {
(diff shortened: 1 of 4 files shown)


cyberw:
LGTM. One small thing, could you maybe add a comment describing what the extra css actually does? Or perhaps just name the file something more to the point instead of `extend.css`?

fabito:
Sorry I couldn't think of a more interesting example...

PR #1873, opened 2021-09-03
Fix Regression in Full History CSV Percentiles

TaylorSMarks:
The docstring for _stats_history_data_rows says it should be outputting `*current* stats` (emphasis not added by me). In the original commit from December 3rd, 2019, it was properly using the `get_current_response_time_percentile` function.

On August 18th, 2020, the code was all reformatted and it looks like a regression was accidentally introduced when they switched over to using `get_response_time_percentile` instead.

This PR simply fixes that regression, switching back to using `get_current_response_time_percentile`.

The bug is easiest to observe if you have a spikey/bursty test shape. During bursts, latency increases. During quieter times, latency goes back down. This is properly reflected in the Locust Web UI's graph tab, but the full history CSV file doesn't show that - instead there's only a slight decline in latency between bursts, because it's showing the 95th percentile since the test began rather than just for the past 10 seconds.

I suspect the bug went just over a year without being noticed because it's unusual to either (1) use a custom load shape for a test and (2) to closely examine the full history CSV file.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -843,11 +843,15 @@ class StatsCSV:
         ]
 
     def _percentile_fields(self, stats_entry, use_current = False):
+        print('In _percentile_fields')
         if not stats_entry.num_requests:
+            print('N/A')
             return self.percentiles_na
         elif use_current:
+            print('Giving current response times.')
             return [int(stats_entry.get_current_response_time_percentile(x) or 0) for x in self.percentiles_to_report]
         else:
+            print('Giving total response times.')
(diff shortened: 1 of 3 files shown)


cyberw:
Interesting. Can you add a test case? (to make sure it doesnt regress again, and to further explain the issue)

TaylorSMarks:
@cyberw - I’ve fixed the formatting issues that flake8 was complaining about and updated an existing unit test to cover this.

With my changes, the unit test passes. Without my changes, the unit tests will fail because the 95th percentile will never decrease.

cyberw:
Thanks!

PR #1875, opened 2021-09-04
docs: add homebrew install instruction

chenrui333:
relates to:
- https://github.com/Homebrew/homebrew-core/pull/84586

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -11,6 +11,12 @@ Install Locust:
 
     $ pip3 install locust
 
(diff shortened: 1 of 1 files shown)


cyberw:
What does a brew install really add over a pip install? I’m also a bit sceptical because it pins the locust version.

cyberw:
I guess some people might prefer a brew-installation but to me it doesnt really add any value. It makes more sense to install python using brew and then installing locust the regular way.

Issue #1884, opened 2021-09-14
User distribution should happen when new workers comes in


tyge68:
Currently when more workers are added (dynamically based in the workers CPU load on kubernetes), if the swarming is over, then new workers are detected ready, but there is redistribution of the users on the new workers to take some load on the other workers.

Only a manually re-swarming can force redistribute the users among the workers.

### Describe the bug
The new workers are ready but not used

### Expected behavior
If a new worker is added, it should automatically redistribute the load of users on the workers.

### Actual behavior
New worker is ready, but don't get any user / tasks to execute. Other workers are still using all the CPU.

### Steps to reproduce
Setup some workers , and start a load test, wait until the target number of user is reached. Add a new worker to the pool, check the state is ready, but no user is assigned to it, until a "re-swarm" is manually triggered.

### Environment

- OS:
- Python version: 3.9
- Locust version: 2.2.1
- Locust command line that you ran:
- Locust file contents (anonymized if necessary):

mboutet:
This is what's currently implemented in https://github.com/locustio/locust/pull/1809. I only implemented the re-distribution while ramping-up/ramping-down. As described by @tyge68, this means that if workers are added or removed while the load test is at a plateau (i.e. steady state), then the redistribution won't take place until the next ramp-up/down.

I'm in the same boat as @tyge68 and I know some others that would like for this feature to be implemented. I think that in order to satisfy both types of users (those who want it and others who don't), this should be configurable as a CLI flag and/or environment variable and disabled by default.

From an implementation standpoint, I already have a design idea which consists in having a background greenlet similar to `locust.runners.MasterRunner.heartbeat_worker` which periodically (configurable and defaulted to something around ~10s) triggers a re-distribution if needed and if a ramp-up/down is not already in progress.

tyge68:
@cyberw  @mboutet Note I have created a PR, while checking the code , I see there was already some tests related to this issue, I noticed while retesting locally that it rebalance once for the first added worker (after ramp up), but any new worker added again after that are "ignored" and so stay then in "ready" state.

With PR based on what @mboutet  suggested, that would resolve it definitely, but it could be that another solution could solve it too in that part of the code (1).

Also it is interesting that there is indeed a test already existing on this topic (2) in the test_runner.py.  Eventually it would be nice to check that adding a few more  workers a few milliseconds later still rebalance as I observed.

(1) https://github.com/locustio/locust/blob/master/locust/runners.py#L861-L875
[...]

Issue #1883, opened 2021-09-14
MasterRunner target_user_count no longer set for test_start event listeners


jrweldon:
### Describe the bug
I was following the `custom_messages.py` example and noticed that `environment.runner.target_user_count` is no longer resolving to the number of users being created for the test. This functionality appears to have been broken between tags `2.0.0b1` and `2.0.0b2`.

### Expected behavior
The MasterRunner's `target_user_count` should resolve to the number of users specified in the WebUI for the test.
The `custom_messages.py` example for a test with 10 users and 2 workers should log a message from each worker indicating that they received 5 users.
```
master_1  | [2021-09-14 02:48:02,550] d4b6940eebd0/INFO/locust.runners: Sending spawn jobs of 10 users at 1.00 spawn rate to 2 ready clients
master_1  | Thanks for the 5 users!
master_1  | Thanks for the 5 users!
worker_1  | User4
worker_2  | User9
worker_1  | User3
...
```

### Actual behavior
The MasterRunner's `target_user_count` is 0, causing numerous exceptions in the `custom_messages.py` example.
```
master_1  | [2021-09-14 03:01:33,449] 0134f51ea2b2/INFO/locust.runners: Sending spawn jobs of 10 users at 1.00 spawn rate to 2 ready clients
master_1  | Thanks for the 0 users!
master_1  | Thanks for the 0 users!
worker_2  | Traceback (most recent call last):
worker_2  |   File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run
worker_2  |   File "/usr/local/lib/python3.8/site-packages/locust/runners.py", line 1138, in <lambda>
worker_2  |     self.spawning_greenlet = self.greenlet.spawn(lambda: self.start_worker(job["user_classes_count"]))
worker_2  |   File "/usr/local/lib/python3.8/site-packages/locust/runners.py", line 1075, in start_worker
worker_2  |     self.spawn_users(user_classes_spawn_count)
worker_2  |   File "/usr/local/lib/python3.8/site-packages/locust/runners.py", line 221, in spawn_users
worker_2  |     new_users += spawn(user_class, spawn_count)
worker_2  |   File "/usr/local/lib/python3.8/site-packages/locust/runners.py", line 210, in spawn
worker_2  |     new_user = self.user_classes_by_name[user_class](self.environment)
worker_2  |   File "/mnt/locust/custom_messages.py", line 55, in __init__
worker_2  |     self.username = usernames.pop()
[...]

PR #1886, opened 2021-09-15
#1884 User distribution should happen when new workers comes in

tyge68:
- Adding a new argument to enable automatic rebalancing for new workers (after ramp up completed)
- Log message added when rebalancing is triggered

To be checked how to test such code change.

cc @cyberw  @mboutet

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -464,6 +464,13 @@ def setup_parser_arguments(parser):
         dest="equal_weights",
         help="Use equally distributed task weights, overriding the weights specified in the locustfile.",
     )
+    other_group.add_argument(
+        "--enable-rebalancing",
+        action="store_true",
+        default=False,
+        dest="enable_rebalancing",
+        help="Allow to automatically rebalance users if new workers are added or removed during a test run.",
+    )
 
     user_classes_group = parser.add_argument_group("User classes")
     user_classes_group.add_argument(
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -601,6 +601,7 @@ class MasterRunner(DistributedRunner):
         self.master_bind_host = master_bind_host
         self.master_bind_port = master_bind_port
         self.spawn_rate: float = 0
+        self.spawning_completed = False
 
         self.clients = WorkerNodes()
         try:
@@ -637,6 +638,9 @@ class MasterRunner(DistributedRunner):
 
         self.environment.events.quitting.add_listener(on_quitting)
 
+    def rebalancing_enabled(self) -> bool:
+        return self.environment.parsed_options and self.environment.parsed_options.enable_rebalancing
+
     @property
     def user_count(self) -> int:
         return sum(c.user_count for c in self.clients.values())
@@ -649,6 +653,7 @@ class MasterRunner(DistributedRunner):
         return warning_emitted
 
     def start(self, user_count: int, spawn_rate: float, **kwargs) -> None:
+        self.spawning_completed = False
         num_workers = len(self.clients.ready) + len(self.clients.running) + len(self.clients.spawning)
         if not num_workers:
             logger.warning(
@@ -744,6 +749,7 @@ class MasterRunner(DistributedRunner):
             timeout.cancel()
 
         self.environment.events.spawning_complete.fire(user_count=sum(self.target_user_classes_count.values()))
+        self.spawning_completed = True
 
         logger.info("All users spawned: %s" % _format_user_classes_count_for_log(self.reported_user_classes_count))
 
@@ -831,7 +837,8 @@ class MasterRunner(DistributedRunner):
                     client.user_classes_count = {}
                     if self._users_dispatcher is not None:
                         self._users_dispatcher.remove_worker(client)
-                        # TODO: If status is `STATE_RUNNING`, call self.start()
+                        if self.rebalancing_enabled() and self.state == STATE_RUNNING and self.spawning_completed:
+                            self.start(self.target_user_count, self.spawn_rate)
                     if self.worker_count <= 0:
                         logger.info("The last worker went missing, stopping test.")
                         self.stop()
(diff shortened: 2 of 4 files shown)


cyberw:
I only had a quick look, but I dont see anything bad about this.

report.html
```diff
@@ -0,0 +1,353 @@
+<!DOCTYPE html>
```
mboutet, line 1: Was this file committed inadvertently?

mboutet:
I know I said yesterday that the background task was my preferred solution, but looking back at the code, I remembered I started to implement it in a more efficient fashion.

It is more efficient because rebalancing only takes place whenever a new worker joins or if a worker goes missing instead of always rebalancing even if there's nothing to rebalance. It is also faster to react because it acts immediately upon a worker joining or missing.

The logic for a worker joining goes here:
https://github.com/locustio/locust/blob/bb7734404be4c44144022fffb0152fcb57606e66/locust/runners.py#L880-L883

The logic for a worker that goes missing goes here:
https://github.com/locustio/locust/blob/bb7734404be4c44144022fffb0152fcb57606e66/locust/runners.py#L834

mboutet:
Also note that because of a regression I introduced in https://github.com/locustio/locust/pull/1809 and reported in https://github.com/locustio/locust/issues/1883, the `target_user_count` on the master is currently not updated until the end of the ramp-up/down. This should be fixed in an upcoming PR. I mention it because it means it could potentially cause a rebalancing to be messed up.

tyge68:
@mboutet  indeed it would make more sense to implement it in the event, currently I have added a new integration test, it would currently fail without my changes, but I will try to fix it in the event code instead.

mboutet:
I saw in one of your latest commit that you removed the `--enable-rebalancing` flag. However, I believe it is still needed because some people might not want the rebalancing behaviour.

tyge68:
@mboutet  ok sure I will add it back

tyge68:
@cyberw  @mboutet  I think it should be ready for final review now, let me know if the last changes and the added test is good enough.

mboutet:
Can you address the comment I left: https://github.com/locustio/locust/pull/1886#discussion_r709174516

locust/runners.py
```diff
@@ -637,6 +638,9 @@ def on_quitting(environment, **kw):
 
         self.environment.events.quitting.add_listener(on_quitting)
 
+    def rebalancingEnabled(self):
```
mboutet, line 641:
```suggestion
    @property
    def rebalancing_enabled(self) -> bool:
```

locust/runners.py
```diff
@@ -877,10 +883,8 @@ def client_listener(self):
                     "Client %r reported as ready. Currently %i clients ready to swarm."
                     % (worker_node_id, len(self.clients.ready + self.clients.running + self.clients.spawning))
                 )
-                # if self.state == STATE_RUNNING or self.state == STATE_SPAWNING:
-                #     # TODO: Necessary now that UsersDispatcher handles that?
-                #     # balance the load distribution when new client joins
-                #     self.start(self.target_user_count, self.spawn_rate)
+                if self.rebalancingEnabled() and self.state == STATE_RUNNING and self.spawning_completed:
```
mboutet, line 886:
```suggestion
                if self.rebalancing_enabled and self.state == STATE_RUNNING and self.spawning_completed:
```

locust/argument_parser.py
```diff
@@ -464,6 +464,13 @@ def setup_parser_arguments(parser):
         dest="equal_weights",
         help="Use equally distributed task weights, overriding the weights specified in the locustfile.",
     )
+    other_group.add_argument(
+        "--enable-rebalancing",
+        action="store_true",
+        default=False,
+        dest="enable_rebalancing",
+        help="Allow to automatically rebalance users if new workers are added after ramp up completed.",
```
mboutet, line 472:
```suggestion
        help="Allow to automatically rebalance users if new workers are added or removed during a test run.",
```

mboutet:
@tyge68, the re-balancing also needs to be implemented when a worker goes missing: 
https://github.com/locustio/locust/blob/bb7734404be4c44144022fffb0152fcb57606e66/locust/runners.py#L834

Also, I'll have to revisit my code, because according to this:
https://github.com/locustio/locust/blob/a278fb3992093b1dc25cb856757dec6dfc97f53a/locust/runners.py#L875
I already implemented the rebalance when a worker joins. Apparently, it doesn't work, otherwise @tyge68 would have not reported the issue. I'll try to look into this soon.

locust/test/test_runners.py
```diff
@@ -717,7 +717,7 @@ def addWorker():
             def checkRebalancedTrue():
```
mboutet, line 717: In Python, `snake_case` is preferred. There's some old library such as `unittest` that uses `camelCase` (i.e. `self.assertTrue`), but this is only because the standard library needs to keep backward compatibility.

mboutet: Ok, this code was already like that. Sorry. Nevermind then

tyge68:
@mboutet I have applied the last suggested changes regarding the snake_case, and cleanup the commit history.
you can recheck the new test eventually , that would also be a good one to try to fix the original line of code that should have worked.

cyberw:
LGTM, except for the remaining camelCase method (checkRebalancedTrue) in the test

cyberw:
LGTM, except for the remaining camelCase method (checkRebalancedTrue) in the test

tyge68:
@cyberw indeed, fixed casing issue :)

cyberw:
Just waiting for any comments from @mboutet ...

mboutet:
Sorry for not getting back sooner. The changes and the test look good. I haven't tested the branch locally though.

mboutet:
Thanks @tyge68

cyberw:
Thanks! I'll trust the tests to report any errors (and the poor sods running latest master/prerelease version :)

mboutet:
I'm one of them 😜

PR #1889, opened 2021-09-21
Release docker image for arm64.

odidev:
The following file has been modified:
Modified .github/workflows/tests.yml file to build and push the image for both amd64 and arm64 platforms.

Signed-off-by: odidev odidev@puresoftware.com

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,11 +55,19 @@ jobs:
         with:
           username: locustbuild
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
-      - uses: docker/build-push-action@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: build and publish image
+        id: docker_build
+        uses: docker/build-push-action@v2
         with:
           build-args: |
             BUILDKIT_CONTEXT_KEEP_GIT_DIR=1
           context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
           push: ${{ github.ref == 'refs/heads/master' && github.repository_owner == 'locustio' }}
           tags: locustio/locust:master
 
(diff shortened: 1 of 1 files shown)


cyberw:
Great to get a arm64 build. 

But the PR changes other things that should not change (when to push) and also needs to change the step `docker_tagged`, so that release versions (tagged `latest` and x.y.z, by git tag) are also built for ARM.

cyberw:
Also, in order to release on arm64 I think we need to run at least some basic tests on that platform as well.

cyberw:
@odidev ?

.github/workflows/tests.yml
```diff
@@ -55,12 +55,19 @@ jobs:
...
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
-      - uses: docker/build-push-action@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: build and publish image
+        id: docker_build
+        uses: docker/build-push-action@v2
+        if: ${{ github.ref == 'refs/heads/master' && github.repository_owner == 'locustio' }}
```
cyberw, line 65: The reason for having this if statement for the push-setting rather than the whole step is that we can see any errors *before* merging. Is it necessary to move it?

.github/workflows/tests.yml
```diff
@@ -55,12 +55,19 @@ jobs:
...
         with:
-          build-args: |
-            BUILDKIT_CONTEXT_KEEP_GIT_DIR=1
```
cyberw, line 61: The reason I had to set this flag was so that the .git folder is available for docker (without it, it setuptools_scm cannot detect the version and will say something stupid like 0.0.1dev1234 instead of 2.2.4dev123)

odidev:
@cyberw , I have added changes as per your suggestions and pushed the commit and am now checking into adding tests for arm64.

cyberw:
@odidev Did you find a way to run the tests on arm? I can merge this without the tests, but it would be nice to have them...

odidev:
> 
> 
> @odidev Did you find a way to run the tests on arm? I can merge this without the tests, but it would be nice to have them...

@cyberw,  

I have tried to run the test cases on arm64 but the test cases are failing for arm64. 

**Commit link:**  https://github.com/odidev/locust/commit/eb24c0fb42e81ff01a23649901923586cd6347ea
 
**GitHub action link:** https://github.com/odidev/locust/runs/3814915067?check_suite_focus=true 

Could you suggest some pointers on these issues and how can I get details of failed cases? 
If changes looks good to you, I will amend these changes as well.

cyberw:
I want to try adding 3.10 support, so merging this first makes sense, even without tests...

Issue #1915, opened 2021-10-26
Exception then test stops if run in headless mode and autoquit != -1


EzR1d3r:
### Describe the bug
Exception then test stops
venv\lib\site-packages\locust\main.py", line 327, in stop_and_optionally_quit
    web_ui.stop()
AttributeError: 'NoneType' object has no attribute 'stop'

### Expected behavior
Stop without exception

### Steps to reproduce
Run test in destributed mode with headless = true and autoquit = 0.

### Environment

- OS:
- Python version: 3.9
- Locust version: 2.4.0
- Locust command line that you ran:
- Locust file contents (anonymized if necessary):

EzR1d3r:
> Did you run with headless AND autostart? Those settings are not meant to be combined, as headless already implies autostart...

Yes, I run it with headless and autostart. But I think more friendly if user dont need keep in mind, what if he changes settings headless, he needs also change autostart to work programm correctly. )

cyberw:
I guess so, but that is not the intended way (and most commonly would happen because the user is confused and thinks they *need* to specify autostart) so if we want to support that then we should at least log a warning. 

If you add that I can merge.

cyberw:
Or maybe we dont need to warn, but we should at least invert that if-statement to check for headless instead of autostart. There should be no difference at all in what code gets executed when someone has added --autostart in addition to --headless.

EzR1d3r:
> Or maybe we dont need to warn, but we should at least invert that if-statement to check for headless instead of autostart. There should be no difference at all in what code gets executed when someone has added --autostart in addition to --headless.

Already added a warning. ) But I will think about this proposal.

PR #1916, opened 2021-10-26
Fix issue #1915

EzR1d3r:
Fix the case when locust stops with exception in headless mode with autoquit != -1


cyberw:
Did you see my comment in the issue?

EzR1d3r:
> Did you see my comment in the issue?

Now yes, sorry for long answer.

PR #1931, opened 2021-11-11
Ensure terminal is restored even after sys.exit or something similar 

cyberw:
(still doesnt help with something like os._exit or sigkill)


cyberw:
@DennisKrone can you make this nicer? :)

cyberw:
Replacing this with a new PR after feedback from @DennisKrone

PR #1935, opened 2021-11-16
fix master runner not close rpc server

lizhaode:


cyberw:
That test was added by @delulu about 2 years ago. Maybe he/she can explain it?

lizhaode:
@cyberw  
I don't know why in `test_test_stop_event` unit test,  `master.quit()` will raise error like this 
```python
  File "/home/runner/work/locust/locust/locust/test/test_runners.py", line 909, in test_test_stop_event
    master.quit()
  File "/home/runner/work/locust/locust/locust/runners.py", line 825, in quit
    self.server.send_to_client(Message("quit", None, client.id))
  File "/home/runner/work/locust/locust/locust/util/exception_handler.py", line 13, in wrapper
    return function(*args, **kwargs)
  File "/home/runner/work/locust/locust/locust/rpc/zmqrpc.py", line 29, in send_to_client
    raise RPCError("ZMQ sent failure") from e
locust.exception.RPCError: ZMQ sent failure
```

could kindly help me, thanks

cyberw:
Sorry, I dont have time atm. Lets hope @delulu responds.

cyberw:
I'm going to close this due to inactivity soon. Sorry there hasnt been any clarification on what that test does. Do you have a specific issue you are trying to fix with this change?

lizhaode:
> I'm going to close this due to inactivity soon. Sorry there hasnt been any clarification on what that test does. Do you have a specific issue you are trying to fix with this change?

the rpc server will not close when use locust as a library, so I add `self.server.close()` in quit def

cyberw:
I see. Can you add a test case that addresses your use case? You can change delulus test case from

    master.quit()

to something like

   self.assertRaises(RpcError, master.quit)

lizhaode:
Thanks for help. 

But I try to understand this test case. 

That found in this test case: 

1. it start a master runner 
2. start a worker runner
3. assert something
4.  master send quit event
5. call `master.quit()` func

The error is occurred by `self.server.send_to_client` func. Maybe in step 2, master and worker runner start a real rpc server and client?

So, it should not raise error?

cyberw:
Aha, I thought it was due to the special code in MockedRpcServerClient ( https://github.com/locustio/locust/blob/master/locust/test/test_runners.py#L89), but I guess that's not where the error is coming from. 

Then I dont have any ideas, sorry.

lizhaode:
> Aha, I thought it was due to the special code in MockedRpcServerClient ( https://github.com/locustio/locust/blob/master/locust/test/test_runners.py#L89), but I guess that's not where the error is coming from.
> 
> Then I dont have any ideas, sorry.

I found the reason maybe.

1st call `master_env.events.quitting.fire`

then call `master.quit()`, the client is closed already, so the quit func can not send rpc message

locust/test/test_runners.py
```diff
@@ -906,7 +906,6 @@ def _(*args, **kwargs):
             # give time for users to generate stats, and stats to be sent to master
             sleep(0.1)
             master_env.events.quitting.fire(environment=master_env, reverse=True)
-            master.quit()
```
cyberw, line 909: Wouldnt it make more sense to remove the other line instead?

lizhaode: I guess this test case is test quit event?

cyberw: Yes, but the quit event would normally not be triggered by the user directly.

lizhaode: changed

cyberw:
thx!

Issue #1939, opened 2021-11-17
Possibility to set the exact number of users to spawn (instead weight)


EzR1d3r:
### Is your feature request related to a problem? Please describe.

I needed a special query that would be executed once every 30 seconds, regardless of the total number of users created. The only solution I came up with is to create one instance of a user with such a request. To do this, it was necessary to count his weight according to the formula

```
@events.test_start.add_listener
def on_test_start(environment: Environment, **kwargs) -> None:
    SingleUserUser.weight = 1
    CommonUser.weight = environment.runner.target_user_count - 1
```
This is not very convenient not very obvious.

### Describe the solution you'd like

Just a field (like weight) "count" or "fix_count", which describes the exactly count of user instances. I guess weight parameter in that case have to be ignored.

### Additional context

If you find it a useful figure I can try to do

cyberw:
Looks like a reasonable approach. If someone makes a PR with a few appropriate tests I’ll be happy to review.

EzR1d3r:
I was thinking about the realization of this feature.

The approach of @marcinh is pretty laconic and understandable. But here is the some limit - the fixed count users spawns first. In some cases that behavior may not be wanted.

May be we can recalculate all the weights together? For example the approach based on desired instance count of every class. For example:

User1.fix_count = 5
User2.weight = 1
User3.weight = 2
tagret_count = 11

sum_weights_non_fix = User1.weight + User2.weight # 3

user1_weight = User1.fix_count # 5
user2_weight = (tagret_count - user1_weight) * (User2.weight/sum_weights_non_fix) # 2
user3_weight = (tagret_count - user1_weight) * (User3.weight/sum_weights_non_fix) # 4

Further act as usually: normalizing, smooth, etc...

It is possible to slightly increase the weight of fixed users to generate a little faster, so we can be sure the total number would be enough for them. Controlling the count somewhere higher in the hierarchy, has not yet been precisely determined, **_add_users_on_workers** may be.

EzR1d3r:
> Looks like a reasonable approach. If someone makes a PR with a few appropriate tests I’ll be happy to review.

After diving into code I realized what my previous comment was pretty senseless. I have deleted it. )) After experiments, I also realized that in the current implementation of user spawning, trying to generate them more evenly, they still have to give large weights, although I achieved some success and even wrote part of the tests, my implementation close to generating fixed users at first. So I back to the @marcinh  realization, with some modification to resolve case with restore fixed users count after ramp-down and ramp-up.

[...]

Issue #1583, opened 2020-10-06
independent LoadTestShapes for users


stanislawskwark:
### Describe the solution you'd like
I have a locustfile with 2 users and I would like to have independent LoadTestShapes for them. 
Some examples of usage:
constant number of users A and number of users B changing over time
initially big number of users A and small number of users B, and opposite at the end of test
more types of users with independent shapes

### Describe alternatives you've considered
I've been thinking about adjusting weights dynamically, but didn't really come up with any nice solution

### Additional context
example of shape I would like to get:
![image](https://user-images.githubusercontent.com/46684704/95177482-84e96100-07be-11eb-8969-08126e9ad9c3.png)

cyberw:
Makes sense. If it can be done without complicating things too much, I think it is a good idea.

stanislawskwark:
@max-rocket-internet I've seen you are the one who implemented LoadTestShapes functionality, could you guide me a little in this topic? I don't have any experience with development of locust

max-rocket-internet:
Hi @stanislawskwark, the `LoadTestShape` class is manipulating the user count during a test on the master. This user count is kind of "global", not per `HttpUser`. I've not touched any of the `HttpUser` or weights code so can't really say how it could be implemented but I think not in the same way as `LoadTestShape`. Sorry I can't offer you any help here 🙂

aek:
Maybe this could be done like the TaskSet nested class or tasks class member with a (list/set/dict) of task/tasksets

Something like this:
```python
class CustomShape1(LoadTestShape):
    def tick(self):
        pass

class CustomUser1(HttpUser):
    shape = CustomShape1

class CustomUser2(HttpUser):
    @shape
    class CustomShape2(LoadTestShape):
        def tick(self):
            pass
```
To use that design, the method `shape_worker` should be modified in order to look at the environment user_classes to retrieve custom user shapes to spawn different user count at different spawn_rate. All of this specified at the User level.
I think this could be a good fine grained way to manage user distributions where all the Users should be spawned at least once. 

Related to:
#1618 
#1635
#1621

suiluj:
Are there any plans on this feature? I just started with locust and look for a solution to define different custom load shapes for different users.

PR #1964, opened 2021-12-25
fixed_count: ability to spawn a specific number of users (as opposed to just using weights)

EzR1d3r:
Issue #1939
Here is the overall working option, but noticed a problem with the display 'Ratio per User class' and 'Total ratio' on web UI.
I'm going to redo the logic a little bit, get_task_ratio_dict it will recalculate the ratio based on the number of users at the moment. I'll make it a separate commit in same branch. Maybe there are other ideas?

### DIFF
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -248,6 +248,7 @@ class UsersDispatcher(Iterator):
         current_user_count_target = min(
             self._current_user_count + self._user_count_per_dispatch_iteration, self._target_user_count
         )
+
         for user in self._user_generator:
             if not user:
                 self._no_user_to_spawn = True
@@ -309,6 +310,8 @@ class UsersDispatcher(Iterator):
         user_count = 0
         while user_count < target_user_count:
             user = next(user_gen)
+            if not user:
+                break
             worker_node = next(worker_gen)
             users_on_workers[worker_node.id][user] += 1
             user_count += 1
@@ -366,18 +369,25 @@ class UsersDispatcher(Iterator):
         # Spawn users
         while True:
             if self._try_dispatch_fixed:
+                self._try_dispatch_fixed = False
+                current_fixed_users_count = {u: self._get_user_current_count(u) for u in fixed_users}
                 spawned_classes = set()
-                while True:
+                while len(spawned_classes) != len(fixed_users):
                     user_name = next(cycle_fixed_gen)
                     if not user_name:
                         break
-                    if self._get_user_current_count(user_name) >= fixed_users[user_name].fixed_count:
-                        spawned_classes.add(user_name)
-                    else:
+
+                    if current_fixed_users_count[user_name] < fixed_users[user_name].fixed_count:
+                        current_fixed_users_count[user_name] += 1
+                        if current_fixed_users_count[user_name] == fixed_users[user_name].fixed_count:
+                            spawned_classes.add(user_name)
                         yield user_name
-                    if len(spawned_classes) == len(fixed_users):
-                        break
-                self._try_dispatch_fixed = False
+
+                        # 'self._try_dispatch_fixed' was changed outhere,  we have to recalculate current count
+                        if self._try_dispatch_fixed:
+                            current_fixed_users_count = {u: self._get_user_current_count(u) for u in fixed_users}
+                            spawned_classes.clear()
+                            self._try_dispatch_fixed = False
 
             yield next(cycle_weighted_gen)
 
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -4,7 +4,7 @@ import math
 import time
 from collections.abc import Iterator
 from operator import attrgetter
-from typing import Dict, Generator, List, TYPE_CHECKING, Tuple, Type
+from typing import Dict, Generator, List, TYPE_CHECKING, Optional, Tuple, Type
 
 import gevent
 import typing
@@ -98,6 +98,10 @@ class UsersDispatcher(Iterator):
 
         self._rebalance = False
 
+        self._try_dispatch_fixed = True
+
+        self._no_user_to_spawn = False
+
     @property
     def dispatch_in_progress(self):
         return self._dispatch_in_progress
@@ -132,6 +136,9 @@ class UsersDispatcher(Iterator):
                 if self._rebalance:
                     self._rebalance = False
                     yield self._users_on_workers
+                if self._no_user_to_spawn:
+                    self._no_user_to_spawn = False
+                    break
 
         while self._current_user_count > self._target_user_count:
             with self._wait_between_dispatch_iteration_context():
@@ -242,12 +249,17 @@ class UsersDispatcher(Iterator):
             self._current_user_count + self._user_count_per_dispatch_iteration, self._target_user_count
         )
         for user in self._user_generator:
+            if not user:
+                self._no_user_to_spawn = True
+                break
             worker_node = next(self._worker_node_generator)
             self._users_on_workers[worker_node.id][user] += 1
             self._current_user_count += 1
             self._active_users.append((worker_node, user))
             if self._current_user_count >= current_user_count_target:
-                return self._users_on_workers
+                break
+
+        return self._users_on_workers
 
     def _remove_users_from_workers(self) -> Dict[str, Dict[str, int]]:
         """Remove users from the workers until the target number of users is reached for the current dispatch iteration
@@ -264,8 +276,16 @@ class UsersDispatcher(Iterator):
                 return self._users_on_workers
             self._users_on_workers[worker_node.id][user] -= 1
             self._current_user_count -= 1
+            self._try_dispatch_fixed = True
             if self._current_user_count == 0 or self._current_user_count <= current_user_count_target:
                 return self._users_on_workers
+
+    def _get_user_current_count(self, user: str) -> int:
+        count = 0
+        for users_on_node in self._users_on_workers.values():
+            count += users_on_node.get(user, 0)
+
+        return count
 
     def _distribute_users(
         self, target_user_count: int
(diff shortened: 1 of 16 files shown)


EzR1d3r:
Here is the overall working option, but noticed a problem with the display 'Ratio per User class' and 'Total ratio' on web UI.
I'm going to redo the logic a little bit, get_task_ratio_dict it will recalculate the ratio based on the number of users at the moment. I'll make it a separate commit in same branch. Maybe there are other ideas?

cyberw:
We should probably mention it somewhere in the documentation. 

@mboutet what do you think?

EzR1d3r:
I made the changes about which I wrote, it seems to me that it turned out quite appropriate. I made two separate functions for counting the ratio of classes and tasks, as can be seen from unit tests, the results of the calculation remained the same, only a little changed approach.

EzR1d3r:
@cyberw Should I do something else or request ready to review?

cyberw:
Add a mention at some appropriate point in the docs (.rst files). I’m hoping @mboutet will review so I dont have to :)

mboutet:
I'll try to look at this first week of January. Sorry for the delay.

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,239 @@ class User3(User):
...
 
+class TestRampUpUsersFromZeroWithFixed(unittest.TestCase):
+    class Case:
+        def __init__(self, fixed_counts: Tuple[int], weights: Tuple[int], target_user_count: int) -> None:
+            self.fixed_counts = fixed_counts
+            self.weights = weights
+            self.target_user_count = target_user_count
+
+    def case_handler(self, cases: List[Case], expected: Dict[str, int], user_classes=List[User]):
+        assert len(cases) == len(expected), "Different len of 'cases' and 'expected'"
```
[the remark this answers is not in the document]
EzR1d3r: I am not very familiar with asserts, but I only wanted to mark what it is not the test case but clue for developer if he make a mistake with test adding.

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,239 @@ class User3(User):
...
+
+        class User4(User):
+            ...
+
+        class User5(User):
+            ...
+
+        self.case_handler(
+            cases=[
+                self.Case(fixed_counts=(1, 1), weights=(1, 1, 1), target_user_count=5),
```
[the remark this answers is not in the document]
EzR1d3r: Only final distribution. Since the same algorithm is used as for ordinary users, and it is tested above.

locust/test/test_dispatch.py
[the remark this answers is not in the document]
cyberw: self.assertEqual(len(cases), len(expected)) is appropriate

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,239 @@ class User3(User):
...
+                        list(users_dispatcher)
+
+                        for user_class in user_classes:
+                            if user_class.fixed_count:
+                                self.assertEqual(
+                                    users_dispatcher._get_user_current_count(user_class.__name__),
+                                    user_class.fixed_count,
+                                )
+
+
```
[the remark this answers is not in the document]
EzR1d3r: I've looked into test_main and a little bit confused, what exactly you want to test? Basic spawning with fixed users? ie check the output is ok and corresponds to the users data?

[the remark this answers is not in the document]
cyberw: Yes

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
 
 
+class TestRampUpUsersFromZeroWithFixed(unittest.TestCase):
+    class Case:
+        def __init__(self, fixed_counts: Tuple[int], weights: Tuple[int], target_user_count: int):
+            self.fixed_counts = fixed_counts
+            self.weights = weights
+            self.target_user_count = target_user_count
+
+    def case_handler(self, cases: List[Case], expected: Dict[str, int], user_classes=List[User]):
```
mboutet, line 3301: Did you mean?
```suggestion
    def case_handler(self, cases: List[Case], expected: Dict[str, int], user_classes: List[User]):
```

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+
+                        for user_class in user_classes:
+                            if user_class.fixed_count:
+                                self.assertNotEqual(
+                                    users_dispatcher._get_user_current_count(user_class.__name__),
+                                    user_class.fixed_count,
+                                )
+
+                        # Ramp-up go back to `target_user_count` and ensure
+                        # what the fixed users return to their counts
```
mboutet, line 3514: Unless I misunderstood the sentence, I think the "what" does not belong in that sentence:
```suggestion
                        # the fixed users return to their counts
```

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+
+                        for user_class in user_classes:
+                            if user_class.fixed_count:
+                                self.assertEqual(
+                                    users_dispatcher._get_user_current_count(user_class.__name__),
+                                    user_class.fixed_count,
+                                )
+
+                        # Ramp-down to go to `down_to_count`
+                        # and ensure what the fixed users was decreased too
```
mboutet, line 3499: Unless I misunderstood the sentence, I think the "what" does not belong in that sentence:
```suggestion
                        # and ensure the fixed users was decreased too
```

mboutet:
I think adding an integration test similar to https://github.com/locustio/locust/blob/cce0d29b90e5e9f9dad7a7b96b25a393923c4a16/locust/test/test_runners.py#L957
would be good.

locust/dispatch.py
```diff
@@ -98,6 +98,10 @@ def __init__(self, worker_nodes: "List[WorkerNode]", user_classes: List[Type[Use
 
         self._rebalance = False
 
+        self._try_dispatch_fixed = True
+
+        self._no_user_to_spawn = False
```
mboutet, line 103: Do you also need to set these in https://github.com/locustio/locust/blob/cce0d29b90e5e9f9dad7a7b96b25a393923c4a16/locust/dispatch.py#L145
?

mboutet:
Can you also add a small stress test like https://github.com/locustio/locust/blob/cce0d29b90e5e9f9dad7a7b96b25a393923c4a16/locust/test/test_dispatch.py#L1934
? This is just to make sure there's no significant performance regression.

locust/dispatch.py
```diff
@@ -307,26 +327,58 @@ def _user_gen(self) -> Generator[str, None, None]:
...
+                    user.__name__,
+                    round(target_min_weight * value / min([u[1] for u in users])),
+                )
+                for user, value in users
+            ]
+            generation_length_to_get_proper_distribution = sum(
+                normalized_val[1] for normalized_val in normalized_values
+            )
+            gen = smooth(normalized_values)
+            # Instead of calling `gen()` for each user, we cycle through a generator of fixed-length
```
mboutet, line 352: Just to be consistent with the vertical spacing above :) 
```suggestion
            gen = smooth(normalized_values)

            # Instead of calling `gen()` for each user, we cycle through a generator of fixed-length
```

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+                user_class.weight, user_class.fixed_count = 1, 0
+
+            case = cases[case_num]
+            self.assertEqual(
+                len(case.fixed_counts) + len(case.weights),
+                len(user_classes),
+                msg="Invalid test case or user list.",
+            )
+
+            fixed_users_list = user_classes[: len(case.fixed_counts)]
```
mboutet, line 3316: No need to use hungarian notation
```suggestion
            fixed_users = user_classes[: len(case.fixed_counts)]
```

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+
+        class User2(User):
+            ...
+
+        class User3(User):
+            ...
+
+        self.case_handler(
+            cases=[
+                self.Case(fixed_counts=(), weights=(1, 1, 1), target_user_count=3),
```
mboutet, line 3346: This case makes the test's name (`test_ramp_up_2_weigted_user_with_1_fixed_user`) a little misleading, no?

mboutet:
I'll have to take another look later, but it looks good at first glance. The unit tests you added in `test_dispatch.py` seem to cover the new functionality adequately 👍🏼.

locust/test/test_dispatch.py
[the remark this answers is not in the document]
EzR1d3r: Added

locust/test/test_dispatch.py
EzR1d3r: Yes, thx. Fixed.

locust/test/test_dispatch.py
EzR1d3r: I guess this subject jumped from my native language and not needed in eng ))) Removed

locust/test/test_dispatch.py
EzR1d3r: Removed

locust/dispatch.py
EzR1d3r: I guess no, **self._no_user_to_spawn** resets to False almost immediately after checking in `_dispatcher`
**self._try_dispatch_fixed** will be True if we stop to spawn somewhere on fixed users, so when we continue it is fine. If we spawned all fixed and it is False, it is also ok. The only case we need to reset it is ramp-down. Also I guess nothing changes if to set this fields, except a superfluous work in _try_dispatch_fixed if it was False.

locust/dispatch.py
EzR1d3r: Done.

locust/test/test_dispatch.py
EzR1d3r: Fixed.

locust/test/test_dispatch.py
EzR1d3r: Yep, move this case to **test_ramp_up_various_count_weigted_and_fixed_users**

EzR1d3r:
I made some fixups, also will add the tests @mboutet asked.

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+        def __init__(self, fixed_counts: Tuple[int], weights: Tuple[int], target_user_count: int):
+            self.fixed_counts = fixed_counts
+            self.weights = weights
+            self.target_user_count = target_user_count
+
+    def case_handler(self, cases: List[Case], expected: Dict[str, int], user_classes: List[User]):
+        self.assertEqual(len(cases), len(expected))
+
+        for case_num in range(len(cases)):
+            # Reset to defaul values
```
mboutet, line 3305: Small typo
```suggestion
            # Reset to default values
```

locust/test/test_dispatch.py
```diff
@@ -3291,6 +3291,241 @@ class User3(User):
...
+                {"User1": 2, "User2": 499, "User3": 499},
+                {"User1": 100, "User2": 450, "User3": 450},
+                {"User1": 960, "User2": 20, "User3": 20},
+                {"User1": 9990, "User2": 5, "User3": 5},
+                {"User1": 100, "User2": 0, "User3": 0},
+            ],
+            user_classes=[User1, User2, User3],
+        )
+
+    def test_ramp_up_various_count_weigted_and_fixed_users(self):
```
mboutet, line 3368: It could also be a good to add a test for ramping-down a mix of weighted and fixed users (i.e. `test_ramp_down_various_count_weigted_and_fixed_users`). If you implement this ramp-down test in the current test class, make sure to also rename the class to something like `TestRampUpAndDownUsersWithFixedCound`

mboutet:
@EzR1d3r, the PR looks good. The tests in `test_dispath.py` gives me confidence that the feature works as intended. Also, all the existing tests pass, so there should not be any regression 👍🏼.

After the last comments are resolved, I think this will be good to go.

locust/test/test_dispatch.py
EzR1d3r: done

locust/test/test_dispatch.py
EzR1d3r: Yeah, you're right.
The name changed.

EzR1d3r:
It was a good idea to add TestLargeScale, so I found a bug with **_distribute_users** function and I had to remake the realisation of fixed users spawning. Within the _distribute_users iterating over the user_gen not add the users in **_users_on_workers** instantly and we cant monitore the count of the fixed users.

EzR1d3r:
I see the tests fails. But my locals tests are ok. As I can see, they need a little bit more time.

EzR1d3r:
I see the tests fail. But my locals tests are ok.
`self.assertIn("Shutting down (exit code 0), bye.", output)`

cyberw:
> I see the tests fail. But my locals tests are ok. `self.assertIn("Shutting down (exit code 0), bye.", output)`
> 
> upd.: in some reason, is missed the tail of "Shutting down (exit code 0)**, bye.**

I recently removed the ”, bye.” in master. Do thr same in your test and you should be fine.

cyberw:
Looks pretty good to me. 

Just needs an entry somewhere in the documentation (.rst files). One question, why did you add the _actual in the method names? Is that the right word? What would be the ”not actual” ratio? Maybe _current is more appropriate? (on mobile right now, sorry if it would have become obvious by reading the whole PR at once)

Anything left to do in your view, @mboutet ?

EzR1d3r:
> Looks pretty good to me.
> 
> Just needs an entry somewhere in the documentation (.rst files).

I will look.
>One question, why did you add the _actual in the method names? Is that the right word? What would be the ”not actual” ratio? Maybe _current is more appropriate? (on mobile right now, sorry if it would have become obvious by reading the whole PR at once)
> 
> Anything left to do in your view, @mboutet ?

I guess, you right. 'Actual' in relation to previous 'static' behaviour. But if you dont know the previous, it has no sense, right?
I can change to 'current' or smth better or may be just remove at all?

cyberw:
> I guess, you right. 'Actual' in relation to previous 'static' behaviour. But if you dont know the previous, it has no sense, right? I can change to 'current' or smth better or may be just remove at all?

Remove sounds good!

EzR1d3r:
Added some docs (last commit) and removed 'actual'.

mboutet:
@cyberw, this looks good!

@EzR1d3r, good job on that feature!

cyberw:
Awesome stuff, thanks! This will automatically become part of the next prerelease build, and I'll do a full release in a couple days.

EzR1d3r:
Thx you too for review and accepting this feature!

domik82:
@EzR1d3r  thanks for wrapping it up.
@cyberw @mboutet  - I guess work we did on user distribution controlled by the master is paying off. 

Just to confirm the functionality we described here: https://github.com/locustio/locust/issues/1583  should work now?

mboutet:
I don't believe independant load test shapes has been implemented.

cyberw:
No, independent load shapes is not implemented. That ticket is closed because it was stale.  Dont know why it was referenced..

I guess in some cases fixed user count helps a little.

PR #1976, opened 2022-01-20
Pass tags and exclude-tags to workers. 

### LINKED ISSUE #1882: --tags / --exclude-tags seems not taken in consideration when running distributed unless they are specified in every worker
I'm not sure this is a bug, it's just quite inconvenient to have to specify --tags and --excluded-tags for each worker. It's indeed working if the tags are specified at wokrer level (the `--tags` and `--exlcuded-tags` is completely ignored for the master)

### Describe the bug
I run the master with `--tags  process_text` because I want to only run one one of the tasks

### Expected behavior
Only the tasks I specify in `--tags` are taken in account, or the ones that are in `--excluded-tags` are excluded 

### Actual behavior
All the tasks are executed 

### Steps to reproduce
Only the tasks that are selected/excluded should be only selected / excluded

### Environment

- OS: Linux (Centos, Ubuntu)
- Python version: 3.9.6
- Locust version: 2.2.0 (on all machines)
- Locust command line that you ran: `locust --config yomite-1h-50u-master.conf --tags process_text` See [config](https://github.com/lfoppiano/grobid-superconductors-performance/blob/main/yomite-1h-50u-master.conf). 
- Locust file contents (anonymized if necessary): See [here](https://github.com/lfoppiano/grobid-superconductors-performance/blob/main/locustfiles/locustfile_java.py). The data is not available, though.

cyberw:
This required moving the call to filter_tasks_by_tags from Environment init to Runner.start/start_worker. Fixes #1882

### DIFF
--- a/locust/test/test_main.py
+++ b/locust/test/test_main.py
@@ -770,7 +770,7 @@ class SecondUser(HttpUser):
     @task
     def task1(self):
         print("task1")
-    
+
     @tag("tag2")
     @task
     def task2(self):
--- a/locust/test/test_main.py
+++ b/locust/test/test_main.py
@@ -777,7 +777,6 @@ class SecondUser(HttpUser):
         print("task2")
 """
         )
-        print(content)
         with mock_locustfile(content=content) as mocked:
             proc = subprocess.Popen(
                 [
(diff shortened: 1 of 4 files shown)


locust/test/test_main.py
```diff
@@ -757,3 +757,69 @@ def on_test_stop(environment, **kwargs):
...
+    def task1(self):
+        print("task1")
+
+    @tag("tag2")
+    @task
+    def task2(self):
+        print("task2")
+"""
+        )
+        print(content)
```
stasfilin, line 780: Do we really need this `print` here?

cyberw: Good catch

Issue #1961, opened 2021-12-17
Incorrect "All users spawned" log messages when running distributed


cyberw:
When running distributed, we print the wrong numbers, e.g. 

    All users spawned: {"UserSubclass": 2} (2 total users)

When the desired (and actual) final user count was 3.

Here's a test case that exhibits this (not 100% of the time, but almost) https://github.com/locustio/locust/commit/4f7b37ab187a7c084e3c0fda6d23f2adf10345c3

Maybe we are printing the numbers from the previous step, or we must wait until the next worker report comes in before printing? It seems to only happens with at least 5 workers (!?)

This issue seems to have been around for a while (maybe since 2.0)

@mboutet , do you have time to take a look?

cyberw:
I did some additional testing and was able to reproduce with as few as **one** worker. Then the output is:

`All users spawned: {} (0 total users)`

mboutet:
I'll have time in January. I'll set a reminder on my side.

PR #1977, opened 2022-01-23
Fixing issue #1961 with incorrect "All users spawned" log messages wh…

EzR1d3r:
Issue #1961 I saw this bug on a remote machine. I can't debug there, but I think the reason is that 100ms sometimes is not enough for workers (I guess, the last report from last worker) to send reports (or may be to receive and handle on master side). PR requires polish, but you can see the essence.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -749,32 +749,35 @@ class MasterRunner(DistributedRunner):
         # when the user count is really at the desired value.
         timeout = gevent.Timeout(self._wait_for_workers_report_after_ramp_up())
         timeout.start()
+        msg_prefix = "All users spawned"
         try:
             while self.user_count != self.target_user_count:
-                gevent.sleep()
+                gevent.sleep(0.01)
         except gevent.Timeout:
-            pass
+            msg_prefix = (
+                "Spawning is complete and report waittime is expired, but not all reports recieved from workers"
+            )
         finally:
             timeout.cancel()
 
         self.environment.events.spawning_complete.fire(user_count=sum(self.target_user_classes_count.values()))
         self.spawning_completed = True
 
-        logger.info("All users spawned: %s" % _format_user_classes_count_for_log(self.reported_user_classes_count))
+        logger.info("%s: %s" % (msg_prefix, _format_user_classes_count_for_log(self.reported_user_classes_count)))
 
     @functools.lru_cache()
     def _wait_for_workers_report_after_ramp_up(self) -> float:
         """
         The amount of time to wait after a ramp-up in order for all the workers to report their state
-        to the master. If not supplied by the user, it is 100ms by default. If the supplied value is a number,
+        to the master. If not supplied by the user, it is 1000ms by default. If the supplied value is a number,
         it is taken as-is. If the supplied value is a pattern like "some_number * WORKER_REPORT_INTERVAL",
         the value will be "some_number * WORKER_REPORT_INTERVAL". The most sensible value would be something
         like "1.25 * WORKER_REPORT_INTERVAL". However, some users might find it too high, so it is left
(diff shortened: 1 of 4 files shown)


locust/runners.py
```diff
@@ -747,32 +747,33 @@ def start(self, user_count: int, spawn_rate: float, wait=False) -> None:
         # when the user count is really at the desired value.
         timeout = gevent.Timeout(self._wait_for_workers_report_after_ramp_up())
         timeout.start()
+        msg_prefix = "All users spawned"
         try:
             while self.user_count != self.target_user_count:
                 gevent.sleep()
```
cyberw, line 753: We might want to change this to gevent.sleep(0.01) or something, to prevent busy waiting. What do you think @mboutet ?

cyberw:
Would this delay the logging (as well as  spawning_complete.fire) by a whole second? That is a bit unfortunate, but maybe unavoidable..

EzR1d3r:
> Would this delay the logging (as well as spawning_complete.fire) by a whole second? That is a bit unfortunate, but maybe unavoidable..
The delay for a second will only be if the worker does not make a report on time. In most cases 100 ms is enough. But unfortunately, it is not known how much it takes when a delay occurs, whether the worker reports at all. 

> We might want to change this to gevent.sleep(0.01) or something, to prevent busy waiting. 
Do you want to sleep(0.01) to reduce the CPU load while we wait?

cyberw:
> > Would this delay the logging (as well as spawning_complete.fire) by a whole second? That is a bit unfortunate, but maybe unavoidable..
> 
> The delay for a second will only be if the worker does not make a report on time. In most cases 100 ms is enough. But unfortunately, it is not known how much it takes when a delay occurs, whether the worker reports at all.

Cool, then it is a minor issue.

> > We might want to change this to gevent.sleep(0.01) or something, to prevent busy waiting.
> 
> Do you want to sleep(0.01) to reduce the CPU load while we wait?

Yes please.

If you do that, and resolve the conflicts I'd be happy to merge.

EzR1d3r:
Of course, I'll add edits. I think the report error message will not be superfluous. However, if the bug is repeated again, a deeper analysis will be needed, why the last report sometimes is missed.

locust/runners.py
mboutet: Makes sense

EzR1d3r:
@cyberw @mboutet 
I'm not very familiar with gevent and I found a strange thing. In this place sleep takes 100ms instead of wanted 1 second.
https://github.com/locustio/locust/blob/87c3dd10754ae23042408f577dc7a0c7b5fa7d95/locust/runners.py#L840

So the test falls because master miss all workers (for ~300 ms instead of 3 second **HEARTBEAT_LIVENESS**) in **heartbeat_worker** while he wait reports for 1 second (was 0.1 second before).
https://github.com/locustio/locust/blob/87c3dd10754ae23042408f577dc7a0c7b5fa7d95/locust/test/test_runners.py#L1897

cyberw:
Hmm.. I am pretty sure there is no way for a sleep to take a *shorter* amount of time than specified...

mboutet:
> Hmm.. I am pretty sure there is no way for a sleep to take a _shorter_ amount of time than specified...

Yes, that's certainly not possible. It could take longer if the event loop is busy, but not less.

EzR1d3r:
This is possible if several lines above make a patch. ))) Didn't guess to check it out yesterday.
I rightly understand - is this just to speed up the test?
https://github.com/locustio/locust/blob/87c3dd10754ae23042408f577dc7a0c7b5fa7d95/locust/test/test_runners.py#L1896

cyberw:
Yup!

cyberw:
Awesome. Thanks!

PR #1978, opened 2022-01-24
Repair broken hyperlinks in documentation

jeroenhendricksen:
A small PR to fix currently not working/visible hyperlinks under the [More information](http://docs.locust.io/en/stable/running-cloud-integration.html#more-information)-section

I'm not familiar with how `:ref:` statements work in `.rst` files, so if they are important for these links (I don't think so), please feel free to improve on my PR.

### DIFF
--- a/docs/running-cloud-integration.rst
+++ b/docs/running-cloud-integration.rst
@@ -22,9 +22,9 @@ The prerequisites are:
 More information
 ===================
 
-- :ref:`Example <https://github.com/locustio/locust/blob/master/examples/terraform/aws/README.md>`
+- `Example <https://github.com/locustio/locust/blob/master/examples/terraform/aws/README.md>`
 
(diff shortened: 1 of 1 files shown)


cyberw:
Lgtm, thanks!

cyberw:
Hmm... actually, it doesnt look so good? https://docs.locust.io/en/latest/running-cloud-integration.html

jeroenhendricksen:
> Hmm... actually, it doesnt look so good? https://docs.locust.io/en/latest/running-cloud-integration.html

I see. Let me try to fix it for you.

jeroenhendricksen:
The only two things I can think of to explain this anomaly is:
- The link should have been postfixed with a `_`, like this:  

    ```
    `Helm chart <https://helm.sh/>`_
    ```
- Or links in a list are required to use a `:ref:`

Issue #1969, opened 2022-01-12
Locust while running as library with grpc client not outputting stats


saurabhsharma721:
### Describe the bug
Grpc client not creating stats while running in library mode

### Expected behavior
Stats should be created and available both in csv files and in api.

### Actual behavior
The Csv files generated and logs created will have zero stats output

   locust test have completed running
   ```
   Test run completed.
   requests after run : 0
   Failures after run : 0
   ```

### Steps to reproduce

The code is available here in github repo 

https://github.com/saurabhsharma721/grpc_locust_stats_repro_prj

### Environment

- OS: mac
- Python version: 3.8.5
- Locust version: (please dont file issues for anything but the most recent release or prerelease builds): 2.24
- Locust file contents (anonymized if necessary): Code available here : https://github.com/saurabhsharma721/grpc_locust_stats_repro_prj/blob/main/grpc/locust_driver.py

I see we have similar question on SO but want to check where to place request events inside the test. A code snippet will help

cyberw:
Interesting. I dont have time to look at it but hopefully someone will. Or maybe dig into the code base yourself.

saurabhsharma721:
I am trying to see diff between non ui and ui code and so far nothing

cyberw:
You should also update locust, just in case that matters.

BonelessPi:
I had an almost identical issue when running Locust as a library, although I was using SFTP clients instead of grpc. I resolved my issue by using the .events attribute of the Environment object rather than the locust.events module. I do not know if this will work for you, but it may be worth a try.

In your case, I would suggest modifying your Locustfile as so:
Line 26:
    def __init__(self, environment, stub):
        self.env = environment
        ...
Line 50:
    self.env.events.request.fire(**request_meta)
Line 69:
    self.client = GrpcClient(environment, stub)

I hope this works for you!

saurabhsharma721:
The above solution worked for me. Thanks @BonelessPi

cyberw:
Huh. Seems like a bug in the example GrpcClient. Anyone up for making a PR?

PR #1982, opened 2022-01-26
Fire requests through Environment.events

BonelessPi:
Example Grpc client not reporting stats when Locust is running as a library.
Fixed by firing requests through Environment.events.request.fire instead of locust.events.request.fire

Related Issue: https://github.com/locustio/locust/issues/1969


cyberw:
Thx!

Issue #1981, opened 2022-01-24
Error when executing Environment.runner.start_shape


linh07:
### Describe the bug

I'm using locust as library and experimenting with custom load shape.

### Expected behavior

Load increasing by 10 users every 30 seconds.

### Actual behavior

The following error is shown when I run the code:
```
  File "\path_to_locustfile", line 44, in <module>
    env.runner.start_shape()
  File "\path_to_file\locust\runners.py", line 384, in start_shape
    self.environment.shape_class.reset_time()
TypeError: reset_time() missing 1 required positional argument: 'self'
```

```
  File "\path_to_file\locust\runners.py", line 389, in shape_worker
    new_state = self.environment.shape_class.tick()
TypeError: tick() missing 1 required positional argument: 'self'
```

### Steps to reproduce
 
This is my code:
```
import gevent
import math
from locust import HttpUser, task, between, constant
from locust.env import Environment
from locust.stats import stats_printer, stats_history
from locust.log import setup_logging
from locust.shape import LoadTestShape

setup_logging("INFO", None)


class StepLoadShape(LoadTestShape):

    step_time = 30
    step_load = 10
    spawn_rate = 10
    time_limit = 600

    def tick(self):
        run_time = self.get_run_time()

        if run_time > self.time_limit:
            return None

        current_step = math.floor(run_time / self.step_time) + 1
        return (current_step * self.step_load, self.spawn_rate)


class User(HttpUser):
    wait_time = constant(1)

    @task
    def my_task(self):
        self.client.get("/")


env = Environment(user_classes=[User], shape_class=StepLoadShape, host="https://docs.locust.io/")
env.create_local_runner()
gevent.spawn(stats_printer(env.stats))
gevent.spawn(stats_history, env.runner)
env.runner.start_shape()
gevent.spawn_later(600, lambda: env.runner.quit())
env.runner.greenlet.join()
```

### Environment

- OS: Windows 10 Pro
- Python version: Python 3.9
- Locust version: 2.6.0
- Locust command line that you ran: `python /path_to_file`
- Locust file contents (anonymized if necessary):

PR #1983, opened 2022-01-26
hardening Environment.shape_class for distinct usage

uddmorningsun:
Hardening for issue https://github.com/locustio/locust/issues/1981, make it more clear if the user made a mistake.

Signed-off-by: Chenyang Yan <memory.yancy@gmail.com>

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -98,6 +98,11 @@ class Environment:
                     ", ".join(map(methodcaller("fullname"), self.user_classes))
                 )
             )
+        if self.shape_class is not None and not isinstance(self.shape_class, LoadTestShape):
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


cyberw:
Does it fix that issue or just make it more clear if the user made a mistake? Looks like a reasonable change, I'm just wondering. And maybe it is meaningful to add a test case?

uddmorningsun:
Making it more clear, prevent usage error.

> And maybe it is meaningful to add a test case?

OK, I will add test case happily and rebase commit later.

uddmorningsun:
@cyberw I have updated and rebased it, you could be ready to review.

cyberw:
thx!

Issue #1897, opened 2021-10-06
Configurable heartbeat values for workers


nathan-beam:
### Is your feature request related to a problem? Please describe.

Duplicating/Re-opening #1787
I keep running into heartbeat timeouts and there's only so much multi-threading and multi-processing I can do.  It'd be extremely helpful to be able to configure this.

### Describe the solution you'd like

IMO the option to set this from the command line isn't crucial. If we're worried about too many advanced settings being exposed, I think a good option (at least for now, until a real overhaul of the settings takes place) is to just have a setter for heartbeat liveness in the `MasterRunner` class and a setter for interval on the `DistributedRunner`

```python
@events.init.add_listener
def on_locust_init(environment, **_kwargs):
    if isinstance(environment.runner, MasterRunner):
        environment.runner.set_heartbeat_liveness(30) # Set 30 second timeout on heartbeats
    environment.runner.set_heartbeat_interval(10) # 10 second heartbeat check/update
    some_long_task()
```

### Describe alternatives you've considered

Manually overriding the constant works, but is certainly less than ideal
```python 
import locust.runners 
locust.runners.HEARTBEAT_LIVENESS = 30
```

Issue #1843, opened 2021-08-06
Workers go down with message: failed to send heartbeat, setting state to missing.


roquemoyano-tc:
### Describe the bug
every time that I run locust I'm always getting:  locust-worker-xxxx failed to send heartbeat, setting state to missing.

### Expected behavior
worker in running status

### Actual behavior
I'm using the oficial helm chart  to run locust, when I run the python code after some time the workers changes their status to missing and I'm not able to finish running the test.

### Environment

- OS: AKS
- Python version: python version included in helm chart
- Locust version: 1.5.0* to 2.0.0*
- Locust command line that you ran: I open the Gui to run it

mboutet:
@roquemoyano-tc You need to provide the complete logs from both the master and at least one of the workers. Share them as [gists](https://docs.github.com/en/github/writing-on-github/editing-and-sharing-content-with-gists/creating-gists) as to not paste wall of logs in this issue.

roquemoyano-tc:
sorry, in this link are the master and worker logs

https://gist.github.com/roquemoyano-tc/f23dc4a4c8c17da30fa6c101c55a0ad9

cyberw:
Can you explain the slave log?
What is this for example "Adding new paired device"

Also attach your locustfile.

amaanupstox:
@cyberw  
#!interpreter [optional-arg]
# -*- coding: utf-8 -*-

"""
This contains test cases to load test foo service via gRPC call with locust.
"""

# Built-in/Generic Imports
import logging
import sys
from locust import task, constant
from locust.contrib.fasthttp import FastHttpUser
# Libs
from lib.locust.service_defination import stopwatch
# Own modules

from utility import utils
from utility.logger import Logger

logging = Logger().get_app_logger()
backoffice_customer_profile_service = BackOfficeCustomerProfileService()


class GRPCBackOfficeClient:

    @stopwatch
    def email_mobile_list(self):
        try:
            response = backoffice_customer_profile_service.get_email_mobile_list()
            assert response.metadata.success, "response should be true"

        except (KeyboardInterrupt, SystemExit):
            logging.error("Interrupted by keyboard............")
            sys.exit(0)

  
class GRPCBackOfficeLocust(FastHttpUser):
    host = "https://{}".format(utils.get_properties("backoffice-service", "url"))
    grpc_backoffice_client = GRPCBackOfficeClient()
    wait_time = constant(0)

    def on_start(self):
[...]

Issue #1787, opened 2021-06-14
Enable to define the heartbeat liveness from the cmdline


motybz:
I found out that a long time ago the service had an ability to define the heartbeat liveness from the cmdline but[ this commit](https://github.com/locustio/locust/commit/d5ec17e8ad631046b57eac469a1a04ec337bb7c0) removes this ability.

I think that in order to be a generic service the "heartbeat liveness" should be configurable.
In the current behavior, if I have a bug that slows the process I can't even understand it from the test results.

### Describe the solution you'd like

Add the following flag
`--heartbeat-liveness`

### Describe alternatives you've considered
I made a workaround and it works for me
`import locust.runners 
locust.runners.HEARTBEAT_LIVENESS = 30
`

cyberw:
We have a couple of these "internal" settings I think. they definitely deserve command line settings, but the list of settings is already super long. 

Perhaps we could divide settings into "normal" and "advanced"? I'm not sure what is the best way to it and I am sure I dont have time to do it myself, but maybe someone else could investigate.

PR #1988, opened 2022-01-28
add a configuration option for specifying a worker heartbeat interval

bioker:


bioker:
I have noticed that there is no option to configure this parameter.

I have faced a situation when the heartbeat fails but workers seem to be healthy while doing this lab project: https://github.com/bioker/mysql-lab/blob/f85ea857d0c4b29261569eadcca6b5344ac7eed1/docker-compose.yaml#L4.

Running the project with the custom docker image I build from my version shows that the increased heartbeat helps to avoid such problems and the test is still working well.

Maybe it's worth checking why it fails in the first place, but it seems harder than adding this parameter and I suppose it can be useful in some flaky environments anyway so here is the PR.
I hope you will find it useful too.

Also seems to be related to the following issues:
https://github.com/locustio/locust/issues/1787
https://github.com/locustio/locust/issues/1843

bioker:
Ah, have followed the issue history and noticed this feature request so linking it also:
https://github.com/locustio/locust/issues/1897

bioker:
This change represents how I have used the updated version: https://github.com/bioker/mysql-lab/commit/eedbb5a31f717acbc128ebf4ba3a05fe88d01458

cyberw:
Hmm... those values are already customizable, and I'm not sure they are worthy of a command line setting.

```
from locust import runners
runners.HEARTBEAT_INTERVAL = 2
```

bioker:
Ok, I see, thanks for the example

PR #1992, opened 2022-01-29
fix --html report in web mode

### LINKED ISSUE #1944: --html doesnt work in web mode
### Describe the bug

There is no any setting description about the html option in doc. Only said it "Store HTML report file". No any value example.
I tried "true", "html_prefix_like_csv" and it doesn't work. Can any one tell me what value should I set and where would be the report saved, and is there any way to customized the report html format?

### Environment

- OS: MacOS
- Python version: 3.9.0
- Locust version: 2.5.0
- Locust command line that you ran: locust
- Locust file contents (anonymized if necessary): i set html = report.html in locust.config

uddmorningsun:
If locust is started by `--headful` argument(web mode), `main_greenlet.join()` will serve forever. `get_html_report` will be inaccessable.

If running locust will be CTRL+C, `get_html_report` will not be received by catching the exception.

In order to keep simple, `get_html_report` is dealt with `shutdown()` according to `--html` argument value.

Signed-off-by: Chenyang Yan <memory.yancy@gmail.com>

Edit: Fixes #1944

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -120,7 +120,7 @@ class WebUI:
                 self.auth.init_app(self.app)
             else:
                 raise AuthCredentialsError(
-                    "Invalid auth_credentials. It should be a string in the following format: 'user.pass'"
+                    "Invalid auth_credentials. It should be a string in the following format: 'user:pass'"
                 )
         if environment.runner:
             self.update_template_args()
@@ -174,6 +174,8 @@ class WebUI:
                 self._swarm_greenlet.kill(block=True)
                 self._swarm_greenlet = None
             environment.runner.stop()
+            if self.greenlet is not None:
+                self.greenlet.kill(block=True)
             return jsonify({"success": True, "message": "Test stopped"})
 
         @app.route("/stats/reset")
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -397,7 +397,7 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
     stats_group.add_argument(
         "--html",
         dest="html_file",
-        help="Store HTML report file",
+        help="Store HTML report to file path specified",
         env_var="LOCUST_HTML",
     )
 
--- a/locust/main.py
+++ b/locust/main.py
@@ -457,6 +457,12 @@ def main():
         logger.info("Got SIGTERM signal")
         shutdown()
 
+    def save_html_report():
+        html_report = get_html_report(environment, show_download_link=False)
+        logger.info("writing html report to file: %s", options.html_file)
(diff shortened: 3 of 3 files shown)


cyberw:
command line help text is a bit verbose, "Store HTML report to file path specified" should be enough. 

If you havent already, please test this in "regular" mode (run a test from the gui, stop it from the gui) as well as using `--autostart`

uddmorningsun:
> "Store HTML report to file path specified"

It's OK

I've forgot stop button in web ui:sweat_smile:. But there's a problem in `locust.web.py:WebUI.stop`, only runner is stopped, but there has always been found greenlet process. If sending `/stop`, you will  see process hang up. Maybe add `self.greenlet.kill(block=True)`?

And, what do you think about dealing with `ctrl-c` for `get_html_report`

cyberw:
Good question. Unfortunately I'm not completely up to speed on those things, and I dont have time to dig into it, so your guess is as good as mine :) (I will of course be happy to review, once it is working)

> but there has always been found greenlet process

I dont understand that sentence...

uddmorningsun:
Sorry, my broken English😅 It means there's a bug with `/stop` in web.py. Process will hang up in terminal window while clicking stop button in web ui.

I updated, could you help me review it?

locust/web.py
```diff
@@ -174,6 +174,8 @@ def stop():
                 self._swarm_greenlet.kill(block=True)
                 self._swarm_greenlet = None
             environment.runner.stop()
+            if self.greenlet is not None:
+                self.greenlet.kill(block=True)
```
cyberw, line 178: Hmm... I'm worried this might cause issues with repeated gui runs. Have you tried that?

uddmorningsun: What does meaning of `repeated gui runs`? I guess that different `--web-port` to run?

I have tried that it's normal to start again after sending the stop button, even for different ports.

cyberw: > I have tried that it's normal to start again after sending the stop button, even for different ports.

That's all I meant :) Thanks!

cyberw:
Yea, that didnt work very well, causing the issue #1995 I'm reverting that part.

chy8803:
> Yea, that didnt work very well, causing the issue #1995 I'm reverting the greenlet killing part.

This PR caused the master pod restarting and resetting the data. After clicking Stop button UI, /stats/report becomes unaccessible since the process is stopped.

locust/web.py
uddmorningsun: Oops, it's my fault :neutral_face: for the issue #1995. Again, I understood the meaning of `with repeated gui runs`, run multiple tests in one running locust environment.

Issue #1994, opened 2022-02-03
'Tasks' section remains empty for html on v 2.7.0


eliezershindler:
I found this to be working on v2.5.1 but on 2.7.0, the 'tasks' section remains empty in the html report generated remains empty:
![image](https://user-images.githubusercontent.com/34885341/152351852-d31cfbd4-1251-4585-8aee-796006afdd9c.png)

cyberw:
@tyge68 Could I bother you to have a look?

tyge68:
@cyberw it works for me on the 2.7.1 
<img width="1091" alt="Screenshot 2022-02-03 at 14 58 35" src="https://user-images.githubusercontent.com/8066508/152357190-d6069ba9-259f-45c0-902e-5beacc227453.png">

@eliezershindler  Can you please retry with latest 2.7.1 ? if that still occurs, you could please provide more details how to reproduce it ?

eliezershindler:
It is still happening for me. Note I'm talking about the html file that is defined in locust.conf as html = results.html. It's a static file. I'm running a test from the command line, not looking at the web UI. The conf file looks like:

locustfile = locustfile.py
headless = true
users = 1
spawn-rate = 1
run-time = 300s
html = results.html
only-summary = true
csv = results.csv

I define the host elsewhere. When I open up the file in chrome, the 'Tasks' section is empty. Also tried Safari:
![image](https://user-images.githubusercontent.com/34885341/152397707-e9385f76-c283-4679-ae72-d8475a3a8803.png)

tyge68:
@eliezershindler so it's the offline report that is not working, I see it tries to get the tasks from the locust server, while it actually have the information in the HTML already. Not sure why it was then working already in the past.
Anyway the solution would be to load it from the data-tasks attribute

eliezershindler:
@tyge68 was that directed at me? If so, could you elaborate what the data-tasks attribute is?

tyge68:
@eliezershindler  I see this issue was introduced by this PR  (1), @EzR1d3r  do you remember why you had to change from initTasks to updateTasks ? it now doesn't work anymore in context of "saved" reports.

@cyberw would it be ok to store a constant in the report.html so it could be used to define if that is using static data from DOM or read from /tasks ?

(1) https://github.com/tyge68/locust/commit/6eeb5b82453d54f770f589b76202f7c5b9184b00

tyge68:
[...]

PR #1997, opened 2022-02-04
#1994 Fixing to fallback in case of local execution

tyge68:
In order to allow the rendering of Tasks on downloaded reports, it need to handle the case get /tasks is not fulfilled.

cc @cyberw @EzR1d3r


cyberw:
thx

PR #2001, opened 2022-02-06
Issue 1994 and static report bug

EzR1d3r:
Issue #1994 
A little corrected @tyge68 [fix](https://github.com/locustio/locust/commit/b14464c167da9b4939c65dbac2590e126ccaf1bd), now it seems to me more obvious the behavior that we expect in every case. Also fixed another bug: since now ratio is considered in dynamics, then after stopping ratio was considered for the case when there are no more active users, and this is clearly not what we want to see in the report.
@cyberw @eliezershindler

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -7,7 +7,7 @@ from .stats import sort_stats
 from .user.inspectuser import get_ratio
 from html import escape
 from json import dumps
-from .runners import MasterRunner
+from .runners import MasterRunner, STATE_STOPPED, STATE_STOPPING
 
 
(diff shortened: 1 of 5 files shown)


tyge68 (approved):
@EzR1d3r thanks, it looks good, I did a quick smoke test with basic.py example, and could see the downloaded report with last values for the final ratios.

cyberw:
Good stuff!

eliezershindler:
confirmed working on 2.7.3 thanks.

PR #2004, opened 2022-02-09
added tags/exclude_tags to web UI

JontiH:
My company has started using locust to loadtest stuff. its great! But it was annoying to contantly change locustfile to test different stuff, and tags didn't work in the UI. So I made some changes to get tags to change in UI between runs.

I tried to make it non-intrusive. The one big thing I added was turning include_in_web_ui from default true to default false. I didn't really understand why you only allow extra args that are both in include_in_web_ui and NOT in default configs, which included tags. So I got rid of that clause too.

In web.py, I just assume input is a string, and try to match it to a list or bool if the variable type corresponds to that. Otherwise I just accept set the value. I am maybe not thinking of some corner case, but it seems like the vars are all bool, lists, str or int, so I don't see it being such a big issue. Worst case the variable sent doesn't work, but that's how it was before.

Anyway, I am a noob when it comes to contributing to OSS, so I'm not sure what else I need to do to submit something. Feel free to change anything you want / tell me what to fix, I hope this is useful to you guys in some way!


JontiH:
also I can squash/rebase the commits if that's an issue

cyberw:
Hi! Yea, some of the implementation choices there were a bit random. Your changes dont seem unreasonable, but it is kind of a breaking change. I dont know if @Trouv , who originally implemented tags, might have some input?

There are some code style & unit tests that failed because of your changes so you'd need to fix that, but because Github Actions config was broken in master for a couple days, you need to rebase on master to see it.

I pushed your changes to a test branch: https://github.com/locustio/locust/runs/5131529689?check_suite_focus=true, you can see the errors there too.

Trouv:
It's been a while since I worked on it, but I went ahead and took a look.

As far as tags supporting types other than strings, I don't remember that being important, so it can probably go (though, as you said, technically a breaking change).

Splitting the string from the web UI on space probably isn't ideal, I could very easily see users wanting a space in their tag names (which I think they could deal with when running CLI by just using quotes around the tag name). It makes a little more sense to me to split by comma and comma+space.

I remember some of the weirder logic w/ tags being concerned with nested tasksets. I think maybe the logic around only filtering once might've had something to do with that, but I'm not sure. I'd just say make sure to be careful to not have any regressions with tags in nested tasksets here.

It looks like I was originally planning on allowing users to change tags without restarting locust, though at the time the use case I had in mind was running locust as a library. For some reason, we decided against it. Actually, you can see me talking about it in the last paragraph of this comment (as well as heyman's responses) here: https://github.com/locustio/locust/pull/1358#issuecomment-621964058. Food for thought.

cyberw:
👍 we have to at least add a couple unit tests for modifying tags. 3 cases spring to mind: standalone, distributed and as-a-library.

JontiH:
okay, I will fix the style issues.

@Trouv so I agree with you regarding commas vs spaces as a delimiter. I think I just used that because in your documentation you list `--tags tag1 tag2` and so I assumed that was the case. I can fix that.

As far as tags being both strings and lists,I did that because if you run locust with tags specified in the cli, then the value gets parsed as a list. But if you add the input blindly from from the webUI  (where the initial type is None so you would have nothing to go off of) then the type is a string. That's why in _filter_tasks_by_tags I check for both the input being a string and a list.

I did this because I wanted the code in web.py to be generically usable for any parsed_option going through the UI, and I wasn't sure what types they would have. So I see 3 options:

a) either we use what I have written now
b) we make it so that there is a specific case for tags (like hosts and spawn_rate), where they are always stored as a list (or None)
c) assume that all parsed_options that go through the webUI are acceptable to be stored as lists of strings.

option b seems less elegant and I don't see any reason why c) wouldn't work, but I would like to hear your thoughts on this.

Lastly, your point about nested tasksets is good. I didn't like the 
`        if getattr(self, "_tasks_filtered", False):
            return  # only filter once
        self._tasks_filtered = True
`
code, because it was attached to users, and so each user would only be able to filter once in their lifetime. But you are right, removing it leaves issues with possible loops.

I think a good idea would be to use this _tasks_filtered attribute, but reset it outside of the method before each filter job is called by env.py. That way, users can refilter each time the environment calls, but they also only filter once per recursive iteration, so there won't be any loops.

Let me know if all that makes sense, and I will keep working on it. Thank you so much for your replies!

JontiH:
> +1 we have to at least add a couple unit tests for modifying tags. 3 cases spring to mind: standalone, distributed and as-a-library.

would these all use the UI? Or should there be a cli/library based way to change tags too?

treussart:
Your feature seems very interesting to me. It would greatly improve the interest of the UI for the non-technical people.

.

cyberw:
> > +1 we have to at least add a couple unit tests for modifying tags. 3 cases spring to mind: standalone, distributed and as-a-library.
> 
> would these all use the UI? Or should there be a cli/library based way to change tags too?

Damn, I missed your reply. Yes, the standalone and distributed tests should use the UI. Look at test_main.py / StandaloneIntegrationTests.test_custom_arguments() for inspiration.

shedfull:
Just a +1 from me. I'd be very interested in having this feature in locust and could possibly help - although my Python skills are rudimentary

PR #2010, opened 2022-02-11
fixed load/users getting distributed to missing worker

### LINKED ISSUE #2008: Load being distributed to missing workers too Version 2.7.4.dev14
When running Locust in a distributed mode in Kubernetes, if the pod restarts then the worker status gets changed to missing. I can see that  spawning distributes the users equally to all workers, but in this case it distributes to missing workers as well which leads to reduced spwaning rate.
Kindly help!!
I have made some changes by adding Kubernetes support natively, have shared the details in slack channel. kindly suggest on how we can take this further.

radhakrishnaakamat:
fixed load/users getting distributed to the missing workers. 
I am new to open source commit and python. 
Please let me know if anything needs to be changed.
fixes #2008 #2010

### DIFF
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -16,6 +16,8 @@ from locust import User
 if TYPE_CHECKING:
     from locust.runners import WorkerNode
 
+HEARTBEAT_DEAD = -10
+
 
 # To profile line-by-line, uncomment the code below (i.e. `import line_profiler ...`) and
 # place `@profile` on the functions/methods you wish to profile. Then, in the unit test you are
@@ -199,6 +201,13 @@ class UsersDispatcher(Iterator):
         if len(self._worker_nodes) == 0:
             # TODO: Test this
             return
+        self._prepare_rebalance()
+
+    def prepare_rebalance(self) -> None:
+        """
+        This method will remove missing workers after the HEARTBEAT_DEAD interval has passed from getting users during distribution
+        """
+        self._worker_nodes = [w for w in self._worker_nodes if w.heartbeat > HEARTBEAT_DEAD]
         self._prepare_rebalance()
 
     def _prepare_rebalance(self) -> None:
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -16,8 +16,6 @@ from locust import User
 if TYPE_CHECKING:
     from locust.runners import WorkerNode
 
-HEARTBEAT_DEAD = -10
-
 
 # To profile line-by-line, uncomment the code below (i.e. `import line_profiler ...`) and
 # place `@profile` on the functions/methods you wish to profile. Then, in the unit test you are
@@ -201,13 +199,6 @@ class UsersDispatcher(Iterator):
         if len(self._worker_nodes) == 0:
             # TODO: Test this
             return
-        self._prepare_rebalance()
-
-    def prepare_rebalance(self) -> None:
-        """
-        This method will remove missing workers after the HEARTBEAT_DEAD interval has passed from getting users during distribution
-        """
-        self._worker_nodes = [w for w in self._worker_nodes if w.heartbeat > HEARTBEAT_DEAD]
         self._prepare_rebalance()
 
     def _prepare_rebalance(self) -> None:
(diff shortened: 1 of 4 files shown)


cyberw:
Can you add a unit test? You can override/shorten the timeout in the tests to reduce the time they take by using pytests mock feature (there should be some examples among the tests)

locust/test/test_dispatch.py
```diff
@@ -3273,6 +3273,29 @@ class User5(User):
...
+
+        worker_nodes = [WorkerNode(str(i + 1)) for i in range(3)]
+        worker_nodes[1].heartbeat = HEARTBEAT_DEAD - 1
+
+        users_dispatcher = UsersDispatcher(worker_nodes=worker_nodes, user_classes=user_classes)
+
+        users_dispatcher.new_dispatch(target_user_count=9, spawn_rate=3)
+        users_dispatcher._wait_between_dispatch = 0
+        users_dispatcher.prepare_rebalance()
+        list(users_dispatcher)
```
cyberw, line 3296: what does this line do? :)

locust/test/test_dispatch.py
```diff
@@ -3273,6 +3273,29 @@ class User5(User):
...
+        self.assertIs(len(users_dispatcher._worker_nodes), 2, "Expected worker 2 will be removed from worker nodes")
```
cyberw, line 3297: Can you also validate that the remaining workers get the right number of users?

radhakrishnaakamat:
It asserts to Count of active workers in dispatch.. Here the missing worker
should be removed after heartdead interval.. Added comment to code and
added more sensible message in assert

locust/test/test_dispatch.py
radhakrishnaakamat: added the test case

locust/test/test_dispatch.py
radhakrishnaakamat: re-organized the code. it validates the workers available for redistribution

locust/dispatch.py
```diff
@@ -16,6 +16,8 @@
 if TYPE_CHECKING:
     from locust.runners import WorkerNode
 
+HEARTBEAT_DEAD = -10
```
cyberw, line 19: import from runners instead?

locust/runners.py
```diff
@@ -59,6 +59,7 @@
 CPU_MONITOR_INTERVAL = 5.0
 HEARTBEAT_INTERVAL = 1
 HEARTBEAT_LIVENESS = 3
+HEARTBEAT_DEAD = -10
```
cyberw, line 62: Feels weird that this is a negative number?

locust/dispatch.py
radhakrishnaakamat: I tried that, but was getting error during initialization.
Will try the reverse tomorrow..

Could you help me with any resource on how to setup debugging in distributed mode.. That would be really helpful for me to contribute in a better way.. I am resorting to print statement for debugging(This is my first python project)

locust/runners.py
radhakrishnaakamat: It symbolic to time with no response, as current heartbeat logic subtracts 1 for each missed heartbeat. Didn't want to introduce another variable to maintain skipped heart beat.

locust/dispatch.py
cyberw: Sorry, I dont have much to contribute with debugging distributed. You know how to run an individual unit test, right?

radhakrishnaakamat: Yes, initially used to run tox.. later found out the way to run individually.

mboutet:
Hi @radhakrishnaakamat, can you link this PR to the corresponding issue #2008?

radhakrishnaakamat:
Fixes #<2010> #<2008>

radhakrishnaakamat:
@mboutet thanks for the pointers in #2008, I was able to fix it in the right places in a better way. 
@cyberw could you kindly review it once again, I have updated the tests and tox worked fine too. Thanks!!

cyberw:
👍👍

PR #2024, opened 2022-02-21
Added option for handling an exception from execute_task

mirkochip:
I'm willing to introduce this little change in order to handle the chain of exceptions raised when a Locust Task workflow would be manually driven.

A concrete example occurs when raising manually the `InterruptTaskSet`, to skip and reschedule immediately a task: this would lead to series of not handled exceptions, reaching back its parent caller, defined in `execute_task` function.

Such chain of exceptions is printed in stdout, with the following traceback:

```
[2022-02-21 12:26:46,925] venus/ERROR/common.base.load_testing_base: Returned HTTP status code 503: Service Temporary Unavailable
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/clients.py", line 21, in raise_for_status
    Response.raise_for_status(self)
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error for url: https://my_faulty_server.com

During handling of the above exception, another exception occurred:

  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/task.py", line 297, in run
    self.execute_next_task()
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/task.py", line 322, in execute_next_task
    self.execute_task(self._task_queue.pop(0))
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/task.py", line 436, in execute_task
    task(self.user)
locust.exception.InterruptTaskSet

The above exception was the direct cause of the following exception:

  File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/users.py", line 171, in run_user
    user.run()
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/users.py", line 139, in run
    self._taskset_instance.run()
  File "/home/mirkochip/miniconda3/envs/locust/lib/python3.9/site-packages/locust/user/task.py", line 307, in run
    raise RescheduleTaskImmediately(e.reschedule) from e
locust.exception.RescheduleTaskImmediately: True
2022-02-21T12:26:46Z <Greenlet at 0x7f514991dae0: run_user(<get_invoices.GetInvoices object at 0x7f51498be9a0)> failed with RescheduleTaskImmediately
```

Setting the flag, defined as class attribute `handle_exceptions = True` in the class containing the Locust task definition, would "replace" the traceback reported above with this single line of log:

`[2022-02-21 12:27:11,018] venus/INFO/root: Exception InterruptTaskSet occurred.`

This would make the console log more fluid and cleaner.

### DIFF
--- a/locust/test/test_locust_class.py
+++ b/locust/test/test_locust_class.py
@@ -8,6 +8,7 @@ from locust.env import Environment
 from locust.exception import CatchResponseError, RescheduleTask, RescheduleTaskImmediately, StopUser
 
 from .testcases import LocustTestCase, WebserverTestCase
+from ..user.task import DefaultTaskSet
 
 
 class TestTaskSet(LocustTestCase):
@@ -430,6 +431,46 @@ class TestTaskSet(LocustTestCase):
 
         MyUser(Environment()).run()
         self.assertTrue(MyTaskSet.raised_attribute_error)
+
+    def test_execute_task_raise_exception(self):
+        class MySubTaskSet(TaskSet):
+            constant(1)
+
+            @task()
+            def a_task(self):
+                self.user.sub_locust_task_executed = True
+                self.interrupt()
+
+        class MyTaskSet(DefaultTaskSet):
+            tasks = [MySubTaskSet]
+
+        self.locust.tasks = MyTaskSet.tasks
+        self.sub_locust_task_executed = False
+        loc = MyTaskSet(self.locust)
+
+        with self.assertRaises(RescheduleTaskImmediately):
+            loc.execute_task(loc.get_next_task())
+        self.assertTrue(self.locust.sub_locust_task_executed)
+
+    def test_execute_task_exception_handled(self):
+        class MySubTaskSet(TaskSet):
+            constant(1)
+
+            @task()
+            def a_task(self):
+                self.user.sub_locust_task_executed = True
+                self.interrupt()
+
+        class MyTaskSet(DefaultTaskSet):
+            tasks = [MySubTaskSet]
+
+        self.locust.tasks = MyTaskSet.tasks
+        self.locust.handle_exceptions = True
+        self.sub_locust_task_executed = False
+        loc = MyTaskSet(self.locust)
+
+        loc.execute_task(loc.get_next_task())
+        self.assertTrue(self.locust.sub_locust_task_executed)
 
 
 class TestLocustClass(LocustTestCase):
--- a/locust/user/task.py
+++ b/locust/user/task.py
@@ -446,7 +446,10 @@ class DefaultTaskSet(TaskSet):
             else:
                 # task is a function
                 task(self.user)
-        except (InterruptTaskSet, ) as e:
+        except (
+            InterruptTaskSet,
+            RescheduleTaskImmediately,
+        ) as e:
             if not self.user.handle_exceptions:
                 raise
             logging.info(f"Exception of type {type(e).__name__} occurred while executing the task.")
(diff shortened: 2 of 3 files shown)


cyberw:
I'm not sure the abbreviated version is better, and it would need unit tests. 

I personally never use TaskSets, but if you can find someone who does and they agree this is a good thing then maybe, but as it stands I'm -1 for this change.

mirkochip:
> 

Hi @cyberw, I would have, ideally, solved this soft/cosmetic "issue" overriding `sys.excepthook` just in the test side of the things, without touching `locustio`. Unfortunately this wasn't possible because it looks like it doesn't work in imported Python modules (At the time when `sys.excepthook` is called, all module members are already set to `None`, please checkout here: https://bugs.python.org/issue11705)

If this change could still make sense, I can provide unit tests and updated changelog along with it: I just wanted to wait for a "triage" first, in order to understand if there's still a little chance to proceed further with it.

Any other option which could lead to suppress and handle somehow the chained exceptions is more than welcome!

Thank you!

mirkochip:
> > 
> 
> Hi @cyberw, I would have, ideally, solved this soft/cosmetic "issue" overriding `sys.excepthook` just in the test side of the things, without touching `locustio`. Unfortunately this wasn't possible because it looks like it doesn't work in imported Python modules (At the time when `sys.excepthook` is called, all module members are already set to `None`, please checkout here: https://bugs.python.org/issue11705).
> 
> If this change could still make sense (to me yes, since I could get rid of tons of logs appended into stdout, for test running within a no well defined time frame), I can provide unit tests and updated changelog along with it: I just wanted to wait for a "triage" check first, in order to understand if there's still a little chance to proceed further with it.
> 
> Any other option which could lead to suppress and handle somehow the chained exceptions is more than welcome!
> 
> Thank you!



> I'm not sure the abbreviated version is better

With this solution, the "abbreviated version" is gonna be anyway kept optional, and by default will be disabled.

cyberw:
> If this change could still make sense (to me yes, since I could get rid of tons of logs appended into stdout, for test running within a no well defined time frame), I can provide unit tests and updated changelog along with it: I just wanted to wait for a "triage" check first, in order to understand if there's still a little chance to proceed further with it.

👍 Makes sense! What you could do is just overwrite DefaultTaskSet/TaskSet.execute_task(), but that's not very pretty I guess..

mirkochip:
> +1 Makes sense! What you could do is just overwrite DefaultTaskSet/TaskSet.execute_task(), but that's not very pretty I guess..

Not very pretty indeed, but it could fit for my current needs, thanks.

In case this change would not be relevant for some other Locust users, I can close this PR.

cyberw:
> > +1 Makes sense! What you could do is just overwrite DefaultTaskSet/TaskSet.execute_task(), but that's not very pretty I guess..
> 
> Not very pretty indeed, but it could fit for my current needs, thanks.
> 
> In case this change would not be relevant for any other Locust users, I can close this PR.

I think that makes sense. If we ever do a rewrite of TaskSets (imho they are unnecessarily complex atm) maybe we will simplify this somehow.

mirkochip:
> I think that makes sense. If we ever do a rewrite of TaskSets (imho they are unnecessarily complex atm) maybe we will simplify this somehow.

Alright, just in case.. I added unit tests and the related now documented `handle_exceptions` default attribute.

PR #2025, opened 2022-02-21
Ensure users are distributed evently across hosts during ramp up

cyberw:


cyberw:
@mboutet @DennisKrone What do you guys think?

mboutet:
Makes sense.

I think you'll also need to add this new sorting logic to:
https://github.com/locustio/locust/blob/d34c680b4251e4ab2ca7ca74b8ab0d9de847f053/locust/dispatch.py#L187-L212

DennisKrone:
Looks good!

PR #2033, opened 2022-02-25
Add test_stopping event

### LINKED ISSUE #2031: Introduce test_stopping event
I noticed multiple times that it would be good to have additional event test_stopping to be fired before actual stopping procedure to notify locust test that teardown procedure is about to happen - unlike test_stopped that is fired after all users are stopped.

marcinh:
Adding test_stopping event - fired before stopping users.
It resolves #2031

### DIFF
--- a/examples/test_data_management.py
+++ b/examples/test_data_management.py
@@ -7,9 +7,10 @@
 # 4. User start
 # 5. Inside a task
 # ...
-# 6. User stop
-# 7. Test run stop
-# 8. (not shown in this example) Locust quit
+# 6. Test run stopping
+# 7. User stop
+# 8. Test run stop
+# 9. (not shown in this example) Locust quit
 #
 # try it out by running:
 #  locust -f test_data_management.py --headless -u 2 -t 5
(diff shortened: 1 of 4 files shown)


locust/test/test_runners.py
```diff
@@ -2224,11 +2230,15 @@ def my_task(self):
         with mock.patch("locust.rpc.rpc.Server", mocked_rpc()) as server:
             master = self.get_runner(user_classes=[TestUser])
 
-            run_count = [0]
+            run_count = [0, 0]
```
cyberw, line 2233: Using a list here is more or less a hack for modifying values outside the function. I think this can be done in a more obvious way. even using a (properly named) global is preferable.

cyberw:
Looks good apart from my comment about the test. Can you add this event to this example? https://github.com/locustio/locust/blob/master/examples/test_data_management.py

examples/test_data_management.py
```diff
@@ -60,11 +60,16 @@ def _(environment, **_kwargs):
         ).json()["data"]
 
 
-@events.test_stop.add_listener
+@events.test_stopping.add_listener
```
cyberw, line 63: Could you update the comment at the start of the file too!

marcinh: like this?
```
# 6. Test run stopping
# 7. User stop
# 8. Test run stop
# 9. (not shown in this example) Locust quit
````

PR #2039, opened 2022-03-03
Remove Host dialog box if env set

treussart:
A feature to choose if the Host form is displayed if env var (LOCUST_HOST) is set.

My requirement is that I am using Locust with an SDK for my target system and it is he who sets the target via an environment variable. 
So I don't want the Locust user to be able to change the target in the UI, because that won't have any effect.

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -527,6 +527,15 @@ def ui_extra_args_dict(args=None) -> Dict[str, str]:
     locust_args = default_args_dict()
 
     parser = get_parser()
+    parser.add_argument(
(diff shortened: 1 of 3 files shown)


cyberw:
Niche feature with just cosmetic effect, not worth another parameter, sorry.

treussart:
This may not be a useful case for you, but think about the community that uses your tool, I find this answer not very open-minded.

cyberw:
I get that you feel this way, but Locust is hairy enough as it is with tons of parameters and legacy features few people use. Again, sorry.

Issue #1947, opened 2021-11-29
Locust does not stop all users


marcinh:
### Describe the bug
When there is more than 1 user defined in locust file, Locust does not stop users correctly - there are always 10%-20% left in running state. I'm adding the scenario, that causes problem - there is a background task that periodically checks the number of users and print sit to console.

### Expected behavior
All users stopped correctly

### Actual behavior
Not all users stopped

### Steps to reproduce
Run scenario attached (containing 2 user classes)

### Environment

- OS: Windows/MAC
- Python version: 3.9
- Locust version: 2.5
- Locust command line that you ran: locust -f bug_scenario.py
- Locust file contents (anonymized if necessary):

```
from logging import getLogger
from random import randrange

from gevent import sleep, spawn
from locust import task, constant_pacing, SequentialTaskSet, events, User
from locust.env import Environment

logger = getLogger('Scenario')


def add_request_event(request_type, name, duration, length=0, response=None, context=None, exception=None):
    events.request.fire(
        request_type=request_type,
        name=name,
        response_time=duration,
        response_length=length,
        response=response,
        context=context or {},
        exception=exception
    )


def run_environment_checker(environment: Environment):
    while True:
        user_count = environment.runner.user_count
        logger.info(f'User count: {user_count}')
        sleep(3)


@events.init.add_listener
def hadle_environment_checker(environment: Environment, **_):
    spawn(lambda: run_environment_checker(environment))


class TestScenario(SequentialTaskSet):

    def __init__(self, parent):
        super().__init__(parent)

    def on_start(self):
        logger.info('on_start')

    @task
    def open_homepage(self):
        add_request_event(
            request_type='PUT',
            name='3d',
            duration=randrange(1000),
            length=randrange(1000)
        )

    @task
    def open_dashboard(self):
        add_request_event(
            request_type='GET',
            name='2d',
            duration=randrange(1000),
            length=randrange(1000)
        )

    @task
    def get_open_positions(self):
        add_request_event(
            request_type='POST',
            name='1d',
            duration=randrange(1000),
            length=randrange(1000)
        )

    def on_stop(self):
        logger.info('on_stop')


class ExampleUser(User):
    tasks = [TestScenario]
    wait_time = constant_pacing(10)


class ExampleUser2(ExampleUser):
    pass
```

marcinh:
So I updated on_start and on_stop method to log when they're called, and in this casescenario on_stop is not called for all users and additionally after hitting stop button I can see increased number of request.

[...]

PR #2041, opened 2022-03-04
Bugfix/1947 locust does not stop all users

marcinh:
Change a way of selecting users to stop
Add unit test

It resolves https://github.com/locustio/locust/issues/1947


cyberw:
Brilliant, thanks!

PR #2059, opened 2022-03-27
Support sharing connection pools between users

### LINKED ISSUE #1925: Make each virtual user to share same connections

- I need to simulate the real load on server. If I use mutiple connections, it will make the server to test got extra unnecessary load than real environment, which leads to unprecise result.

- Make each virtual user to share same connections.

- write a SDK to solve it.

### Additional context

- I found an old un-solved issue:https://github.com/locustio/locust/issues/334. about it.

miedzinski:
This PR adds support for sharing connection pools between users. It is done by exposing lower level pool (`urllib3.PoolManager` for `HttpUser`, `geventhttpclient.client.HTTPClientPool` for `FastHttpUser`).

Example usage:

```python
from geventhttpclient.client import HTTPClientPool
from locust import FastHttpUser, HttpUser
from urllib3 import PoolManager


# Both users will be limited to 10 concurrent connections at most.

class UserA(HttpUser):
    pool_manager = PoolManager(maxsize=10, block=True)


class UserB(FastHttpUser):
    client_pool = HTTPClientPool(concurrency=10)
```

Fixes #1925.

### DIFF
--- a/docs/increase-performance.rst
+++ b/docs/increase-performance.rst
@@ -42,7 +42,7 @@ FastHttpUser class
 --------------------
 
 .. autoclass:: locust.contrib.fasthttp.FastHttpUser
-    :members: network_timeout, connection_timeout, max_redirects, max_retries, insecure
+    :members: network_timeout, connection_timeout, max_redirects, max_retries, insecure, concurrency, client_pool
 
 
 FastHttpSession class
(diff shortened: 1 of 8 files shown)


cyberw:
Cool stuff. A few test cases would be nice ideally one that actually verifies somehow that connections are really pooled. And we’d need to add it to the docs somewhere.

miedzinski:
@cyberw PTAL.

I have no idea why mypy reported that error, because I didn't touch `HttpSession.request_name`, but adding a type hint fixed it.

miedzinski:
Now black is failing due to https://github.com/psf/black/issues/2964.

cyberw:
Thx! I think merging will get us the fix needed for >3.6 builds.

PR #2062, opened 2022-03-29
Move lint tests to their own tox environments

kurtmckee:
In this model, linters are only run once.


cyberw:
Can you show me where the linters are actually executed? I dont see it in the log here on GH so I think maybe something is missing...

kurtmckee:
@cyberw I missed the CI update on the first attempt. I've added the linters to `tests.yaml` so they'll run as separate checks.

cyberw:
Awesome, thanks!

PR #2070, opened 2022-04-11
modernized build

mgor:
i mentioned in a previous PR #2040 that there was problems adding locust as a git dependency; pip or pip-tools got confused that there was a `pyproject.toml`, but no `build-system` section available. all it contained was settings for `black`.

there have been other projects having the [same problem](https://snarky.ca/what-the-heck-is-pyproject-toml/).

the underlying problem seems to have been fixed now, so it is indeed possible to have locust as a git dependency (which I realized after implementing this...). but there are probably other benefits of moderizing the build setup to be PEP517 and PEP518 compatible.

since most stuff already was present in `setup.cfg`, it was quite an easy transition to move away from `setup.py`, however editable installs is not yet supported by `pyproject.toml` and hence a bare bone `setup.py` is needed.

most work had to be done in the github workflow and tox settings.

regarding the changes of `command` to `python -m <module>` in `test.yml`:
https://snarky.ca/why-you-should-use-python-m-pip/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.9-slim as base
 
 FROM base as builder
-RUN apt update && apt install -y git 
+RUN apt-get update && apt-get install -y git 
 # there are no wheels for some packages (geventhttpclient?) for arm64/aarch64, so we need some build dependencies there
 RUN if [ -n "$(arch | grep 'arm64\|aarch64')" ]; then apt install -y --no-install-recommends gcc python3-dev; fi
 RUN python -m venv /opt/venv
--- a/locust/test/test_runners.py
+++ b/locust/test/test_runners.py
@@ -1213,7 +1213,7 @@ class TestMasterWorkerRunners(LocustTestCase):
                     0, test_shape.get_current_user_count(), "Shape is not seeing stopped runner user count correctly"
                 )
             self.assertDictEqual(master.reported_user_classes_count, {"FixedUser1": 0, "FixedUser2": 0, "TestUser": 0})
-            sleep(0.5)
+            sleep(1.0)
             self.assertEqual(STATE_STOPPED, master.state)
 
     def test_distributed_shape_with_stop_timeout(self):
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ description = Developer friendly load testing framework
 long_description = file: README.md
 long_description_content_type = text/markdown
 license = MIT
-obsoletes-dist = locustio
+obsoletes_dist = locustio
 classifiers =
     Topic :: Software Development :: Testing :: Traffic Generation
     Development Status :: 5 - Production/Stable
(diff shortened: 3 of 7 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:3.9-slim as base
 
 FROM base as builder
-RUN apt update && apt install -y git 
+RUN apt-get update && apt-get install -y git
```
mgor, line 4: to get rid of:
```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

locust/test/test_runners.py
```diff
@@ -1213,7 +1213,7 @@ def tick(self):
                     0, test_shape.get_current_user_count(), "Shape is not seeing stopped runner user count correctly"
                 )
             self.assertDictEqual(master.reported_user_classes_count, {"FixedUser1": 0, "FixedUser2": 0, "TestUser": 0})
-            sleep(0.5)
+            sleep(1.0)
```
mgor, line 1216: test would fail occasionally with state being `stopping` instead of `stopped`.

setup.cfg
```diff
@@ -11,7 +11,7 @@ description = Developer friendly load testing framework
 long_description = file: README.md
 long_description_content_type = text/markdown
 license = MIT
-obsoletes-dist = locustio
+obsoletes_dist = locustio
```
mgor, line 14: `obsoletes-dist` is deprecated.

setup.cfg
```diff
@@ -35,6 +35,22 @@ packages = find:
...
+    pyzmq >=22.2.1
+    geventhttpclient >=1.5.1
+    ConfigArgParse >=1.0
+    psutil >=5.6.7
+    Flask-BasicAuth >=0.2.0
+    Flask-Cors >=3.0.10
+    roundrobin >=0.0.2
+    typing-extensions >=3.7.4.3
+    Jinja2 <3.1.0
+    pywin32;platform_system=='Windows'
```
mgor, line 53: i was unable to get it working with `pywin32` being an  `extras_require` as in the `setup.py` case, however this *should* be the same(?). pywin32 only needed when running on windows.
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

cyberw:
LGTM. Have you tested it on windows? I have a windows box somewhere, but I can't be bothered to fire it up :)

tox.ini
```diff
@@ -24,7 +25,8 @@ allowlist_externals =
     timeout
     grep
 commands =
-    coverage run -m unittest discover []
+    python3 -m pip install .
```
cyberw, line 28: Why is this needed? (probably a stupid question...)

mgor: there's no stupid questions :)

tox would run `python setup.py install` for building a sdist, and since we'd like to build via pyproject.toml, `skipsdist = True` was added, and the dependencies installed with pip install.

see https://stackoverflow.com/questions/18962403/how-do-i-run-tox-in-a-project-that-has-no-setup-py

mgor:
> LGTM. Have you tested it on windows? I have a windows box somewhere, but I can't be bothered to fire it up :)

If you mean the changed way pywin32 is included?

I could've actually just tested it myself.. without writing an obscure comment like that, my bad.

on "linux" (Windows 10 WSL2 with ubuntu 21.04):
```➜ pip install .
➜ pip freeze | grep pywin32
➜ 
```

on windows 10:
```
(locust-venv) C:\NoBackup\locust [feature/pep_517_518 ≡]> pip3 install .
(locust-venv) C:\NoBackup\locust [feature/pep_517_518 ≡]> pip3 freeze | Select-String pywin32

pywin32==303
```

cyberw:
I’m gonna give Heyman a day or two to chime in, but it looks all good to me!

heyman:
Looks good to me 👍

PR #2072, opened 2022-04-13
Ensure `test_start` is run to completion on worker

### LINKED ISSUE #1986: `test_start` event triggered multiple times on workers
### Describe the bug

When running locust in distributed mode, the `test_start` handler gets triggered every time new users are spawned, rather than once running once at the start of the test.

This only seems to occur if there is a `test_start` handler which takes longer than the interval between spawn events (I think 1s?).

This is problematic if the test needs to load test data from an external source and share it between the users.

To give an example of a concrete use case:
- I want the test to access load logs from an external source and replay them.
- These logs are valid for the lifetime of a test (rather than process or user) - so they can be shared between users on a worker.
- When I trigger the test the log data-store gets hammered by duplicated requests (load testing the wrong thing! :sweat_smile:).

### Expected behavior

The `test_start` event handlers should be triggered once when the test begins, allowing e.g. expensive test data loading operations to occur only once.

### Actual behavior

The `test_start` event handlers are triggered multiple times, with all but the final handler being killed prematurely.

### Steps to reproduce

Given the following locustfile:

```python
import time

from locust import HttpUser, events, runners, task


COUNT = 0

def test_start_worker(environment, **kwargs):
    if not isinstance(environment.runner, runners.WorkerRunner):
        return
    global COUNT
    COUNT += 1
    print(f"'test_start' triggered {COUNT} times.")
    time.sleep(1)  # Same result with `gevent.sleep`. Less than 1 second works fine, >=1 second causes the described behaviour.


class HelloWorldUser(HttpUser):
    @task
    def hello_world(self):
        self.client.get("/hello")
        self.client.get("/world")
```

And the following `docker-compose.yml`
```yaml
version: '3'

services:
  master:
    image: locustio/locust:2.5.1
    ports:
    - "8089:8089"
    volumes:
    - ./:/mnt/locust
    command: -f /mnt/locust/locustfile.py --master -H http://master:8089 --headless -u 10 -r 1 --run-time 1m --expect-workers 1

  worker:
    image: locustio/locust:2.5.1
    volumes:
    - ./:/mnt/locust
    command: -f /mnt/locust/locustfile.py --worker --master-host master
```

Run
```shell
docker-compose up -d
```

And the logs for the worker will display:

```
worker_1  | [2022-01-27 11:56:19,677] 26ff2afeca30/INFO/locust.main: Starting Locust 2.5.1
worker_1  | 'test_start' triggered 1 times.
worker_1  | 'test_start' triggered 2 times.
worker_1  | 'test_start' triggered 3 times.
worker_1  | 'test_start' triggered 4 times.
worker_1  | 'test_start' triggered 5 times.
worker_1  | 'test_start' triggered 6 times.
worker_1  | 'test_start' triggered 7 times.
worker_1  | 'test_start' triggered 8 times.
worker_1  | 'test_start' triggered 9 times.
worker_1  | 'test_start' triggered 10 times.
```

### Environment

- OS: Ubuntu 21.10
- Python version: 3.9.9
- Locust version: 2.5.1
- Locust command line that you ran: As described by `docker-compose.yml`
- Locust file contents (anonymized if necessary): Shown above
- docker-compose version 1.27.4
- Docker version 20.10.7

mboutet:
Fixes #1986

I tested with the steps indicated in https://github.com/locustio/locust/issues/1986#issue-1116128138 and only a single `'test_start' triggered 1 times.` is printed.

TODO:
- [x] Write a test case

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1102,15 +1102,6 @@ class WorkerRunner(DistributedRunner):
         """
         self.target_user_classes_count = user_classes_count
         self.target_user_count = sum(user_classes_count.values())
-        if self.worker_state != STATE_RUNNING and self.worker_state != STATE_SPAWNING:
-            self.stats.clear_all()
-            self.exceptions = {}
-            self.cpu_warning_emitted = False
-            self.worker_cpu_warning_emitted = False
-            self.environment._filter_tasks_by_tags()
-            self.environment.events.test_start.fire(environment=self.environment)
-
-        self.worker_state = STATE_SPAWNING
 
         for user_class in self.user_classes:
             if self.environment.host:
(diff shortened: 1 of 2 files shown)


mboutet:
@jacksmith15 FYI

cyberw:
Nice! Thanks!

PR #2082, opened 2022-04-28
Add table linkage, you can see the data of the three tables at the sa…

helloNice:

During the test, I want to see the corresponding values of the three tables at the same time. Now I can only use the mouse to point to the corresponding table one by one, which is very inconvenient.

I would like to point to any of the tables and see the values in the three graphs at the same time.

I saw that the table we used was echart, and I checked the api document for the multi-table linkage display function.
### Additional context

I want locust can be like this.
image


huangjj27:
The instant charts have linkages, but the "download report" 's charts have no linkages.

PR #2083, opened 2022-04-29
FastHttpUser improvements (including a rename of parameter "url" to "path")

mgor:
Sorry for the long "intro", which might seem irrelevant at first... but bear with me :)

In "my" project I had a case where I wanted to create a group of [async] requests, by running them in their own `greenlet`.
Normally it uses the `requests` client via `HttpUser`, but the way `requests` is using the `ConnectionPool` from `urllib3` is not thread-safe[0], which caused the requests to run more or less sequential instead.

This lead me to using a `FastHttpSession` for each async request, which worked great for that part. But instead ran into some typing issues, e.g.:

```python
class FastResponse(CompatResponse):
  headers = None
...
```
Since `geventhttpclient` isn't typed, this causes `mypy` to think that `headers` is always `None`, causing it to get type `Never`, making it hard to actually use these attributes without using `type: ignore`-comments.

The other ~problem~ inconvenience I had, was that `locust.contrib.fasthttp.FastResponse` was so different from `requests.models.Response`, causing "my" code that handles the response (for logging purposes) to get a bunch of special cases depending on what type of response it was. I've tried my best to make it easier to use `FastHttpSession` as a drop-in replacement for `HttpSession`.

One thing that I haven't changed, is that `locust.clients.HttpSession.request` has `url` as argument, where as `locust.contrib.fasthttp.FastHttpSession.request` has `path`. I would love to change that... but thought that it might be a too big breaking change(?).

Also, once again I had problems with `test_runners::test_distributed_shape_with_fixed_users`. I increased it in a previous PR from `0.5` to `1.0`, but seems like it sometimes is not enough, so now I changed so `master.status` is polled in a `gevent.Timeout` with 3 seconds, instead of having a static sleep so the test *shouldn't* fail when it sometimes takes a little longer. That check would take at minimum 0.1 seconds and at maximum 3 seconds when it's actually is a problem.

[0] https://github.com/psf/requests/issues/1871

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -139,7 +139,7 @@ class FastHttpSession:
     def request(
         self,
         method: str,
-        path: str,
+        url: str,
         name: str = None,
         data: str = None,
         catch_response: bool = False,
@@ -156,7 +156,7 @@ class FastHttpSession:
         Returns :py:class:`locust.contrib.fasthttp.FastResponse` object.
 
         :param method: method for the new :class:`Request` object.
-        :param path: Path that will be concatenated with the base host URL that has been specified.
+        :param url: path that will be concatenated with the base host URL that has been specified.
             Can also be a full URL, in which case the full URL will be requested, and the base host
             is ignored.
         :param name: (optional) An argument that can be specified to use as label in Locust's
@@ -179,7 +179,7 @@ class FastHttpSession:
             content will not be accounted for in the request time that is reported by Locust.
         """
         # prepend url with hostname unless it's already an absolute URL
-        url = self._build_url(path)
+        built_url = self._build_url(url)
 
         start_time = time.time()  # seconds since epoch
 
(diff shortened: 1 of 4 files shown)


mgor:
ehm, really strange that two test cases in `test_dispatch.py` failed, only on python 3.7... i couldn't see that it would be related to any of the changes?

heyman:
Nice! This looks like an improvement to me.

> One thing that I haven't changed, is that locust.clients.HttpSession.request has url as argument, where as locust.contrib.fasthttp.FastHttpSession.request has path. I would love to change that... but thought that it might be a too big breaking change(?).

I agree that it would be a nice change. As long as we don't do it in a patch release, and clearly document it in the Docs' Changelog, I think it's worth doing. It should be a simple fix changing`path` to `url` for anyone using `request()` with keyword arguments (code calling `request()` without keyword args for `method` and `path` should be compatible).

cyberw:
This is really odd. The `timeout` keyword stopped working now, but I cant understand why. It broke with the new release of gevenhttpclient, i see no relevant changes there. https://github.com/locustio/locust/runs/6991527717?check_suite_focus=true

mgor:
Strange. I can have a look on Monday, on my way to our midsommer destination right now.

cyberw:
Awesome, thanks. Glad midsommar!

cyberw:
As far as I can tell, geventhttpclient never supported `timeout` as a parameter, but it just happened to not complain about it earlier (just ignoring it). I will remove its use from our tests.

mgor:
oh crap, I totally forgot to have a look at this today as promised (came back to a full calendar). 
I can take a quick look now.

cyberw:
I think I have it figured out, no worries :)

mgor:
seems like this is the commit that caused `timeout` to slip-through in `urlopen` to `_make_request`.
`**kwargs` was added to the call to `_make_request`, but that method does not have a "catch-all" `**kwargs` in the arguments.

https://github.com/geventhttpclient/geventhttpclient/commit/20f85d7a99911adb4dcb36dfafc3887e2f3c6ebd#diff-89faf154761b75cf1fc552eb81207dfe2bc3adc61329e93dea127ca65b458f58R430

cyberw:
👍 

Hmm... I guess at the very least I should bump the minor version of geventhttpclient. Or should we do something else?

cyberw:
@gwik was it intentional that geventhttpclient should ignore extra parameters?

Issue #2075, opened 2022-04-19
Output is not saved to CSV when using LoadTestShape


eldaduzman:
### Describe the bug
When I use `LoadTestShape` the output is not fully saved to the csv file.

Running the following script:

Then open the file 1_stats_history.csv

### Expected behavior
Aggregated results per second

### Actual behavior
Only one second is listed
```
Timestamp,User Count,Type,Name,Requests/s,Failures/s,50%,66%,75%,80%,90%,95%,98%,99%,99.9%,99.99%,100%,Total Request Count,Total Failure Count,Total Median Response Time,Total Average Response Time,Total Min Response Time,Total Max Response Time,Total Average Content Size

1650379714,0,,Aggregated,137.100000,0.000000,320,480,560,620,1100,1600,2500,3100,4000,4100,4100,4894,0,320.0,520.2181464853838,185.3720999788493,4064.4731000065804,16707.96240294238

```

### Environment

- OS:  Windows
- Python version: 3.10.2
- Locust version:2.8.6
- Locust command line that you ran: locust -f locust\locustfile.py  --headless --csv=1
- Locust file contents (anonymized if necessary):
```
"""locust file"""
import math
from typing import Generator

from locust import HttpUser, task, constant, LoadTestShape


class WebsiteUser(HttpUser):
    """users container"""

    wait_time = constant(1)
    host = "https://google.com"
    data: Generator

    @task(1)
    def simple_get(self):
        """performs get"""
        self.client.get("")


class StepLoadShape(LoadTestShape):
    """
    A step load shape


    Keyword arguments:

        step_time -- Time between steps
        step_load -- User increase amount at each step
        spawn_rate -- Users to stop/start per second at every step
        time_limit -- Time limit in seconds

    """

    step_time = 10
    step_load = 50
    spawn_rate = 50
    time_limit = 50

    def tick(self):
        run_time = self.get_run_time()

        if run_time > self.time_limit:
            return None

        current_step = math.floor(run_time / self.step_time) + 1
        return (current_step * self.step_load, self.spawn_rate)

```

max-rocket-internet:
Sorry for the delay (I have too many github notifications), I will look next week 🙂

max-rocket-internet:
I made a [quick PR](https://github.com/locustio/locust/pull/2085) but think it's too simple to work 🤔

PR #2085, opened 2022-05-02
Move CSV stats printer gevent spawn up a few lines

max-rocket-internet:
Resolves https://github.com/locustio/locust/issues/2075

The problem is that if `--headless` is used with `LoadTestShape` then the  `stats_csv_writer` greenlet is never spawned. By moving the `gevent.spawn(stats_csv_writer.stats_writer)` part up a few lines I think it solves the problem? Let's wait for the tests 🙏


max-rocket-internet:
> All checks have passed

Alllriiight 🎉

cyberw:
Nice

PR #2088, opened 2022-05-04
print_stats table width fix for #2084

mgor:
the width of the `print_stats` table was increased in a previous PR, which resulted in line wraps in terminal sizes that previously didn't have it (see discussion in #2084).

this PR tries to solve that. in the case of a big terminal window causing  `STATS_NAME_WIDTH` to be at max 80 characters, the table is now 1 character less wide than the original implementation (before #2084), 159 columns vs 158 columns.

this is an comparision of the output of the tables with this PR and the original implementation with a 119 column wide terminal:
image

the original implementation would allow a request name of maximum 34 characters in the `DELETE` request type case, and 37 characters in the `GET` request type case.

with the new implementation, the maximum request name length, with 119 columns, would always be 32 characters.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -528,8 +528,8 @@ class StatsEntry:
             "%-"
             + str(STATS_TYPE_WIDTH)
             + "s %-"
-            + str(STATS_NAME_WIDTH - STATS_TYPE_WIDTH)
-            + "s %7d %12s | %7d %7d %7d %7d | %7.2f %11.2f"
+            + str((STATS_NAME_WIDTH - STATS_TYPE_WIDTH) + 4)
+            + "s %7d %12s |%7d %7d %7d%7d | %7.2f %11.2f"
         ) % (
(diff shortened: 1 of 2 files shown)


mgor:
@cyberw, i have another commit (not pushed), that would give an additional 4 characters in request name (totalting to a maximum request name at a fixed 36 characters in a 119 column wide terminal), by "compacting"[0] some other columns, and changing `Median` to its abbrevation `Med`.

[0] affects when `Avg` or `Med` gets maximum values of 7 fixgures

cyberw:
LGTM (this time I tested it properly with my setup :P )

I think compacting the response time columns as you suggested makes sense. Response times in excess of 100 seconds is kind of rare (at least more rare than request names longer than 32 characters)

Please push it here, and if it looks good I'll merge it as one PR.

mgor:
great! here comes an additional 4 characters for request name :)

cyberw:
Nice! Thanks!

Issue #2091, opened 2022-05-05
Users with `fixed_count` not being relocated after rebalance


andydunstall:
### Describe the bug
Running Locust with multiple workers, where `UserA` has `fixed_count = 1` and `UserB` has no weight attributes.

As expected when starting 4 workers and 8 users, 1 `UserA` is started with 7 `UserB`s. Though when the worker `UserA` is running on is terminated, Locust rebalances such that theres now 8 `UserB`s and no `UserA`s. The same issue seems sometimes occur when just scaling up the number of workers, triggering a rebalance.

Adding some logs to the code, looks like `self._try_dispatch_fixed` is `False` on the rebalance ([dispatch.py#L387](https://github.com/locustio/locust/blob/master/locust/dispatch.py#L387)) - which is only set to `True` when it first runs or when the number of users is reduced.

I'm happy have a go at fixing this - though adding issue for visibility.

### Steps to reproduce
 
```python
# locustfile.py
class UserA:
  fixed_count = 1
  # ...

class UserB:
  # ...
```

1. Add 4 workers
2. Add 1 user: This adds a single `UserA` as expected
4. Add 7 more users: This adds 7 `UserB`s as expected
5. Terminate the worker running the single `UserA`
6. After the rebalance now have 8 `UserB`s and no `UserA`s
7. Adding 2 more users (taking total to 10) still has no `UserA`s

```
All users spawned: {"UserA": 1, "UserB": 0} (1 total users)
All users spawned: {"UserA": 1, "UserB": 7} (8 total users)
All users spawned: {"UserA": 0, "UserB": 10} (10 total users)
```

### Environment

- OS: `python:3.9` container
- Locust version: running `master`
- Locust command line that you ran: `/usr/local/bin/python /usr/local/bin/locust --master --locustfile .../locustfile.py`
- Locust file contents (anonymized if necessary)
```python
class UserA:
  fixed_count = 1
  # ...

class UserB:
  # ...
```

cyberw:
@EzR1d3r anything to add?

EzR1d3r:
@cyberw @andydunstall  Hi! It is not known issue. I dont remember rebalance in details right now. Does a master request from workers remained users before rebalancing? Overall, changing **self._try_dispatch_fixed** to _True_ looks reasonable.  And I guess, we need a test for this case, may be in locust/test/test_main.py?

andydunstall:
[...]

PR #2093, opened 2022-05-08
dispatch: rebalance users with a fixed count

andydunstall:
Fixes https://github.com/locustio/locust/issues/2091

* Resets the user worker count before rebalancing since `_user_gen` compares the expected number of fixed count users with the users in `self._users_on_workers`
* Sets `self._try_dispatch_fixed = True` to include fixed users in the rebalance
* Extends `TestAddWorker` and `TestRemoveWorker` to add a test for fixed count (which both fail without this commit) - these just copy `test_remove_worker_during_ramp_up` and `test_add_worker_during_ramp_up` with a fixed user

### DIFF
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -221,6 +221,14 @@ class UsersDispatcher(Iterator):
         we started from 0 user. So, if we were currently running 500 users, then the `_distribute_users` will
         perform a fake ramp-up without any waiting and return the final distribution.
         """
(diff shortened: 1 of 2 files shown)


EzR1d3r:
Looks fine. @cyberw @mboutet can you review too?

cyberw:
Lgtm, but @mboutet is the user distribution expert :)

mboutet:
LGTM also! Thank you all.

Issue #2036, opened 2022-03-02
Multiple resetting connection after RPCError


marcinh:
### Description
When there is something wrong with communication master-worker (RPCError caught)), connection is being reset (new RPC server created) multiple times (actually after every heartbeat interval) until new message from worker appears. So when there is only single worker and RPCError is caught - connection is being reset infinitely.

### Expected behavior
After catching RPCError connection should be reset once

### Actual behavior
Connection is being reset every single second (heartbeat interval) until new message from worker arrives or new worker appears

### Steps to reproduce
Setup locust master + single worker, trigger RPCError communication error somehow.

### Possible solution
Currently connection_reset() method in MasterRunner is called in heartbeat_worker thread when connection_broken flag is set. Once this flag is set to True connection reset bill be done in a loop until this flag is set to false (what is done in client listener thread). So the possible solution is to move setting connection_broken flag to false into connection_reset() method itself

```
    def reset_connection(self):
        logger.info("Reset connection to worker")
        try:
            self.server.close()
            self.server = rpc.Server(self.master_bind_host, self.master_bind_port)
            self.connection_broken = False
        except RPCError as e:
            logger.error(f"Temporary failure when resetting connection: {e}, will retry later.")
```

### Environment

- OS: Centos
- Python version: 3.9
- Locust version: 2.8.2

PR #2096, opened 2022-05-12
Fix multiple resetting connection after RPCError

Nosibb:
Fix for issue [2036](https://github.com/locustio/locust/issues/2036)

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -887,6 +887,7 @@ class MasterRunner(DistributedRunner):
         try:
             self.server.close()
             self.server = rpc.Server(self.master_bind_host, self.master_bind_port)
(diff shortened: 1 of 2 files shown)


locust/runners.py
```diff
@@ -905,7 +906,6 @@ def client_listener(self):
                 self.connection_broken = True
                 gevent.sleep(FALLBACK_INTERVAL)
                 continue
-            self.connection_broken = False
```
cyberw, line 908: I dont really know this code. Can you ping the person who wrote it and see if s/he thinks this makes sense?

Nosibb: Hi @delulu what do you think about it?

PR #2101, opened 2022-05-20
fix: client_listener will raise an KeyError error when received a cli…

BirdLearn:
…ent_stopped message from a client with a nonexistent node_id

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -582,9 +582,13 @@ class WorkerNodes(MutableMapping):
         self._worker_nodes[k] = v
 
     def __delitem__(self, k: str) -> None:
+        if k not in self._worker_nodes:
+            return None
         del self._worker_nodes[k]
 
     def __getitem__(self, k: str) -> WorkerNode:
+        if k not in self._worker_nodes:
+            return None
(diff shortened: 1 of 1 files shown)


locust/runners.py
```diff
@@ -938,6 +942,8 @@ def client_listener(self):
                 #    warnings.warn("The worker node's clock seem to be out of sync. For the statistics to be correct the different locust servers need to have synchronized clocks.")
             elif msg.type == "client_stopped":
                 client = self.clients[msg.node_id]
+                if client is None:
+                    continue
```
cyberw, line 946: Can you log a warning here?

locust/runners.py
```diff
@@ -977,6 +983,8 @@ def client_listener(self):
             elif msg.type == "quit":
                 if msg.node_id in self.clients:
                     client = self.clients[msg.node_id]
+                    if client is None:
+                        continue
```
cyberw, line 987: Can you log a warning here?

locust/runners.py
BirdLearn: Of course, I will resubmit the pull request after modifying the code

PR #2102, opened 2022-05-20
Stop client_listener from raising a KeyError when receiving a client_stopped message from unknown worker

BirdLearn:
fix: client_listener will raise an KeyError error when received a client_stopped message from an unknown worker.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -937,15 +937,17 @@ class MasterRunner(DistributedRunner):
                 # if abs(time() - msg.data["time"]) > 5.0:
                 #    warnings.warn("The worker node's clock seem to be out of sync. For the statistics to be correct the different locust servers need to have synchronized clocks.")
             elif msg.type == "client_stopped":
-                if msg.node_id in self.clients:
-                    client = self.clients[msg.node_id]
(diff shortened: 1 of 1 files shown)


cyberw:
We need to at least log a warning when it happens. Also, to avoid the deep nesting, can you reverse the if-statement to be:

```
if not msg.node_id in self.clients:
    continue
```

BirdLearn:
> We need to at least log a warning when it happens. Also, to avoid the deep nesting, can you reverse the if-statement to be:
> 
> ```
> if not msg.node_id in self.clients:
>     continue
> ```

When I modified the code, I found that the same judgment was made on other conditional branches, such as "heartbeat" and "quit", so, in order to keep the code style consistent, I made the same code modification.

In addition, I found that after processing the message, there is a check_stopped() call, if I use "continue " to skip this loop, it will not execute to check_stopped(), I am worried about other side effects.

cyberw:
Not calling check_stopped in this case should be fine (this is only relevant if a worker has actually stopped, so skipping it is a).

In this case I think readability (and minimizing the changes to maintain git blame history) is more important than consistency.

But the logging is the most important.

PR #2104, opened 2022-05-20
Added explicit type check for headers in fasthttp.py

lundstrj:
I accidentally tried to use supply a set instead of a dictionary and it took me way longer than it should have to figure out what was wrong. These two lines of code in locust would have gone a long way to help me out.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -187,6 +187,8 @@ class FastHttpSession:
             context = {**self.user.context(), **context}
 
         headers = headers or {}
+        if type(headers) is not dict:
+            print("headers is assumed to be a dictionary. You tried to use a: %s instead. That won't work (sorry about that)." % type(headers))
         if auth:
             headers["Authorization"] = _construct_basic_auth_str(auth[0], auth[1])
         elif self.auth_header:
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -187,8 +187,7 @@ class FastHttpSession:
             context = {**self.user.context(), **context}
 
         headers = headers or {}
-        if type(headers) is not dict:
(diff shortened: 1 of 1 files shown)


cyberw:
Well.. as there's no way to recover from this error and we dont want to add checks for all every sort of parameter. I dont really think this is worth it.

This should have been caught by a linter, but for some reason, mine (pylint+vscode language server) don't see a problem passing something else, maybe I can fix that...

lundstrj:
Yeah I mean if you pass a non dictionary locust will crash and that's that. I just wanted it to be easier for the user to find their input error faster. I did this:
`headers = {"Content-Type:application/json"}`

Instead of this:
`headers = {"Content-Type":"application/json"}`

And I am ashamed of how long it took me to spot my error.
This assert would have helped. Perhaps it's too expensive to check the type of headers in every request?

As much as I like to develop in Python I also find myself practicing software constraints a lot more than I do in other more less forgiving languages. I try to assert all of my assumptions on the input at the beginning of my functions, if I can do so reasonably quickly/affordably.

cyberw:
This ticket made me go back and revisit our type hinting. Turns out, mypy was unable to automatically detect the type of FastHttpUser.client, so this type of error could not have been detected. 

I fixed that now in https://github.com/locustio/locust/commit/4ba878139cfd645887809f8002e8ed734a4c80be

So now (if you use latest master, at the next release if you dont), you can use mypy to find these kinds of errors. I highly recommend running that - in most editors you can run this on-save, so you'll never have to run into this kind of problem.

Here's what it should look like in vscode for instance:
image

lundstrj:
@cyberw Alright, cool man. My PR was rejected BUT it seems I had a positive impact on the code base anyway and at the end of the day, that is what I wanted.

PR #2106, opened 2022-05-24
additional typing improvements

mgor:
when doing some "integration" work with `RequestStats`, i noticed that there was a lot of stuff that was missing types and had inferred `Any`.

so this PR started out with the goal to add typing to `locust.stats` which took me down a rabbit hole of also adding types in `locust.env`, `locust.web`, `locust.runners` and `locust.user`.

i had to get creative in some places, which added a bit more code than just the types, mostly due to how many variables has an initial value of `None`.

also, maybe a bit unnecessary... but `StatsEntry` had `serialize` and `unserialize`, while `StatsError` for the same thing had `to_dict` and `from_dict`, so changed it so they have the same names.

### DIFF
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -49,7 +49,7 @@ class UsersDispatcher(Iterator):
             from 10 to 100.
     """
 
-    def __init__(self, worker_nodes: "List[WorkerNode]", user_classes: List[Type[User]]):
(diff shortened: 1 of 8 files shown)


cyberw:
Awesome. Just let me test it against my own locustfiles first.

cyberw:
thx

Issue #2066, opened 2022-04-04
Support for TLS authentication when using `FastHttpUser`


jumaffre:
### Is your feature request related to a problem? Please describe.

The `FastHTTPUser` does not support TLS client authentication (using a private key/certificate).

### Describe the solution you'd like

It should be possible to set client TLS session authentication when using the `FastHttpUser`. Ideally, it should be possible to access the underlying SSL context object, or supply it when creating the user. 

### Describe alternatives you've considered

`HttpUser` supports this feature but is obviously slower. 

### Additional context

With `HttpUser`, when targeting a HTTPS server that makes use of TLS client authentication, the following works OK, but does not work (`401` response code) when switching to `FastHttpUser`:
```python
from locust import FastHttpUser, task,

class Submitter(HttpUser):

    @task()
    def submit(self):
        opts = self.environment.parsed_options
        self.client.get(
            "endpoint",
            cert=(opts.cert, opts.key),
            verify=opts.ca,
        )
```

renato-farias:
Hey @jumaffre,

I know this is not straightforward but you try something like that (worked for me):

```python
import gevent
from locust import task
from locust.contrib.fasthttp import FastHttpUser

def _create_default_context():
    context = gevent.ssl.create_default_context()
    context.check_hostname = True
    context.verify_mode = gevent.ssl.CERT_REQUIRED
    context.load_cert_chain(certfile="yourpempath", keyfile="yourkeypath")
    context.load_verify_locations(cafile="yourcapath")
    return context

class ApiLocust(FastHttpUser):
    def __init__(self, environment):
        super().__init__(environment=environment)
        self.client.client.clientpool.client_args['ssl_context_factory'] = _create_default_context

    @task()
    def submit(self):
        self.client.get('endpoint')
```

PR #2113, opened 2022-06-13
add support for custom SSLContext when using FastHttpUser

renato-farias:
As per https://github.com/locustio/locust/issues/2066, currently there's no option for setting up certificates when using `FastHttpUser`. This PR intends to add a new `ssl_context_factory` param to enabling this functionality by defining a custom `Callable` returning a SSLContext when implementing the user.

As the callable will be called within the class context if it is defined outside the `ApiLocust` class we need to fake an arg there. The two example on how implementing can be found below.

**Option 1**

```python
import gevent
from locust import task
from locust.contrib.fasthttp import FastHttpUser

def my_custom_context(*args):
    context = gevent.ssl.create_default_context()
    context.check_hostname = True
    context.verify_mode = gevent.ssl.CERT_REQUIRED
    context.load_cert_chain(certfile="yourpempath", keyfile="yourkeypath")
    context.load_verify_locations(cafile="yourcapath")
    return context

class ApiLocust(FastHttpUser):
    ssl_context_factory = my_custom_context
    
    def __init__(self, environment):
        super().__init__(environment=environment)

    @task()
    def submit(self):
        self.client.get('/')
```

**Option 2**

```python
import gevent
from locust import task
from locust.contrib.fasthttp import FastHttpUser

class ApiLocust(FastHttpUser):
    
    def __init__(self, environment):
        super().__init__(environment=environment)
   
    def ssl_context_factory(self):
        context = gevent.ssl.create_default_context()
        context.check_hostname = True
        context.verify_mode = gevent.ssl.CERT_REQUIRED
        context.load_cert_chain(certfile="yourpempath", keyfile="yourkeypath")
        context.load_verify_locations(cafile="yourcapath")
        return context

    @task()
    def submit(self):
        self.client.get('/')
```

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -5,7 +5,7 @@ import json
 import json as unshadowed_json  # some methods take a named parameter called json
 from base64 import b64encode
 from urllib.parse import urlparse, urlunparse
-from ssl import SSLError
+from ssl import SSLError, SSLContext
 import time
 from typing import Optional, Tuple, Dict, Any, Union
 
@@ -75,14 +75,16 @@ class FastHttpSession:
         user: Optional[User],
         insecure=True,
         client_pool: Optional[HTTPClientPool] = None,
-        ssl_context: Optional[gevent.ssl.SSLContext] = None,
+        ssl_context: Optional[SSLContext] = None,
         **kwargs,
     ):
         self.environment = environment
         self.base_url = base_url
         self.cookiejar = CookieJar()
         self.user = user
-        if ssl_context and self._check_ssl_context(ssl_context):
+        if ssl_context:
+            if not isinstance(ssl_context, SSLContext):
+                raise TypeError('You must provide a valid SSLContext.')
             ssl_context_factory = lambda: ssl_context
         elif insecure:
             ssl_context_factory = insecure_ssl_context_factory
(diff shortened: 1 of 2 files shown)


locust/contrib/fasthttp.py
```diff
@@ -107,6 +110,13 @@ def __init__(
             # store authentication header (we construct this by using _basic_auth_str() function from requests.auth)
             self.auth_header = _construct_basic_auth_str(parsed_url.username, parsed_url.password)
 
+    def _check_ssl_context(ssl_context):
```
cyberw, line 113: Not sure this deserves its own method :)

cyberw:
Lgtm, apart from the style issues & my comment

renato-farias:
> Lgtm, apart from the style issues & my comment

Thanks for reviewing it @cyberw . Pushed some fixes.

locust/contrib/fasthttp.py
```diff
@@ -75,13 +75,18 @@ def __init__(
...
     ):
         self.environment = environment
         self.base_url = base_url
         self.cookiejar = CookieJar()
         self.user = user
-        if insecure:
+        if ssl_context:
+            if not isinstance(ssl_context, SSLContext):
+                raise TypeError("You must provide a valid SSLContext.")
+            ssl_context_factory = lambda: ssl_context
```
cyberw, line 88: This looks a little strange to me. Doesnt it make more sense to just pass the function instead of a value that we then make into a function anyway? I guess we can't check the type as neatly at runtime, but I think it is unnecessary (especially with a type hint it to indicate it must return an SSLContext)

cyberw, line 88: Doesnt it make more sense to just pass a function instead of an instantiated SSLContext that we make a fake function for? (I made a comment and then deleted it, but after some more thought I think my initial comment was correct :)

renato-farias: Surely it does. I was trying to make use of `ssl_context` as an actual context then the caller could set it accordingly, but making it a factory keeps the way it does in the background. I will rename it to `ssl_context_factory` and type-hint it as Callable. Thanks for reviewing it.

cyberw:
Nice! But the test doesnt really check anything, does it? I guess it could be hard to validate that the context was actually used in the request, but we could at least check that the function was called (perhaps by setting a value there that we assert on)

Issue #2111, opened 2022-06-09
"Stop" hang on "stopping" state  when there are more than one workers in distributed mode.


xllalitha:
Observed an  issue after clicking "stop" load test from web console of master node. "stop" status change to "stopping" status and hang there without reaching "stopped" state. When there is only 1 worker,  it reaches properly to "stopped" state.. I had to restart locust master again to restart load test when it was hung on "stopping" state. Issue repeats when I have more than one worker.

I experienced this issue first when running master and workers on AWS fargate containers .Interestingly, there were no errors logged in master and worker containers  after pressing stop button. With just single worker container, Stop status properly move to "stopped" on web console master. Having more than 1 worker cause this issue.

Later I could  reproduce this issue on local  machine without even docker. 

When I run locust master and 2 workers from my local MacBook itself, I am experiencing this issue. 
There are no errors thrown after clicking "stop".  With 1 worker, it work fine and move to "stopped" status.

How to reproduce in local MacBook

1. Start Locust master from one terminal with below command
 
  locust -f my_locustfile.py  --master

2. Start 2 locust workers from another 2 terminals with below command

  locust -f my_locustfile.py  --worker.

 (Ignored "--master-host"  parameter here as it is local run. In AWS Fargate setup, we have that parameter)

After this master web console(http://0.0.0.0:8089/) shows 2 workers connected, also load test work fine.  When click on "stop" it changes to "stopping" and never come to "stopped".

PR #2116, opened 2022-06-15
fix: stopping state when running more than one worker node.

renato-farias:
As per the issue https://github.com/locustio/locust/issues/2111 when running locust in distributed mode with more than 1 worker when the job is stopped the state is kept as STOPPING. Doing some investigation I realised after it removes a client from the list it gets back as a `ready` one, thus when the `check_stopped()` is called it checks the state for all clients, as the operation is fast, before checking the second client the first one is already `ready` and pass the `update_state(STATE_STOPPED)`

This PR intends to fix this bug hence to fix the continuously RPS in the graph even with the test "stopped" (issue with the graph shown bellow).

image

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -970,7 +970,6 @@ class MasterRunner(DistributedRunner):
                 gevent.sleep(FALLBACK_INTERVAL)
                 continue
             msg.node_id = client_id
-            logger.debug(f"Received a message: {msg.type}")  # noisy?
             if msg.type == "client_ready":
                 if not msg.data:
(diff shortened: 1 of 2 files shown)


cyberw:
Cool! Yes, the debug logging is a bit noisy :) (I wish we had a trace level though...)

@mboutet LGTY?

renato-farias:
> Could you have a look to see if it could be added as a test?

Added a test.

cyberw:
Nice! I'll give @mboutet a couple days to chime in, but it lgtm.

mboutet:
LGTM also. If all tests are green, then it's good for me.

PR #2120, opened 2022-06-22
More robust handling of ZMQ/RPC errors

solowalker27:
When using Boomer as a worker runner, I encountered some occasional errors on the master when receiving reports from the workers. In that situation, current Locust would break itself by attempting to reset the entire RPC server. Firstly, various issues prevented that from completing successfully. This broke the entire test instead of just fixing the one worker. Secondly, that isn't always necessary for the master to do. In most cases, subsequent reports from the workers are fine. In some other cases, the worker with the bad message resetting its connection to the master can resolve it. This will now reserve the master resetting its server to times when the master gets an error attempting to send to a worker, or if there's any other unexpected `RPCError` that gets thrown.

### DIFF
--- a/locust/exception.py
+++ b/locust/exception.py
@@ -54,6 +54,22 @@ class RPCError(Exception):
     """
 
 
+class RPCSendError(Exception):
+    """
+    Exception when sending message to client.
+
+    When raised from zmqrpc, sending can be retried or RPC can be reestablished.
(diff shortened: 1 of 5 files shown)


cyberw:
Lgtm. Will merge (probably as squash) once we sort out the `timeout`-issue...

Issue #2126, opened 2022-07-04
HTML Report does not correctly escape statistics data


herrmanntom:
### Describe the bug
In case your statistics include html, or html-like content, then the html report breaks in various ways.
For example, one could:
- include "place holders" in custom names for requests that happen to have angle brackets in them (`name="/api/user/<userId>/info"`)
- custom error messages might contain (parts of) the request text, which then might contain html (`response.failure("failed with payload: ' + response.text)`)

### Expected behavior
all html entities within user controlled parts of the statistics must be properly encoded, so that report can be viewed correctly with such data inside

### Actual behavior
data is not encoded, report breaks

### Steps to reproduce
Mark a response as failed with this message:
`response.failure("<title>")`
then generate html report
Look at "failure" table. It will break after this entry. Also remaining content after this line is broken (so graphs are missing etc)

### Environment

- OS: Fedora 36
- Python version: 3.9
- Locust version: 2.10.1
- Locust command line that you ran: locust -f testcases.py
- Locust file contents (anonymized if necessary): (see above for relevant parts)

cyberw:
Sounds troublesome. PRs welcome.

herrmanntom:
Should be a matter of putting
https://docs.python.org/3/library/html.html#html.escape
in the right places.

herrmanntom:
turns out, the exception display in the default UI also does not escape the traceback correctly... I added another PR for this

PR #2127, opened 2022-07-05
Escape user supplied data in html report (#2126)

herrmanntom:
This is a quick fix for issue #2126, possibly not the most elegant solution.

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -80,6 +80,8 @@ def get_html_report(environment, show_download_link=True):
         "report.html",
         int=int,
         round=round,
+        escape=escape,
+        str=str,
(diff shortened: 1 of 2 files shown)


cyberw:
LGTM. Have you tested a few cases? Like spaces in filenames etc?

cyberw:
I think the problems on GH are real errors. Can you check if your package versions are different from the ones listed in the build? (or just try installing the latest ones yourself)

herrmanntom:
Maybe I did it somehow wrong before. Now everything seems to be fixed.

cyberw:
Thx!

Issue #2132, opened 2022-07-13
Html report:  table should be sortable


Likqez:
### Is your feature request related to a problem? Please describe.

Exploring the report data can be quite hard, especially if it contains hundreds of requests.
Its nearly impossible to search for e.g. the request with the highest response time or the one with the largest response.

### Describe the solution you'd like

A sorting mechanism to order the columns by clicking the respective table headers.
Like the index page, whose table is able to be sorted already.

Edit:
~~It should be possible to reuse the sorting mechanism from the main page.~~

PR #2134, opened 2022-07-14
Test Report: Implement table sorting

Likqez:
This PR implements the following feature request #2132.

It enables sorting the statistic tables by clicking on the respective table headers.
To make the headers clickable, the class `sortable` needs to be applied. This makes it possible to ex-/include columns from being sort-able.

I only excluded the error messages as well as the stacktraces from being sort-able from now on.
Changes are welcome.

### DIFF
--- a/locust/templates/report.html
+++ b/locust/templates/report.html
@@ -129,7 +129,7 @@
                         <tr>
                             <th class="sortable">Method</th>
                             <th class="sortable">Name</th>
-                            <th>Error</th>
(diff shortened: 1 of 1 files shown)


cyberw:
LGTM. But why not have error messages sortable?

Likqez:
Seemed unnecessary to me. Added it with the latest commit.

cyberw:
Thx!

PR #2136, opened 2022-07-21
Log an error for every failed attempt to connect to master

gdm85:
The connection timeout and number of attempts are hardcoded (60 attempts with an interval of 5 seconds each), so a complete failure of the worker to connect to master will take very long.

These log lines will allow to troubleshoot issues with the connection to master.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1336,6 +1336,14 @@ class WorkerRunner(DistributedRunner):
         self.client.send(Message("client_ready", __version__, self.client_id))
         success = self.connection_event.wait(timeout=CONNECT_TIMEOUT)
         if not success:
+            if self.retry < 3:
+                logger.debug(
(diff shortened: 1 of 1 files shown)


cyberw:
Maybe make it debug the first 2 times and then warning level? Dont wanna spam the user in what is a normal scenario (master is often not available at the very start), and error should be reserved for catastrophic fails.

gdm85:
Sure, I will change it as I am not familiar with how log levels are used in Locust and this change should indeed not break consistency.

> Dont wanna spam the user in what is a normal scenario (master is often not available at the very start)

"2" is not a magic number that will cover all situations, I'd rather have everything with same level but will do as you suggested for now. I would probably make retries and delay between retries configurable (in a separate PR), but at least with this one the on-going failures will stop being completely concealed.

cyberw:
Lgtm, except for the formatting/black issue.

gdm85:
Oh, my bad, I had missed that; fixed now.

PR #2146, opened 2022-08-02
Patch 1

balaji2711:
I have updated the locust usage that was not mentioned in the documentation. Kindly review and merge this pull request.

### DIFF
--- a/docs/what-is-locust.rst
+++ b/docs/what-is-locust.rst
@@ -8,7 +8,7 @@ You define the behaviour of your users in regular Python code, instead of being 
 
 This makes Locust infinitely expandable and very developer friendly.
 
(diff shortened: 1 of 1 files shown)


cyberw:
Not sure these things need this much focus in the docs (although we should definitely mention that other protocols can be supported, if it isnt already mentioned). Websockets is not even part of locust-core, so mentioning it without referencing locust-plugins is just confusing.

balaji2711:
Ok, As you said, At least we can mention the supported protocols and also we can mention about the custom tasks in the document, It's not mentioned anywhere in the docs. If required, we can provide the sample snippet as well.

balaji2711:
Hi @cyberw, 
Could you please reject this pull request if that's not useful?

cyberw:
> Hi @cyberw,
> Could you please reject this pull request if that's not useful?

Ok! I’m sure the documentation can be improved in other places, so if you find other things feel free to open a new PR.

Issue #2135, opened 2022-07-20
"New test" with different number of users


bsypniewski:
It seems likes a bug to me, but please correct me if I'm wrong. It seems that after clicking on "new test" and giving lower amount of users to be created, Locust still uses some of the users created on first run.

### Describe the bug
Example:
When i start locust and create (total) 5 users with 5 users per sec, I get correct amount of users/requests. That is correct. Then i click Stop (in the UI).
After clicking on "New test" and specifying lower amount of users(total 1 and 1 per sec), Locust still starts 4 users in total and execute tasks - but in console i can see that 1 user was spawned

Code:
```
import logging
import time
import uuid
from locust import task, HttpUser, SequentialTaskSet, constant_pacing, task, events
from locust.exception import StopUser

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger("locust")

class UserTasks(SequentialTaskSet):
    wait_time = constant_pacing(1)

    @task
    def test_api_one(self):
        logger.info(f"Task1: {self.parent.user_id}")
        time.sleep(1)

    @task
    def test_api_two(self):
        logger.info(f"Task2: {self.parent.user_id}")
        time.sleep(1)
    
    @task
    def test_api_three(self):
        self.interrupt(False)

class WebsiteUser(HttpUser):
    host = "http://local"
    wait_time = constant_pacing(5)
    tasks = [UserTasks]

    def on_start(self):
        self.user_id = str(uuid.uuid4())
        logger.info(f"START USER: {self.user_id}")

    def on_stop(self):
        logger.info(f"STOP USER: {self.user_id}")
```

### Expected behaviour
I would expect that if i run new test with lower number of users - from 5 to 1 users, Locust will just create 1 user and run (spawn) only that one.

You can see it on the screenshot:

### Environment

- OS: docker with python 3.9
- Python version: 3.9
- Locust 2.9.0
- locust -f temp.py

![Screenshot 2022-07-20 at 16 21 47](https://user-images.githubusercontent.com/83775196/180020920-6ceb807a-26b3-4de6-b849-e65629ce54fb.png)

cyberw:
Wow, that looks crazy. Can you update to latest and see if it makes a difference?

bsypniewski:
Hi.
I got the same results on latest (v.2.10.1)

cyberw:
Any difference if you remove `test_api_three`?

cyberw:
Very strange. I’ll take a look.

[...]

PR #2150, opened 2022-08-05
Fix issue #2135

prestonhale:
Fully fixes the issue in #2135 for me. Not sure what the purpose of maintaining the user_dispatcher was in the first place so take this with a grain of salt. If you do want to dig deeper the issue seems to be the `_dispatch` generator.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -488,10 +488,9 @@ class LocalRunner(Runner):
         if self.state != STATE_INIT and self.state != STATE_STOPPED:
             self.update_state(STATE_SPAWNING)
 
-        if self._users_dispatcher is None:
-            self._users_dispatcher = UsersDispatcher(
-                worker_nodes=[self._local_worker_node], user_classes=self.user_classes
-            )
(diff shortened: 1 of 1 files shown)


cyberw:
It seems to break a test, but maybe the test is wrong?

mboutet:
@cyberw see https://github.com/locustio/locust/pull/2152 for the appropriate fix and reproducing test.

As for `test_stop_timeout_with_ramp_down`, it is an odd test indeed. Even when it was introduced in ded61e0eeb64857b4b0ecfb3fd34ad8196abaa60, the test name doesn't make much sense. It does not strike me as test for validating the `stop_timeout` as the name implies. If it was the case, I would expect the task to sleep for more than two seconds.

Moreover, the docstring I added in d9bb87f1680ee491b82dabba2a03b9f1b0a70293 is outdated and misleading. I'll remove it in a separate PR. 

However, I'm not sure it should be removed as the test itself makes sense (apart from its name).

cyberw:
This should be closed then, I suppose.

PR #2152, opened 2022-08-07
fix: Ensure new test starts with specified number of users after previous test has been stopped

### LINKED ISSUE #2135: "New test" with different number of users

mboutet:
Fixes #2135

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -407,6 +407,8 @@ class Runner:
 
         self.stop_users(self.user_classes_count)
 
+        self._users_dispatcher = None
+
         self.update_state(STATE_STOPPED)
 
         self.cpu_log_warning()
--- a/locust/test/test_runners.py
+++ b/locust/test/test_runners.py
@@ -768,6 +768,61 @@ class TestLocustRunner(LocustRunnerTestCase):
         sleep(1)
         self.assertEqual(0, runner.user_count)
 
+    def test_user_count_starts_from_specified_amount_when_creating_new_test_after_previous_step_has_been_stopped(self):
+        """Test for https://github.com/locustio/locust/issues/2135"""
+
+        class MyUser1(User):
+            wait_time = constant(0)
+
+            @task
+            def my_task(self):
+                pass
+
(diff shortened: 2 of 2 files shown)


cyberw:
👍👍

PR #2156, opened 2022-08-12
Add example launch.json for debugging the whole Locust runtime in vscode

SamPosh:
In case user want to run locust directly from debugger


cyberw:
Thanks!

PR #2161, opened 2022-08-15
Allows weights of users to be 0 when set outside of class

saltekar:
Helps use case where you want to set the weight from the UI (custom arguments).

### DIFF
--- a/locust/test/test_dispatch.py
+++ b/locust/test/test_dispatch.py
@@ -3591,6 +3591,72 @@ class TestRampUpUsersFromZeroWithFixed(unittest.TestCase):
                                 )
 
 
+class TestWeight(unittest.TestCase):
+    def test_set_weight_outside_class(self):
+        class User1(User):
+            pass
+
+        class User2(User):
+            pass
+
+        User1.weight = 0
+
+        worker_node1 = WorkerNode("1")
+        worker_node2 = WorkerNode("2")
+        users_dispatcher = UsersDispatcher(worker_nodes=[worker_node1, worker_node2], user_classes=[User1, User2])
+
+        users_dispatcher._distribute_users(
+            target_user_count=1_000
+        )
+
+    def test_set_weight_inside_class(self):
+        class User1(User):
+            weight = 0
+            pass
+
+        class User2(User):
+            pass
+
+        worker_node1 = WorkerNode("1")
+        worker_node2 = WorkerNode("2")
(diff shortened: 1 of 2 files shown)


locust/test/test_dispatch.py
```diff
@@ -3591,6 +3591,72 @@ class User5(User):
...
+        users_dispatcher = UsersDispatcher(worker_nodes=[worker_node1, worker_node2], user_classes=[User1, User2])
+
+        users_dispatcher._distribute_users(
+            target_user_count=1_000
+        )
+
+    def test_set_weight_inside_one_class(self):
+        class User1(User):
+            weight = 0
+            pass
```
AlexMooney, line 3649:
```suggestion
            weight = 0
```

locust/test/test_dispatch.py
```diff
@@ -3591,6 +3591,72 @@ class User5(User):
...
+        users_dispatcher = UsersDispatcher(worker_nodes=[worker_node1, worker_node2], user_classes=[User1, User2])
+
+        users_dispatcher._distribute_users(
+            target_user_count=1_000
+        )
+
+    def test_set_weight_outside_class_to_one(self):
+        class User1(User):
+            weight = 0
+            pass
```
AlexMooney, line 3631:
```suggestion
            weight = 0
```

AlexMooney (commented):
We can remove a couple of these `pass` statements, and we should assert the mix of users comes out how we expect after tampering with the weights.

locust/test/test_dispatch.py
```diff
@@ -3591,6 +3591,85 @@ class User5(User):
...
+        worker_node2 = WorkerNode("2")
+        worker_node3 = WorkerNode("3")
+        users_dispatcher = UsersDispatcher(worker_nodes=[worker_node1, worker_node2, worker_node3], user_classes=[User1, User2, User3])
+
+        users_dispatcher.new_dispatch(target_user_count=30, spawn_rate=0.5)
+        users_dispatcher._wait_between_dispatch = 0
+        iterations = list(users_dispatcher)
+
+        # Here returns: {"User1": 0, "User2": 10, "User3": 0}
+        self.assertDictEqual(iterations[-1]["1"], {"User1": 0, "User2": 30, "User3": 0})
```
saltekar, line 3635: Looks like a bug where it's deciding how many users to add based on the number of classes rather than the weights those classes have

PR #2164, opened 2022-08-18
Fix broken link in cpu warning message.

conghuiw:
change https://docs.locust.io/en/stable/running-locust-distributed.html to https://docs.locust.io/en/stable/running-distributed.html


cyberw:
thx

PR #2165, opened 2022-08-20
Example had inconsistent labeling

aathan:
The example inconsistently labeled worker and master in the comments and/or registered the wrong messages to the wrong runners.


cyberw:
Thx

PR #2173, opened 2022-08-27
fix(style): about dialog cannot be opened in the startup page 

alterhu2020:
Fix these issues:
* When you startup the locust web page first time, in the main page, we cannot open the about page when click the bottom link "About"
* In the "Start new load test" and "Edit running load test" page, the top left close button is not formatted correct(disorderred) , so I corrected the close link position to the same as in the "About" dialog page. see the following screenshot:

1. Modified screenshot, which can show the "About" dialog in the startup page:
image

2. The "close" button in the "Start new load test" dialog page:
image

2. The "close" button in the "Edit running load test" dialog page:
image


cyberw:
Good stuff, thanks!

PR #2175, opened 2022-08-28
fix: when run from web with `-f a_file_folder/` caused page title is None as  `locustfile` argument is None

alterhu2020:
Steps:
1. run the locust with command `-f a_file_folder/` the `locustfile` argument is a folder, so in the main.py it will return `locustfile` as `None`
2. When start the test, then the web page title `Locust for None` shows, we need to correct the `locustfile` variable at this situation

### DIFF
--- a/locust/test/test_web.py
+++ b/locust/test/test_web.py
@@ -348,6 +348,9 @@ class TestWebUI(LocustTestCase, _HeaderCheckMixin):
         self.assertEqual(self.environment.host, "https://localhost")
         self.assertListEqual(["User1", "User2"], response.json()["user_classes"])
 
+        self.assertIsNotNone(self.environment.locustfile, 'verify locustfile is not empty')
+        self.assertEqual(self.environment.locustfile, "User1,User2", "Verify locustfile variable used in web ui title")
+
         # stop
         gevent.sleep(1)
         response = requests.get("http://127.0.0.1:%i/stop" % self.web_port)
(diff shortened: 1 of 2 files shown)


cyberw:
Wouldnt this cause issues on repeated runs? I mean wouldnt the second run show the same users as the first even if the user had changed the selection?

alterhu2020:
> Wouldnt this cause issues on repeated runs? I mean wouldnt the second run show the same users as the first even if the user had changed the selection?

Hi @cyberw  the above code can show the changed user when we switch to another selected user. But here comes another issue, we need to refresh the page to make the changes take affect. We need to change the `locustfile` in this function: `locust.web.WebUI.update_template_args` as well.

please take consider this issue should we change the code in above location as well ? thanks

alterhu2020:
> Wouldnt this cause issues on repeated runs? I mean wouldnt the second run show the same users as the first even if the user had changed the selection?

Hi @cyberw  the above code can show the changed user when we switch to another selected user. But here comes another issue, we need to refresh the page to make the changes take affect. 
So we need to change the `locustfile` variable in this function: `locust.web.WebUI.update_template_args` as well.

please take consider this issue should we change the code in above location as well ? thanks

cyberw:
I dont have many opinions on the UI, and never use it myself :) So whatever works and doesnt show incorrect info…

cyberw:
Need to fix the code formatting though...

cyberw:
This is probably a good fix, you just need to run Black to format your code..

PR #2193, opened 2022-09-11
stats summary refactoring

SamPosh:
Added get_stats_summary and get_percentile_stats_summary . This will return the stats as string , So that the string can be used in other places. I am using this to record it in Reportportal.
currently print_stats and print_percentile_stats functions are directly record the result into console_logger. But these functions return the same results into string so that it can be used by other log handler

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -791,7 +791,7 @@ def print_stats(stats: RequestStats, current=True) -> None:
 
 def get_stats_summary(stats: RequestStats, current=True) -> List[str]:
     """
-        stats summary will be returned as string
+        stats summary will be returned as list of string
     """
     name_column_width = (STATS_NAME_WIDTH - STATS_TYPE_WIDTH) + 4  # saved characters by compacting other columns
     summary = list()
@@ -817,7 +817,7 @@ def print_percentile_stats(stats: RequestStats) -> None:
 
 def get_percentile_stats_summary(stats: RequestStats) -> List[str]:
     """
-        Percentile stats summary will be returned as string
+        Percentile stats summary will be returned as list of string
     """
     summary = []
     summary.append("Response time percentiles (approximated)")
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -784,14 +784,15 @@ def setup_distributed_stats_event_listeners(events: Events, stats: RequestStats)
 
 
 def print_stats(stats: RequestStats, current=True) -> None:
-    stat_summary = get_stats_summary(stats,current)
+    stat_summary = get_stats_summary(stats, current)
     for summary in stat_summary:
         console_logger.info(summary)
     console_logger.info("")
 
+
 def get_stats_summary(stats: RequestStats, current=True) -> List[str]:
     """
-        stats summary will be returned as list of string
+    stats summary will be returned as list of string
     """
     name_column_width = (STATS_NAME_WIDTH - STATS_TYPE_WIDTH) + 4  # saved characters by compacting other columns
     summary = list()
@@ -815,9 +816,10 @@ def print_percentile_stats(stats: RequestStats) -> None:
         console_logger.info(summary)
     console_logger.info("")
 
+
 def get_percentile_stats_summary(stats: RequestStats) -> List[str]:
     """
-        Percentile stats summary will be returned as list of string
+    Percentile stats summary will be returned as list of string
     """
     summary = []
     summary.append("Response time percentiles (approximated)")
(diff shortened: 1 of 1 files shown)


cyberw:
Did you end up just duplicating the code and not using the new functions from anywhere else? This doesnt make sense.

And `current` parameter is never used? 

Also, you need to fix flake8 & black formatting errors.

SamPosh:
> Did you end up just duplicating the code and not using the new functions from anywhere else? This doesnt make sense.
> 
> And `current` parameter is never used?
> 
> Also, you need to fix flake8 & black formatting errors.

print_stats_summary is directly printing to console_logger. So I refactored that function into get_stats_summary and called it from print_stats but the unit test is checking the length of string. Since it is single string the length is always 1. So it is failed. May be I will try to create it as list of string then call it from print_stats so that it won't be duplicate.

For black/flake8 formatting, as of now I am doing this from vscode web editor which doesn't have formatting option.  I am trying to setup local dev environment and try to format it.

cyberw:
> print_stats_summary is directly printing to console_logger. So I refactored that function into get_stats_summary and called it from print_stats but the unit test is checking the length of string. Since it is single string the length is always 1. So it is failed. Due to that I duplicated the function. May be I will try to create it as list of string then call it from print_stats so that it won't be a duplicate.

Not sure what you mean, but duplication is bad :)

> For black/flake8 formatting, as of now I am doing this from vscode web editor which doesn't have formatting option. I am trying to setup local dev environment and try to format it.

Yea... tbh, that kind of sounds like you'll have a lot of work ahead of you to make this reach the quality required for merge. You might consider having this as a fork, or trying to implement it inside your own locustfile or support lib.

SamPosh (commented):
Duplication is removed

SamPosh:
> > print_stats_summary is directly printing to console_logger. So I refactored that function into get_stats_summary and called it from print_stats but the unit test is checking the length of string. Since it is single string the length is always 1. So it is failed. Due to that I duplicated the function. May be I will try to create it as list of string then call it from print_stats so that it won't be a duplicate.
> 
> Not sure what you mean, but duplication is bad :)
> 
> > For black/flake8 formatting, as of now I am doing this from vscode web editor which doesn't have formatting option. I am trying to setup local dev environment and try to format it.
> 
> Yea... tbh, that kind of sounds like you'll have a lot of work ahead of you to make this reach the quality required for merge. You might consider having this as a fork, or trying to implement it inside your own locustfile or support lib.

I removed the duplicate.

cyberw:
Looks a lot better now. Just fix the formatting and typing issues and it might be ready.

SamPosh:
> Looks a lot better now. Just fix the formatting and typing issues and it might be ready.

I have fixed the type hint issue.

SamPosh:
> Looks a lot better now. Just fix the formatting and typing issues and it might be ready.

formatting also done. type hint is fixed as well

locust/stats.py
```diff
@@ -784,25 +784,47 @@ def on_worker_report(client_id: str, data: Dict[str, Any]) -> None:
...
+        summary.append(r.to_string(current=current))
+    summary.append(separator)
+    summary.append(stats.total.to_string(current=current))
+    return summary
 
 
 def print_percentile_stats(stats: RequestStats) -> None:
-    console_logger.info("Response time percentiles (approximated)")
+    stat_summary = get_percentile_stats_summary(stats)
+    for summary in stat_summary:
```
cyberw, line 815: skip intermediate "summary" variable

locust/stats.py
```diff
@@ -784,25 +784,47 @@ def on_worker_report(client_id: str, data: Dict[str, Any]) -> None:
 
 
 def print_stats(stats: RequestStats, current=True) -> None:
+    stat_summary = get_stats_summary(stats, current)
+    for summary in stat_summary:
```
cyberw, line 788: skip intermediate "start_summary" variable. Also, maybe "line" is a more appropriate name than "summary"?

locust/stats.py
```diff
@@ -784,25 +784,47 @@ def on_worker_report(client_id: str, data: Dict[str, Any]) -> None:
...
+        console_logger.info(summary)
+    console_logger.info("")
+
+
+def get_percentile_stats_summary(stats: RequestStats) -> List[str]:
+    """
+    Percentile stats summary will be returned as list of string
+    """
+    summary = []
+    summary.append("Response time percentiles (approximated)")
```
cyberw, line 825: combine this with the line above

cyberw:
👍 👍

PR #2195, opened 2022-09-12
Fix / Custom Messages and update examples/docs

samuelspagl:
#### Background
I was trying to implement the new locust version, running into a few errors.
In our implementation we rely on the messaging feature, when I saw that when registering a message a class ´CustomMessageListener` is now expected I tried to change my code accordingly. 

#### The Error
Beforehand it worked, but after these changes it didn't. 
After some research if found out that using a class with a `__call__` method, still needs proper instantiating.

#### Solution
Therefore I changed the runner code from
```python
self.custom_messages[message_type](environment=self.environment, msg=msg)
```
to
```python
listener = self.custom_messages[msg.type]()
listener(environment=self.environment, msg=msg)
```
and it worked as expected.

I also updated the examples and the documentation to the latest state regarding this feature.

Happy to hear your comments / ideas / input.  Thanks a lot in advance. 

Best Sam

### DIFF
--- a/docs/running-distributed.rst
+++ b/docs/running-distributed.rst
@@ -88,24 +88,26 @@ order to coordinate data. This can be easily accomplished with custom messages u
 .. code-block:: python
 
     from locust import events
-    from locust.runners import MasterRunner, WorkerRunner
+    from locust.runners import MasterRunner, WorkerRunner, CustomMessageListener
 
     # Fired when the worker receives a message of type 'test_users'
-    def setup_test_users(environment, msg, **kwargs):
-        for user in msg.data:
-            print(f"User {user['name']} received")
-        environment.runner.send_message('acknowledge_users', f"Thanks for the {len(msg.data)} users!")
+    class SetupTestUsers(CustomMessageListener):
+        def __call__(self, environment, msg, **kwargs):
+            for user in msg.data:
+                print(f"User {user['name']} received")
+            environment.runner.send_message('acknowledge_users', f"Thanks for the {len(msg.data)} users!")
 
     # Fired when the master receives a message of type 'acknowledge_users'
-    def on_acknowledge(msg, **kwargs):
-        print(msg.data)
+    class OnAcknowledge(CustomMessageListener):
+        def __call__(super, msg, **kwargs):
+            print(msg.data)
 
     @events.init.add_listener
     def on_locust_init(environment, **_kwargs):
         if not isinstance(environment.runner, MasterRunner):
-            environment.runner.register_message('test_users', setup_test_users)
+            environment.runner.register_message('test_users', SetupTestUsers)
         if not isinstance(environment.runner, WorkerRunner):
-            environment.runner.register_message('acknowledge_users', on_acknowledge)
+            environment.runner.register_message('acknowledge_users', OnAcknowledge)
 
     @events.test_start.add_listener
     def on_test_start(environment, **_kwargs):
--- a/examples/custom_messages.py
+++ b/examples/custom_messages.py
@@ -1,26 +1,29 @@
 from locust import HttpUser, task, events, between
-from locust.runners import MasterRunner, WorkerRunner
+from locust.runners import MasterRunner, WorkerRunner, CustomMessageListener
+from locust.env import Environment
 
 usernames = []
 
 
-def setup_test_users(environment, msg, **kwargs):
-    # Fired when the worker receives a message of type 'test_users'
-    usernames.extend(map(lambda u: u["name"], msg.data))
-    environment.runner.send_message("acknowledge_users", f"Thanks for the {len(msg.data)} users!")
+class SetupTestUsers(CustomMessageListener):
+    def __call__(self, environment: "Environment", msg, **kwargs):
+        # Fired when the worker receives a message of type 'test_users'
+        usernames.extend(map(lambda u: u["name"], msg.data))
+        environment.runner.send_message("acknowledge_users", f"Thanks for the {len(msg.data)} users!")
 
 
-def on_acknowledge(msg, **kwargs):
-    # Fired when the master receives a message of type 'acknowledge_users'
-    print(msg.data)
+class OnAcknowledge(CustomMessageListener):
+    def __call__(self, msg, **kwargs):
+        # Fired when the master receives a message of type 'acknowledge_users'
+        print(msg.data)
 
 
 @events.init.add_listener
 def on_locust_init(environment, **_kwargs):
     if not isinstance(environment.runner, MasterRunner):
-        environment.runner.register_message("test_users", setup_test_users)
+        environment.runner.register_message("test_users", SetupTestUsers)
     if not isinstance(environment.runner, WorkerRunner):
-        environment.runner.register_message("acknowledge_users", on_acknowledge)
+        environment.runner.register_message("acknowledge_users", OnAcknowledge)
 
 
 @events.test_start.add_listener
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -581,7 +581,7 @@ class LocalRunner(Runner):
         """
         logger.debug(f"Running locally: sending {msg_type} message to self")
         if msg_type in self.custom_messages:
-            listener = self.custom_messages[msg_type]
+            listener = self.custom_messages[msg_type]()
             msg = Message(msg_type, data, "local")
             listener(environment=self.environment, msg=msg)
         else:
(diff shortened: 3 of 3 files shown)


samuelspagl:
Hmm thinking about it, and reading more I think it could be that the code changes in the PR are not necessary. 
When using `@CustomMessageListener` as a decorator for a given function it should work the way it was probably intended to be. 

Is my assumption correct? 

If yes, then I would remove the code changes and just update the examples and documentation. 


Best

cyberw:
Ping @nathan-beam who created them in the first place

samuelspagl:
Okay, so I did some more testing and if possible I would like you @cyberw or somebody else to confirm my results. 
My test were always against a stock locust 2.12.0.

 #### First Test
This is the `custom_messages.py` example from the example folder. running it with `locust -f examples/custom_messages.py --master` / `locust -f examples/custom_messages.py --worker` should work just fine, despite that we are ignoring the type hinting right now.

```python
from locust import HttpUser, task, events, between
from locust.runners import MasterRunner, WorkerRunner, CustomMessageListener
from locust.env import Environment

usernames = []

def SetupTestUsers(self, environment: "Environment", msg, **kwargs):
    # Fired when the worker receives a message of type 'test_users'
    usernames.extend(map(lambda u: u["name"], msg.data))
    environment.runner.send_message("acknowledge_users", f"Thanks for the {len(msg.data)} users!")

def OnAcknowledge(self, msg, **kwargs):
    # Fired when the master receives a message of type 'acknowledge_users'
    print(msg.data)


@events.init.add_listener
def on_locust_init(environment, **_kwargs):
    if not isinstance(environment.runner, MasterRunner):
        environment.runner.register_message("test_users", SetupTestUsers)
    if not isinstance(environment.runner, WorkerRunner):
        environment.runner.register_message("acknowledge_users", OnAcknowledge)


def on_test_start(environment, **_kwargs):
    # When the test is started, evenly divides list between
    # worker nodes to ensure unique data across threads
    if not isinstance(environment.runner, WorkerRunner):
        users = []
        for i in range(environment.runner.target_user_count):
            users.append({"name": f"User{i}"})

        worker_count = environment.runner.worker_count
        chunk_size = int(len(users) / worker_count)

        for i, worker in enumerate(environment.runner.clients):
            start_index = i * chunk_size

            if i + 1 < worker_count:
                end_index = start_index + chunk_size
            else:
                end_index = len(users)

            data = users[start_index:end_index]
            environment.runner.send_message("test_users", data, worker)


class WebsiteUser(HttpUser):
    host = "http://127.0.0.1:8089"
    wait_time = between(2, 5)

    def __init__(self, parent):
        self.username = usernames.pop()
        super().__init__(parent)

    @task
    def task(self):
        print(self.username)
```

#### Second Test: 
Importing the `CustomMessageListener` and using it as a class decorator. 

```python
@CustomMessageListener
def SetupTestUsers(self, environment: "Environment", msg, **kwargs):
    # Fired when the worker receives a message of type 'test_users'
    usernames.extend(map(lambda u: u["name"], msg.data))
    environment.runner.send_message("acknowledge_users", f"Thanks for the {len(msg.data)} users!")

@CustomMessageListener
def OnAcknowledge(self, msg, **kwargs):
    # Fired when the master receives a message of type 'acknowledge_users'
    print(msg.data)
```

This results in the following error: 
`TypeError: Can't instantiate abstract class CustomMessageListener with abstract method __call__`

#### Third Test: 
Making the events to subclasses.
```python 
class SetupTestUsers(CustomMessageListener):
    def __call__(self, environment: "Environment", msg, **kwargs):
        # Fired when the worker receives a message of type 'test_users'
        usernames.extend(map(lambda u: u["name"], msg.data))
        environment.runner.send_message("acknowledge_users", f"Thanks for the {len(msg.data)} users!")

class OnAcknowledge(CustomMessageListener):
    def __call__(self, msg, **kwargs):
        # Fired when the master receives a message of type 'acknowledge_users'
        print(msg.data)
```
After starting the test this results in:
`TypeError: SetupTestUsers.__init__() takes exactly one argument (the instance to initialize)
2022-09-15T07:02:10Z <Greenlet at 0x1059c0040: <bound method WorkerRunner.worker of <locust.runners.WorkerRunner object at 0x10599b1f0>>> failed with TypeError` 



### Solving 
So I think there are two solutions to that. One is the current solution in this PR, which I think is still a bit bulky and not that pretty. 
The second one would be to go with the class decorator, probably as it was intended to be. So after some reading, we could probably change the `CustomMessageListener` from:
```python 
class CustomMessageListener(Protocol):
    @abstractmethod
    def __call__(self, environment: "Environment", msg: Message) -> None:
        ...
```
to 
```python
class CustomMessageListener(Protocol):
    def __init__(self, f):
        self.f = f  # Store the function

    def __call__(self, environment: "Environment", msg: Message) -> None:
        self.f(environment, msg)
```

This should make 
```python
@CustomMessageListener
def someEvent(self, environment: "Environment", msg, **kwargs):
    ...
```
work, and also should probably be the only thing that would need to be changed if I understood the concept correctly. 

Thanks a lot in advance!
Really looking forward to your comments :)

cyberw:
I had a quick look and really dont quite understand this. Why do you need to use an object method? And wouldnt this be a breaking change for everyone who uses simple methods?

cyberw:
Also, the current implementation in `examples/custom_messages.py` works fine for me.

```
locust -f custom_messages.py  --master --headless
[2022-09-15 20:43:45,011] .../INFO/root: Waiting for workers to be ready, 0 of 1 connected
Type     Name                                # reqs      # fails |    Avg     Min     Max    Med |   req/s  failures/s
--------|----------------------------------|-------|-------------|-------|-------|-------|-------|--------|-----------
--------|----------------------------------|-------|-------------|-------|-------|-------|-------|--------|-----------
         Aggregated                               0     0(0.00%) |      0       0       0      0 |    0.00        0.00

[2022-09-15 20:43:45,084] .../INFO/locust.runners: Worker ..._c52b8c5e8da741dcaf6e941446bb65a9 (index 0) reported as ready. 1 workers connected.
[2022-09-15 20:43:45,085] .../INFO/locust.runners: Worker ..._c52b8c5e8da741dcaf6e941446bb65a9 (index 0) reported as ready. 1 workers connected.
[2022-09-15 20:43:46,016] .../INFO/locust.main: No run time limit set, use CTRL+C to interrupt
[2022-09-15 20:43:46,016] .../INFO/locust.main: Starting Locust 2.11.2.dev15
[2022-09-15 20:43:46,017] .../INFO/locust.runners: Sending spawn jobs of 1 users at 1.00 spawn rate to 1 ready workers
Thanks for the 1 users!
[2022-09-15 20:43:46,095] .../INFO/locust.runners: All users spawned: {"WebsiteUser": 1} (1 total users)
```

```
locust -f custom_messages.py --worker
[2022-09-15 20:43:45,085] .../INFO/locust.main: Starting Locust 2.11.2.dev15
User0
User0

```

samuelspagl:
Okay sorry for the late response. Probably I didn't make myself clear, or couldn't explain what I was thinking. Sorry for that. 

> Also, the current implementation in `examples/custom_messages.py` works fine for me.

Same for me. And that is what I was talking about. 
So right now everything is working. And that's great. But also right the type hinting class `CustomMessageListener` for the messages, which was implemented a while ago, is being ignored. 

Right now when trying to be **type accurate**, if you would at the `@CustomMessageListner` decorator or wrap it into the `CustomMessageListener` class it would fail. 

Did I make it clearer? 😅

cyberw:
ah, ok, I understand a little better now. But you are suggesting a change breaking backwards compatibility? And the unit tests are failing?

samuelspagl:
> ah, ok, I understand a little better now. But you are suggesting a change breaking backwards compatibility? And the unit tests are failing?

The unit tests were failing because, my previous change was not backwards compatible. And I didn't adjust the tests that time.

cyberw:
> So the change in this pull request is not the change I would request. The change rn is more than it needs to be.
> 
> I think changing
> 
> ```
> class CustomMessageListener(Protocol):
>     @abstractmethod
>     def __call__(self, environment: "Environment", msg: Message) -> None:
> ```
> 
> to
> 
> ```
> class CustomMessageListener(Protocol):
>     def __init__(self, f):
>         self.f = f  # Store the function
> 
>     def __call__(self, environment: "Environment", msg: Message) -> None:
>         self.f(environment, msg)
> ```
> 
> Should be the only necessary change, and also retain backwards compatibility.
> 
> If okay with you, I would try that out and commit it to this (or another) PR, as you wish. :)

Cool! Yes, lets do a separate PR that doesnt include all of my confusion (but reference this one). Ping `@nathan-beam` there as well.

PR #2200, opened 2022-09-14
changing loggers in worker runner to log waiting for master

alvaro-ajv:
Adding logger for the worker when waiting for the master to connect and also when the maximum of retries were exeeded

cc: @cyberw


cyberw:
Nah, too spammy for info logging, especially if you run lots of workers and aggregate the logs (like with locust-swarm) Use `-L debug` if you need to know.

Issue #2192, opened 2022-09-11
Class picker incorrectly populates Parsed Options


Vino4:
### Describe the bug
When using the Web UI's class picker, Parsed Options are incorrectly populated

### Expected behavior
Each name of a picked class is available as a single item in the `environment.parsed_options.user_classes` list 
Example when selecting the class "PurchaseUser":
![image](https://user-images.githubusercontent.com/14026415/189515433-b0991191-2685-49ec-b52b-e1c6eb0b8e9d.png)

### Actual behavior
When choosing classes through class picker, the name of the class is broken into characters. Each character is present as its own item in the `environment.parsed_options.user_classes` list.
Example when selecting the class "PurchaseUser":
![image](https://user-images.githubusercontent.com/14026415/189515340-10c3a26e-a6bd-4321-878e-3300e372cd94.png)

### Steps to reproduce
- Start locust with the --class-picker option and **without** specifying a user class through the CLI
- Select a class from the class picker menu and start the test
- Access the `environment.parsed_options.user_classes` list in the `test_start` listener
```
@events.test_start.add_listener
def on_test_start(environment, **__):
    user_classes = environment.parsed_options.user_classes
    log.info(user_classes)
```

### Environment

- OS: MacOS Big Sur (11.4)
- Python version: 3.7.9
- Locust version: 2.12.0
- Locust command line that you ran: `locust --class-picker`
- Locust file contents (anonymized if necessary):
```
import logging
import time
from locust import HttpUser, task, between, events

log = logging.getLogger(__name__)

@events.test_start.add_listener
def on_test_start(environment, **_):
    user_classes = environment.parsed_options.user_classes
    log.info(user_classes)

class PurchaseUser(HttpUser):
    wait_time = between(1, 5)

    @task
    def hello_world(self):
        self.client.get("/hello")
        self.client.get("/world")

    @task(3)
    def view_items(self):
        for item_id in range(10):
            self.client.get(f"/item?id={item_id}", name="/item")
            time.sleep(1)

    def on_start(self):
        self.client.post("/login", json={"username":"foo", "password":"bar"})
```

cyberw:
@mikenester Could you take a look?

mikenester:
Whoops! Yeah i'll take a look asap

mikenester:
I think this is happening when `/swarm` [updates environment.parsed_options](https://github.com/locustio/locust/blob/master/locust/web.py#L198). I'll try to get a fix in today or tomorrow

[...]

PR #2201, opened 2022-09-14
Fix parsed options user classes when using class picker

mikenester:
Fix for https://github.com/locustio/locust/issues/2192

When `user_classes` is in the `/swarm` payload, `swarm` will update `environment.parsed_options.user_classes` to include the selected user_classes


cyberw:
Great!

Issue #2204, opened 2022-09-16
Editing a running test in the Web UI with class-picker restarts user count back at 0


edmundlam:
### Describe the bug
When I edit a running load test in the Web UI with `class-picker` active, it restarts the users back at 0 users and ramps up to the desired count. 

This is different from the behaviour without `--class-picker`, where if I edit to go from 10 to 20 users, it will merely add 10 more users on top of the existing 10.

Checking the code in `web.py`, it seems that if class-picker is active, the code will always restart the runners.

``` python
def swarm() -> Response:
            assert request.method == "POST"

            [...]

            # Loading UserClasses & ShapeClasses if Locust is running with UserClass Picker
            if self.userclass_picker_is_active:

                [...]

                # Stopping Runners and setting user dispatchers to None so that they
                # are recreated with the appropriate UserClasses
                self._stop_runners()
```

I'm wondering if we can skip the restarting of the runners if the user classes requested in the `POST /swarm` request are the same as those currently running. Especially since the "Edit" form for a running task doesn't let you change the class of users!

If this is something that you would agree to, I would be willing to open a PR to contribute the change. Thanks!

### Expected behavior
When using --class-picker, increasing from 10 to 20 will add 10 more users

### Actual behavior
When using --class-picker, increasing from 10 to 20 users will start from 0 again, and ramp up to 20 users.

### Steps to reproduce
1. Run locust with --class-picker
2. Start a test with X users
3. Edit the test to set it to X + 3 users
4. Problem is that users start back from 0, instead of staying at X and adding 3 users.

You can also run without --class-picker to see that the expected behaviour occurs when `--class-picker` is not set. 

### Environment

- OS: MacOS
- Python version: 3.9.7
- Locust version: 2.12.0
- Locust command line that you ran: `locust -f tests/load/locustfile.py --config tests/load/locust.conf` # with `class-picker = true` in the conf
- Locust file contents (anonymized if necessary):

cyberw:
@mikenester ?

edmundlam:
If it helps, this was the monkeypatch I am using to acheive the expected behaviour for my usage. I haven't tested it thoroughly though. 

``` diff
diff --git a/locust/web.py b/locust/web.py
index de688030..452dbf85 100644
--- a/locust/web.py
+++ b/locust/web.py
@@ -174,6 +174,8 @@ class WebUI:
                         # Defaulting to using all available user classes
                         user_classes = self.environment.available_user_classes

[...]

PR #2207, opened 2022-09-18
Bug Fix for User Class Count reset when editing a running test and using --class-picker

mikenester:
Fix for https://github.com/locustio/locust/issues/2204. It turns out that stopping the runners wasn't necessary. I couldn't think of a way to test the bug, but am happy to add a test if anyone has any thoughts.


cyberw:
thx!

PR #2208, opened 2022-09-19
Fix CustomMessageListener class, and update example and documentation

samuelspagl:
This is an implementation of the changes mentioned in #2195 which therefore can be closed. 

#### Content of the PR
In previous commits there was the `CustomMessageListener` class implemented, to add type hinting to the message listeners.
This PR is fixing the bug of not being able to utilise this typehinting class. It makes sure to enable users to use the `@CustomMessageListener` decorator for their listener functions, while still being backwards compatible. 

#### Changes made
In `runners.py`
```python 
class CustomMessageListener(Protocol):
    @abstractmethod
    def __call__(self, environment: "Environment", msg: Message) -> None:
```
to
```
class CustomMessageListener:
    def __init__(self, f):
        self.f = f  # Store the function

    def __call__(self, environment: "Environment", msg: Message) -> None:
        self.f(environment=environment, msg=msg)
```

#### Additional Changes:
Updated the `custom_messages.py` example and the `running-distributed.rst` documentation file.

### DIFF
--- a/docs/running-distributed.rst
+++ b/docs/running-distributed.rst
@@ -88,15 +88,17 @@ order to coordinate data. This can be easily accomplished with custom messages u
 .. code-block:: python
 
     from locust import events
-    from locust.runners import MasterRunner, WorkerRunner
+    from locust.runners import MasterRunner, WorkerRunner, CustomMessageListener
 
     # Fired when the worker receives a message of type 'test_users'
+    @CustomMessageListener
     def setup_test_users(environment, msg, **kwargs):
         for user in msg.data:
             print(f"User {user['name']} received")
         environment.runner.send_message('acknowledge_users', f"Thanks for the {len(msg.data)} users!")
 
     # Fired when the master receives a message of type 'acknowledge_users'
+    @CustomMessageListener
     def on_acknowledge(msg, **kwargs):
         print(msg.data)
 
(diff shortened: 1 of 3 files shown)


samuelspagl:
ping @nathan-beam, @cyberw

cyberw:
Looking at the updated example I'm not sure about this. This is different from other Locust event handlers work, and I'm not sure if the added complexity (importing a class, tagging the method) is worth it. But lets have Nathan weigh in.

samuelspagl:
Okay :) then let's wait. 

Just to be on the same page: it is fully backwards compatible. So you can also just use it without the decorator. It's just for the type consistency. 

So the other side would be completely deleting the typehint? 

For me it was very confusing to have this not working as expected type hint :)

cyberw:
Right, we shouldnt have broken type hints either, so thats not an option :)

samuelspagl:
So any news about this? :) Just wanna check.

cyberw:
I guess @nathan-beam is asleep. I dont like having to introduce a decorator to the examples. There should be some more user friendly solution. But I’m afk, and probably wont really have the energy once I get back :)

samuelspagl:
Okay, so rn I don't have another solution for this. I honestly think that the decorator is kinda the cleanest method, as it also is not breaking any existing code.

Then I would propose to remove the `CustomMessageListener` class completely. That way it will be type consistent again.
I can prepare a PR, if you want that. :)

cyberw:
I think that might be a good idea. That class never really made sense to me :) If you make a PR I’ll be sure to review it.

PR #2214, opened 2022-09-24
Added milliseconds in the stats.py file

balaji2711:
I would like to have milli-seconds as well in the report. So I have changed the code (line number:886). Kindly review and approve the changes.

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -883,7 +883,7 @@ def stats_history(runner: "Runner") -> None:
             break
         if runner.state != "stopped":
             r = {
-                "time": datetime.datetime.utcnow().strftime("%H:%M:%S"),
+                "time": datetime.datetime.utcnow().strftime("%H:%M:%S.%f")[:-3],
                 "current_rps": stats.total.current_rps or 0,
                 "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
                 "response_time_percentile_95": stats.total.get_current_response_time_percentile(0.95) or 0,


cyberw:
Lgtm. But can you look into adding a unit test? (that at least verifies the length of the time stamp)

balaji2711:
Hi @cyberw,
I have written the unit tests, Could you please add them to the respective file -

import datetime
r = {
                "time": datetime.datetime.utcnow().strftime("%H:%M:%S.%f")[:-3],
                "current_rps": stats.total.current_rps or 0,
                "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
                "response_time_percentile_95": stats.total.get_current_response_time_percentile(0.95) or 0,
                "response_time_percentile_50": stats.total.get_current_response_time_percentile(0.5) or 0,
                "user_count": runner.user_count or 0,
            }
if len(r.get('time')) <=12:
    print('Timestamp is within the length')
else:
    assert False, "Timestamp is not within the length"

cyberw:
No time right now. It should fit somewhere in test_stats.py (https://github.com/locustio/locust/blob/master/locust/test/test_stats.py)

balaji2711:
Hi @cyberw, 

Ok, Thanks for letting me know the file name.
I have written the unit test in the test_stats.py and raised the pull request -
https://github.com/locustio/locust/pull/2215

cyberw:
Actually, this does not make sense after all. Because we're only logging every 5 seconds, having milliseconds in the timestamp is more noice/hard to read than valuable.

cyberw:
This might make sense as an option sometimes (e.g. if  HISTORY_STATS_INTERVAL_SEC < 1, but I dont even know if that works)

PR #2215, opened 2022-09-24
Added Unit test to validate the timestamp

balaji2711:
Added Unit test to validate the timestamp (def test_time(self)), Please refer to the line from 104 to 109. With reference of this pull request - https://github.com/locustio/locust/pull/2214


cyberw:
That tests strftime, you’re supposed to test Locust… (the stats_history method)

cyberw:
See other PR

PR #2217, opened 2022-09-30
GraphQL example for locust

sanin-mohd:
Hi, After some R&D  I have successfully integrated GraphQL with locust . I hope this will be helpful for someone. pls validate the PR


cyberw:
Cool stuff! But I think it makes more sense as a stand-alone example in examples/ , with a reference from the documentation (rst files)

PR #2219, opened 2022-10-03
Locust custom

kbhatia2910:


heyman:
Hi! Please provide a detailed description of any pull-requests. From a quick peek It's not obvious to me what the PR does, and if it's something we would like to merge into Locust. All pull-requests should also include tests and documentation.

PR #2221, opened 2022-10-12
Delete the `CustomMessageListener` class for type consistency

samuelspagl:
This PR is the conclusion of #2208. 
It deletes the `CustomMessageListener` class for type consistency and adds `Callable` type hint instead.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -27,7 +27,8 @@ from typing import (
     Tuple,
     Type,
     Any,
-    cast, Callable,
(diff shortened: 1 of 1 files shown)


samuelspagl:
@cyberw ping

I used `Callable` instead of `any` as I think it is a bit more concise. Hope that's okay. 
Looking forward to your comments.

cyberw:
Definitely good enough. Thanks!

PR #2227, opened 2022-10-16
Update on README for deployment on local enviroment

JonanOribe:
A set of instructions about how to deploy Locust locally


cyberw:
Thats just instructions for using venv? Using venv is entirely optional and the instructions are not specific to locust, so I dont think they belong in the manual. Maybe add the to the installation page in the wiki and link them from manual?

JonanOribe:
You are right, I'm going to add this to the wiki.Thanks!

PR #2228, opened 2022-10-16
URL link on the host name for fast navigation to the API

JonanOribe:
I have added a direct link from Locust to the API indicated by the user so that the user can move directly to it


cyberw:
No js code formatting changes please. If you want to do this, make sure the only diff is what you have actually changed (adding the link)

JonanOribe:
Ok, I will correct the formatting and make an upload

cyberw:
Thx

Issue #2243, opened 2022-11-03
'Namespace' object has no attribute 'stop_timeout' in version 2.13.0


klazuka:
### Describe the bug

After upgrading to 2.13.0, we started seeing the following error:
```
  File "/path/to/app/venv/lib/python3.9/site-packages/locust/env.py", line 84, in __init__
    self.stop_timeout = float(parsed_options.stop_timeout)
AttributeError: 'Namespace' object has no attribute 'stop_timeout'
```

We are using Locust as a library, and when creating the Locust `Environment`, we are passing in a `Namespace` object for `parsed_options`. In old versions of Locust, we did not need to pass `stop_timeout`, but now it appears that we do. Is this intentional?

### Expected behavior
Existing code should continue to work after a minor version upgrade.

### Actual behavior
An `AttributeError` exception is raised from within Locust.

### Steps to reproduce

```
from argparse import Namespace
from locust.env import Environment
parsed_options = Namespace(timeout=42)
Environment(parsed_options=parsed_options)
```

^ the above works in 2.12.0 but fails in 2.13.0

### Environment

- OS: Linux
- Python version: 3.9.11
- Locust version: 2.13.0
- Locust command line that you ran: n/a
- Locust file contents (anonymized if necessary): n/a

cyberw:
Dont construct your Environment that way. Pass arguments explicitly: `Environment(stop_timeout=42)`

PR #2247, opened 2022-11-07
Update documentation for Environment.parsed_options

klazuka:
Improves documentation to avoid misuse of the `parsed_options` argument to `Environment`.
See #2243


cyberw:
Thx

PR #2248, opened 2022-11-09
Added a better working docker command for Windows users

MagnusNordboe:
I struggled with several errors using the supplied Docker commands on Windows using WSL, so I have added an alternative Docker command that gave the desired result.


cyberw:
👍👍

Issue #2256, opened 2022-11-16
Extend typing of "task" property of a locust user


samuelspagl:
### Describe the bug

Creating a dictionary of `Callable,int` for the `tasks` attribute of a locust user is failing common python linters. (for example "mypy")

### Expected behavior

The type hinting should be inline with the comment:
```
    Collection of python callables and/or TaskSet classes that the Locust user(s) will run.

    If tasks is a list, the task to be performed will be picked randomly.

    If tasks is a *(callable,int)* list of two-tuples, or a {callable:int} dict,
    the task to be performed will be picked randomly, but each task will be weighted
    according to its corresponding int value. So in the following case, *ThreadPage* will
    be fifteen times more likely to be picked than *write_post*::
```

### Actual behavior

Resulting mypy error:
```bash
Mypy: Incompatible types in assignment (expression has type "Dict[Callable[..., None], Any]", base class "User" defined the type as "List[Union[TaskSet, Callable[..., Any]]]") [assignment]
```

### Steps to reproduce
 
I'm using as IDE and also have the `mypy` plugin active, therefore this error is shown in the IDE itself.
But it should also be possible to just start the linting process for these lines.
```python
class SomeUser(FastHttpUser):
    tasks = {
        some_task_1: 12,
        some_task_2: 4,
        some_task_3: 6
    }
...
```

### Solution

Extend the typing in the `user` class with the following: 
```python
from
    tasks: List[TaskSet | Callable]  = []
to
    tasks: List[TaskSet | Callable]  | dict[Callable, int] | Tuple = []
```

### Environment

- OS: MacOS
- Python version: 3.10
- Locust version: 2.13.0

Looking forward to your comments. :)

cyberw:
Lgtm, PR welcome.

samuelspagl:
kk I'll make one :)

PR #2258, opened 2022-11-17
Fix: Enhance typing of the tasks attribute in the User class

samuelspagl:
#### Description
This PR fixes a minor typing issue in the `User` class, it is related to the issue #2256.

#### The change
The code is changed from: 
```python
from
    tasks: List[TaskSet | Callable]  = []
to
    tasks: List[TaskSet | Callable | Tuple]  | dict[Callable, int] = []
```

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -245,6 +245,8 @@ class Environment:
             u.weight = 1
             user_tasks: List[TaskSet | Callable] = []
             tasks_frontier = u.tasks
+            isinstance(tasks_frontier, list)
+
             while len(tasks_frontier) != 0:
                 t = tasks_frontier.pop()
                 if isinstance(t, TaskHolder):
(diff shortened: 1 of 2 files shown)


samuelspagl:
Okay so I saw that `mypy` runs now into an error in the `assign_equal_weights` function in the `env.py`.
If I understood it correctly then somewhere in the code the `Dict[Callable, int]` is being converted into a `List[Callable]`. Therefore I added a `isinstance(tasks_frontier, list)` to give mypy the information that it doesn't need to be confused by the other types. 

Hope that's fine :)

samuelspagl:
@cyberw there is one mypy error left could you take a look at it? 
```
locust/env.py:222: error: Argument 1 to "filter_tasks_by_tags" has incompatible type "Type[User]"; expected "Type[TaskHolder[Any]]"
```

I needed to add a `type: ignore` to one line. If needed I can elaborate why. 😊

Issue #2250, opened 2022-11-13
Mass monkeypatching leads to boto3 incompatibility and infinite recursion in pytest (when used as lib)


inpv:
### Describe the bug

I know this [issue](https://github.com/gevent/gevent/issues/1016) has been referenced before, but it seems that re-patching all modules with the exception of Python's native SSL with:
```
from gevent import monkey
monkey.patch_all(ssl=False)
```
in the beginning of a Python test file doesn't lead to any positive results when using locust as library in pytest if boto3 is also installed.

I wanted to introduce pytest load testing in a PR to a repo that I cloned. Chose locust for integration and extension capabilities.
Wrote a simple [function](https://github.com/inpv/cvat/blob/inpv_test_task_mass_user_register/tests/python/rest_api/test_load_register.py) to use locust as library based on the example in the docs, checked that the test works as expected, but if I try to launch all REST API tests with locust installed, half of them fail because of mass monkeypatching that locust introduces in the [init](https://github.com/locustio/locust/blob/b64b9a6082451be74f94bf96406315fa819fcd1c/locust/__init__.py#L11) file. So I disabled ssl monkeypatching explicitly like I wrote above. Now this
```
.env/lib/python3.9/site-packages/locust/__init__.py:11
  /home/inpv/Development/cvat/.env/lib/python3.9/site-packages/locust/__init__.py:11: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util (/home/inpv/Development/cvat/.env/lib/python3.9/site-packages/urllib3/util/__init__.py)', 'urllib3.util.ssl_ (/home/inpv/Development/cvat/.env/lib/python3.9/site-packages/urllib3/util/ssl_.py)']. 
    monkey.patch_all()
```
warning is gone, but there is another:
```
.env/lib/python3.9/site-packages/locust/__init__.py:11
  /home/inpv/Development/cvat/.env/lib/python3.9/site-packages/locust/__init__.py:11: MonkeyPatchWarning: Patching more than once will result in the union of all True parameters being patched
    monkey.patch_all()
```

[...]

PR #2262, opened 2022-11-20
Add conditional monkeypatching for threads (see issue #2250)

inpv:
Mass monkeypatching via `monkey.patch_all()` in some setups did lead to double patching and [incompatibilities](https://github.com/locustio/locust/issues/2250) e.g. with `boto3` modules, hence there was a [proposition](https://github.com/locustio/locust/issues/2250#issuecomment-1315065631) to patch only the `thread` module if it wasn't patched yet.

### DIFF
--- a/locust/__init__.py
+++ b/locust/__init__.py
@@ -8,7 +8,8 @@ if os.getenv("LOCUST_PLAYWRIGHT", False):
 
 from gevent import monkey
 
-monkey.patch_all()
+if not monkey.is_module_patched('thread'):
+    monkey.patch_thread()
 
 from ._version import version as __version__
 from .user.sequential_taskset import SequentialTaskSet
--- a/locust/__init__.py
+++ b/locust/__init__.py
@@ -8,8 +8,8 @@ if os.getenv("LOCUST_PLAYWRIGHT", False):
 
 from gevent import monkey
 
-if not monkey.is_module_patched('thread'):
-    monkey.patch_thread()
+if monkey.is_module_patched("ssl"):
+    monkey.patch_all(ssl=False)
 
 from ._version import version as __version__
 from .user.sequential_taskset import SequentialTaskSet


cyberw:
I dont think that is a good fix, it deadlocked all tests :)

What I meant was "only apply the patches that have not yet been applied", your change makes it so that everything except threading is never patched.

inpv:
> I dont think that is a good fix, it deadlocked all tests :)
> 
> What I meant was "only apply the patches that have not yet been applied", your change makes it so that everything except threading is never patched.

Ok. I had a solution like this,
```Python
import inspect

[monkey._patch_module(module) for module in inspect.getfullargspec(monkey.patch_all).args
 if not monkey.is_module_patched(module)]
```
but it seems that `gevent`'s single-module patching function doesn't match the possible imports:
```
ImportError while importing test module '~/Development/cvat/tests/python/rest_api/test_load_register.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/python/rest_api/test_load_register.py:5: in <module>
    import locust
.env/lib/python3.9/site-packages/locust/__init__.py:12: in <module>
    [monkey._patch_module(module) for module in inspect.getfullargspec(monkey.patch_all).args
.env/lib/python3.9/site-packages/locust/__init__.py:12: in <listcomp>
    [monkey._patch_module(module) for module in inspect.getfullargspec(monkey.patch_all).args
.env/lib/python3.9/site-packages/gevent/monkey.py:462: in _patch_module
    gevent_module, target_module, target_module_name = _check_availability(name)
.env/lib/python3.9/site-packages/gevent/monkey.py:448: in _check_availability
    gevent_module = getattr(__import__('gevent.' + name), name)
E   ModuleNotFoundError: No module named 'gevent.dns'
```

and if I use it this way:
```Python
[monkey.patch_module(module, 'gevent.' + str(module)) for module in inspect.getfullargspec(monkey.patch_all).args
 if not monkey.is_module_patched(module)]
```
`gevent` raises a `BadImplements` error:
```
tests/python/rest_api/test_load_register.py:5: in <module>
    import locust
.env/lib/python3.9/site-packages/locust/__init__.py:12: in <module>
    [monkey.patch_module(module, 'gevent.' + str(module)) for module in inspect.getfullargspec(monkey.patch_all).args
.env/lib/python3.9/site-packages/locust/__init__.py:12: in <listcomp>
    [monkey.patch_module(module, 'gevent.' + str(module)) for module in inspect.getfullargspec(monkey.patch_all).args
.env/lib/python3.9/site-packages/gevent/monkey.py:400: in patch_module
    raise _BadImplements(source_module)
E   gevent.monkey._BadImplements: Module 'gevent.socket' has a bad or missing value for __implements__
```
so the solution is probably either creating a list with module names beforehand or mapping module names to `patch_x` methods.

What am I missing here?

cyberw:
Not sure. if only threading causes issues with double patching (docs are kind of vague on whether it is allowed to double patch), then maybe you can just do `patch_all(thread=False)` in that case. That reduces the change as well.

inpv:
> Not sure. if only threading causes issues with double patching (docs are kind of vague on whether it is allowed to double patch), then maybe you can just do `patch_all(thread=False)` in that case. That reduces the change for everyone else as well.

The error emerged not because of `thread`, but because of `ssl` double patching, which is stated in the warnings in my bug [report](https://github.com/locustio/locust/issues/2250#issue-1446883951):
```
.env/lib/python3.9/site-packages/locust/__init__.py:11
  /home/inpv/Development/cvat/.env/lib/python3.9/site-packages/locust/__init__.py:11: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util (/home/inpv/Development/cvat/.env/lib/python3.9/site-packages/urllib3/util/__init__.py)', 'urllib3.util.ssl_ (/home/inpv/Development/cvat/.env/lib/python3.9/site-packages/urllib3/util/ssl_.py)']. 
    monkey.patch_all()
```
and
```
.env/lib/python3.9/site-packages/locust/__init__.py:11
  /home/inpv/Development/cvat/.env/lib/python3.9/site-packages/locust/__init__.py:11: MonkeyPatchWarning: Patching more than once will result in the union of all True parameters being patched
    monkey.patch_all()
```
For example, the solution below works in my case:
```Python
if monkey.is_module_patched("ssl"):
    monkey.patch_all(ssl=False)
```

heyman:
I think a solution could be to introduce an environment variable that'll disable all monkey patching if it's set, and it'll then be up to the user to patch stuff themselves. I like this because it won't break any existing code.

cyberw:
The updated PR is still broken. If there is no sensible way to iterate over all things that can be patched then something like this is more appropriate (and backwards compatible). 

```
if monkey.is_module_patched("ssl"):
    monkey.patch_all(ssl=False)
else:
    monkey.patch_all()
```

I dont think we need an env var/setting for this because it is kind of obvious what the user wants (patch all things that have not yet been patched, but not more).

heyman:
> I dont think we need an env var/setting for this because it is kind of obvious what the user wants (patch all things that have not yet been patched, but not more).

There might be cases (such as the original issue in #2250) where a third-party lib is incompatible with monkey patching. Then it might make sense to provide a way to disable it entirely and let the user handle the monkey patching themselves (Locust itself should only need socket monkey patching, I think, but auto-patching all the stuff should provide a better user experience in most cases).

cyberw:
I think my suggestion sovlves the original issue and has no real down sides (nobody should ever _want_ to re-patch stuff)

A settings for bypassing monkey patching entirely might be useful in some cases, but it isnt needed for this issue. The third party lib in this case is not incompatible with monkey patching, just with applying monkey patching _twice_.

PR #2269, opened 2022-12-06
Stats in json to stdout

AndersSpringborg:
I extended the cli with a `--json` argument. This argument prints the result of a run in json, after a run is finished. It blocks other INFO logs, so it's easy to interoperable with other programs. Below is an example with `jq`

```shell
❯ python -m locust -f examples/locustfile.py --host http://localhost:80 --headless -u 10 -t 4s --json
[
    {
        "name": "/",
        "method": "GET",
        "last_request_timestamp": 1670363796.9102132,
        "start_time": 1670363793.665749,
        "num_requests": 8,
        "num_none_requests": 0,
        "num_failures": 0,
        "total_response_time": 61.882097012130544,
        "max_response_time": 18.852641980629414,
        "min_response_time": 3.2629000197630376,
        "total_content_length": 4920,
        "response_times": {
            "19": 1,
            "9": 1,
            "4": 3,
            "10": 1,
            "3": 1,
            "8": 1
        },
        "num_reqs_per_sec": {
            "1670363793": 1,
            "1670363794": 2,
            "1670363795": 1,
            "1670363796": 4
        },
        "num_fail_per_sec": {}
    }
]
```
```shell
❯ python -m locust -f examples/locustfile.py --host http://localhost:80 --headless -u 10 -t 4s --json | jq .[0].response_times
{
  "11": 2,
  "10": 1,
  "4": 4,
  "9": 1
}
```
Locust: 2.13.2.dev11
Python: 3.10.8

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -492,7 +492,7 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
         "--json",
         default=False,
         action="store_true",
-        help="Prints the final stats in JSON format to stdout. Usefull for parsing the results in other programs/scripts."
+        help="Prints the final stats in JSON format to stdout. Usefull for parsing the results in other programs/scripts.",
     )
 
     log_group = parser.add_argument_group("Logging options")
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -488,6 +488,12 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
         help="Store HTML report to file path specified",
         env_var="LOCUST_HTML",
     )
+    stats_group.add_argument(
+        "--json",
+        default=False,
+        action="store_true",
+        help="Prints the final stats in JSON format to stdout. Usefull for parsing the results in other programs/scripts."
+    )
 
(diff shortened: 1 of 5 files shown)


cyberw:
Nice!

I think it makes more sense to send the output to stdout (using print()) instead of stderr (using logging). That removes the need to disable logging too.

Can you also add a test case? (either an integration test in test_main.py, or in test_stats.py or whatever it is called)

AndersSpringborg:
Yeah I'll do that. That seems easier to maintain too.

I'll make some tests too

cyberw:
I think the user specifying `--skip-log-setup` or `--loglevel ERROR` when you need to suppress stderr logging is better. Changing the log level as a side effect of setting `--json` is not what I would expect, as a user.

AndersSpringborg:
I've made some integration tests, and deleted the part about skipping logging.

Merry Christmas

cyberw:
Cool stuff, thanks!

meeech:
@cyberw @AndersSpringborg out of curiosity, why not matching ENV_VAR for this option? Any particular reason? Would a pr to add that be welcome?

Issue #2260, opened 2022-11-18
UnboundLocalError after receiving ZMQ corrupted message


marcinh:
[PR 2120](https://github.com/locustio/locust/pull/2120) introduced problem with wrong handling of ZMQ corrupted message what leads to:
UnboundLocalError: local variable 'client_id' referenced before assignment.

```
locust-master: 2022-11-11T07:03:39.637+0000 |CRITICAL|locust.runners           |  Unhandled exception in greenlet: <Greenlet at 0x7fab1317dae0: <bound method MasterRunner.client_listener of <locust.runners.MasterRunner object at 0x7fab0767d580>>>
locust-master: Traceback (most recent call last):
locust-master: File "/root/.local/share/virtualenvs/locust-1T7N9san/lib/python3.9/site-packages/locust/rpc/zmqrpc.py", line 44, in recv_from_client
locust-master: addr = data[0].decode()
locust-master: UnicodeDecodeError: 'utf-8' codec can't decode byte 0x86 in position 9: invalid start byte
locust-master: The above exception was the direct cause of the following exception:
locust-master: Traceback (most recent call last):
locust-master: File "/root/.local/share/virtualenvs/locust-1T7N9san/lib/python3.9/site-packages/locust/runners.py", line 960, in client_listener
locust-master: client_id, msg = self.server.recv_from_client()
locust-master: File "/root/.local/share/virtualenvs/locust-1T7N9san/lib/python3.9/site-packages/locust/rpc/zmqrpc.py", line 46, in recv_from_client
locust-master: raise RPCReceiveError("ZMQ interrupted or corrupted message") from e
locust-master: locust.exception.RPCReceiveError: ZMQ interrupted or corrupted message
locust-master: During handling of the above exception, another exception occurred:
locust-master: Traceback (most recent call last):
locust-master: File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run
locust-master: File "/root/.local/share/virtualenvs/locust-1T7N9san/lib/python3.9/site-packages/locust/runners.py", line 962, in client_listener
locust-master: logger.error(f"RPCError when receiving from client: {e}. Will reset client {client_id}.")
locust-master: UnboundLocalError: local variable 'client_id' referenced before assignment
[...]

PR #2271, opened 2022-12-07
Fix: Ask worker to reconnect if master gets a broken RPC message

marcinh:
#2260

Follow up - restore reconnect when client_id is known


cyberw:
Thx!

marcinh:
Are you planning to do a release in near future?

cyberw:
Done!

PR #2272, opened 2022-12-09
GRPC example - rewrite using interceptor

zifter:
It's better to use interceptor instead of old approach described in old example
I'm using suggested interceptor and it's working pretty good.

### DIFF
--- a/docs/testing-other-systems.rst
+++ b/docs/testing-other-systems.rst
@@ -42,9 +42,10 @@ Dummy server to test:
 
 .. literalinclude:: ../examples/grpc/hello_server.py
 
-gRPC client, base User and example usage:
+gRPC client, base GrpcUser, interceptor for sending events to locust and example usage:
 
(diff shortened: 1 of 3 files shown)


zifter:
@cyberw thanks for review!
Generally speaking, it's the same wrapper pattern, but implemented  natively in grpc library. A bit more information about grpc [interceptors](https://grpc.io/blog/grpc-web-interceptor/). It's definitely not a callback, because interceptors wraps the original request, and a set of interceptors become a chain during a method call. So, how is working with grpc should know that pattern.

I didn't remove `import grpc.experimental.gevent as grpc_gevent`, it's [moved](https://github.com/locustio/locust/pull/2272/files#diff-5e6469a21b168ea4f8bcd10b34e98dcbfdffa73956f1534e54da7dedaa1dcc52R7) because of correct import order in python.

Yes, I have tested with many users locally, it's working.

cyberw:
Ah, I misread. Looks nice now!

PR #2276, opened 2022-12-16
Fix exception grouping for requests with both catch_response and name arguments

ianmetcalf:


cyberw:
Lgtm! @heyman ? Can you add some more context to the PR? (describe the problem it fixes in the title)

heyman:
:+1: Looks good!

PR #2278, opened 2022-12-21
Fix some format problems in document

THUzxj:
Side menu preview: 
image


cyberw:
Nice! I think I'll want to reorder some things, but overall this is a big improvement. Thanks!

PR #2284, opened 2023-01-07
add is_secret option for custom args to be shown in the web UI masked

mzhukovs:
Thought it would be nice, especially for live demo purposes, to be able to quickly indicate when adding custom arguments that are to be exposed in the web UI which should be password masked text inputs, e.g. when you want to use the Web UI to run the load tests under different user logins, so the user password text input field should be a password input.


locust/templates/index.html
```diff
@@ -103,14 +103,14 @@ <h2>Start new load test</h2>
                     </div>
                     {% if extra_options %}<label>Custom parameters:</label>{% endif %}
                     {% for key, value in extra_options.items() %}
-                        {% if not ((value is none) or (value is boolean)) %}
+                        {% if not ((value[0] is none) or (value[0] is boolean)) %}
                             <label for="{{key}}">{{key}}</label>
-                            <input type="text" name="{{key}}" id="{{key}}" class="val" value="{{value}}" /><br>
+                            <input type="{{'password' if value[1] else 'text'}}" name="{{key}}" id="{{key}}" class="val" value="{{value[0]}}" /><br>
```
cyberw, line 108: This is a little too weird/un-obvious. Maybe use a dict instead of a tuple?

cyberw:
Makes sense. I’m thinking maybe the option should just be `password` or `is_password` instead of `is_secret`?

mzhukovs:
> Sounds useful. I’m thinking maybe the option should just be`is_password` instead of `is_secret`?

My thought was "secret" is a lot more general than "password" - e.g. what if the user wanted to have the API Key as a masked input? It's not really a password, but it is a secret - so feels a little more intuitive for use with parser.add_argument() but of course I defer to you.

locust/templates/index.html
mzhukovs: I agree @cyberw , it even felt wrong as I was typing it. How about a NamedTuple instead? I also threw in help_text there to display in same style as the other default inputs displayed in the UI if provided. Please refer to new commit 64e930399124f19810550561b3cd791b5e6567cd 

Thank you for your consideration

mzhukovs:
Also, is it just me or does the CSS for the footer need to be updated to make it sticky (without overlapping with actual content)?

cyberw:
> > Sounds useful. I’m thinking maybe the option should just be`is_password` instead of `is_secret`?
> 
> My thought was "secret" is a lot more general than "password" - e.g. what if the user wanted to have an API Key as a masked input? It's not really a password, but it is a secret - so feels a little more intuitive for use with parser.add_argument() but of course I defer to you.

because i’m used to html is_password made more sense to me :) but it doesnt really matter.

cyberw:
@heyman what do you think?

cyberw:
> Also, is it just me or does the CSS for the footer need to be updated to make it sticky (without overlapping with actual content)?

Idk :)

mzhukovs:
> > Also, is it just me or does the CSS for the footer need to be updated to make it sticky (without overlapping with actual content)?
> 
> Idk :)

Yes definitely, please refer to latest commit 52fe0746391cb3abc4297562c55d1a381a582b78 to fix this - also I noticed the "About" popup (from the link in the footer) was getting covered up by the header so adjusted the z-index to prevent that as well

heyman:
Sounds reasonable! I'd prefer `is_secret` over `is_password` I think. 

We should also add something about this under the [Custom arguments](https://docs.locust.io/en/stable/extending-locust.html#custom-arguments) section in the documentation for extending Locust (which I also think we should rename to "Extending Locust" instead of "Event hooks" btw, see my comment at: https://github.com/locustio/locust/commit/94694864b090e34c22902a5172c0803a5115cf41#r95553244)

cyberw:
I can merge this now if there is nothing more to add.

mzhukovs:
> Sounds reasonable! I'd prefer `is_secret` over `is_password` I think.
> 
> We should also add something about this under the [Custom arguments](https://docs.locust.io/en/stable/extending-locust.html#custom-arguments) section in the documentation for extending Locust (which I also think we should rename to "Extending Locust" instead of "Event hooks" btw, see my comment at: [9469486#r95553244](https://github.com/locustio/locust/commit/94694864b090e34c22902a5172c0803a5115cf41#r95553244))

Yep this was already done, please refer to this part of the first commit: https://github.com/locustio/locust/pull/2284/commits/e71eb8fe7fdfc55ad10d1085dc4360788f62b045#diff-0a5d25e3b717d98c55f177095b87d64583ccfb2e076c91c03e4afcbf155864dd

Which the docs file docs/extending-locust.rst references.

heyman:
> Yep this was already done, please refer to this part of the first commit: [e71eb8f#diff-0a5d25e3b717d98c55f177095b87d64583ccfb2e076c91c03e4afcbf155864dd](https://github.com/locustio/locust/commit/e71eb8fe7fdfc55ad10d1085dc4360788f62b045#diff-0a5d25e3b717d98c55f177095b87d64583ccfb2e076c91c03e4afcbf155864dd)
> 
> Which the docs file docs/extending-locust.rst references.

Ah, yes, I forgot that the example file is inlined into the documentation :+1:

PR #2285, opened 2023-01-11
add events when initialize Environment

keegoo:
Add events when initialize `Environment(user_classes=[MyUser], events=events)`, otherwise the event codes in the example(as following) won't work

```python
# ...
env.events.init.fire(environment=env, runner=runner, web_ui=web_ui)
```

### DIFF
--- a/examples/use_as_lib.py
+++ b/examples/use_as_lib.py
@@ -1,5 +1,5 @@
 import gevent
-from locust import HttpUser, task
+from locust import HttpUser, task, events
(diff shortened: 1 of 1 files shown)


keegoo:
This is my first PR to the repo, please feel free to let me know if I missed anything, thanks!

xaviergmail:
I was deep in the rabbit hole wondering why I wasn't getting any stats when running a funky distributed setup. This just saved my bacon, thank you!

Issue #2188, opened 2022-09-08
locust k8s operator


AbdelrhmanHamouda:
### Is your feature request related to a problem? Please describe.
Need for flexible, secure and easy to use distributed performance testing solution while enabling keeping in mind points like; cloud governance, k8s resource management, CI compatibility. 

### Describe the solution you'd like
A k8s operator

### Describe alternatives you've considered
HELM. However, this is not optimal as it requires a lot of manipulation and dealing with a lot of different k8s resource types not to mention lack of governance when deployed in an organisation. 

 
### Additional context
In my company, we were interested in cloud based distributed performance testing. We ended up choosing “locust” as a tool and I wrote a kubernetes operator for it. As far as I know and searched, this simply doesn’t exist for locust today.The operator have been running on AWS EKS for months now and it is very stable and performant.

The project was created to provide:
- A friendly user interface "A simple k8s custom resource based on a CRD"
- Ability to run a distributed performance testing
- Only ask for needed cluster permissions (security) 
- Exposing a Prometheus endpoint for test metrics
- Maintain cloud infrastructure governance i.e. when used in an organisation,  it is possible to maintain the governance over the allocated / created resources. 
- Cloud native by design 
- Enablement of performance testing in CI / CD pipelines

The project was inspired after doing an industry study / survey. 
The project was influenced in functionality by the k6 k8s operator.

As the code really doesn't include anything exclusive to the company, Open sourcing it is not a problem and it is something that I would happily do. 

What is the "catch"? 
The only "catch" is that the project is written in Java. That being said, from a user perspective, the interface purely through k8s and users are not exposed to any java code.  This is similar to how k8s and most operators are written in Go but non of k8s users need to know anything about Go. 

Looking forward to your feedback.

OneCricketeer:
I see you listed Helm as an alternative, but that doesn't really address other solutions to run distributed load tests such as [`k6` on k8s](https://k6.io/blog/running-distributed-tests-on-k8s/)

AbdelrhmanHamouda:
Before selecting Locust, we actually did a market study to see which tool would best fit our needs. K6 was on the table and it almost made the cut if i'm being honest. However "and this is a big one",  when it came down to distributed performance testing, locust was much better with supporting functionality that come out of the box. k6 "at least at the time of the study" was really focused on single instance tests.

AbdelrhmanHamouda:
[...]

PR #2288, opened 2023-01-16
docs (#2188):  Add Locust Kubernetes Operator

AbdelrhmanHamouda:
locustio/locust#2188 

- Update `Running a distributed load test on Kubernetes` docs section to include Locust Kubernetes Operator. 
- Small reformat of the section to clearly list all possibilities (HELM & Operator)  in a structured manner.

### DIFF
--- a/docs/running-in-docker.rst
+++ b/docs/running-in-docker.rst
@@ -39,10 +39,21 @@ official Locust docker image as a base image::
 
 
 Running a distributed load test on Kubernetes
-=============================================
+==============================================
 
-The easiest way to run Locust on Kubernetes is to use a `Helm chart <https://helm.sh/>`_.
+HELM
+*****
+
+One of the ways to run Locust on Kubernetes is to use a `Helm chart <https://helm.sh/>`_.
 
 There is a good helm chart here: `github.com/deliveryhero/helm-charts <https://github.com/deliveryhero/helm-charts/tree/master/stable/locust>`_.
 
-Note: this Helm chart is a separate project, and not supported by Locust maintainers.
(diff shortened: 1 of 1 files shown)


docs/running-in-docker.rst
```diff
@@ -39,10 +39,21 @@ official Locust docker image as a base image::
...
 
 Running a distributed load test on Kubernetes
-=============================================
+==============================================
 
-The easiest way to run Locust on Kubernetes is to use a `Helm chart <https://helm.sh/>`_.
+HELM
+*****
+
+One of the ways to run Locust on Kubernetes is to use a `Helm chart <https://helm.sh/>`_.
```
AbdelrhmanHamouda, line 47: This rephrasing of the statement is purely an opinion and  I fully understand if Locust want to state that HELM is the easiest way to run distributed tests. If the rephrasing is not okay, please let me know and I will restore the original phrasing. 

Opinion justification: 
HELM is an incredible tool and the provided chart does a lot to make sure that the processes is easy. However, the user is not really "shielded" from Kubernetes and have to be more aware many other moving peaces in order to get things right. That in my opinion makes it so the term "easiest" may cause a bit of misguidance.

AbdelrhmanHamouda:
Hello @cyberw , I attempted to follow the contribution docs as close as i could.  Things look good locally. Please let me know if something is missing. 

Also, I have a proposal to move the "Running in Kubernetes" section to it's own dedicated page. The topic is too big "and irrelevant IMHO" to be under Running in Docker.

cyberw:
Great stuff. Yes, splitting it to a separate page makes sense.

PR #2290, opened 2023-01-18
Add IPv6 support for zmq connections

jcrapuchettes:
This is a small fix to allow distributed testing in a IPv6 only network like fly.io.


cyberw:
That didnt seem to work, the tests just hang...

jcrapuchettes:
OK, I'll look into this.

jcrapuchettes:
I made a small change to check if the platform running locust supports IPv6 before enabling IPv6 for the zmq socket.

cyberw:
Would this work in a mixed environment? For example if the master has both ipv4 and ipv6 addresses but the workers only have ipv4 (or vice versa)?

jcrapuchettes:
I'm not really sure where to go since that if statement didn't work. Thank you for working with me.

PR #2293, opened 2023-01-25
Enable download of (aggregated) history from web

mielverkerken:
## Description
Enable downloading of CSV with aggregated history through the web ui.
Currently the web ui allows for downloading following data:
- request statistics csv
- failures csv
- exceptions csv
- report html
- full history csv when enabled in configuration `--csv-full-history`

## Changes
Allow the download of the aggregated history csv in the web ui when `--csv` is enabled but not `--csv-full-history`.


cyberw:
Looks reasonable. Kindly take a look at the failed format check. I dont really use this feature myself so I dont have much input. Maybe @heyman ?

PR #2297, opened 2023-02-16
fix: docs describing running without web UI had improper flag -f. Cor…

adriangonciarz:
Docs describing running in headless mode had dangling `-f` flag which caused error
Actually there are 4 CSV files generated in headless mode.


cyberw:
thx

PR #2305, opened 2023-02-24
Allow to change worker id pattern

zifter:
At that moment, worker ID equals to pattern:
**HOSTNAME_UUID**
for example:
**locust-worker-5f754b69b5-vsjs8_9fe2c4e93f654fdbb24c02b15259716c**

When we use locust in Kubernetes, each worker runs in a separate pod with a unique hostname (it's guaranteed by k8s).
If a worker suddenly restarts inside a pod, the new worker ID will be generated, because there is a new uuid created.
It's not convenient - master becomes a bit crazy and sees the worker with the old ID as missing and with the new ID as running simultaneously. Why is it a new worker, if it's actually the same?
In Kubernetes, it's better to use hostname only for worker ID, so, these changes make it possible to generate any convenient worker id based on pattern and autofill variables.

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -12,6 +12,7 @@ version = locust.__version__
 
 
 DEFAULT_CONFIG_FILES = ["~/.locust.conf", "locust.conf"]
+WORKER_ID_PATTERN = "%(hostname)s_%(uuid)s"
 
 
 class LocustArgumentParser(configargparse.ArgumentParser):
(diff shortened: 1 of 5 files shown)


cyberw:
Worker ID is meant to be a a unique identifier for the worker process. Reusing it for a new process might be helpful for your particular case, but it could also create very subtle bugs, and I dont think this is something we want to support.

zifter:
@cyberw definitely, Worker ID must be unique. In Kubernetes, it's enough to have only hostname, because it's unique name of pod.
I understand, that it can lead to some stumble bugs and it's easy to do not support case with process restarting, but maybe it's better to fix them, if we face with bugs like that?
I already use this changes in our load testing, sot if any cases I find, I will fix.
Default behavior is not changes at all.

cyberw:
Sorry, still doesnt make sense to me. If a process is restarted in a container it _is_ a new worker, faking that it is not has the potential to open up a can of worms (for example, if locust master sends some command to workers on initial connect, new worker processes started with the same id would never receive that command).

Even if the default behaviour is unchanged, this would require us to support the new behaviour. Perhaps you can maintain this as a fork.

zifter:
@cyberw ok, got it. Thanks for your replay!
I will look at all inconvenients with default worker id.

Issue #2307, opened 2023-02-26
"Download as PNG" text gets cut off


allison-strandberg:
### Describe the bug
When hovering over the download icon on a chart, text appears reading "Download as PNG". This text is cut off by the right edge of the chart.

### Expected behavior
The full "Download as PNG" text should be visible.

### Actual behavior
This text is cut off by the right edge of the chart.
<img width="1440" alt="Screen Shot 2023-02-26 at 2 44 36 PM" src="https://user-images.githubusercontent.com/55810428/221436500-b49437f0-741e-4ed6-8613-08d30c8a81c3.png">

### Steps to reproduce
After running a test in the Web UI, go to the Charts tab.
Hover over a chart's download icon in the top right.

### Environment

- OS: Mac OS Monterey Version 12.5
- Python version: 3.9.13
- Locust version: (please dont file issues for anything but the most recent release or prerelease builds) 2.14.1
- Locust command line that you ran: `locust`
- Locust file contents (anonymized if necessary):
```python
from locust import HttpUser, task

class PostmanEchoUser(HttpUser):
    host = "https://postman-echo.com"

    @task
    def hello_world(self):
        self.client.get("/get")
```

PR #2308, opened 2023-02-26
Prevent Download as PNG text from getting cut off

allison-strandberg:
Resolves https://github.com/locustio/locust/issues/2307.

This prevents the "Download as PNG" text, which displays when hovering over the download icon, from getting cut off.
This is achieved by placing the label to the left of the icon instead of below the icon.

Before:
Screen Shot 2023-02-26 at 2 44 36 PM

After:
Screen Shot 2023-02-26 at 3 17 56 PM


cyberw:
Nice! Thanks.

Issue #2311, opened 2023-03-01
Ability to specify percentiles you need for response time chart


A1BOCO:
### Is your feature request related to a problem? Please describe.

The software I'm currently testing by locust requires 98% < 1.0 sec and 99.5 % < 2.0 sec response time performance. The performance requirement will vary based on software or project etc.. It would be better that percentiles on performance char is configurable not only median and 95% percentile. 
Benefit of having this function on chart page is,  we can see necessary percentile info in response time chart dynamically every minutes.   

### Describe the solution you'd like

Ideal solution would be to add new command line argument like attachment. If the parameters are not specified, default value median and 95% are used 
# locust -f locust.py --percentile1 0.8 --percentile2 0.99

![image](https://user-images.githubusercontent.com/79035645/222012451-ed5d14d3-1f4d-492a-b9dc-ac22ca5cac25.png)

then displays like this
1. Performance chart
![locust2](https://user-images.githubusercontent.com/79035645/222014123-06fd8ab5-0f5d-4ed6-9e56-6e48e6ccfc44.png)

2. report page
![image](https://user-images.githubusercontent.com/79035645/222013106-dcb004e6-3bca-4922-ae9a-84f32778dad9.png)

### Describe alternatives you've considered

applying to stats page is alternatives, but it's overall percentiles not time series 

![locust4](https://user-images.githubusercontent.com/79035645/222014520-27ef74fe-d056-42ab-b7aa-c83c4ccd2523.png)

### Additional context

I have updated locust for the ability on my local repo folked from orignal.

cyberw:
Sounds reasonably useful. I’d prefer `—percentiles 0.8,0.99` as the parameter style though. PR welcome.

cyberw:
Note that for csv reports you can overwrite locust.stats.PERCENTILES_TO_REPORT https://docs.locust.io/en/stable/configuration.html#customization-of-statistics-settings

A1BOCO:
Thank you for the reply. I'll take this style "--percentiles 0.8,0.99" and will send PR.

PR #2313, opened 2023-03-03
Add PERCENTILES_TO_CHART param in stats.py to make the Response Time Chart configurable

A1BOCO:
Added --percentiles command line argument which is for response time chart in chart page and download report page.
Default value is 95% and 50%. 
example . # locust -f locust_file.py --percentiles 0.99,0.8

This is from issue [#2311](https://github.com/locustio/locust/issues/2311)

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -4,6 +4,7 @@ import pathlib
 import datetime
 from itertools import chain
 from .stats import sort_stats
+from . import stats as stats_module
 from .user.inspectuser import get_ratio
 from html import escape
 from json import dumps
@@ -75,7 +76,6 @@ def get_html_report(environment, show_download_link=True):
         "per_class": get_ratio(environment.user_classes, user_spawned, False),
         "total": get_ratio(environment.user_classes, user_spawned, True),
     }
-    from .stats import PERCENTILES_TO_CHART
 
     res = render_template(
         "report.html",
@@ -95,8 +95,8 @@ def get_html_report(environment, show_download_link=True):
         show_download_link=show_download_link,
         locustfile=environment.locustfile,
         tasks=dumps(task_data),
-        percentile1=PERCENTILES_TO_CHART[0],
-        percentile2=PERCENTILES_TO_CHART[1],
+        percentile1=stats_module.PERCENTILES_TO_CHART[0],
+        percentile2=stats_module.PERCENTILES_TO_CHART[1],
     )
 
     return res
--- a/locust/html.py
+++ b/locust/html.py
@@ -8,6 +8,7 @@ from .user.inspectuser import get_ratio
 from html import escape
 from json import dumps
 from .runners import MasterRunner, STATE_STOPPED, STATE_STOPPING
+from .stats import PERCENTILES_TO_CHART
 
 
 def render_template(file, **kwargs):
@@ -94,12 +95,8 @@ def get_html_report(environment, show_download_link=True):
         show_download_link=show_download_link,
         locustfile=environment.locustfile,
         tasks=dumps(task_data),
-        percentile1=options.percentiles.split(",")[0]
-        if options and options.percentiles and len(options.percentiles.split(",")) >= 1
-        else 0.95,
-        percentile2=options.percentiles.split(",")[1]
-        if options and options.percentiles and len(options.percentiles.split(",")) >= 2
-        else 0.5,
+        percentile1=PERCENTILES_TO_CHART[0],
+        percentile2=PERCENTILES_TO_CHART[1],
     )
 
     return res
--- a/locust/html.py
+++ b/locust/html.py
@@ -94,8 +94,8 @@ def get_html_report(environment, show_download_link=True):
         show_download_link=show_download_link,
         locustfile=environment.locustfile,
         tasks=dumps(task_data),
-        percentile1=options.percentile1 if options.percentile1 else 0.95,
-        percentile2=options.percentile2 if options.percentile2 else 0.5,
+        percentile1=options.percentiles.split(',')[0] if options and options.percentiles and len(options.percentiles.split(',')) >= 1 else 0.95,
+        percentile2=options.percentiles.split(',')[1] if options and options.percentiles and len(options.percentiles.split(',')) >= 2 else 0.5,
     )
 
     return res
--- a/locust/html.py
+++ b/locust/html.py
@@ -19,7 +19,7 @@ def render_template(file, **kwargs):
 
 def get_html_report(environment, show_download_link=True):
     stats = environment.runner.stats
-
+    options = environment.parsed_options
     start_ts = stats.start_time
     start_time = datetime.datetime.utcfromtimestamp(start_ts).strftime("%Y-%m-%d %H:%M:%S")
 
@@ -94,6 +94,8 @@ def get_html_report(environment, show_download_link=True):
         show_download_link=show_download_link,
         locustfile=environment.locustfile,
         tasks=dumps(task_data),
+        percentile1=options.percentile1 if options.percentile1 else 0.95,
+        percentile2=options.percentile2 if options.percentile2 else 0.5,
     )
 
     return res
--- a/locust/html.py
+++ b/locust/html.py
@@ -20,7 +20,7 @@ def render_template(file, **kwargs):
 
 def get_html_report(environment, show_download_link=True):
     stats = environment.runner.stats
-    options = environment.parsed_options
+
     start_ts = stats.start_time
     start_time = datetime.datetime.utcfromtimestamp(start_ts).strftime("%Y-%m-%d %H:%M:%S")
 
--- a/locust/html.py
+++ b/locust/html.py
@@ -94,8 +94,12 @@ def get_html_report(environment, show_download_link=True):
         show_download_link=show_download_link,
         locustfile=environment.locustfile,
         tasks=dumps(task_data),
-        percentile1=options.percentiles.split(',')[0] if options and options.percentiles and len(options.percentiles.split(',')) >= 1 else 0.95,
-        percentile2=options.percentiles.split(',')[1] if options and options.percentiles and len(options.percentiles.split(',')) >= 2 else 0.5,
+        percentile1=options.percentiles.split(",")[0]
+        if options and options.percentiles and len(options.percentiles.split(",")) >= 1
+        else 0.95,
+        percentile2=options.percentiles.split(",")[1]
+        if options and options.percentiles and len(options.percentiles.split(",")) >= 2
+        else 0.5,
     )
 
     return res
--- a/locust/html.py
+++ b/locust/html.py
@@ -8,7 +8,6 @@ from .user.inspectuser import get_ratio
 from html import escape
 from json import dumps
 from .runners import MasterRunner, STATE_STOPPED, STATE_STOPPING
-from .stats import PERCENTILES_TO_CHART
 
 
 def render_template(file, **kwargs):
@@ -76,6 +75,7 @@ def get_html_report(environment, show_download_link=True):
         "per_class": get_ratio(environment.user_classes, user_spawned, False),
         "total": get_ratio(environment.user_classes, user_spawned, True),
     }
+    from .stats import PERCENTILES_TO_CHART
 
     res = render_template(
         "report.html",
--- a/locust/main.py
+++ b/locust/main.py
@@ -110,13 +110,29 @@ def main():
         sys.stderr.write("The --slave/--expect-slaves parameters have been renamed --worker/--expect-workers\n")
         sys.exit(1)
 
-    if options.percentile1 and (float(options.percentile1) < 0 or float(options.percentile1) >1):
-        sys.stderr.write("The --percentile1/--percentile2 parameters need to be between 0 to 1. Eg : 0.95 \n")
-        sys.exit(1)
+    def isfloat(parameter):
+        if not parameter.isdecimal():
+            try:
+                float(parameter)
+                return True
+            except ValueError:
+                return False
+        else:
+            return False
 
-    if options.percentile2 and (float(options.percentile2) < 0 or float(options.percentile2) >1):
-        sys.stderr.write("The --percentile1/--percentile2 parameters need to be between 0 to 1. Eg : 0.95 \n")
-        sys.exit(1)
+    if options.percentiles:
+
+        if any([not isfloat(percentile) for percentile in options.percentiles.split(',')]):
+            sys.stderr.write("The --percentiles parameter need to be float.  Eg 0.95 \n")
+            sys.exit(1)
+
+        if any((0 > float(pecentile) or 1 < float(pecentile)) for pecentile in options.percentiles.split(',')):
+            sys.stderr.write("The --percentiles parameter need to be value between. 0 <= percentile <= 1 Eg 0.95 \n")
+            sys.exit(1)
+
+        if len(options.percentiles.split(',')) >= 3:
+            sys.stderr.write("The --percentiles parameter can be specified up tp 2 parameters by comma separators. Eg 0.95,0.5   \n")
+            sys.exit(1)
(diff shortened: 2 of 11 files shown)


locust/test/test_main.py
```diff
@@ -1136,6 +1136,66 @@ def task1(self):
             self.assertIn("No tasks defined on MyUser", stderr)
             self.assertEqual(1, proc.returncode)
 
+    def test_invalid_percentilesrange_options(self):
```
cyberw, line 1139: Maybe add a positive test too?

locust/web.py
```diff
@@ -332,7 +332,17 @@ def failures_stats_csv() -> Response:
         def request_stats() -> Response:
             stats: List[Dict[str, Any]] = []
             errors: List[StatsErrorDict] = []
-
+            options = self.environment.parsed_options
```
cyberw, line 335: Doesn't `environment` work? (instead of `self.environment`). But anyways, like I mentioned in my PR comment, I think this should be a "hidden" parameter, like with the other stats.py settings (like this one https://github.com/locustio/locust/blob/5cf5da22b6774a537e14b08ed68acb9dfc4a8e98/locust/stats.py#L134). Would that be ok for you?

locust/test/test_main.py
A1BOCO: Sure, I'll add positive test

locust/web.py
A1BOCO: Does "hidden" parameter mean, should the parameter be outside of WebUI class?  or is it ok as instance variable under def __init__

like 
image
web.py

cyberw: Hidden in the sense that it is just a variable on the stats module, set from inside the locustfile, not an argument. As i linked here: https://docs.locust.io/en/stable/configuration.html#customization-of-statistics-settings

A1BOCO: Let me clarify, 

stats.py has default value like this,

image

then if command line argument percentiles have value, the parameter is overwritten like this when it is used.
image

Is it what you expect?

cyberw: Almost :) what I mean is dont have a command line parameter at all, and only expose it on stats.py. There are too many command line options already, and there needs to be some effort in cleaning them up before adding new ones (maybe grouping all stats-related stuff in an ”advanced” block somehow? The --help text is already ridiculously long, and I dont want to make it longer.

A1BOCO: Thank you for the explanation, I got what you intend. I thought command line argument is easier to change config test by test. But anyway, I have removed command line arguments of percentiles and percentiles values for response time chart is from stats.PERCENTILES_TO_CHART param. I have pushed the code, please review it.

locust/html.py
```diff
@@ -19,7 +20,7 @@ def render_template(file, **kwargs):
 
 def get_html_report(environment, show_download_link=True):
     stats = environment.runner.stats
-
+    options = environment.parsed_options
```
cyberw, line 23: This isnt needed any more, right?

locust/main.py
```diff
@@ -100,6 +100,24 @@ def main():
             user_classes[key] = value
             available_user_classes[key] = value
 
+    if len(stats.PERCENTILES_TO_CHART) > 2:
+        sys.stderr.write("stats.PERCENTILES_TO_CHART parameter can be specified up tp 2 parameters \n")
```
cyberw, line 104: Use logging.error instead

locust/stats.py
```diff
@@ -888,8 +890,10 @@ def stats_history(runner: "Runner") -> None:
                 "time": datetime.datetime.now(tz=datetime.timezone.utc).strftime("%H:%M:%S"),
                 "current_rps": stats.total.current_rps or 0,
                 "current_fail_per_sec": stats.total.current_fail_per_sec or 0,
-                "response_time_percentile_95": stats.total.get_current_response_time_percentile(0.95) or 0,
-                "response_time_percentile_50": stats.total.get_current_response_time_percentile(0.5) or 0,
+                "response_time_percentile_95": stats.total.get_current_response_time_percentile(PERCENTILES_TO_CHART[0])
```
cyberw, line 893: Variables needs to be changed. Probably use a list: response_time_percentile[0] instead of response_time_percentile_95.

cyberw: (If that cannot be done, it could be response_time_percentile_first/second or something)

locust/html.py
A1BOCO: Right, It's not needed. I'll delete this.

locust/main.py
A1BOCO: Sure, I'll do like this

    if len(stats.PERCENTILES_TO_CHART) > 2:
        logging.error("stats.PERCENTILES_TO_CHART parameter can be specified up tp 2 parameters \n")
        sys.exit(1)

locust/stats.py
A1BOCO: Sure, will use the variable

locust/main.py
cyberw: 👍

locust/stats.py
cyberw: 👍

cyberw:
Once you're done, rename the PR to something more appropriate, now that we're not using a command line argument

cyberw:
Nice. I think you accidentally commented out the import of pty in test_main.py.

Add it to the documentation at the bottom of docs/configuration.rst, and then we're pretty much done!

A1BOCO:
I have corrected test_main.py and format of stats.py. Also update configuration.rst. Please review them.

locust/html.py
```diff
@@ -75,6 +75,7 @@ def get_html_report(environment, show_download_link=True):
         "per_class": get_ratio(environment.user_classes, user_spawned, False),
         "total": get_ratio(environment.user_classes, user_spawned, True),
     }
+    from .stats import PERCENTILES_TO_CHART
```
cyberw, line 78: group this import with the one on line 6

locust/main.py
```diff
@@ -100,6 +100,26 @@ def main():
             user_classes[key] = value
             available_user_classes[key] = value
 
+    from .stats import PERCENTILES_TO_CHART
```
cyberw, line 103: group this import with the one on line 16

locust/static/locust.js
```diff
@@ -219,8 +219,8 @@ function update_stats_charts(){
         responseTimeChart.chart.setOption({
             xAxis: {data: stats_history["time"]},
             series: [
-                {data: stats_history["response_time_percentile_50"], markLine: createMarkLine()},
-                {data: stats_history["response_time_percentile_95"]},
+                {data: stats_history["response_time_percentile_2"], markLine: createMarkLine()},
+                {data: stats_history["response_time_percentile_1"]},
```
cyberw, line 223: maybe we want to reorder these?

locust/templates/report.html
```diff
@@ -214,8 +216,8 @@ <h2>Final ratio</h2>
             responseTimeChart.chart.setOption({
                 xAxis: {data: stats_history["time"]},
                 series: [
-                    {data: stats_history["response_time_percentile_50"]},
-                    {data: stats_history["response_time_percentile_95"]},
+                    {data: stats_history["response_time_percentile_2"]},
+                    {data: stats_history["response_time_percentile_1"]},
```
cyberw, line 220: reorder (or rename)

locust/test/test_main.py
```diff
@@ -191,6 +192,49 @@ def my_task(self):
...
+            PERCENTILES_TO_CHART[1] = 0.4
+            class TestUser(User):
+                wait_time = constant(3)
+                @task
+                def my_task(self):
+                    print("running my_task()")
+        """
+            )
+        ) as file_path:
+            proc = subprocess.Popen(["locust", "-f", file_path], stdout=PIPE, stderr=PIPE, text=True)
```
cyberw, line 211: Add something to the locustfile that actually makes requests, add `--autostart` to make actually run, and check the exit code or something.

locust/web.py
```diff
@@ -383,16 +384,17 @@ def request_stats() -> Response:
                 truncated_stats += [stats[-1]]
 
             report = {"stats": truncated_stats, "errors": errors[:500]}
+            from .stats import PERCENTILES_TO_CHART
```
cyberw, line 387: group imports

locust/web.py
```diff
@@ -536,6 +538,9 @@ def update_template_args(self):
         if self.environment.available_shape_classes:
             available_shape_classes += sorted(self.environment.available_shape_classes.keys())
 
+        from .stats import PERCENTILES_TO_CHART
```
cyberw, line 541: group imports

locust/web.py
```diff
@@ -536,6 +538,9 @@ def update_template_args(self):
...
+
+        # print(PERCENTILES_TO_CHART)
         self.template_args = {
```
cyberw, line 544: accidentaly comitted comment?

locust/html.py
A1BOCO: Thank you for the comment. importing PERCENTILES_TO_CHART at 6  does not work. What I mean by that,
stats. PERCENTILES_TO_CHART default is [0.95, 0.5] as list in stats.py. When passing new value from locust execution file like stats. PERCENTILES_TO_CHART = [0.8,0.3] as list for instance, when PERCENTILES_TO_CHART was read from html.py file, default [0.95, 0.5] was used not new value [0.8,0.3].  But importing PERCENTILES_TO_CHART at line 78 which is right before PERCENTILES_TO_CHART is used, new value  [0.8,0.3] is used by html.py.  I think this is caused by pass by reference etc.. Actually, passing new value as PERCENTILES_TO_CHART[0] = 0.8 works, even if importing PERCENTILES_TO_CHART at line 6. I would like to suggest to use PERCENTILES_TO_CHART_1 and  PERCENTILES_TO_CHART_2 in stats.py not as list. This should work.I hope this explanation makes sense to you. If you have any suggestion, let me know.

locust/main.py
A1BOCO: Thank you for the comment. This is same situation as I explained html.py

Importing PERCENTILES_TO_CHART at 16 does not work. What I mean by that,
stats. PERCENTILES_TO_CHART default is [0.95, 0.5] as list in stats.py. When passing new value from locust execution file like stats. PERCENTILES_TO_CHART = [0.8,0.3] as list for instance, when PERCENTILES_TO_CHART was read from main.py file, default [0.95, 0.5] was used not new value [0.8,0.3]. But importing PERCENTILES_TO_CHART at line 78 which is right before PERCENTILES_TO_CHART is used, new value [0.8,0.3] is used by main.py. I think this is caused by pass by reference etc.. Actually, passing new value as PERCENTILES_TO_CHART[0] = 0.8 works, even if importing PERCENTILES_TO_CHART at line 6. I would like to suggest to use PERCENTILES_TO_CHART_1 and PERCENTILES_TO_CHART_2 in stats.py not as list. This should work.I hope this explanation makes sense to you. If you have any suggestion, let me know.

locust/static/locust.js
A1BOCO: Sure, will reorder this

locust/templates/report.html
A1BOCO: Sure, will reorder this

locust/test/test_main.py
A1BOCO: I'll update test_main

locust/web.py
A1BOCO: This is same as I explained for html.py

locust/web.py
A1BOCO: That's my mistake, I'll remove it

locust/main.py
cyberw: import the whole stats module and access it as `stats.PERCENTILES_TO_CHART` and it should work!

cyberw: Like here: https://github.com/locustio/locust/blob/master/locust/main.py#L245

A1BOCO: Yes, that worked! thank you.

locust/static/locust.js
```diff
@@ -235,7 +235,7 @@ function update_stats_charts(){
 
 // init charts
 var rpsChart = new LocustLineChart($(".charts-container"), "Total Requests per Second", ["RPS", "Failures/s"], "reqs/s", ['#00ca5a', '#ff6d6d']);
-var responseTimeChart = new LocustLineChart($(".charts-container"), "Response Times (ms)", ["Median Response Time", "95% percentile"], "ms");
+var responseTimeChart = new LocustLineChart($(".charts-container"), "Response Times (ms)", [(percentile1 * 100).toString() + "% percentile" , (percentile2 * 100).toString() + "% percentile"], "ms");
```
cyberw, line 238: Can you change this to `..th percentile` instead of `..% percentile`?

A1BOCO: Updated locust.js. also I have updated report.html

image

locust/test/test_main.py
```diff
@@ -191,6 +192,58 @@ def my_task(self):
...
+            proc = subprocess.Popen(
+                ["locust", "-f", file_path, "--web-port", str(port), "--autostart"], stdout=PIPE, stderr=PIPE, text=True
+            )
+            gevent.sleep(1)
+            proc.send_signal(signal.SIGTERM)
+            stdout, stderr = proc.communicate()
+            self.assertIn("Starting web interface at", stderr)
+            try:
+                response = requests.get(f"http://0.0.0.0:{port}/")
+                self.assertEqual(200, response.status_code)
```
cyberw, line 221: This doesnt really test anything? Any error would be caught..

locust/test/test_main.py
```diff
@@ -191,6 +192,58 @@ def my_task(self):
...
+                wait_time = constant(3)
+                @task
+                def my_task(self):
+                    print("running my_task()")
+        """
+            )
+        ) as file_path:
+            proc = subprocess.Popen(["locust", "-f", file_path, "--autostart"], stdout=PIPE, stderr=PIPE, text=True)
+            gevent.sleep(1)
+            proc.send_signal(signal.SIGTERM)
```
cyberw, line 242: No need to send signal, I think you can remove it.

locust/test/test_main.py
A1BOCO: I'm expecting 200 response, I think I can remove try/exception

locust/test/test_main.py
A1BOCO: Will check

locust/test/test_main.py
A1BOCO: I made mistake, I'll check this test again

A1BOCO: I wanted to check normal behavior in this test,  I tried requests.get(f"http://0.0.0.0:{port}/") and self.assertEqual(0, proc.returncode) to check, but none of them worked by running unittest. How can I check normal case? 
Anyway, test_webserver line 172 is actually testing normal case. Can I remove this test?

cyberw: proc.send_signal(signal.SIGTERM) will kill locust. Move that statement to the end of the test.

A1BOCO: That worked, I have pushed update. Thank you for helping a lot.

locust/stats.py
```diff
@@ -133,6 +133,8 @@ def resize_handler(signum: int, frame: Optional[FrameType]):
 
 PERCENTILES_TO_REPORT = [0.50, 0.66, 0.75, 0.80, 0.90, 0.95, 0.98, 0.99, 0.999, 0.9999, 1.0]
 
+PERCENTILES_TO_CHART = [0.95, 0.50]
```
cyberw, line 136: Maybe put lowest percentile first?

A1BOCO: Done, now PERCENTILES_TO_CHART = [0.50, 0.95]

cyberw:
Nice!

PR #2317, opened 2023-03-14
Update helper text

rafaelhdr:
Just a tiny helper text update that I got confused when coding. I hope it is all good, but let me know if you think I should change anything :)


cyberw:
thx

PR #2324, opened 2023-03-20
Fix: show host url

zifter:


cyberw:
Can you show a before-after screenshot? Sorry, I’m lazy/busy :(

zifter:
Actually, nothing changed, but now host is not empty after page reloading:)

cyberw:
Ah!

PR #2326, opened 2023-03-22
Typos fixes in comments

JoaoGFarias:


cyberw:
First fix is definitely an improvement, but the second change makes no sense :)
`
# if the response isn't even json, but RestUser will already have been marked it as a failure, so we don't have to do it again`

PR #2333, opened 2023-04-12
Update writing-a-locustfile.rst

sumanchapai:
Clarify


cyberw:
I dont think this needs further clarification. A class inheriting a subclass of a class (User in this case) is commonly said to inherit the base class, it is just an indirect inheritance.

PR #2336, opened 2023-04-13
Put dependencies/package information in pyproject.toml instead of setup.cfg (PEP621)

JE-Chen:
Add pep621_pyproject.toml

* This TOML file complies with PEP517 & PEP518 & PEP621

This link is for the PYPI package generated by this TOML file.

* https://pypi.org/project/locust-je-test-pep621/2.15.2.dev8/

I don't want to overwrite old pyproject.toml so i call this file pep621_pyproject.toml

PEP621
* https://peps.python.org/pep-0621/


cyberw:
Nice. But go ahead and add it to pyproject.toml, otherwise it will not be tested before merging etc. We definitely dont want to have both at the same time. Is this compatible with Python 3.7 btw?

You'll need to merge this with latest master to get the build working (because of this https://github.com/locustio/locust/pull/2337)

JE-Chen:
> Is this compatible with Python 3.7 btw?
Sure, but we need upgrade build & setuptools (setuptools>=61) first

cyberw:
> PEP621 allows pyproject.toml to generate the package on its own, but CI/CD requires setup.cfg so I didn't delete it.

Hm. If we'd have to maintain duplicate dependencies in setup.cfg I dont think this is worth it.

JE-Chen:
Look like we don't need  setup.cfg & setup.py anymore

cyberw:
That just ensures pip cache uniqueness, an optimization for faster (repeated) builds. I think you can replace 'setup.cfg' with 'pyproject.toml'

cyberw:
Makefile needs to be adjusted too, as this doesnt work anymore:

```
build:
	rm -f dist/* && python3 setup.py sdist bdist_wheel
```

cyberw:
While you are at it, you can nuke `examples/vagrant/vagrant.sh`. 

It contains a reference to setup.py which will stop working now. And the file hasnt been updated in 11 (!) years :)

cyberw:
Hmm...
image

JE-Chen:
We need "pip install --upgrade build" first

cyberw:
> We need "pip install --upgrade build" first

Yea, I was in a virtualenv and forgot to install it. Oops. Do you wanna remove vagrant.sh or should I?

cyberw:
Good stuff, thanks!

PR #2346, opened 2023-05-16
Duplicated titles

Daemo00:
RIght now in https://docs.locust.io/en/latest/ the edited titles are shown duplicated: 
image


cyberw:
thx

PR #2350, opened 2023-05-22
Fixed the users dispatching when it got infinite loop (issue-2345)

EzR1d3r:
I currently do not have not a local configured unittests, so this is WIP MR for checks.


cyberw:
Nice! Whats a good title for the PR? Something like ”Fix problem with fixed_count in combination with LoadShapes”?

EzR1d3r:
> Nice! Whats a good title for the PR? Something like ”Fix problem with fixed_count in combination with LoadShapes”?

Done. =)

cyberw:
👍 thx!

Issue #1363, opened 2020-05-04
Ability to exclude response time of failed requests from time based metrics


vstepanov-lohika-tix:
### Is your feature request related to a problem? Please describe.
Currently Locust stats for time based metrics (ART, median, percentiles) include response times from all requests, including failed.
From my experience, In most cases it much affects metrics for successful requests and can make it completely incorrect.

**Examples:**
1) 100 successful requests with ART 1000 ms + 20 failed requests with 503 error (Service Temporary Unavailable) and fast response time 5 ms
In this case ART in report will be "improved" by failed requests (~830 ms vs actual 1000), and the more failed requests in test, the better result will be in results.
2) 100 successful requests with ART 1000 ms, and 5 failed requests with 504 Server error in 60 seconds (because of proxy gateway issues)
In this case, successful results will be much worsened by failures - (3.8 sec vs actual 1 sec)

### Describe the solution you'd like
In order to keep old behavior by default and not break compatibility, there might be additional command line argument, like _--exclude-failed-response-time_ or _--exclude-failures_  
With using this parameter, response times from failed requests must be excluded for calculation of time based metrics.
The number of failures and failures rate shouldn't be affected.

_StatsEntry_ class should be extended with response_times_success & response_times_failures fields, that should be used with calculation functions in case of using exclude failures parameter.

Functions/properties  to be refactored (at least):
_calculate_response_time_percentile_
_avg_response_time_
_median_response_time_
_log_response_time_

### Describe alternatives you've considered
Any additions and suggestions are welcomed.

heyman:
If we choose to just add a flag for excluding failed requests from response times, I believe it can be achieved by modifying this function:

https://github.com/locustio/locust/blob/0f6f2170331a10f6e0427e947bf91aab6a797b91/locust/runners.py#L61-L63

to just set `response_time=None` in the call to `log_request()` (if the option has been set in the `Environment` instance).

Unless there's some issue with that that I'm forgetting.

vstepanov-lohika-tix:
If no need to keep failed response times in separate collection and to use it somehow, it must be much easier, as suggested @heyman

vidyli:
why this is closed? 

![image](https://user-images.githubusercontent.com/75243787/110775435-de881700-8299-11eb-8f7e-ce89df7ecda0.png)

locust report set minimum response time to the time of the 503 error response , how can we avoid this?

Prometheousl:
Why is this closed? This would be very useful.

[...]

PR #2353, opened 2023-05-29
Issue #1363: Adding option for exlusion time of failed requests from time metrics

vs-odessa:
This PR revives the request from https://github.com/locustio/locust/issues/1363 (_If possible, please reopen it_)

It allows adding the parameter  `--exсlude-failures` to locust runner to ignore response times of failed requests from time-based metrics (min,max, avg, percentiles etc.)

This helps to exclude abnormal values (too big or too small) from time statistics produced by failed requests. (For example, when getting gateway timeout errors equal to 60 seconds or more)
At the same time, it keeps stats for the total requests number and failures number.

**Example:** 
9 passed requests with 1 second response time + 1 failed request with 5 seconds response time

Without option:
`Avg time = (9*1 + 1*5) / 10 = 1.4 sec, Max time = 5 sec`

With option `--exclude-failures`:
`Avg time = (9*1) / 9 = 1 sec, Max time = 1 sec`

Total/Failed = 10/1 in both cases

**Update:** In a second commit added a feature for setting or overriding global setting for exclusion failures by passing keyword parameter to specified request(s).

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -488,6 +488,14 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
         env_var="LOCUST_ONLY_SUMMARY",
     )
     stats_group.add_argument(
+        "--exclude-failures",
+        action="store_true",
+        default=False,
+        dest="exclude_failures_time",
+        help="Excludes response times of failed requests from time-based metrics (but not from total/failed requests count)",
+        env_var="EXCLUDE_FAILURES",
+    )
+    stats_group.add_argument(
         "--reset-stats",
         action="store_true",
         help="Reset statistics once spawning has been completed. Should be set on both master and workers when running in distributed mode",
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -492,7 +492,8 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
         action="store_true",
         default=False,
         dest="exclude_failures_time",
-        help="Excludes response times of failed requests from time-based metrics (but not from total/failed requests count)",
+        help="Excludes response times of failed requests from time-based metrics (but not from total/failed requests count)."
+             "This option can be set or overridden for specified requests by passing 'exclude_failures=True/False' keyword parameter to the request",
         env_var="EXCLUDE_FAILURES",
     )
     stats_group.add_argument(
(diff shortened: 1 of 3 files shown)


vs-odessa:
@heyman, @cyberw - please review

vs-odessa:
I'm not sure that it can be solved by adding a custom handler - we can add our own actions, but it will not discard a core handler that writes all requests to stats. Correct me if I am wrong, but looks like a custom class should be created from Runner with an overriding on_request method, and this will require a significant refactoring of existing projects. 

Let me add a few words about the motivation for this feature with some examples.

In my experience - I have at least a third project where this feature would be extremely useful, and I saw similar demands from other engineers.
It is true, that most of the tools have such behaviour as Locust by default, but I guess all of them came from times when synchronous HTTP requests were >99% of load tools' purpose.
In modern event-based systems, there is very often a situation when you need to check asynchronous transactions.
In this case, we are interested not only in a single request time, but need also to check a status of a background service task produced by the initial request - until the service task is finished.
Usually, such flow is wrapped as "locust request" to get its duration and other metrics. This flow includes retries of some requests with a timeout, and if something went wrong we get failed transaction with a time equal to timeout.
This significantly exceeds the time for successful operations and distorts a picture of the positive system performance.
For the same reason, you can't rely on minimum/maximum value if any failures occurred in requests.
And conversely, various client & server errors often may give unusually fast timings. In this case, we get a curious situation: the greater number of failures - the better performance metrics appear.

Summing up, I believe it's nice to have this option from the box, and it will be often used. It shouldn't break any old functionality and adds a very slight overhead in the handling of failed requests only.

cyberw:
I get that you have a use case. Can it be worked around? If not then the PR may be justified, otherwise I kinda disagree.

vs-odessa:
I've analyzed and played for a while with a locust code trying to implement a workaround, and in my view, it looks unlikely feasible :(
Environment class and Runner class, that manages requests stats - both are core features and they are not managed when running Locust in a standalone mode.
In theory, perhaps there's some workaround way for "running as a library" mode, but it also looks too complicated and hard maintainable.
If I overlooked something and you see an idea for a solution, please point me in the right direction for it.

vs-odessa:
Possible workaround using response validation:
`        with self.client.request(method, url, name=name, catch_response=False) as response:
            if 'should-fail' in url:
                response.request_meta['response_time'] = None`

PR #2355, opened 2023-06-02
Allow selecting user classes using LOCUST_USER_CLASSES env var

ibarbech:
This pull request introduces an improvement to Locust, enabling the specification of `user_classes` through the `LOCUST_USER_CLASSES` environment variable. Currently, `user_classes` can only be defined via command-line arguments, limiting flexibility. With this enhancement, users can conveniently configure a list of user classes separated by spaces, making it easier to simulate specific user behaviors when running Locust in a Docker container.

**Details:**
In the existing implementation of Locust, when executing it within a Docker container, `user_classes` must be specified using command-line arguments. This approach can be cumbersome, requiring modifications to the command every time different user classes need to be tested. This pull request addresses this limitation by introducing support for user_classes through the `LOCUST_USER_CLASSES` environment variable.

To utilize this new feature, users can now set the `LOCUST_USER_CLASSES` environment variable with a list of user class names separated by spaces. For example, if the desired user classes are "UserClass1", "UserClass2", and "UserClass3", the environment variable would be set as follows: `LOCUST_USER_CLASSES="UserClass1 UserClass2 UserClass3"`.

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -583,7 +583,8 @@ Only the LOCUSTFILE (-f option) needs to be specified when starting a Worker, si
         "user_classes",
         nargs="*",
         metavar="UserClass",
-        help="Optionally specify which User classes that should be used (available User classes can be listed with -l or --list)",
+        help="Optionally specify which User classes that should be used (available User classes can be listed with -l or --list). LOCUST_USER_CLASSES environment variable can also be used to specify User classes",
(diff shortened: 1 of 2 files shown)


cyberw:
Nice! Can you add it to the help text? And maybe a test case?

ibarbech:
Updated!

cyberw:
thx!

PR #2365, opened 2023-06-21
Fix passing host to user class when debugging

LKajan:
HttpUser needs the host parameter but if not set directly it was not passed to the user class in `run_single_user` method like it is in runners.

### DIFF
--- a/locust/debug.py
+++ b/locust/debug.py
@@ -22,7 +22,7 @@ class PrintListener:
 
     def __init__(
         self,
-        env: locust.env.Environment,
+        env: Environment,
         include_length=False,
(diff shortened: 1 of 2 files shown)


LKajan:
I changed the statement order a bit to avoid the mypy error. I hope this is ok too.
Added the locustfile also to the Environment to better match the actual runner situation.

cyberw:
The build is either really slow, or the test is sometimes hanging?

LKajan:
Yep, somehow the actions are stuck. All the tests seems to be passed but actions won't finish.

cyberw:
Thx!

Issue #2371, opened 2023-07-15
Dropdown menu in web UI for custom arguments


ktrnka:
### Is your feature request related to a problem? Please describe.

We have a number of common configuration options in our use of Locust. Some of them have a short list of valid options and it's a bit of a hassle to go over to our repo documentation and/or code to figure out the valid options.

### Describe the solution you'd like

I'd like to specify argparse choices when configuring a custom argument and have that affect the web UI. Example:
```
@events.init_command_line_parser.add_listener
def _(parser):
    parser.add_argument("--my-argument", type=str, env_var="LOCUST_MY_ARGUMENT", default="dev", help="It's working", choices=["dev", "staging", "prod")
```

Currently, this will validate the option values on the command line and print the valid options on error. It just doesn't affect the web UI.

### Describe alternatives you've considered

Since Locust uses argparse/configargparse for this, it seems best to leverage their implementation. I didn't really consider alternatives that didn't match argparse.

### Additional context

If you'd like I'm happy to submit a PR for this.

cyberw:
Makes sense! PR welcome!

cyberw:
Fixed

PR #2372, opened 2023-07-17
Web UI dropdown for custom args with choices

ktrnka:
Addresses #2371 

This adds web UI support for a dropdown list in custom arguments:
Screen Shot 2023-07-16 at 7 34 44 PM

Code:
```
parser.add_argument("--env", choices=["dev", "staging", "prod"], default="dev")
```

argparse already checks those choices on command-line args, this PR just makes the web UI behave similarly.

The CSS width is different for select (340) vs input (set to 328, actual 340) to get them to be the same actual width. I'm not exactly sure why but it looks like it has to do with differences in how padding is handled between the two tags.

# Testing
I ran locally with `scripts/run-local-web.sh` on OS X in a virtual env with Python 3.11.3

I wasn't able to get the tox tests running locally (I tried on master before making changes but it failed). Sorry about that! Hopefully the github actions aren't too burdensome.

### DIFF
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -194,7 +194,7 @@ On windows:
 Custom arguments
 ----------------
 
-You can add your own command line arguments to Locust, using the :py:attr:`init_command_line_parser <locust.event.Events.init_command_line_parser>` Event. Custom arguments are also presented and editable in the web UI.
+You can add your own command line arguments to Locust, using the :py:attr:`init_command_line_parser <locust.event.Events.init_command_line_parser>` Event. Custom arguments are also presented and editable in the web UI. If `choices` are specified for the argument, they will be presented as a dropdown in the web UI.
 
 .. literalinclude:: ../examples/add_command_line_argument.py
     :language: python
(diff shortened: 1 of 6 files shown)


cyberw:
Nice! Can you add a note about this feature in the documentation (.rst files), as well as a test case (add this type of extra arg to one of the existing gui args tests, I hope such a test exists :)

ktrnka:
> Nice! Can you add a note about this feature in the documentation (.rst files), as well as a test case (add this type of extra arg to one of the existing gui args tests. I hope such a test exists :)

Makes sense, yeah I'll take a shot at that when I get home later.

I'll see what I can do with the GUI tests... there's an existing custom args test but it doesn't test the GUI itself, it tests that parse_args works as expected and that Flask can receive those values. There are a couple others though that test the HTML results and I'll see if I can combine the custom arg test with one of those.

ktrnka:
Ok I think this works! I was able to run the test locally and confirm that it failed while in development then passed when done.

I chose regex to test the generated html because it's less brittle than assertIn. I considered doing an xpath-based test which would be more robust, but would add a single-use dependency to the test suite.

cyberw:
👍👍

Issue #2381, opened 2023-08-12
Ability to exclude response time of failed requests from time based metrics #1363


nmondal:
### Is your feature request related to a problem? Please describe.

Currently Locust stats for time based metrics (ART, median, percentiles) include response times from all requests, including failed.
From my experience, In most cases it much affects metrics for successful requests and can make it completely incorrect.

Examples:

100 successful requests with ART 1000 ms + 20 failed requests with 503 error (Service Temporary Unavailable) and fast response time 5 ms
In this case ART in report will be "improved" by failed requests (~830 ms vs actual 1000), and the more failed requests in test, the better result will be in results.
100 successful requests with ART 1000 ms, and 5 failed requests with 504 Server error in 60 seconds (because of proxy gateway issues)
In this case, successful results will be much worsened by failures - (3.8 sec vs actual 1 sec)

### Describe the solution you'd like

In order to keep old behavior by default and not break compatibility, there might be additional command line argument, like --exclude-failed-response-time or --exclude-failures
With using this parameter, response times from failed requests must be excluded for calculation of time based metrics.
The number of failures and failures rate shouldn't be affected.

StatsEntry class should be extended with response_times_success & response_times_failures fields, that should be used with calculation functions in case of using exclude failures parameter.

Functions/properties to be refactored (at least):
calculate_response_time_percentile
avg_response_time
median_response_time
log_response_time

### Describe alternatives you've considered

Any additions and suggestions are welcome!.

### Additional context

https://github.com/locustio/locust/issues/1363
The original issue was reported in 2021. I am doing perf testing and tuning from 2010+, and I concur.
It was closed because of lack of volunteers. We would fix it.
As  a community we must support this, and I am personally trying to fix it, with a PR to be updated soon.

@vstepanov-lohika-tix

nmondal:
Added the PR:
https://github.com/locustio/locust/pull/2382

nmondal:
Hello @cyberw can you please help us with this?

vs-odessa:
@nmondal, there was an attempt at the implementation:
https://github.com/locustio/locust/pull/2353/files
After discussions, it was rejected due to low demand for the feature.

oumsofiane1:
+1

cyberw:
https://github.com/locustio/locust/pull/2382 is still open but needs work.

amard33p:
[...]

PR #2382, opened 2023-08-12
Feature : --exclude-failed-response-time.

nmondal:
Issue / Feature:
https://github.com/locustio/locust/issues/2381

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -71,8 +71,6 @@ class Environment:
         """If set, only tasks that are tagged by tags in this list will be executed. Leave this as None to use the one from parsed_options"""
         self.exclude_tags = exclude_tags
         """If set, only tasks that aren't tagged by tags in this list will be executed. Leave this as None to use the one from parsed_options"""
-        self.stats = RequestStats()
-        """Reference to RequestStats instance"""
         self.host = host
         """Base URL of the target system"""
         self.reset_stats = reset_stats
@@ -94,6 +92,12 @@ class Environment:
         """
         self.parsed_options = parsed_options
         """Reference to the parsed command line options (used to pre-populate fields in Web UI). When using Locust as a library, this should either be `None` or an object created by `argument_parser.parse_args()`"""
+        self.exclude_failed_response_time = False
+        if self.parsed_options.exclude_failed_response_time:
+            self.exclude_failed_response_time = True
+        self.stats = RequestStats(exclude_failed_response_time=self.exclude_failed_response_time)
+        """Reference to RequestStats instance"""
+
         self.available_user_classes = available_user_classes
         """List of the available User Classes to pick from in the UserClass Picker"""
         self.available_shape_classes = available_shape_classes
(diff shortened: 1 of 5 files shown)


nmondal:
Hello @cyberw  can you please help us with this?

cyberw:
sorry, on holiday

cyberw:
I'm back :)

I think this could be ok. But you'd need to add a test case that verifies the feature. And remove the shortcut name (`-x`), as it could conflict with some other parameter that might be added in the future. I have a few questions as well...

locust/env.py
```diff
@@ -94,6 +92,12 @@ def __init__(
         """
         self.parsed_options = parsed_options
         """Reference to the parsed command line options (used to pre-populate fields in Web UI). When using Locust as a library, this should either be `None` or an object created by `argument_parser.parse_args()`"""
+        self.exclude_failed_response_time = False
+        if self.parsed_options and self.parsed_options.exclude_failed_response_time:
+            self.exclude_failed_response_time = True
+        self.stats = RequestStats(exclude_failed_response_time=self.exclude_failed_response_time)
+        """Reference to RequestStats instance"""
```
cyberw, line 99: I dont understand this comment.

cyberw:
Hmm.. Iooking more closely at the code it is just not good enough. Too many oddities, and I'd need to spend too much time trying to wrap my head around it. I'm not sure if the old PR worked (https://github.com/locustio/locust/pull/2353), but it was a lot cleaner.

locust/env.py
nmondal: I did not write it, this is probably a copy paste issue for restructuring.

nmondal:
> Hmm.. Iooking more closely at the code it is just not good enough. Too many oddities in the code, and I'd need to spend too much time trying to wrap my head around it. I'm not sure if the old PR worked (#2353), but it was a lot cleaner.

It won't work. 
Given stats has no clue what to exclude, it won't even know.
It would only work for the case of the option is set to true, and even then it would have all metrics computed wrong.
For example avg, percentile all of the computation would be broken as well as total.
😢 

My first change was exactly the same, and I realized it won't work.

locust/env.py
```diff
@@ -148,7 +152,13 @@ def create_worker_runner(self, master_host: str, master_port: int) -> WorkerRunn
         """
         # Create a new RequestStats with use_response_times_cache set to False to save some memory
         # and CPU cycles, since the response_times_cache is not needed for Worker nodes
-        self.stats = RequestStats(use_response_times_cache=False)
```
amard33p, line 151: Again, we can simplify it to:
```
self.stats = RequestStats(
    use_response_times_cache=False,
    exclude_failed_response_time=getattr(self.parsed_options, "exclude_failed_response_time", False),
)

Alex-TG001:
Hello, I'm trying to make our final metrics based on what we get from filtering out failed requests. If I changed the logic of the on_requests function like this, would it works?
These codes are in /locust/runners.py
image

PR #2393, opened 2023-09-05
feat(shape): support user base custom load shape

noirbizarre:
This PR improves the abstract custom base shape detection by using the same pattern than the base `User` class: abstract base classes should define the `abstract`attribute to `True`.

This change is entirely backward compatible and tested.

This PR fixes the following case:
- define a reusable `LoadTestShape` in your common modules (let's say `lib.BaseShape`)
- create a locust file, create a child shape (cf. below)
- run the bench

This will result in:
- without this change, the imported base class is picked as shape, the child class is ignored.
- with this change, the imported abstract base is ignored and the child class is selected as shape

`lib.py`
```python

class BaseShape(LoadTestShape):
	abstract = True

	some_class_attr = None # to be provided by child implementation

	def tick(self):
        # Do something with `self.__class__.some_class_attr`
```
`locusfile`
```python
from lib import BaseShape

class MyShape(BaseShape):
    some_class_attr = ["my test data"]
```

Currently, `BaseShape` will be chosen as `shape_class` and `MyShape` will be silently ignored (unless you specify it on the command line, or you rely on the class picker in which `BaseShape` will appear and shouldn't).

With this change, `BaseShape` is seen as abstract and is ignored from shape candidates, won't appear in the class picker and `MyShape` will be chosen if not explicitly told.


cyberw:
Nice!

PR #2395, opened 2023-09-05
refactor(shape): allows custom shapes to reuse run-time, spawn-rate and users parameters

noirbizarre:
Currently, it is not possible for a custom `LoadTestShape` to reuse the built-in `run_time`, `num_users` and `spawn_rate` parameters, even if they need/use it (they are rather classical load testing parameters).
There must be an undocumented reason somewhere for this choice.

This pull request allows custom `LoadTestShape` to reuse those command line parameters if explicitly told with the `reuse_parameters` attribute.

So with this change, the following code will work:

```python
class MyCustomShape(LoadTestShape):

    reuse_parameters = True

    def tick(self):
        expected_run_time = self.runner.environment.parsed_options.run_time
        # Do something with this expected run time
        ...
        return None
```

This change is tested, documented and backward compatible. The WebUI swarm form behavior has been adapted to it (`user_count` and `spawn_rate` fields are not disabled with `reuse_parameters = True`).

**Note**: ~~build is failing on github while succeeded in `tox` locally (for all Python version).~~ I added some output to the test in case of failure to understand why it fails on github.

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -38,6 +38,15 @@ except ModuleNotFoundError:
 version = locust.__version__
 
 
+# Options to ignore when using a custom shape class without `use_common_options=True`
+# See: https://docs.locust.io/en/stable/custom-load-shape.html#use-common-options
+COMMON_OPTIONS = {
+    "num_users": "users",
+    "spawn_rate": "spawn-rate",
+    "run_time": "run-time",
+}
+
+
 def create_environment(
     user_classes,
     options,
@@ -214,10 +223,17 @@ See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-numb
         available_shape_classes=available_shape_classes,
     )
 
-    if shape_class and (options.num_users or options.spawn_rate):
+    if (
+        shape_class
+        and not shape_class.use_common_options
+        and any(getattr(options, opt, None) for opt in COMMON_OPTIONS)
+    ):
         logger.warning(
-            "The specified locustfile contains a shape class but a conflicting argument was specified: users or spawn-rate. Ignoring arguments"
+            "--run-time, --users or --spawn-rate have no impact on LoadShapes unless the shape class explicitly reads them. "
+            "See: docs.locust.io/en/stable/custom-load-shape.html#use-common-options"
         )
+        ignored = [f"--{arg}" for opt, arg in COMMON_OPTIONS.items() if getattr(options, opt, None)]
+        logger.warning(f"The following option(s) will be ignored: {', '.join(ignored)}")
 
     if options.show_task_ratio:
         print("\n Task ratio per User class")
@@ -381,9 +397,6 @@ See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-numb
 
             # start the test
             if environment.shape_class:
-                if options.run_time:
-                    sys.stderr.write("It makes no sense to combine --run-time and LoadShapes. Bailing out.\n")
-                    sys.exit(1)
                 try:
                     environment.runner.start_shape()
                     environment.runner.shape_greenlet.join()
--- a/docs/custom-load-shape.rst
+++ b/docs/custom-load-shape.rst
@@ -78,6 +78,7 @@ Adding the element ``user_classes`` to the return value gives you more detailed 
             {"duration": 30, "users": 50, "spawn_rate": 10, "user_classes": [UserA, UserB]},
             {"duration": 60, "users": 100, "spawn_rate": 10, "user_classes": [UserB]},
             {"duration": 120, "users": 100, "spawn_rate": 10, "user_classes": [UserA,UserB]},
+        ]
 
         def tick(self):
             run_time = self.get_run_time()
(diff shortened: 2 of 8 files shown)


cyberw:
I thought I commented on this but cant seem to find it now. I think this makes sense, but can be done without the extra parameter and just make it the default behavior. Maybe just log an info level message about the fact that `run-time` etc will not be automatically picked up by the shape.

noirbizarre:
Oh, I chose this path because I thought you wanted to keep the actual behavior but making it the default is even better !
I'll rebase this PR and update it, making it the default behavior 👌🏼

cyberw:
yea, I kind of changed my mind. hmm... I guess there is no way of having it both ways... (getting a clear error message when e.g. run time was specified but not actually read/respected by load shape, without having to add the new parameter)

Hmm... let me think about it a bit more.

cyberw:
Hmm. I had forgotten about the web ui, where it becomes really strange to show the parameters by default.

cyberw:
Yea. I think my first idea was the most appropriate (we had that discussion in an issue, right? link it from the PR by adding it in the PR description please)

noirbizarre:
You might be mistaking me for someone else about the in-PR discussion. But it means there is already a discussion about it (I will try to find it back).

So which way is it ? Default with proper logging or opt-in like it is already ?

For the `reuse-parameters`, I would take any better proposal. Just `use-parameters` ?

cyberw:
> You might be mistaking me for someone else about the in-PR discussion. But it means there is already a discussion about it (I will try to find it back).

Maybe I'm just imagining things  :)

> So which way is it ? Default with proper logging or opt-in like it is already ?
>
> For the `reuse-parameters`, I would take any better proposal. Just `use-parameters` ?

Can we just check the shape class for the `reuse_parameters` flag? (and crash if it is not set and the user combines it with e.g. `--run-time`) Do we really need a command line flag?

cyberw:
Uhm. I completely misread your description and thought you were introducing a command line flag.

cyberw:
This change looks good (I dont know what happened when I read it the first time, maybe reviewing too many PR:s at once:)

Two minor things:
* The name of the "is_shape" parameter in web.py doesnt make sense any more. Does it make sense to call it something more like "hide_run_parameters"?
* update the error message "It makes no sense to combine --run-time and LoadShapes. Bailing out." to something that mentions the ability to use --run-time in the shape using reuse_parameters

cyberw:
--run-time, --users and --spawn-rate dont really have a name, but we could call them "common options" (thats what they are called in the help message, and it differentiates them somewhat from "custom arguments" which is handled differently in the UI and other places).

Perhaps we should use "hide_common_options" in web UI and "use_common_options" as class variable name?

noirbizarre:
I agree, options is more concise.
I am working on it, I'll choose between hide/reuse when I get the big picture. If opted for reuse for backward compatibility, but keeping those 3 common or base parameters by default and opting-out when not needed make sense to me. Web UI and cli are 2 different cases, I'll see what makes more sense for the 2

(I am on it, PR should be updated soon)

cyberw:
> If opted for reuse for backward compatibility, but keeping those 3 common or base parameters by default and opting-out when not needed make sense to me. 

Not 100% sure I understand what you mean here, but looking forward to your updated PR :)

noirbizarre:
PR updated with:
- chosen wording (`use_common_options`/`hide_common_options`)
- better error message with deep link to the doc

noirbizarre:
Last question because we have different behavior for `--run-time` vs `--users` and `--spawn-rate` and there is a contradictory statement:
- Ignoring parameters https://github.com/noirbizarre/locust/blob/shape/parameters/locust/main.py#L217-L220
- Then bailing out if headless and runtime specified: https://github.com/noirbizarre/locust/blob/shape/parameters/locust/main.py#L217-L220
- 
I think this is the same pattern and we should have a warning for the 3, choose to bail or ignore but do the same in both case.

Are you OK for:
- `log.warning` for the 3 (instead of `sys.stderr.write`)
- ignoring instead of bailing out for the 3
- do it once in early parameter validation (given the WebUI also ignore those parameters if provided with `use-common-options`)

noirbizarre:
PR updated with those changes.

Sample log output (with a custom logger I use, but the message is the same):
Capture d’écran du 2023-09-15 15-35-34

On terminal supporting hyperlinks, the link to the doc is clickable.

locust/main.py
```diff
@@ -38,6 +38,15 @@
 version = locust.__version__
 
 
+# Options to ignore when using a custom shape class without `use_common_options=True`
+# See: https://docs.locust.io/en/stable/custom-load-shape.html#use-common-options")
+IGNORABLE_OPTIONS = {
```
cyberw, line 43: How do you feel about renaming this variable to COMMON_OPTIONS?

noirbizarre: Done 👍🏼 

In fact, I hesitated between the 2 but then I added a docstring to be explicit so `COMMON_OPTIONS` is better

cyberw:
I agree with the change regarding warning instead of failing. But we should adjust the first two log lines accordingly.

How about merging them into one line:
"Note: --run-time, --users or --spawn-rate have no impact on LoadShapes unless the shape class explicitly reads them. See: https://docs.locust.io/en/stable/custom-load-shape.html#use-common-options"

locust/main.py
cyberw: My thoughts exactly :)

noirbizarre:
I splitted them on 2 lines to avoid having a log line spanning on multiple lines but yes, I can group the first 2 log statements into a single. However I would not add the `Note:`  prefix given it's a warning statement and it will be presented as:
```
WARNING: --run-time, --users or --spawn-rate have no impact on LoadShapes unless the shape class explicitly reads them. See: docs.locust.io/en/stable/custom-load-shape.html#use-common-options
```

PR updated accordingly

cyberw:
Excellent!

PR #2404, opened 2023-09-22
Feature: Separate fetching and waiting times for response

ringus1:
Adding just a tiny bit more granularity for metadata returned when executing requests. Similarly as browsers do this and provide info when server first responded and how long it took to actually fetch the content.

image

### DIFF
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -132,7 +132,7 @@ class HttpSession(requests.Session):
 
         # use stream=True to avoid closing socket and downloading content at the beginning
         # we'll download content if actually requested a bit later
-        # 
+        #
         # a bit of a hack to actually estimate fetching time
         stream = kwargs.pop("stream", False)
 
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -133,7 +133,7 @@ class HttpSession(requests.Session):
         start_time = time.time()
         start_perf_counter = time.perf_counter()
         response = self._send_request_safe_mode(method, url, **kwargs)
-        response_time = (time.perf_counter() - start_perf_counter) * 1000
+        server_response_time = time.perf_counter()
 
         request_before_redirect = (response.history and response.history[0] or response).request
         url = request_before_redirect.url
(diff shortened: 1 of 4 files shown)


cyberw:
I like this as a feature, but it needs more work + tests. also needs to be backwards compatible, which I dont think this is?

ringus1:
I mean, sure it's not a fully ready "feature" per se, as it should be integrated with GUI, etc. But it is backwards compatible definately. It just adds additional kwargs to `request_metadata`, that's it. Can be used by people manually. And used in future PRs to extend UI.

ringus1:
The only "incompatible" change would be removing cast to `int` of _response_time_ in `FastHttpSession`. But standard `HttpSession` returns `float` normally. So just wanted to make it standarized.

cyberw:
I see. I misread the diff to imply that response_time was only set in certain conditions, sorry. 

I'm not sure about the naming

I wont have time to review this any time soon, but if you add unit tests I can check it out later.

ringus1:
Ok, will do. Thank you.

locust/clients.py
```diff
@@ -130,10 +130,16 @@ def request(self, method, url, name=None, catch_response=False, context={}, **kw
         # prepend url with hostname unless it's already an absolute URL
         url = self._build_url(url)
 
+        # use stream=True to avoid closing socket and downloading content at the beginning
+        # we'll download content if actually requested a bit later
+        #
+        # a bit of a hack to actually estimate fetching time
+        stream = kwargs.pop("stream", False)
```
ringus1, line 137: @cyberw I really hate this one, but it seems standard `requests.Session` fetches the content eagerly when `stream=False` is used. So the only way to actually estimate `fetching` is to use `stream=True` during first request. Then consume `content` normally later.

cyberw: Does setting stream=True affect the headers sent, as this implies?https://stackoverflow.com/questions/60343944/how-does-requests-stream-true-option-streams-data-one-block-at-a-time
If it does then I dont think this change is ok...

Issue #2356, opened 2023-06-06
Unable to install locust for Poetry managed project in python 3.11


DrowningSushi:
### Describe the bug
When running `poetry add locust` to install locust from pypi, I see the following error related to locust dependencies, namely `flask-basicauth` and `roundrobin`

```
  Hash for flask-basicauth (0.2.0) from archive Flask_BasicAuth-0.2.0-py3-none-any.whl not found in known hashes (was: sha256:d1a370b5364c280ecbf45bbc294c5c9135a6d96ebf35bb515f8f2d54913fd3cc)

  at ~/Library/Application Support/pypoetry/venv/lib/python3.11/site-packages/poetry/installation/executor.py:673 in _validate_archive_hash
      669│         archive_hash: str = "sha256:" + get_file_hash(archive)
      670│         known_hashes = {f["hash"] for f in package.files}
      671│
      672│         if archive_hash not in known_hashes:
    → 673│             raise RuntimeError(
      674│                 f"Hash for {package} from archive {archive.name} not found in"
      675│                 f" known hashes (was: {archive_hash})"
      676│             )
      677│

  • Installing roundrobin (0.0.4): Failed

  RuntimeError

  Hash for roundrobin (0.0.4) from archive roundrobin-0.0.4-py3-none-any.whl not found in known hashes (was: sha256:97f155e1c2813126fee90725381073fb68c73e040807f8b46dd2ef0110a04a3a)

  at ~/Library/Application Support/pypoetry/venv/lib/python3.11/site-packages/poetry/installation/executor.py:673 in _validate_archive_hash
      669│         archive_hash: str = "sha256:" + get_file_hash(archive)
      670│         known_hashes = {f["hash"] for f in package.files}
      671│
      672│         if archive_hash not in known_hashes:
    → 673│             raise RuntimeError(
      674│                 f"Hash for {package} from archive {archive.name} not found in"
      675│                 f" known hashes (was: {archive_hash})"
      676│             )
      677│

```

### Expected behavior
This issue did not occur when we were previously using python 3.8. Only after upgrading to python 3.11 have we seen this issue.

### Actual behavior
Sharing more detailed logged events from terminal. Can confirm that we are trying to install the latest version of locust

```
❯ poetry add -G dev locust
Using version ^2.15.1 for locust

Updating dependencies
Resolving dependencies... (3.1s)

Writing lock file

Package operations: 12 installs, 0 updates, 0 removals

  • Installing zope-event (4.6)
[...]

PR #2405, opened 2023-09-22
[Feature/2396] Proposal for Re-Writing the Web UI

andrewbaldwin44:
### Changes Proposed

- Add a new boolean argument `modern_ui` (`--modern-ui`, `LOCUST_MODERN_UI`). When `true`, will use the new frontend
- Have `UIExtraArgOptions` returned as a dict. This is needed as otherwise the NamedTuple is returned as an Array when JSON stringified. This is tested and working in the old web ui
- Modify the response from `/stats/reset` to be JSON. I propose this change to be consistent with all other responses from the API (without this change we may see console errors in the new UI as it expects responses to always be JSON)
- Add React frontend using Typescript and Material UI. The UI supports all existing features from the old UI
- Add an example of how it can be possible to extend upon the new modern UI
- Add documentation for developing the dashboard

### Demo

Light and dark mode:
image
image

Advanced options and custom arguments:
image

About Modal:
image

 Running Test:
image
image
image
image
image
image
image

Exceptions:
image

Second Test Markers:
image

Class Picker:
image

Custom Shape:
image

Custom Shape with Class Picker:
image

Multiple Hosts:
image
image

Full Requests Statistics History:
image

With Workers:
image
image
image

With examples/extend_modern_web_ui.py
image
image


andrewbaldwin44:
@heyman @KasimAhmic I know both of you suggested using Vite for this project, would either of you be able to assist in changing the Webpack code to Vite? As I mentioned, I am not at all familiar with the tool. I tried with the following config:
```
import reactSwcPlugin from '@vitejs/plugin-react-swc';
import { defineConfig, splitVendorChunkPlugin } from 'vite';
import checkerPlugin from 'vite-plugin-checker';
import tsconfigPaths from 'vite-tsconfig-paths';

// https://vitejs.dev/config/
export default defineConfig({
  root: 'src',
  publicDir: '../public',
  plugins: [
    tsconfigPaths(),
    reactSwcPlugin(),
    splitVendorChunkPlugin(),
    checkerPlugin({
      typescript: true,
      eslint: {
        lintCommand: 'eslint --ext .ts,.tsx',
      },
    }),
  ],
  build: {
    outDir: '../dist',
    emptyOutDir: true,
  },
  server: {
    port: 4000,
  },
});
```
This unfortunately would not load any images properly (I even tried directly importing, such as they suggest [here](https://vitejs.dev/guide/assets.html)).

I additionally couldn't seem to find any replacement in Vite for the Webpack `watch` command. Since the locust instance will read from the `dist`, we need to have the build emitted for development while running  a locust instance.

Speaking of development - I propose in Webpack to use two different html files, one for developing using the Webpack dev server (it uses test data for the template arguments so we do not need to have a Locust server running) and one for production / developing with a Locust instance. Could the same be easily achieved in Vite?

KasimAhmic:
Checked out your code and got it working with the config below:

```typescript
import reactSwcPlugin from '@vitejs/plugin-react-swc';
import { defineConfig, splitVendorChunkPlugin } from 'vite';
import checkerPlugin from 'vite-plugin-checker';
import tsconfigPaths from 'vite-tsconfig-paths';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    reactSwcPlugin(),
    tsconfigPaths(),
    splitVendorChunkPlugin(),
    checkerPlugin({
      typescript: true,
      eslint: {
        lintCommand: 'eslint --ext .ts,.tsx',
      },
    }),
  ],
  build: {
    outDir: '../dist',
    emptyOutDir: true,
  },
  server: {
    port: 4000,
    open: './dev.html'
  },
});
```

Main changes are:
1. Removed the root and publicDir options
2. Added the `open` option pointing to the `dev.html` file

In addition to the config, I also moved the `index.html` and `dev.html` files from the `src` directory to the `dashboard` directory as per Vite's requirements. Similarly, you will need to add the following line in each HTML file.

```html
<script type="module" src="/src/index.tsx"></script>
```

Because we specified the dev file in the server configuration, this should cover your use case of having two HTML files as I believe Vite will still use `index.html` on production builds.

Regarding watch mode, Vite does have it [built in](https://vitejs.dev/guide/build.html#rebuild-on-files-changes). You can use it like so: `vite build --watch`

And lastly, with regard to images, when loading images from the `public` directory, you would reference them as if the images are available at the root of the site like so:

```diff
- <img height='52' src='/assets/logo.png' width='52' />
+ <img height='52' src='/logo.png' width='52' />
```

image

andrewbaldwin44:
@cyberw There is one other thing that I believe is missing here. I believe for the UI to be properly bundled when `pip install locust` is ran, we need to have some additional configuration in the `pyproject.toml`. The output of the project, `dashboard/dist` is being gitignored, meaning it will not exist when others install the project. Do you know of some way that we could have setup tools build the `dist` on install? Or do we need to store the `dist` in the project? It seems like when we were using SASS the bundled CSS was being uploaded to github right?  I am unfortunately not so familiar with this side of the Python workflow

andrewbaldwin44:
Hey @KasimAhmic thanks so much for the assist! Looks like the `--watch` flag works great and I was able to configure the HTML file for dev / prod using the following:
```
import reactSwcPlugin from '@vitejs/plugin-react-swc';
import { defineConfig, splitVendorChunkPlugin } from 'vite';
import checkerPlugin from 'vite-plugin-checker';
import tsconfigPaths from 'vite-tsconfig-paths';

const PRODUCTION = process.env.PRODUCTION === 'true';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    reactSwcPlugin(),
    tsconfigPaths(),
    splitVendorChunkPlugin(),
    checkerPlugin({
      typescript: true,
      eslint: {
        lintCommand: 'eslint ./src/**/*.{ts,tsx}',
      },
    }),
  ],
  server: {
    port: 4000,
    open: PRODUCTION ? './index.html' : './dev.html'
  },
});
```

The images do work when running the dev server, however there seems to be an issue with how the assets are placed in the `dist`. Try running `PRODUCTION=true yarn vite build --watch` in one terminal and then run a locust instance in another: `locust -f locust.py --modern-ui`

KasimAhmic:
@andrewbaldwin44 Couple things

1. I ran into the same issue with the production build actually. This feels like it might be a Flask configuration issue though, I saw something about Flask only checking a `static` directory for static assets. Is it possible to change this?

2. You can get rid of this line: `const PRODUCTION = process.env.PRODUCTION === 'true';` and replace the value passed to `defineConfig` with a function that takes a configuration object as an argument. For example:

```typescript
import reactSwcPlugin from '@vitejs/plugin-react-swc';
import { defineConfig, splitVendorChunkPlugin } from 'vite';
import checkerPlugin from 'vite-plugin-checker';
import tsconfigPaths from 'vite-tsconfig-paths';

// https://vitejs.dev/config/
export default defineConfig((config) => ({
  plugins: [
    reactSwcPlugin(),
    tsconfigPaths(),
    splitVendorChunkPlugin(),
    checkerPlugin({
      typescript: true,
      eslint: {
        lintCommand: 'eslint ./src/**/*.{ts,tsx}',
      },
    }),
  ],
  server: {
    port: 4000,
    open: config.mode === 'production' ? './index.html' : './dev.html'
  },
}));
```

3. The server configuration is only used in development mode so I think you should be able to delete the conditional open configuration and just leave it as `open: 'dev.html'`. Could be wrong, never did something like this before.

cyberw:
> @cyberw There is one other thing that I believe is missing here. I believe for the UI to be properly bundled when `pip install locust` is ran, we need to have some additional configuration in the `pyproject.toml`. The output of the project, `dashboard/dist` is being gitignored, meaning it will not exist when others install the project. Do you know of some way that we could have setup tools build the `dist` on install? Or do we need to store the `dist` in the project? It seems like when we were using SASS the bundled CSS was being uploaded to github right? I am unfortunately not so familiar with this side of the Python workflow

I dont know the best way to do this off hand, and I'm kinda busy atm. Let me know if you're *really* stuck on it and I'll have a look.

cyberw:
@andrewbaldwin44 For adding the files, you should just need to include them in `MANIFEST.in`, like this:

`recursive-include dashboard/dist *`

cyberw:
The failures page is missing the "type" column? (but lets change the column name to "Message" or "Error message", while we are at it)
image
old school:
image

andrewbaldwin44:
Thanks everyone! @cyberw updated now with the missing "Message" column and the missing line in the `MANIFEST.in`
image

Thanks @KasimAhmic you were correct, I have Flask configured to server assets from `/assets` and Webpack was copying the assets to `dist/assets` but Vite was not. The dashboard is now using Vite for compiling 👍

cyberw:
Cool stuff! Can you add a basic test case in test_web.py as well?

andrewbaldwin44:
@cyberw Sure, I can add a test to ensure the correct behavior is happening with the `--modern-ui` flag. For testing the new web ui, we should do it using [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/). That part will take time, do you prefer me to do it in this feature, or in a future ticket?

andrewbaldwin44:
@cyberw Updated now to support `run_time` from CLI as default arg (`spawn_rate` was already supported in the new UI)
image

cyberw:
From a code perspective this looks really good, apart from the flake8 and black fails :) (I really should make them pre-commit checks)

I think we should mention the availability of the modern UI explicitly in the manual. I’m thinking maybe as a note at the end of the web ui section of quickstart? And make it clear that it is experimental.

andrewbaldwin44:
Apologies for the failing pipeline, I committed the new tests a bit too quickly! The pipeline is passing now and I have added a line to the quick start guide mentioning the new web UI as well as a warning that this feature is experimental.

docs/developing-locust.rst
```diff
@@ -117,3 +117,71 @@ Or you can make sass watch for changes to the ``.sass`` files and automatically
...
+- Verify yarn was installed correctly
+
+```bash
+yarn --version
+```
+
+- Next in dashboard, install all dependencies
+
+```bash
+cd dashboard
```
cyberw, line 160: update this

cyberw:
Couple things:
* If you refresh the page, everything disappears. The view is not updated until a new request comes in (I think, or maybe it is some other delay)
* The --users parameter is not prefilled correctly (it is "1" no matter what value I set on command line)
* If I use --autostart the UI still shows the "new" button (when it should show stop/edit)

cyberw:
Also, the reports download page still looks like its the old version:
image

andrewbaldwin44:
@cyberw I've resolved the issues mentioned above. The issue with the users not autofilling was because the value provided by the command line is `num_users`, however the value that needs to be passed to `/swarm` is `user_count`. I think we may need to review the naming of these variables at some point :)

As for the 'modern' html report, I have looked into it and I think there is a fair amount of additional changes or refactoring that would need to be made to the backend for the feature to work (for example the stats and exceptions would need to be json serializable, the logic for generating the percentile data would need to be moved to the backend, and the logic for adding the javascript into the html file would need to be a bit different since the name of the js files will change with every build). After all of this we will probably want to add some additional tests as well. What do you think, would it be better to have this as a separate ticket? Or do you prefer to have everything done together?

cyberw:
If it is a lot of work then we can leave it for later. I’ll do a few more tests tomorrow or so.

cyberw:
I'm still experiencing some issues. If I launch with --autostart (or open a new browser window against a running test), there are no updates at all (except if I reload)..

image

andrewbaldwin44:
@cyberw The stats are now being properly fetched even while users are spawning and the issue with the text escaping in the failures table has now been resolved as well

cyberw:
Nice! Seems to have introduced a new issue though: After starting the test, the UI sometimes switches back to the "start load test" page for a second.

cyberw:
Once those two things are fixed I think we're good to go (as this is still opt-in I'm not too worried about bugs ets). 

Maybe @mquinnfd @KasimAhmic or @heyman would like to chime in. Especially regarding anything structural, or any other things that are important to get right from the start.

KasimAhmic (changes requested):
I've left comments on some of the files and I have some more general notes.

1. I'm curious about the decision to use the `connect()` pattern for the connecting the components to the Redux store. Why did we go this route as opposed to creating selectors for the various pieces of the state?
2. I dislike the use of the `sx` prop. While it is what MUI recommends, I find it a bit limiting at times and it makes the code messier IMO. I would prefer if we switched to another CSS-in-JS solution like [tss-react](https://docs.tss-react.dev/). Open to comments on this one though.

Overall, the PR looks good for a first pass save for a few minor things outlined in the comments. Once those are addressed, I'd have no issues merging this and then additional improvements like using `createApi` for requests and using Redux selectors as opposed to the `connect` approach can be made later.

locust/webui/public/assets/logo.png
mquinnfd, line 1: I wonder if we're rendering a 52px^2 image, that we might use a smaller one than 500px^2 - client probably doesn't need to download half a Mb of the icon

mquinnfd:
I suspect `locust/webui/dist/**` needs to be purged and added to `.gitignore`, seems to be some built artifacts sneaking in to the refs there

locust/webui/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="/assets/favicon.ico" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
```
mquinnfd, line 6: I think we fundamentally want to avoid using externally loaded resources, as much as it's tempting.

Lots of Locust users execute tests inside of private networks, so loading Google fonts will likely make things inconsistent.

We should statically cache whatever fonts we need, or use inbuilt ones IMO

locust/webui/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "license": "MIT",
+  "scripts": {
+    "dev": "vite",
+    "dev:watch": "vite build --watch",
+    "build": "tsc && vite build",
+    "lint": "eslint './src/**/*.{ts,tsx}'",
+    "format": "prettier --write '**/**/*.{ts,tsx}'",
+    "type-check": "tsc"
+  },
+  "devDependencies": {
```
mquinnfd, line 14: How comfortable are you with the stability of these packages?

Leaving carets (^) in, in production is quite optimistic, and we risk builds and tests breaking overnight with no changes because some packages messed up a semver change.

mquinnfd: It might be better to uprev these on some duty cycle when performing maintenance upgrades.

locust/webui/src/api/asyncRequest.ts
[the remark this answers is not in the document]
mquinnfd: I would agree, reimplementing an async client wrapper like this is a little heavy-handed

locust/webui/src/utils/object.ts
```diff
@@ -0,0 +1,36 @@
+export const isEmpty = <T extends Record<string, any>>(object: T) => !Object.keys(object).length;
+
+export function merge<ObjectA, ObjectB>(objectA: ObjectA, objectB: ObjectB) {
```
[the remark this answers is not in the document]
mquinnfd: +1 for just using Lodash

locust/web.py
```diff
@@ -258,7 +273,8 @@ def reset_stats() -> str:
             if environment.runner is not None:
                 environment.runner.stats.reset_all()
                 environment.runner.exceptions = {}
-            return "ok"
+
+            return jsonify({"success": True, "message": "ok"})
```
mquinnfd, line 277: Are we unilaterally changing this return for non-modern-ui clients as well?

As much as the new return type is more sensible, it has the potential to break something, for someone.

We might consider conditionally altering this return response based on the UI being modern or not, for now at least.

locust/webui/src/utils/object.ts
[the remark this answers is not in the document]
andrewbaldwin44: I think lodash may be a little overkill for this project, especially since there is never really a case that we need a deep merge. We could even use JavaScript's built-in `Object.merge`

locust/webui/src/redux/slice/ui.slice.ts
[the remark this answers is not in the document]
andrewbaldwin44: @KasimAhmic Yeah I was just trying to follow a purely 'functional' style, where we never modify any variables directly. But I agree, this way is cleaner and more explicit 👍

andrewbaldwin44:
> I suspect `locust/webui/dist/**` needs to be purged and added to `.gitignore`, seems to be some built artifacts sneaking in to the refs there

@mquinnfd Yes I wasn't sure how this should be done in a Python workflow using `pyproject.toml`. For the dashboard to be available to the users, the static files need to built and bundled with locust. I noticed that even with Sass we were commiting the compiled CSS

locust/webui/package.json
andrewbaldwin44: @mquinnfd We are commiting the `yarn.lock`, so we should not be worried about unexpected updates

locust/web.py
andrewbaldwin44: Yeah the old UI never consumes the response from this endpoint, so I wouldn't be worried about this being a breaking change. However I can add the conditional return, just to be extra safe 👍

andrewbaldwin44:
> I've left comments on some of the files and I have some more general notes.
> 
> 1. I'm curious about the decision to use the `connect()` pattern for the connecting the components to the Redux store. Why did we go this route as opposed to creating selectors for the various pieces of the state?
> 2. I dislike the use of the `sx` prop. While it is what MUI recommends, I find it a bit limiting at times and it makes the code messier IMO. I would prefer if we switched to another CSS-in-JS solution like [tss-react](https://docs.tss-react.dev/). Open to comments on this one though.
> 
> Overall, the PR looks good for a first pass save for a few minor things outlined in the comments. Once those are addressed, I'd have no issues merging this and then additional improvements like using `createApi` for requests and using Redux selectors as opposed to the `connect` approach can be made later.

@KasimAhmic 
1. Yes the use of `connect` is perhaps a personal preference of keeping the concerns separate, rather than polluting the component with hooks. I prefer to keep the components quote 'dumb', in that they should be able to receive a list of props and not need to know where to get the props from. I have worked on projects using `createStructuredSelector` from reselect for example, however this seems to be a feature missing from RTK. Is there ever a time where we may particularly need the memorization from selectors, or do you not find the connect function to be explicit enough?
2. Regarding the use of MUI's `sx`, I was actually opting for this _because_ it is limiting. I wanted to avoid using custom CSS as much as possible in this project (the more custom css is allowed, the more magic numbers pop up, and the less 'standard' things look). You'll notice I limit the sx properly to things such as positioning or flex direction / centering and any spacing used is using the MUI grid system (rather than random pixel values). I think there's not really a need for custom CSS, nor do we really want to encourage it. Not to mention this saves us the use of another library

locust/webui/src/api/asyncRequest.ts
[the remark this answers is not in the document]
andrewbaldwin44: Yeah I agree with what you're saying, the issue is I often find us needing more control over the request and response that most libraries would allow. For example, in projects like this using a python backend, I like to have the variable casing converted, so we don't need to guess when to use camel case and when to use snake case. I can see if we can achieve something similar using RTKs built-in wrapped

locust/web.py
mquinnfd: Yeah I do agree, I just know that [someone out there is calling that](https://www.hyrumslaw.com/) (for some reason) :D

locust/webui/package.json
mquinnfd: Yeah fair enough 👍

mquinnfd:
> > I suspect `locust/webui/dist/**` needs to be purged and added to `.gitignore`, seems to be some built artifacts sneaking in to the refs there
> 
> @mquinnfd Yes I wasn't sure how this should be done in a Python workflow using `pyproject.toml`. For the dashboard to be available to the users, the static files need to built and bundled with locust. I noticed that even with Sass we were commiting the compiled CSS

@andrewbaldwin44 Ah, I get you. Are we just looking for a sort of "on-build" hook in `poetry` that we can use to run the `yarn` build then? I'll have a look at any sensible options for this

mquinnfd:
P.S I'm amazed you got this together so quickly @andrewbaldwin44, this is all looking really great, truly epic effort 👍

cyberw:
A side note: Having a tab for viewing the log/console output of locust in the UI would be an absolute killer feature (maybe even having it light up if there are warning/error loggings etc - it is all too common that people are only looking in the UI and dont see issues that are reported on terminal). But maybe for 2.0 of this UI :)

mquinnfd:
On the build process, I note that Locust currently uses plain [build](https://pypa-build.readthedocs.io/en/latest/index.html) to generate the released `.whl`.

It looks as though this tool is designed to be as simple as possible, so I don't really see pre/post build hooks or ways to extend the behaviour (by design).

I suppose we have a few options, and it depends on the requirements. This is not exhaustive but a few ideas which @cyberw will probably have strong feelings about, having been around during the project's evolution. 

---

1. *Use Poetry*: Given that we already maintain `pyproject.toml` - it might be an idea to consider moving to [poetry](https://python-poetry.org/) (which is close to being the new build/packaging standard). This would allow for running `yarn` as part of a prebuild subprocess [a bit like this](https://github.com/python-poetry/poetry/issues/5539#issuecomment-1126818974). This would change the build process a little bit, we could either just use `poetry` to build the artifact, and publish it in the same way, or use `poetry` to publish it as well. There are obviously a fair few implications of changing the established build procedure for the product, however, and it would need to be thoroughly regression-tested - though it would make the local and remote build/package procedures more unified (as opposed to encapsulating the front end build/dist logic in Github actions).

2. *Separate local and remote builds*: If we care less about local and remote build processes being unified, then in the publishing workflow within Github actions, we could have a multi-step workflow, and pass the artifact between the steps [like so](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow), keeping a node-specific action runner for the `yarn` build and then trying to bundle this into the built Python `.whl` as part of a series of actions. Locally, we could then just adapt the `Makefile` to run both builds as part of the build process, and if necessary for testing, write some bash to do this multi-build stitching process to generate a representative `.whl` file.

3. *Back to setup.py*: Move back to a `setup.py` file, which is less in-vogue, may have incompatibilities with what we're currently doing tools-wise, but would most certainly allow for arbitrary build logic like this.

4. *Store generated files*: Commit the compiled files to the repository, which is a little grim for obvious reasons.

cyberw:
> On the build process, I note that Locust currently uses plain [build](https://pypa-build.readthedocs.io/en/latest/index.html) to generate the released `.whl`.
> 
> It looks as though this tool is designed to be as simple as possible, so I don't really see pre/post build hooks or ways to extend the behaviour (by design).

1. IMO, Poetry is only good for/worth the hassle for _applications_ (where pinning exact dependencies makes sense) not _frameworks_ (where users may add extra dependencies or want to use different versions of dependencies etc).
2. Kind of sad, but possible.
3. Yea, that could totally work. I really dont like some of the limitations in pyproject.toml.
4. 🤢

So I guess what I'm saying is going back to setup.py makes the most sense to me.

mquinnfd:
I'm not the biggest `poetry` fan, but `1.2+` has been pretty good, we switched our production stuff over to it (frameworks, applications, everything). It's only worth the hassle if you're creating packages for publication, etc.

Practically, just reverting to the old `setup.py` would be convenient, but I do feel that in general, the Python community is moving away from `setuptools` etc. which would make this a bit of a step backwards, and potentially hamper future development.

- `pip`'s [own docs](https://pip.pypa.io/en/stable/reference/build-system/setup-py/) about the subject
- `build`'s [own docs](https://pypa-build.readthedocs.io/en/latest/differences.html#fallback-backend) about the subject
- Older [PEP-517](https://peps.python.org/pep-0517/) vs Newer [PEP-518](https://peps.python.org/pep-0518/)

I could find some time to make a PR branch for each system I suppose (`poetry`, `setuptools`), and see which feels better for this use-case?

cyberw:
> Yes I am also not a fan of tracking the dist, however I think it may be out of scope to change the build process in this ticket. What do you suggest @cyberw? Do we create another PR to revert #2336 before merging this one?

Hmm... how much space would it require if we were to commit them (option 4), as a temporary workaround?

andrewbaldwin44:
@cyberw Yeah that's what I had already done, right now we're tracking the dist. It was the only simple possible solution :/

cyberw:
> I could find some time to make a PR branch for each system I suppose (`poetry`, `setuptools`), and see which feels better for this use-case?

Imo, lock files make little sense for frameworks, and poetry is very centered around its lock file.

I'd love a PR for both options, but I feel it is unlikely that poetry will fit my taste :) (rn I dont think it even works, see https://github.com/locustio/locust/issues/2356, but that would be nice to fix of course)

cyberw:
> @cyberw Yeah that's what I had already done, right now we're tracking the dist. It was the only simple possible solution :/

Ah, I thought there was something more needed. In that case we leave it as is for now.

locust/webui/index.html
cyberw: Unless it looks terrible, I think we should use the built in ones.

cyberw:
@andrewbaldwin44 I'd like to make a new release of locust (primarily adding support for Python 3.12) within a few days. Are there any final adjustments we can/should make before merging this, or do you feel it is more than a few days work left and I should just make the release first?

andrewbaldwin44:
@cyberw Sorry I have been a little busy the last couple days, I have some time tonight to fix the bug that you were experiencing, but I may need a little more time for re-writing the API logic as @KasimAhmic and @KasimAhmic suggested

cyberw:
No worries. 

It might even make sense to have this in a release of its own (so we dont confuse any bugs), and it makes sense to have it as "git/prerelease only" for a while too. I'll probably make a new locust release (with all the other stuff) this weekend or so.

andrewbaldwin44:
Hey guys, I have just pushed an update that should resolve all the comments posted above. I will still need some more time for getting the API logic working with RTK, I will need to look into it more next week. I think using RTK may be a bit tricky for certain requests, namely `/stats/requests`, as we want the response from this request to be placed into several different reducer paths, as well as some other logic that needs to be handled. What do we think about using an API solution such as axios to handle this?

locust/test/test_web.py
```diff
@@ -1083,3 +1083,63 @@ def test_request_stats_full_history_csv(self):
...
+        gevent.sleep(0.01)
+        try:
+            response = requests.get("http://127.0.0.1:%i/" % web_ui.server.server_port)
+            self.assertEqual(500, response.status_code)
+            self.assertEqual("Error: Locust Environment does not have any runner", response.text)
+        finally:
+            web_ui.stop()
+
+    def test_index_with_modern_ui(self):
+        self.assertEqual(200, requests.get("http://127.0.0.1:%i/" % self.web_port).status_code)
```
cyberw, line 1121: Since every one of these tests relies on starting a new instance we can save a lot of test run overhead by just combining them into one.

cyberw:
tests on python 3.8 are flaky for some reason. ignore that.

andrewbaldwin44:
Sorry for the delay everyone, all the comments are now resolved and I have replaced the API logic with RTK's `createApi`. Thanks for the feedback! :)

cyberw:
Nice! Can you update the pr description? I think the following is no longer relevant and maybe there are other things to add. 

> Modify the response from /stats/reset to be JSON. I propose this change to be consistent with all other responses from the API (without this change we may see console errors in the new UI as it expects responses to always be JSON)

andrewbaldwin44:
@cyberw Yes once this is merged I can open a PR for updating the HTML report and then a third PR to introduce a suite of tests using React testing library. From there I am happy to support with the community requested features, such as boolean custom arguments and having a tab for viewing the logs :)

SmallMousePP:
How to get this new html report when using headless model?

cyberw:
@mquinnfd Does your offer to implement poetry to build the frontend still stand? If so, I'm more interested now :)

I've heard more good stuff about it (and the bad stuff I heard/inferred was maybe a little misguided).

The one thing I want to make sure is that we dont lose the ability to do an editable install (-e) that can be used even outside the directory/venv.

PR #2411, opened 2023-10-01
Test and explicitly support Python 3.12

cyberw:


cyberw:
I'm going to have to pause this, because 3.12 is only supported in gevent 23.7.0, which in turn requires greenlet 3.0 and that is only available as an RC release (greenlet-3.0.0rc3).

I *could* make this work (it would cause some incompatibilities with Playwright though), but I couldnt make geventhttpclient work (`from gevent.ssl import match_hostname` fails because match_hostname is gone).

PR #2413, opened 2023-10-04
:rocket: Try compatible fork Niquests

Ousret:
This PR showcases how Locust could evolve outside of Requests.

Niquests is supposed to be a (mostly) compatible fork of Requests.

**Here are the biggest pros of this:**
- OS truststore by default, no more certifi!
- Fully type-annotated! 
- HTTP/3 over QUIC.
- HTTP/2 by default.
- Exit Python http.client! in favor of h11.
- Timeout by default.
- Inspect peer certificate, HTTP version, TLS version, cipher, and so on via hook/callback before a request is sent.
- All the features you expose are available in all three protocols.
- Python 3.7+, no sacrifice needed.

I took the time to verify that it works properly as-is.
A minimal patch was required but no feature was tampered with or broken in the process.

Also, there is room for adding features, like spawning requests per protocols (1.1, 2, and 3) could be interesting.
This PR is a "solid" proof of concept, if you want to, we can make this work/distributable.

Finally, I plan to land full support for multiplexed requests in HTTP/2 and HTTP/3 soon! Probably going to be useful here.
https://github.com/jawah/niquests

### DIFF
--- a/locust/clients.py
+++ b/locust/clients.py
@@ -5,12 +5,12 @@ from contextlib import contextmanager
 from typing import Generator, Optional
 from urllib.parse import urlparse, urlunparse
 
-import requests
-from requests import Request, Response
-from requests.adapters import HTTPAdapter
-from requests.auth import HTTPBasicAuth
-from requests.exceptions import InvalidSchema, InvalidURL, MissingSchema, RequestException
-from urllib3 import PoolManager
+import niquests
+from niquests import Request, Response
+from niquests.adapters import HTTPAdapter
+from niquests.auth import HTTPBasicAuth
+from niquests.exceptions import InvalidSchema, InvalidURL, MissingSchema, RequestException
+import urllib3
 
 from .exception import CatchResponseError, LocustError, ResponseError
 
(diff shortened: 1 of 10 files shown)


cyberw:
Well it looks… ok. But I dont see it replacing Requests for the default HttpUser. Maybe it could fit in locust-plugins. I’m curious though: have you encountered a specific limitation running Requests in Locust, or is it more of a nice to have/personal preference?

Ousret:
> Maybe it could fit in locust-plugins.

Noted. You went a bit too fast, right now it is not about moving forward with this, as per the "Try" in the title.
We are collecting intelligence on this as we speak and seek feedback from various concerned actors.

> have you encountered a specific limitation running Requests in Locust, or is it more of a nice to have/personal preference?

Well, the bullet points should be a starter. I used Locust in various cases and it is easily criticized in some aspects like (but not limited to) lack of h2, and h3 (as per modern browsers), and does not _dismiss and/or separate_ connect/handshake delays, ..
Locust is more seen as "a quick preview / rough estimate" but not accurate enough for high-stakes analysis. Unfortunately.
Nice tool nonetheless.

cyberw:
👍 As an option, niquests may make sense, but not as the default.

PR #2420, opened 2023-10-11
[Feature/2419] Modernize the report.html when using the --modern-ui Flag

andrewbaldwin44:
### Changes Proposed:
- Add a method `to_dict` to the `StatsEntry` and `StatsError` classes. This will allow for returning a JSON version of these classes. This will also allow for reducing complexity in the `request_stats` method
- Add three new arguments to `get_html_report`:
  - `static_path`: indicates where the static files can be found
  - `use_modern_ui`: to allow for returning a different template with different template arguments when the `--modern-ui` flag is set
  - `theme`: allows for dark mode
- Remove vendor chunking in vite: although this may offer slight performance benefits, in this project it is not very relevant, and for the downloaded report we need to have inline js (no imports / exports)
- Ensured the new modern template has all the same functionality as the old template, and that the old template continues to work as expected

### Demo
image
image
image
image


cyberw:
Nice! I’m on mobile, but will look into the python code asap. Can you add a unit test? One that downloads the report and checks return code and some (any) content inside it.

PR #2422, opened 2023-10-14
Improve readability of MasterRunner.client_listener

lozovskaya:
improve readability of MasterRunner.client_listener and reduce the cyclomatic complexity from F to D


cyberw:
Uhm. Some of these changes make a bit of sense, but others just make spaghetti code a spaghetti/function hybrid monster.

A better improvement might be structured around putting the handling of each message type into its own function.

PR #2423, opened 2023-10-15
Support argcomplete as an optional dependency

singingwolfboy:
[`argcomplete`](https://github.com/kislyuk/argcomplete) adds tab completion for Python CLIs build with the [`argparse`](https://docs.python.org/3/library/argparse.html) module -- [and it works with `configargparse` as well.](https://github.com/kislyuk/argcomplete/issues/123)  This pull request sets up the Locust CLI to use `argcomplete` if it's installed, but if it's not installed, users can still use the Locust CLI without tab completion. In addition, this PR adds an "extra" requirement named "completion", so if users run `pip install locust[completion]`, then Pip will install Locust _and_ `argcomplete`.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -21,6 +21,27 @@ Installation
 
 3. Done! Now you can :ref:`create your first test <quickstart>`
 
+Optional: CLI tab completion
+----------------------------
+
+Locust uses the argcomplete_ module to provide tab completions for the ``locust`` command line interface.
+This is entirely optional; if you don't care about tab completion, you can skip this section entirely!
+If you do want to enable tab completion, first install the ``argcomplete`` module
+into the same environment where you installed ``locust``:
+
+.. code-block:: console
+
+    $ pip3 install argcomplete
+
+Then add this line to your shell config file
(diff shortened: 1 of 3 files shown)


cyberw:
Nice! Can you add a note about this in the documentation?

singingwolfboy:
> Nice! Can you add a note about this in the documentation?

Sure, I just pushed a commit that does so. Unfortunately, this involves modifying the user's shell, and not everyone knows how to do that correctly, so I did a "happy path" description, and linked to the `argcomplete` documentation in case things don't go well. Does that work for you?

cyberw:
Perfect

cyberw:
@singingwolfboy this doesn't really work when there is no locustfile.py in the current directory. Can you fix it?

Here's the trace output (after running `export _ARC_DEBUG=1`)


```
~ locust --helCould not find 'locustfile.py'. See --help for available options.
Could not find 'locustfile.py'. See --help for available options.
```

cyberw:
Hmm... it really slows down "regular" tab completion (for things like filenames).

cyberw:
@singingwolfboy I'm going to have to revert this because of the above issues being so bad. A fix would be welcome though...

PR #2431, opened 2023-10-21
Refactor - replace list with deque: TaskSet._task_queue & WindowsKeyPoller.captured_chars

tdadela:
TaskSet._task_queue & WindowsKeyPoller.captured_chars use methods responsible for removing/appending elements at beginning & end.
collections.deque performs those in constant time in contrast with list (linear time for appending/removing at beginning).

### DIFF
--- a/locust/input_events.py
+++ b/locust/input_events.py
@@ -1,3 +1,4 @@
+import collections
 from typing import Dict, Callable
 
 import gevent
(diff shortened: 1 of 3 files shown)


cyberw:
Irrelevant for real performance. Not technically worse than the current implementation but not significantly better either.

tdadela:
You've changed your mind about TaskSet._task_queue implementation in #2698 & #2653.
If you want, I can open a new PR with changes in input_events.py.

cyberw:
I’m sorry, I got confused by your changes in input_events. They are not really relevant for performance, but your changes in TaskSet really were, and I should have accepted this PR. If you want to make a new PR for input events I will accept it.

tdadela:
No problem. Thanks for the response.

PR #2432, opened 2023-10-21
refactor(MasterRunner): simplify condition in check_stopped

tdadela:
The current version of condition is hard to read, so I've simplified it.


cyberw:
Thx!

PR #2433, opened 2023-10-25
Add elapsed time information to running load test dashboard

ZeroWal:
I thought it might be useful to see the elapsed time information in the current load test dashboard.
Screenshot attached

Locust - Fork for time elapse

### DIFF
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -12,6 +12,10 @@
         <div class="top-content container">
             <img src="./static/img/logo.png?v={{ version }}" class="logo" />
             <div class="boxes">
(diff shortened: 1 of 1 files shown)


cyberw:
A useful feature indeed! But the current/old UI is being replaced. Can you implement it for the new one instead? https://docs.locust.io/en/stable/developing-locust.html?highlight=Modern#making-changes-to-locust-s-modern-web-ui

ZeroWal:
Sure, I'll try...

cyberw:
I'm closing this, as the old UI is being retired, we're not taking in new changes for it. But a similar change for the new one would be great!

PR #2436, opened 2023-10-27
Modern UI shows wrong (old) hostname when setting hostname in start dialog

andrewbaldwin44:
- Tested overriding a single host
- Tested overriding multiple hosts


cyberw:
Thanks!

PR #2444, opened 2023-11-02
feat(shape): support shape generating users

noirbizarre:
This PR adds support for `TestLoadShape`s generating users.

In practice, this means:
- being able to send new classes to workers
- properly register new classes on runner and environment to avoid runtime errors and broken stats

This allows you to dynamically create users from an external source.

```python
users: list[type[User]] = []

class GeneratorShape(LoadTestShape):
    def get_user(self, index: int) -> type[User]:
        if index >= len(users):
            user = self.runner.register_user(f"User{index}", (BaseUser,), {})
            users.append(user)
        return users[index]

    def tick(self):
        index = trunc(self.get_run_time())
        if index >= 2:
            return None
        return (index + 1, 1, [self.get_user(index)])
```

This will create a new `User` each second (`User0`, `User1`...).
Note that we are using the `register_user(name, bases, attributes)` form which create the class itself,
but you can entirely craft your own class:

```python
users: list[type[User]] = []

class GeneratorShape(LoadTestShape):
    def get_user(self, index: int) -> type[User]:
        if index >= len(users):
            user = types.new_class(f"User{index}", (BaseUser,), {})
            user. __module__ == "my_module"
			setattr(sys.modules[user.__module__], user.__name__, user)
            self.runner.register_user(user)
            users.append(user)
        return users[index]

    def tick(self):
        index = trunc(self.get_run_time())
        if index >= 2:
            return None
        return (index + 1, 1, [self.get_user(index)])
```

Note: those are dumb examples just dynamically setting a class name, in practice, this PR allows to dynamically define:
- ancestors
- attributes including:
	- tasks as methods
	- tasks as `tasks` attribute
	- any of the `wait_time`, `weight`, `fixed_count`... attributes

In fine, this allows dynamic user creation from:
- logs
- HAR traces
- any external specification file (JSON, yaml...)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -131,7 +131,7 @@ Runner classes
 =====================
 
 .. autoclass:: locust.runners.Runner
-    :members: start, stop, quit, user_count
+    :members: start, stop, quit, user_count, register_user, on_new_user_class
 
 .. autoclass:: locust.runners.LocalRunner
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -106,6 +106,7 @@ project = "Locust"
 
 # Intersphinx config
 intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
     "requests": ("https://requests.readthedocs.io/en/latest/", None),
 }
 
--- a/docs/custom-load-shape.rst
+++ b/docs/custom-load-shape.rst
@@ -121,3 +121,67 @@ If you need one or all of those parameters, you can force locust to accept them 
             # Do something with this expected run time
             ...
             return None
+
+
+Generating users
+----------------
+
+With custom load shape, you have the possibility to generate user classes on the fly.
+This allows you to dynamically build user from an external source.
+
+.. code-block:: python
+    
+    users: list[type[User]] = []
+
+    class GeneratorShape(LoadTestShape):
+        def get_user(self, index: int) -> type[User]:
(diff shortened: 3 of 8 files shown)


cyberw:
Hmm. Would this work in a distributed run?

I’m not sure this is a good idea. There should be other ways to solve this. And this really has opening-a-can-of-worms potential.

noirbizarre:
I am totally OK with that: I hesitated to submit this PR as I was sure that pickle+sync on class sharing would not be an easy thing to accept, but I did it anyway because I work on different projects using the same pattern, so I thought it might be useful for others.

Anyway, as it is already working for me except for 2 details, would you accept instead an alternative PR just removing those limitations ? 🙏🏼 

Those would be:
- only raise an exception for locustfile without non-abstract user class if the locustfile has a non-abstract shape in it (given only the shape will be used and if you import it, you might have the user not declared in the locust file)
- delay the warning raised for user classes without tasks and raise it when the class is actually used, dispatched or added to the `(Runner|Environment).user_classes` (instead of on locustfile load today, resulting in warnings raised whether the class is actually used or not)

Those 2 points would allow the user generation for those using it without making the user generation built-in locust.

cyberw:
Hmm... Both of those changes would result in a slightly worse experience for other use cases than yours. Perhaps you can hack your way around it some other way?

PR #2453, opened 2023-11-08
Modern UI: Add sorting to columns on statistics page and downloaded report

andrewbaldwin44:
### Proposal

- Re-implement sorting function from old UI
- When clicking on table header, sort by selected field
- Text color will change after a header has been interacted with, to indicate the table is being sorted
- Reversing the order of columns only applies when clicking on the same field (reverse is reset when clicking on a new column)
- Sorting is applied to stats, failures, and workers table (same as the previous UI)

image


cyberw:
❤️ I’ll have a look this weekend (I’m going to be at PyCon Sweden. doing a talk about Locust on friday! :)

cyberw:
Is there a way to revert to the default sort? I was thinking the third click on the same column could make it revert to default sort (and stop indicating the column as being sorted).

PR #2455, opened 2023-11-09
Add terraform variables

sdarwin:
Hi,
This is a very small detail, but I think it makes the Terraform module a bit easier to use.
In Terraform, if there is a setting which the typical user should probably adjust, or consider adjusting, the place to put that is `variables.tf`. 
By reviewing `variables.tf` you see a list of configurations to modify.
In this case, `ssh_export_pem` and `aws_region`. I needed to modify them as likely most users would also.


cyberw:
makes sense!

PR #2459, opened 2023-11-10
Add Modern UI HTML Report to --html Option

### LINKED ISSUE #2458: Add Modern UI HTML Report to --html Option
### Prerequisites

### Description

When using the `--html` option with `--modern-ui`, the modern HTML template should be used

andrewbaldwin44:
When using the `--html` option together with the `--modern-ui` the modern report template will now be used
- Ensured the report is outputed correctly when using the `--headless` option with both the old and new report
- Ensured the report continues to work properly form the Web UI (old and new)

Fixes #2458


cyberw:
Nice! Is it possible to make it respect dark/bright mode?

andrewbaldwin44:
Good suggestion 👍 I added a flag `--theme` which will allow for controlling the default theme. The flag can be used inside and outside of headless mode, whichever theme is set will be the first theme that you see in the UI (the toggle will still be enabled should you wish to change it)

cyberw:
Nice! But could you maybe use a choices parameter type instead of a string? https://docs.python.org/3/library/argparse.html#choices

andrewbaldwin44:
:+1: 
image

andrewbaldwin44:
I also added one more commit which will detect if a users browser prefers dark or light theme and will use this as the default for the report, this way it is not necessary to set/be aware of the `--theme` variable in order not to be blinded (we were already doing this for the UI but not yet for the HTML report)

cyberw:
> I also added one more commit which will detect if a users browser prefers dark or light theme and will use this as the default for the report in headless mode, this way it is not necessary to set/be aware of the `--theme` variable in order not to be blinded (we were already doing this for the UI but not yet for the HTML report)

Perhaps we dont even need to add the option then? We already have too many options so the help text is kinda unreadable :)

andrewbaldwin44:
Yeah and now that I think of it, perhaps it even makes more sense this way, as the HTML file is something that could be passed around to multiple people and different people might have different preferences on the theme

cyberw:
Nice!

PR #2465, opened 2023-11-14
Ensure to wait a second before next `tick()`

wiatrak2:
When developing a custom `TestShape`, a developer needs to implement `tick()` method, that manages `Users` spawning. As mentioned in the [official documentation](https://docs.locust.io/en/stable/custom-load-shape.html#custom-load-shapes):

> Locust will call the tick() method approximately once per second.

Therefore, a developer may design a `TestShape` with a strong assumption that the number of `Users` will be adjusted with ~1 second interval. Unfortunately, the interval is indeed kept only when no adjustment in `Users` number is made - hence it seems useless, as any interval is meaningful only during some changes. Here's an example of logs from Locust master when a custom `TestShape` was implemented and  the `tick()` method is clearly triggered too often:

locust_logs

the timestamps of log messages show, that 20 `Users` were spawned within ~6 seconds, while the spawn rate is set to `1.0`. In fact I was able to achieve much bigger discrepancies between expected/declared and actual number of users.

In this PR I provide a small adjustment in `Runner`'s code to ensure the ~1 second interval is maintained. Also a small code optimization in handling the `current_tick` Tuple was added to keep the file length almost unchanged :)

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -345,7 +345,6 @@ class Runner:
                 self.shape_last_tick = None
                 return
             elif self.shape_last_tick != current_tick:
-                shape_adjustment_start = time.time()
                 if len(current_tick) == 2:
                     user_count, spawn_rate = current_tick  # type: ignore
                     user_classes = None
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -333,6 +333,7 @@ class Runner:
     def shape_worker(self) -> None:
         logger.info("Shape worker starting")
         while self.state == STATE_INIT or self.state == STATE_SPAWNING or self.state == STATE_RUNNING:
+            shape_adjustment_start = time.time()
             current_tick = self.environment.shape_class.tick() if self.environment.shape_class is not None else None
             if current_tick is None:
                 logger.info("Shape test stopping")
@@ -343,11 +344,13 @@ class Runner:
                 self.shape_greenlet = None
                 self.shape_last_tick = None
                 return
-            elif self.shape_last_tick == current_tick:
-                gevent.sleep(1)
-            else:
+            elif self.shape_last_tick != current_tick:
                 shape_adjustment_start = time.time()
-                user_count, spawn_rate, user_classes = current_tick if len(current_tick) == 3 else (*current_tick, None)  # type: ignore
+                if len(current_tick) == 2:
+                    user_count, spawn_rate = current_tick  # type: ignore
+                    user_classes = None
+                else:
+                    user_count, spawn_rate, user_classes = current_tick  # type: ignore
                 logger.info("Shape test updating to %d users at %.2f spawn rate" % (user_count, spawn_rate))
                 # TODO: This `self.start()` call is blocking until the ramp-up is completed. This can leads
                 #       to unexpected behaviours such as the one in the following example:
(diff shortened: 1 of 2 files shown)


locust/runners.py
```diff
@@ -367,6 +366,9 @@ def shape_worker(self) -> None:
                 #        of each load test shape stage.
                 self.start(user_count=user_count, spawn_rate=spawn_rate, user_classes=user_classes)
                 self.shape_last_tick = current_tick
+                shape_adjustment_time_ms = time.time() - shape_adjustment_start
+                if shape_adjustment_time_ms < 1:
```
cyberw, line 370: `<1` feels strange :)

Can you maybe change it to not use an if statement at all and instead do:

gevent.sleep(Math.max(0, 1 - shape_adjustment_time_ms))

Also, the build seems to be breaking...

wiatrak2: Yep, just noticed the test, sorry! I'll fix that. the `max(..., 0)` seems good, will adjust the commit then :)

wiatrak2: I haven't noticed that overriding `current_tick` leads to unwanted behavior. Fix was commited

locust/runners.py
```diff
@@ -346,11 +346,8 @@ def shape_worker(self) -> None:
             elif self.shape_last_tick == current_tick:
                 gevent.sleep(1)
             else:
-                if len(current_tick) == 2:
-                    user_count, spawn_rate = current_tick  # type: ignore
-                    user_classes = None
-                else:
-                    user_count, spawn_rate, user_classes = current_tick  # type: ignore
+                shape_adjustment_start = time.time()
+                user_count, spawn_rate, user_classes = current_tick if len(current_tick) == 3 else (*current_tick, None)  # type: ignore
```
cyberw, line 350: I think this is less readable than the old way. Does it really need to change or is it unrelated to the issue you are trying to fix?

wiatrak2: You're right, it's not related. I wanted to make it more "pythonic" and readable with my previous change (where I modified `current_tick` when its length was 2), but eventually it indeed got not very clear :) I restored the code from `master` for this part. Also, I changed a bit the `gevent.sleep` usage, to call it once for entire `tick()` handling - seems more understable for me now

wiatrak2:
@cyberw I've also added a dedicated test case that doesn't base on `get_run_time` when modifying the Users count, but relies on 1-sec interval

cyberw:
nice!!

cyberw:
@wiatrak2 is it possible that one of the test cases got a little flaky from this change?  Perhaps it should be adjusted? https://github.com/locustio/locust/actions/runs/6919408511/job/18822743361?pr=2472

cyberw:
@wiatrak2 Can you have a look at the occasionally failing test case? While I do believe your change is good, it did break the test case and I dont have time to analyze why it is breaking, so I might have to revert it...

wiatrak2:
@cyberw it happened in the worst possible moment - I'm on vacation and away from my computer until December 16th :( I won't be able to have a look before

cyberw:
No worries! I can disable the tests for now (I dont think anything is actually broken, and if it is its a small issue)

wiatrak2:
Yep, the change was not a sufficient one. If you can disable the test - let's do it. I will have a (very fresh) look once I'm back to the business ;)

PR #2470, opened 2023-11-17
Tiny optimization of StatsEntry._log_response_time

lailongwei:
Optimize a statistical code

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -389,10 +389,10 @@ class StatsEntry:
         self.total_response_time += response_time
 
         if self.min_response_time is None:
-            self.min_response_time = self.max_response_time = response_time
+            self.min_response_time = response_time
         else:
             self.min_response_time = min(self.min_response_time, response_time)
-            self.max_response_time = max(self.max_response_time, response_time)
+        self.max_response_time = max(self.max_response_time, response_time)
 
         # to avoid to much data that has to be transferred to the master node when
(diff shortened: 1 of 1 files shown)


cyberw:
I’m not sure this code is faster… and it is less readable to me.

lailongwei:
> I’m not sure this code is faster… and it is less readable to me. (This comment was in regard to your second commit, the first one looks nice)

first time commit is OK(I think so, haha).
second time is actually make code unreadable, I will be rollback later.
third time commit, I tested performance, the performance is not much difference, I will be rollback too, the simple(and temporary) test code:
```python
# coding: utf-8

import time
import random
from collections import OrderedDict

test_seq = []
test_od_size = 1000
test_times = 10000
test_od = OrderedDict()
for i in range(test_od_size):
    test_od[i] = random.randint(1, 100)
    test_seq.append(i)


random.shuffle(test_seq)


sum_val = 0
beg_pc = time.perf_counter()
for i in range(test_times):
    for idx in test_seq:
        if idx in test_od:
            sum_val += test_od[idx]
print(f'cost time:{time.perf_counter() - beg_pc}')  # 1.5xx

test_time_idx = 0
beg_pc = time.perf_counter()
for i in range(test_times):
    for idx in test_seq:
        sum_val += test_od.get(idx, 0)
print(f'cost time:{time.perf_counter() - beg_pc}')  # 1.5xx
```

PR #2471, opened 2023-11-17
Modern UI - Failures line chart does not display current number of failures

### LINKED ISSUE #2469: Total RPS Charts | Failures line chart does not display current number of failures
### Prerequisites

### Description

The failures line chart does not show the current number of failures over time. 

I have to refresh the page to get the line chart updated on the chart, The green RPS line does update and show up however the red line goes to 0 and it stays there.

1. Opens the charts tab
2. The next time the chart refreshes the failures line goes to zero and stays there

Screenshot 2023-11-16 at 17 23 00

3. Refresh the page to see the failures.
4. Red line goes to zero and stays there over time.
Screenshot 2023-11-16 at 17 23 11

### Command line

-f my_test.py --autostart --modern-ui

### Locustfile contents

```python3
N/A
```

### Python version

3.11

### Locust version

2.18.3

### Operating system

MacOS

andrewbaldwin44:
Fixes #2469

I propose adding a property `total_fail_per_sec` to `/stats/requests`, which tracks the total number of failures from the aggregated row. We will then use this as the proper value for `currentFailPerSec` in the charts.

The old UI used to use `report.stats[report.stats.length-1]` in the frontend. If you are against making changes to the python code we could do the same, however I am of the opinion that it is logical since we already return `current_rps` and this helps keep the FE "dumb" (it shouldn't really need to know that total stats are always last in the array)


cyberw:
Nice!

PR #2481, opened 2023-11-22
Create any directories as part of the CSV Prefix

sillydan1:
This PR adds the functionality to create any directories provided as part of the --csv / CSV_PREFIX value. e.g. `--csv=some/path/to/my_prefix`.


cyberw:
Nice! Can you also add it to the command line help, so people know the feature is available? And add a test case too!

sillydan1:
I've added the things you requested. I'm not sure if the test case is good enough though.

cyberw:
Nice!

PR #2484, opened 2023-11-24
FastResponse: Use utf-8 instead of empty string as default in response text encoding

MiltiadisKoutsokeras:
FastResponse: Use utf-8 instead of empty string as default in response text encoding, if no response headers or no response encoding are set in the headers. Having no headers already used `utf-8` and having response headers but no information on encoding in them used an empty string which always throws an error [str constructor](https://docs.python.org/3.9/library/stdtypes.html#str).


cyberw:
I like the other PR best (as it more closely matches python-requests)

PR #2485, opened 2023-11-24
FastResponse: Detect response text encoding when no information is present on the response headers.

MiltiadisKoutsokeras:
If no response headers are present or do not contain response encoding, use charset_normalizer.detect to detect the encoding from response content bytes.


cyberw:
I like this one the best, because it matches python-requests. Can you fix the typing issue? The regular tests are unstable at the moment, feel free to ignore that...

PR #2493, opened 2023-11-29
Further relax FastHttpUser's host checking to match HttpUser's

### LINKED ISSUE #2492: FastHttpUser is stricter than HttpUser about the host attribute
### Prerequisites

### Description

As a followup to https://github.com/locustio/locust/issues/1222:

`FastHttpUser` has differences from `HttpUser` but is documented as a simple drop-in replacement for most cases. The fix for https://github.com/locustio/locust/issues/1222 (https://github.com/locustio/locust/pull/1227) [was intended to relax `FastHttpUser`'s host checking to be in line with `HttpUser`'s](https://github.com/locustio/locust/issues/1222#issuecomment-574736455) however there's still a major difference between them: `HttpUser` doesn't validate the host url format, only ensuring it's not `None`:

https://github.com/locustio/locust/blob/68ffae8/locust/user/users.py#L252

I don't think there's a compelling reason for these types to differ in their host checks.

I'm using locust against a websocket service where it also uses the `FastHttpUser` client: in our case we would ideally use `host` to point to a websocket url but `FastHttpUser`'s stricter check disallows this whereas `HttpUser` would.

So I think the stricter checking should be removed entirely to bring them into parity.

### Command line

locust -f locustfiles/locustfile.py  --websocket_url ws://localhost:9150

### Locustfile contents

```python3
.
```

### Python version

3.10

### Locust version

2.19.1

### Operating system

Linux

pjenvey:
Fixes #2492

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -338,8 +338,6 @@ class FastHttpUser(User):
             raise LocustError(
                 "You must specify the base host. Either in the host attribute in the User class, or on the command line using the --host option."
             )
-        if not re.match(r"^https?://[^/]+", self.host, re.I):
-            raise LocustError(f"Invalid host (`{self.host}`), must be a valid base URL. E.g. http://example.com")
 
         self.client: FastHttpSession = FastHttpSession(
             self.environment,


cyberw:
Makes sense!

PR #2494, opened 2023-11-30
Implement Modern Web UI Tests

andrewbaldwin44:
### Proposal

Add test coverage to the modern web UI using vitest and react testing library!


cyberw:
NICE!

PR #2499, opened 2023-12-03
Fix Warnings from the Web UI Tests

andrewbaldwin44:
- Add a config to vitest that allows for using `after` / `before` / `afterAll` / `beforeAll` hooks (and thus silences this warning)
 - Add a width and height to the e-charts container so that it doesn't output any warning either
 - Fix warnings that recommends using `act` when a component re-renders
 - The tests also reveled that `addSpaceToChartsBetweenTests` was using `payload[0]`, which isn't correct. This is changed to a simple empty value
image
image


cyberw:
Nice!

PR #2502, opened 2023-12-04
Add Notifications for Log Viewer

andrewbaldwin44:
### Proposal
- Add a generic notification hook that can subscribe to any data changes. The hook will register that a notification should be shown so long as the page containing the data is not currently being viewed. 
- Refactor the log viewer component to use the notification hook
- Add tests for the log viewer and notification hooks

Recording 2023-12-04 at 13 46 33

The locust file used for testing adds a new log every 5 seconds:
```
class ExampleTest(TaskSet):
    @task
    def get_stuff(self):
        self.client.get("/")
        logger.warning("The --autostart argument is implied by --headless, no need to set both.")

class Example(HttpUser):
    wait_time = constant(5)
    tasks = [ExampleTest]
    host = "http://0.0.0.0:5000"
```


locust/webui/src/components/LogViewer/LogViewer.test.tsx
andrewbaldwin44, line 1: (Moved, not deleted)

cyberw:
Not sure if I'm missing something, but no notification appears for me?
image
image

cyberw:
Can we make the notification icon only appear on WARNING or CRITICAL log levels? Every test is going to log some things on INFO level, no need to notify people about that :)

andrewbaldwin44:
Let me look into that, I'm trying to keep the logic as 'self contained' as I can, so if we would like to have notifications for other tabs in the future, we can

cyberw:
Perfect!

PR #2509, opened 2023-12-07
Chart Average Response Time in Modern UI

andrewbaldwin44:
### Proposal
- Change the logic of how `PERCENTILES_TO_CHART` is being passed to the modern UI. We will now pass an array of all the percentiles we wish to chart. This will allow users to continue extending this variable as they wish, and it will additionally allow for users to see as many different percentiles in the chart as they would like
- Add `totalAvgResponseTime` to response from `/stats/requests` and add it to the chart in the modern UI
- Added a new variable `MODERN_UI_PERCENTILES_TO_CHART` to the stats module. This variable can still optionally be overriden from a locust file
- None of these changes affect the legacy UI

### Screenshots
image

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -125,7 +125,7 @@ def get_html_report(
                 "show_download_link": show_download_link,
                 "locustfile": escape(str(environment.locustfile)),
                 "tasks": task_data,
-                "percentiles_to_chart": stats_module.PERCENTILES_TO_CHART,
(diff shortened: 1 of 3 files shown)


cyberw:
nice! is it possible to change PERCENTILES_TO_CHART default to only 0.95, by checking if modern ui is set? (because we dont need avg AND 50th percentile)

cyberw:
Nice!

PR #2512, opened 2023-12-11
Fixed response failure for unsupported encoding

prince-melvin:
FastHttpUser fails the request if the response can't be converted to text. For example, if the response headers have application/octet stream as response header there is no proper encoding to convert it back to text.

This PR allows the request to succeed by, 
1. Check response.content is null for failure
2. Before attempting to convert to text check if encoding is not None.

(We shall log warning to prompt that response cannot be converted to text and will not be available as a json)

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -378,7 +378,7 @@ class FastHttpUser(User):
         with self.client.request(method, url, catch_response=True, headers=headers, **kwargs) as r:
             resp = cast(RestResponseContextManager, r)
             resp.js = None  # type: ignore
-            if resp.text is None:
+            if resp.content is None:
                 resp.failure(str(resp.error))
             elif resp.text:
                 try:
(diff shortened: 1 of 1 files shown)


locust/contrib/fasthttp.py
```diff
@@ -378,7 +378,7 @@ def rest(
         with self.client.request(method, url, catch_response=True, headers=headers, **kwargs) as r:
             resp = cast(RestResponseContextManager, r)
             resp.js = None  # type: ignore
-            if resp.text is None:
+            if resp.content is None:
                 resp.failure(str(resp.error))
             elif resp.text:
```
cyberw, line 383: does this really solve your issue? wont it just crash on this line instead?

locust/contrib/fasthttp.py
```diff
@@ -467,8 +467,8 @@ def text(self) -> Optional[str]:
                 # No information, try to detect
                 if not self.encoding:
                     self.encoding = str(detect(self.content)["encoding"])
-
-        return str(self.content, self.encoding, errors="replace")
+        if self.encoding != "None":
```
cyberw, line 470: are you intentionally comparing with the _string_ None, or can this be replaced with `if not self.encoding is None:`?

locust/contrib/fasthttp.py
prince-melvin: No it doesn't crash, Since in text property I am trying to avoid encoding the string. The issue comes from there where we try to convert to test using `None` encoding

locust/contrib/fasthttp.py
prince-melvin: Since encoding is a string and in the previous line we have assinged, encoding = str(detect()) it is not actually Null or None it is string `None`

cyberw: Ah. Then that should be changed.

prince-melvin: Done

cyberw:
Thx!

PR #2514, opened 2023-12-12
Fix useSwarmUi Test with Mock Timers

andrewbaldwin44:
- Use mock timers in the `useSwarmUi` test to avoid false test failures
- Remove the dependency on `waitFor` in favor of `runAllTimersAsync`


cyberw:
Nice!

PR #2521, opened 2023-12-23
Support IPv6 for zmq connection between master and worker

simonwu-sn:
This is a follow up for PR https://github.com/locustio/locust/pull/2290.

The solution is based on http://api.zeromq.org/4-2:zmq-setsockopt. 

> Set the IPv6 option for the socket. A value of 1 means IPv6 is enabled on the socket, while 0 means the socket will use only IPv4. When IPv6 is enabled the socket will connect to, or accept connections from, both IPv4 and IPv6 hosts.

The test is done with kind for 3 env, ipv4 only, ipv6 only and dual stack. With help of https://abdelrhmanhamouda.github.io/locust-k8s-operator/

Test result:
ipv4-only  before-patch(success) after-patch(success)
ipv6-only before-patch(fail, the worker can't connect with master) after-patch(success)
dualstack before-patch(success) after-patch(success)

### DIFF
--- a/locust/rpc/zmqrpc.py
+++ b/locust/rpc/zmqrpc.py
@@ -13,6 +13,7 @@ class BaseSocket:
 
         self.socket.setsockopt(zmq.TCP_KEEPALIVE, 1)
         self.socket.setsockopt(zmq.TCP_KEEPALIVE_IDLE, 30)
+        self.socket.setsockopt(zmq.IPV6, 1)
 
     @retry()
     def send(self, msg):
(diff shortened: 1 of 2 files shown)


cyberw:
Seems to have the same issue as the previous PR. I’m all for supporting ipv6, but someone needs to figure out the GH/build issue.

simonwu-sn:
> Collaborator

Do you mean the `make test` will hung?

simonwu-sn:
test fixed.

PR #2534, opened 2024-01-04
Update shape class' runner when Web UI picker is used

wiatrak2:
a shape class' `.runner` member is set by `_create_runner` method of `Environment` class. However, when the UI picker is used, the `shape_class` member of an `Environment` instance is not initialized yet, hence, there's no `Runner` assignment on initialization. Therefore, if you'd like to refer to `self.runner` in your custom test shape (for instance to get the number of users running), you receive an error saying that `self.runner` is `None`. 

Hence, I added the shape class `runner` initialization also when the shape class is selected with UI picker.

### DIFF
--- a/locust/env.py
+++ b/locust/env.py
@@ -103,6 +103,10 @@ class Environment:
         self._validate_user_class_name_uniqueness()
         self._validate_shape_class_instance()
 
+    def update_shape_class_runner(self, runner: Optional[Runner] = None) -> None:
+        if self.shape_class:
+            self.shape_class.runner = runner or self.runner
+
     def _create_runner(
         self,
         runner_class: Type[RunnerType],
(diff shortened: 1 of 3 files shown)


cyberw:
Can you add a test case that fails without this change?

wiatrak2:
There you go :)

cyberw:
I see the need for the change, but `update_shape_class_runner` is a very weird method. Perhaps it can be refactored/renamed to something that makes more sense. Or just remove the method and have the "weird" behavior inline, where it is at least readable.

wiatrak2:
I believe that we should avoid code repetition. Also I think that the action should be encapsulated by a method, as it wouldn't be clean to modify `Environment`'s members from inside the `WebUI`. Thinking about a better name, my candidate is `set_runner_for_shape_class` - what do you think? 😅

cyberw:
How about doing it here: https://github.com/locustio/locust/pull/2534/files#diff-498b8c363e788cc0210ac8600bce7660c60d210cad41af086886b79055c717beR636
, where we are already accessing the shape class directly.

Something like
`shape_class.runner = self.environment.runner`

environment-runner separation is very messy, and while I dont like duplication, I like methods with unclear meaning even less :)

wiatrak2:
your code, your rules ;)

cyberw:
Thx!

PR #2535, opened 2024-01-06
Added response.ok support to FastHttpUser's FastResponse class simila…

PankajKhanwani:
…r to HttpUser's Response

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -500,6 +500,11 @@ class FastResponse(CompatResponse):
         """
         return self._response.get_code() if self._response is not None else 0
 
+    @property
+    def ok(self):
+        """Returns True if :attr:`status_code` is less than 400, False if not."""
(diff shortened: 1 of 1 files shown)


cyberw:
Nice!

cyberw:
Huh. Seems this only fixes it for direct calls, not for ResponseContextManager. So the following still doesnt work for FastHttpUser. Can you take a look?

```
with self.client.get("/", catch_response=True) as resp:
    print(resp.ok)
```

PankajKhanwani:
class FastTestUser(FastHttpUser):
    host = "https://example.com"

    @task
    def get(self):
        with self.client.get("/", catch_response=True) as resp:
            print(resp.__class__)
            print(resp.ok)
            
Output:
`<class 'locust.contrib.fasthttp.ResponseContextManager'>
True
`

cyberw:
Sorry, my bad. I must have accidentally been on a branch or something..

PR #2538, opened 2024-01-10
Web UI Modern Auth

### LINKED ISSUE #2517: Update dependency or remove support for Basic Auth for the Web UI
I'm considering removing support for Http Basic Auth for the locust web UI. It depends on Flask-BasicAuth, which hasnt been updated since 2013 (!)

If someone still needs authentication for the UI, maybe it could be replaced with something like: https://pypi.org/project/Flask-HTTPAuth/

But I dont have time to build it myself.

So... PRs welcome. I will remove it in january if nothing happens.

### Command line

locust -f examples/locustfile.py --web-auth foo:bar

andrewbaldwin44:
Fixes #2517 

image

### Proposal
- Add a new page to the modern UI that allows for users to authenticate with username / password or with a provider
- Remove `flask_basicauth` in favor of `flask_login`. The proposal is to move the responsibility of security and maintainability onto the user. Locust will simply specify which routes need to be protected, but the user will be the one responsible for implementing the protection. This has the added benefit of providing flexibility as to how users authenticate to the app
- Add an example of how authentication can be implemented by the user
- Add documentation on how to implement authentication
- Add tests


examples/auth.py
```diff
@@ -0,0 +1,70 @@
+import os
```
cyberw, line 1: Maybe add a comment at the top describing what the example is actually about. And name the file something more descriptive (maybe webui_auth.py or something)

examples/auth.py
```diff
@@ -0,0 +1,70 @@
...
+from locust import HttpUser, events
+from flask import url_for, Blueprint, request, make_response, redirect, session
+from flask_login import UserMixin, login_user
+
+
+class LocustHttpUser(HttpUser):
+    pass
+
+
+class User(UserMixin):
```
cyberw, line 12: Can this class be named something else? To avoid confusion with locusts own User class.

locust/argument_parser.py
```diff
@@ -351,11 +351,9 @@ def setup_parser_arguments(parser):
     )
     web_ui_group.add_argument(
         "--web-auth",
-        type=str,
-        dest="web_auth",
-        metavar="<username:password>",
-        default=None,
-        help="DEPRECATED. See https://github.com/locustio/locust/issues/2517 Turn on Basic Auth for the web interface. Should be supplied in the following format: username:password ",
+        default=False,
```
cyberw, line 354: I think we should use a new argument name to avoid confusion. if someone says `locust --web-auth ...` there needs to be a clear message about what has happened (and maybe a link to the docs). Maybe the new name should be `use_auth`?

cyberw:
Cool stuff. See comments. Can you have an extra look and ensure there are no remnants (documentation etc) of the old stuff? Pretty sure we can nuke the AuthCredentialsError class for example.

locust/test/test_web.py
```diff
@@ -1082,18 +1083,36 @@ def tearDown(self):
...
+        def load_user(id):
+            return User()
+
+        self.web_ui.login_manager.request_loader(load_user)
+
+        response = requests.get("http://127.0.0.1:%i" % self.web_port)
+        d = pq(response.content.decode("utf-8"))
+
+        self.assertNotIn("authArgs", str(d))
+        self.assertIn("templateArgs", str(d))
```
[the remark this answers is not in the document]
andrewbaldwin44: No the first test basically simulates as if a user is already signed in, and that the app should then be returned. The 2nd test simulates a user that is not signed in and expects the log-in page to be returned.

We can't really call `login_user` from Flask login because we are not in a Flask context. I was thinking this might be the simplest way to provide some basic tests without testing Flask login or testing what should be implemented by the user

locust/argument_parser.py
cyberw: Cool! But can you leave `--web-auth` still in there? 

Like with for example --expect-slaves, using `help=configargparse.SUPPRESS,` and exiting if someone specifies it https://github.com/locustio/locust/blob/648781401dac8b3112f9e818f5ce7cf0903e55c6/locust/main.py#L146

cyberw:
Looking good! Only that one last comment about --web-auth :)

locust/argument_parser.py
andrewbaldwin44: Yup makes sense :+1: 
image

andrewbaldwin44: btw we can't use `--use-auth` because of this
image

cyberw:
Given that the package we use is called flask-login, perhaps it makes more sense to name the argument `--login` instead of `--auth` (and the function parameter name etc. sorry if i'm nit-picking but argument names are really hard to change after the fact :)

andrewbaldwin44:
Yup that makes sense! What do you think about `--web-login`, that way it is clear it is only for the web UI?

locust/web.py
```diff
@@ -69,12 +79,15 @@ def my_custom_route():
...
+    auth_args: Dict[str, Any]
+    """Arguments used to render auth.html for the web UI auth page. Must be used when configuring auth"""
+
     def __init__(
         self,
         environment: "Environment",
         host: str,
         port: int,
-        auth_credentials: Optional[str] = None,
+        use_auth: bool = False,
```
cyberw, line 90: rename this to `web_login`?`

locust/argument_parser.py
```diff
@@ -355,9 +355,16 @@ def setup_parser_arguments(parser):
         dest="web_auth",
         metavar="<username:password>",
         default=None,
-        help="DEPRECATED. See https://github.com/locustio/locust/issues/2517 Turn on Basic Auth for the web interface. Should be supplied in the following format: username:password ",
+        help="DEPRECATED: use --web-login",
```
cyberw, line 358: Change this to configargparse.SUPPRESS and let the error message speak for itself :)

cyberw:
I made two more comments :) 

And also... I have another thought. Maybe we dont even *need* a command line argument to enable it? 

Perhaps we can just rely on whether someone has called environment.web_ui.login_manager.user_loader(load_user)? That should already be 1:1 with our users wanting to enable login, right?

andrewbaldwin44:
I guess we could have the users set their `user_loader` on the `web_ui` (as in `environment.web_ui.user_loader = my_user_loader`). Then we would have to initialize the login_manager in the `auth_required_if_enabled` and enable auth if a user_loader is provided. My issue with doing it this way is we could be limiting a lot of flexibility by not exposing the `login_manager` to the user

cyberw:
I see! Sounds messy. Lets keep the parameter then. Ok to merge?

andrewbaldwin44:
Let's do it :)

cyberw:
Agh. I think I made a mistake when merging. Can you see what is wrong? I merged master (with your fix) into this branch first but must have messed it up... If you dont have time I'll get to it later...

PR #2542, opened 2024-01-11
FIX - HTML reports not saved after run-time limit reached in headless mode

mariaanko:
I had an issue with no reports generated after the tests were done in headless mode (--headless) AND (--run-time) limit was set up.
I was running this in a gitlab pipeline along with my tests (wanted the report as an .html artifact) and had this listener set up:
``` 
   @events.test_stop.add_listener
    def on_test_stop(environment):
        print("tests finished - quitting...")
        os._exit(0) #needed this to stop the pipeline completely...to exit locust
```
...after the run time limit was reached, the listener was called, but reports were not generated...
now it's working with this little fix (before it did not do the html report part):
```
[2024-01-11 12:37:54,831] mariaanko/INFO/locust.main: writing html report to file: soak-report.html
[2024-01-11 12:37:55,169] mariaanko/INFO/locust.main: --run-time limit reached, shutting down
tests finished - quitting...
```
**this may be a really stupid fix, but please take this issue into consideration, and if possible please try to do another fix...**

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -482,6 +482,8 @@ See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-numb
             else:
                 logger.info("--autoquit not specified, leaving web ui running indefinitely")
         else:  # --headless run
+            if options.html_file:
+                save_html_report(options.modern_ui)
             logger.info("--run-time limit reached, shutting down")
             runner.quit()
 


cyberw:
Why do you need to call os._exit(0) in test_stop? You're basically crashing locust and I think that is a rather unreasonable case to support.

mariaanko:
@cyberw thanks for taking this into consideration and taking the time to reply!
the reason is plain simple as for my use-case:
1. I have a docker image set up on gitlab which supports python
2. I have a separate repo where my tests using locust are
3.  In that separate repo - I run a pipeline that takes the docker image and starts the locust script with run-time limit and html parameters as well as other parameters I need...
4. when I start the locust process in that pipeline, the pipeline never ends - it's stuck for ever, unless I get a pipeline run timeout and even then the report.html is not being in artifacts of a pipeline...

That's why I thought this might help to fix it, if you have another ideas on how to stop the locust command completely  - to be able to run other commands - or save the artifact - after the tests were finished, that would be awesome!

cyberw:
Normally a locust process launched with —headless and —run-time should exit. Perhaps there is something preventing it from doing so? A thread or greenlet you launched?

mariaanko:
sorry for the late response...I just tried it and it works now...
explanation:
I had the following task:
```
    @task
    def navigate_to_random_page(self):
        random_url = random.choice(self.WEB_PAGES)
        self.client.get(random_url)
        # self.driver.implicitly_wait(5)
        # time.sleep(1)
```
where I had now commented out the wait and sleep and it exits the locust command finally...
I suppose that the requests could not be finished properly since I was implicitly waiting, or what happened I don't know...

thanks for your help, and please close this PR :( I feel bad now

PR #2547, opened 2024-01-13
Fix Python3.8 typing issues (revert to using Union/Optional instead of pipe-notation)

cyberw:
@tdadela brought this to my attention. I still yearn for the day when we drop support for everything prior to 3.10, but that day is not yet here...


cyberw:
Turns out it was actually fixable! #2548

PR #2548, opened 2024-01-13
Fix mypy configuration & upgrade typing

tdadela:
1. Mypy: check typing compatibility for the oldest supported Python version - currently Python3.8.
2. Fix mypy errors, add from \_\_future__ import annotations & upgrade type annotations using fix-future-annotations.
3. Upgrade mypy, set types-requests as mypy test dependency in tox.


cyberw:
Nice!!

Issue #2546, opened 2024-01-13
Unifying percentile displays in Locust's statistics screens


FooQoo:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

## Summary
I have noticed a lack of consistency in the percentile displays on Locust's UI statistical screens. Specifically, there is a discrepancy in the percentiles displayed between the overall statistics screen (statics) and the screen that records the time series change of statistics (charts).

### Current Display
- Statics screen: Median/90%ile/99%ile
- Charts screen: Median/95%ile

## Proposed Changes
To achieve a more consistent data analysis experience, I propose the following changes:
1. Add 95%ile to the Statics screen.
2. Add 90%ile and 99%ile to the Charts screen.

These changes will unify the percentile displays across both screens, making it easier for users to compare data between different screens.

## Expected Impact
With these changes, users will find it easier to compare statistical data across different screens. This will lead to more comprehensive and accurate performance analysis, and is expected to improve the overall usability of Locust.

cyberw:
Graphing percentiles over time is kind of a weird thing, as the average 95th percentile over a run is not the same as the 95th percentile for the whole test.

And with few requests per time interval, percentiles are likely to be all over the place, especially higher percentiles (that's why we dont use 99th in the graph) 

I kind of think the current choice of percentiles make sense, but I get that they feel inconsistent. Adding 95th percentile to the stats screen by default makes some sense though (btw, these are all configurable https://docs.locust.io/en/stable/configuration.html#customization-of-statistics-settings)

cyberw:
What do you think @andrewbaldwin44 ? Maybe we should replace 90th percentile in the stats with 95th just to make it consistent? There's already so many columns in that page :) Maybe add a link to the customization documentation on that page too, I think it would be really useful...

andrewbaldwin44:
I agree with you @cyberw, if it is already configurable in both the table and the charts, I'm not sure if it makes for us to change it. Especially considering that different users will always have different opinions on what columns should be displayed.

I have another suggestion here, what would you think of allowing the users to select which columns they want to be displayed. I've seen some other load testing tools offer this feature and I think it could be useful to allow the user to focus on the data they care about

cyberw:
I was thinking about that. If it can be done without becoming messy it would be great!

FooQoo:
@cyberw, @andrewbaldwin44

[...]

PR #2550, opened 2024-01-14
Customization Feature for Percentile Display on Statics Page

FooQoo:
#### Overview

This PR introduces a feature that allows users to freely specify which percentiles are displayed on the statics page of Locust. Additionally, minor modifications are made to align the display content of the statics page with that of the charts page.

#### Implementation Details

1.  A new variable `PERCENTILES_TO_STATICS` is added, enabling users to specify the percentiles displayed on the statics page.
2.  The specified values in `PERCENTILES_TO_STATICS` will be reflected in both the existing UI and the modern UI of Locust.

#### Display Modifications

*   Changed "Median (ms)" to "50%ile (ms)".
*   Changed "90%ile (ms)" to "95%ile (ms)".

#### Background

Based on Issue #2546, this enhancement aims to improve the customizability of the statics page and ensure consistency with the charts page. This will enable users to analyze data more effectively.


cyberw:
I was thinking this could be even nicer to have controlled in the UI at run time? 

Is _TO_STATICS an abbreviation of _TO_STATISTICS? Looks weird to my eyes :)

Build fails because of code formatting. 

What do you think @andrewbaldwin44 ?

cyberw:
Does the test actually cover the use of PERCENTILES_TO_STATICS?

andrewbaldwin44:
Hey looking good! I also had the thought (and I'm not sure if this is what you meant @cyberw) that it could be nice to be able to control the columns in the web UI. Something like this:
image
 
The nice part about `PERCENTILES_TO_STATICS` is that it allows users to set the percentiles they care about and potentially saves some performance

locust/stats.py
```diff
@@ -692,9 +698,7 @@ def to_dict(self, escape_string_values=False):
             "max_response_time": proper_round(self.max_response_time),
             "current_rps": self.current_rps,
             "current_fail_per_sec": self.current_fail_per_sec,
-            "median_response_time": self.median_response_time,
```
andrewbaldwin44, line 695: Maybe we should keep median response time? As the code is right now, users don't have any way of adding this statistic if they want to see it

locust/web.py
```diff
@@ -613,11 +614,16 @@ def update_template_args(self):
...
+                "percentiles_to_statistics": stats_module.PERCENTILES_TO_STATISTICS,
             }
         else:
             percentiles = {
                 "percentile1": stats_module.PERCENTILES_TO_CHART[0],
                 "percentile2": stats_module.PERCENTILES_TO_CHART[1],
+                # Convert percentiles to percent notation for use in templates
+                "percentiles_to_statistics": [
+                    normalize_decimal(percentile * 100) for percentile in stats_module.PERCENTILES_TO_STATISTICS
+                ],
```
andrewbaldwin44, line 626: I think we shouldn't have to support any new features on the legacy UI. We could leave this as is and then we wouldn't need to change the templating

andrewbaldwin44:
Yeah maybe we can add some tests for `PERCENTILES_TO_STATISTICS` and some validation to ensure that the statistics provided are valid? Similar to what we do with `PERCENTILES_TO_CHART`

FooQoo:
@cyberw @andrewbaldwin44 

Thank you for your positive feedback and insightful suggestions. Here is my response to the points you have raised:

### Addressed Items
1. **Concern About Variable Name and Build Failure**: These issues have already been addressed in a recent commit. I corrected the variable name from `_TO_STATICS` to `_TO_STATISTICS` and resolved the build failure issue by applying the `ruff` formatter.

### Next Actions
1. **Test Coverage for `PERCENTILES_TO_STATISTICS`**: I plan to add tests to cover this new feature to ensure its proper functioning and integration.

### Regarding Your Suggestions
1. **On 'Median' vs '50%ile' Clarity**: Thank you for your valuable opinion. I am considering keeping 'Median' as the default statistical measure and not including '50%ile' in the percentiles. This aligns with your suggestion and I believe it will make the statistics more user-friendly.

2. **Column Control in Web UI**: I will try implementing the proposed mechanism that allows users to control the display columns directly in the Web UI.

### Additional Question
Regarding the support for the legacy UI, I have already implemented some features. Should I consider removing these to focus solely on the modern UI, or would it be more beneficial to maintain them? Your guidance on this matter would be greatly appreciated.

FooQoo:
Added unit tests for `normalize_decimal` and integration tests, also reverted percentile display from `50%ile` back to `median`.

locust/stats.py
cyberw: I just meant that displaying it by default might not make sense (because it is not a very useful stat, when you have the average...). But I dont really care either way :)

docs/configuration.rst
```diff
@@ -203,4 +203,6 @@ The list of statistics parameters that can be modified is:
 +-------------------------------------------+--------------------------------------------------------------------------------------+
 | PERCENTILES_TO_CHART                      | The list of response time percentiles for response time chart                        |
 +-------------------------------------------+--------------------------------------------------------------------------------------+
+| PERCENTILES_TO_STATISTICS                 | The list of response time percentiles for response time statistics                   |
```
cyberw, line 206: Suggest you change this to "list of response time percentiles to show in the statistics table"

cyberw: instead maybe add something about this and PERCENTILES_TO_CHART being UI settings (just to make it clear)

locust/test/test_util.py
```diff
@@ -30,3 +30,8 @@ def test_rounding_up(self):
         self.assertEqual(4, proper_round(3.5))
         self.assertEqual(5, proper_round(4.5))
         self.assertEqual(6, proper_round(5.5))
+
+    def test_normalize_decimal(self):
+        self.assertEqual(1, normalize_decimal(1))
+        self.assertEqual(0.99, normalize_decimal(0.99))
+        self.assertEqual("a", normalize_decimal("a"))
```
andrewbaldwin44, line 37: What is this function meant to do? From these tests it seems like the function is simply returning the inputted value :)

andrewbaldwin44: But I'm of the opinion that we don't need to continue supporting the legacy UI, so then maybe this function isn't needed either?

andrewbaldwin44:
In `main.py`, we have some validation to ensure the provided `PERCENTILES_TO_CHART` is correct:
```
# line 133
for percentile in stats.PERCENTILES_TO_CHART:
        if not is_valid_percentile(percentile):
            logging.error(
                "stats.PERCENTILES_TO_CHART parameter need to be float and value between. 0 < percentile < 1 Eg 0.95\n"
            )
            sys.exit(1)
```
Perhaps we should have this same validation for `PERCENTILES_TO_STATISTICS`?

FooQoo:
Added a Column Control feature to the Web UI in a PR. Would love your feedback on it!

image

locust/test/test_util.py
FooQoo: @andrewbaldwin44 
I agree that it's probably best not to continue adding new implementations to the legacy UI. 
Regarding the features we've already implemented in the legacy UI, should I remove them?

andrewbaldwin44: In my opinion, I would prefer to remove them. As we are right now, we can consider the legacy UI stable. If we continue to add new features it always has the possibility of introducing bugs. In the interest of not having to support the old UI anymore, I would prefer to leave the code as it is. Let me know if that's ok with you :)

locust/webui/src/components/ViewColumnSelector/ViewColumnSelector.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    );
+
+    const button = screen.getByRole('button');
+    fireEvent.click(button);
+
+    // Initial state check: 'column1' should be on and 'column2' should be off
+    const switch1 = screen.getByLabelText('Column 1');
+    const switch2 = screen.getByLabelText('Column 2');
+    // expect(switch1).toBeChecked();
+    // expect(switch2).not.toBeChecked();
```
andrewbaldwin44, line 49: Should these expects be commented out?

andrewbaldwin44:
Awesome work with the column selector! The code and the tests look really great! :) Just a suggestion - what if we had the `useSelectViewColumns` hook return a new table structure with the hidden columns filtered out? This way we wouldn't have to change the table component

locust/test/test_util.py
FooQoo: OK, I've removed the new features from the legacy UI.
[0cb9a62](https://github.com/locustio/locust/pull/2550/commits/0cb9a62e3beebbf46c8218f593a6c7779bdfa29f)

FooQoo:
@andrewbaldwin44 
Thanks for the feedback and the suggestion! 
I've updated the useSelectViewColumns hook to include the filterStructure function. 

Now it takes a structure argument and returns a new table structure with only the visible columns, filtering out the hidden ones. 
This should indeed help us avoid modifying the table component. Let me know what you think!

[6458cde](https://github.com/locustio/locust/pull/2550/commits/6458cdedb518d9d4b6ff527d51337d896130ac63)

andrewbaldwin44:
Looks awesome, nice work! :100: @cyberw Good with you to merge?

PR #2551, opened 2024-01-14
Ruff set python version and enable more rules

tdadela:
1. Set python target-version for ruff.
2. Enable FA102 rule (Checks for uses of PEP 585- and PEP 604-style type annotations in Python modules that lack the required from __future__ import annotations import for compatibility with older Python versions).
3. Enable pyupgrade rules.
4. Make code compatible with rules.

### DIFF
--- a/locust/user/task.py
+++ b/locust/user/task.py
@@ -423,10 +423,8 @@ class TaskSet(metaclass=TaskSetMeta):
             return random.randint(self.min_wait, self.max_wait) / 1000.0
         else:
             raise MissingWaitTimeError(
-                "You must define a wait_time method on either the %s or %s class"
-                % (
-                    type(self.user).__name__,
-                    type(self).__name__,
+                "You must define a wait_time method on either the {} or {} class".format(
+                    type(self.user).__name__, type(self).__name__
                 )
             )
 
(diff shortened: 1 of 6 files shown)


locust/user/task.py
```diff
@@ -423,10 +423,8 @@ class Tasks(TaskSet):
             return random.randint(self.min_wait, self.max_wait) / 1000.0
         else:
             raise MissingWaitTimeError(
-                "You must define a wait_time method on either the %s or %s class"
-                % (
-                    type(self.user).__name__,
-                    type(self).__name__,
+                "You must define a wait_time method on either the {} or {} class".format(
```
cyberw, line 426: Maybe use regular f-string here too?

cyberw:
Nice! One small comment, then it is good to go.

locust/user/task.py
tdadela: Done

tdadela:
What do you think about also enabling isort-like "unsorted-imports (I001)" rule?
(https://docs.astral.sh/ruff/rules/unsorted-imports/)

cyberw:
> What do you think about also enabling isort-like "unsorted-imports (I001)" rule? (https://docs.astral.sh/ruff/rules/unsorted-imports/)

Sounds good! I tried using isort a little, but in some cases (locustfiles) we need to ensure `locust` is imported before other things (because of gevent monkeypatching). If that can be solved then great!

PR #2552, opened 2024-01-15
Fix pedantic typo in the index.html template

marcelofern:
1 liner

### DIFF
--- a/locust/templates/index.html
+++ b/locust/templates/index.html
@@ -279,7 +279,7 @@
                 and made it work distributed across multiple machines.
             </p>
             <p>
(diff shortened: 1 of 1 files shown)


cyberw:
Nice! But why the comma before the "and"?

cyberw:
nvm, I'll fix it some more myself. thanks for bringing it to my attention :)

marcelofern:
> Nice! But why the comma before the "and"?

@cyberw that's an Oxford comma: https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/

PR #2554, opened 2024-01-15
Ruff enable isort

tdadela:
Enable ruff isort-like import sorting & make code compatible with them.

Alternative solutions:
1. `from .user.users import HttpUser, User  # isort: skip` in `locust/__init__.py` instead of
```py
if TYPE_CHECKING:
    from . import User
```
   in debug.py, dispatch.py, runners.py, shape.py.
2. [Custom Sections and Ordering](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html) instead of excluding folders (pyproject.toml).


cyberw:
looks nice! I guess there is no way to actually lint the examples but "prioritize" imports from locust - that would be even better...

cyberw:
Superseeded by #2556

PR #2556, opened 2024-01-16
Ruff enable isort (with custom sections)

tdadela:
Enable ruff isort-like import sorting & make code compatible with them.

Alternative approach to #2554.


locust/shape.py
```diff
@@ -1,11 +1,14 @@
...
+
 import time
-from typing import ClassVar
 from abc import ABCMeta, abstractmethod
+from typing import TYPE_CHECKING, ClassVar
 
-from . import User
 from .runners import Runner
 
+if TYPE_CHECKING:
```
cyberw, line 9: is this needed, or just a remnant of the first solution?

locust/runners.py
```diff
@@ -49,9 +51,9 @@
     StatsError,
     setup_distributed_stats_event_listeners,
 )
-from . import argument_parser
 
 if TYPE_CHECKING:
+    from . import User
```
cyberw, line 56: did this have to be moved in?

cyberw:
Awesome! Can you just have a look at my comments and maybe test run one or two of the example locustfiles to see that nothing broke there?

tdadela:
There are two independent problems to solve: circular imports and importing locust first in locustfiles.

1. Circular imports in `locust/locust` folder.
(The most important file is `locust/locust/__init__.py`).
    Python executes imports (and other expression) from top to down a file.
    Let's consider the following scenario:
    After sorting imports in `locust/locust` folder (without adding `if TYPE_CHECKING:` and moving some imports here) the `from .user.users import HttpUser, User` statement of the `locust/locust/__init__.py` is located in [line 25](https://github.com/locustio/locust/pull/2556/files/f22ad36015836beb22e777cc854f30a5da87a970#diff-6aef9e9db77131e7dac87e0941b14fc0487b43e95c8562626210d58385eef80bR25), below for example `from .shape import LoadTestShape` ([line 21](https://github.com/locustio/locust/pull/2556/files/f22ad36015836beb22e777cc854f30a5da87a970#diff-6aef9e9db77131e7dac87e0941b14fc0487b43e95c8562626210d58385eef80bR21))
    Let's consider following the one-line script:
    ```py 
    from locust import LoadTestShape
    ```
    `from locust import LoadTestShape` makes interpreter try to import LoadTestShape from `locust/locust/__init__.py` → interpreter opens `locust/locust/__init__.py` and reaches line 21 → opens `shape.py` file → reaches `from . import User` line BEFORE definition of LoadTestShape → `from . import User` means "import User from file `locust/locust/__init__.py`"
    However, it's not possible to import User because User is in line 25 and interpreter currently process line 21.
    I see three solutions to this problem:
    1. Making `from .user.users import HttpUser, User` the first import after `monkey.patch_all()`. It's possible thanks to the use of `# isort: skip`:
    ```py
    from .user.users import HttpUser, User  # isort: skip
    ```
    2. Adding `if TYPE_CHECKING` in some files.
    3. Using `from .user.users import User` instead of `from . import User` in some files.

2. In locustfiles Locust have to be imported first (due to gevent Monkey Patching).
    I see two solutions to this problem:
    1. Not sorting imports in locustfiles (like in #2554).
    2. Defining custom sorting section and custom selection order (see this PR – changes in [pyproject.toml](https://github.com/locustio/locust/pull/2556/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R78), [documentation isort](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html), [documentation ruff](https://docs.astral.sh/ruff/settings/#isort-section-order)).

locust/shape.py
tdadela: Needed, see this [comment](https://github.com/locustio/locust/pull/2556#issuecomment-1902687243).

locust/runners.py
tdadela: Yes (or it depends), see this [comment](https://github.com/locustio/locust/pull/2556#issuecomment-1902687243).

cyberw:
I see! I think option "1i" and "2ii" make the most sense - does that sound reasonable? Can you adjust your second PR? (or make a third one, if you prefer)

cyberw:
Actually. Maybe the `if TYPE_CHECKING` makes more sense, as it shows the intent. I'll merge this as is, if you dont have anything more to add? (thanks for the explanations)

tdadela:
i, ii, iii (and any combination of them) looks good to me.

cyberw:
Huh. The build is failing after merge. Can you have a look? https://github.com/locustio/locust/actions/runs/7621650409/job/20758349498

cyberw:
nvm, I fixed it myself

tdadela:
(PR wasn't up-to-date with master branch. Some imports were added in the meantime.)

PR #2559, opened 2024-01-18
Allow User weight adjustment (and task selection) in UI when running with --class-picker, or on command line with --user-config argument

andrewbaldwin44:
- Extend the class picker in the web UI to allow editing any property on a locust user
- Add a new property to the environment, `available_user_tasks`. This will allow for editing which tasks will be ran
- Add a new parser argument `--user-config`. This argument allows for passing a JSON object or file to configure all properties on a locust user
- Added tests to protect functionality in locust and the web ui
image
image


cyberw:
Looks amazing. Can you add some documentation in configuration.rst?

andrewbaldwin44:
Sorry had to fix a weird hanging test, I ran it a couple times locally using `act` to be sure it works now 👍

andrewbaldwin44:
Tests are good now, sorry about that :)

cyberw:
Great stuff!

Issue #2522, opened 2023-12-27
"Sticky" user types on workers


mgor:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

I'm working on a feature where it would be possible to "stick" a certain user type to a (set of) workers. 

The use case being a user type I have that depends on some native libraries, and said libraries have problem if with in the same process one want to run two different contexts (username + certificate).

The tought is to be able to "tag" a user type, e.g.
```python
class User(metaclass=UserMeta):
  ...
  sticky_tag: Optional[str] = None
  ...
```

This would then be used in `UsersDispatcher::_add_users_on_workers`:

```python
        for user in self._user_generator:
            if not user:
                self._no_user_to_spawn = True
                break
            sticky_tag = self._sticky_users_map.get(user, None)
            worker_node = next(self._worker_node_generator)

            if sticky_tag is not None:
                sticky_worker_id = self._sticky_workers_map.get(sticky_tag, None)
                if sticky_worker_id is None:
                    sticky_worker_id = worker_node.id
                    self._sticky_workers_map.update({sticky_tag: sticky_worker_id})
                else:
                    while worker_node.id != sticky_worker_id:
                        worker_node = next(self._worker_node_generator)
            elif len(self._sticky_workers_map) > 0:  # @TODO: do we need this?
                sticky_worker_ids = self._sticky_workers_map.values()
                while worker_node.id in sticky_worker_ids:
                    worker_node = next(self._worker_node_generator)
            # else: works as before
```

This is right now only a thought experiment, but I wanted feedback whether this was something that would be accepted by the community.

Also, outstanding issues:
- [ ] how does the `while`-loop affect balacing of users
- [ ] how to spread a user type over more than one worker
- [ ] should a non-sticky user type be "allowed" on a worker that has sticky user types
- [ ] ...

I know the dispatcher code is... kind of sacred? ;)

cyberw:
The dispatcher code is not "sacred", but it was contributed by @mboutet and I dont know if he's still around? (so I'm hesitant to change it)

While the dispatcher implementation does seem to hold up for most use cases it is "less deterministic" than the old one and doesnt scale as well with high user, worker and user type counts.

[...]

PR #2568, opened 2024-01-29
User type centric dispatcher

mgor:
A new dispatcher, where applicable, is compatible with the "old" `UserDispatcher` when `fixed_count` is used. The old dispatcher has been renamed `WeightedUserDispatcher`, which also is used as the default one.

The new `FixedUsersDispatcher` main difference is that target user count is a `dict` with the number of user of each type, and not the target total user count.

Another feature is being able to assign user types a [sticky] tag, which will be used to pin certain user types to a set of workers, when running distributed.

The tests in `test_dispatcher.py` has been modified to test both implementations, unless the test case was explicitly testing weights.

For now, there's no way to select which dispatcher to use other than programmatically.

We have been running with the `FixedUsersDispatcher` in our internal project for almost a month, without any major issues.


cyberw:
I'll get back to you, dont have time to review right now, but I havent forgotten :)

locust/runners.py
```diff
@@ -1206,7 +1230,7 @@ def on_report_to_master(client_id: str, data: dict[str, Any]):
         self.environment.events.report_to_master.add_listener(on_report_to_master)
 
         # register listener that sends quit message to master
-        def on_quitting(environment: Environment, **kw) -> None:
+        def on_quitting(environment: Environment, **kw: Any) -> None:
```
cyberw, line 1233: Is the Any needed? Type hinting stuff that the method never uses is kinda meaningless..

locust/runners.py
```diff
@@ -750,13 +772,15 @@ def start(
         self.spawn_rate = spawn_rate
 
         if self._users_dispatcher is None:
-            self._users_dispatcher = UsersDispatcher(
+            self._users_dispatcher = self.environment.dispatcher_class(
                 worker_nodes=list(self.clients.values()), user_classes=self.user_classes
             )
 
+        _user_count = self._users_dispatcher.get_target_user_count() if isinstance(user_count, dict) else user_count
```
cyberw, line 779: can you think of a better name than `_user_count`? This code is very hard to read.

locust/stats.py
```diff
@@ -84,7 +84,7 @@ class StatsErrorDict(StatsBaseDict):
 
 class StatsHolder(Protocol):
     name: str
-    method: str
+    method: str | None
```
cyberw, line 87: Why did this become optional?

locust/runners.py
```diff
@@ -481,7 +493,9 @@ def _start(self, user_count: int, spawn_rate: float, wait: bool = False, user_cl
             self.environment._filter_tasks_by_tags()
             self.environment.events.test_start.fire(environment=self.environment)
 
-        if wait and user_count - self.user_count > spawn_rate:
+        _user_count = sum(user_count.values()) if isinstance(user_count, dict) else user_count
```
cyberw, line 496: Same issue here with the variable naming and readability.

locust/main.py
```diff
@@ -589,7 +594,8 @@ def start_automatic_run():
         elif not options.worker and not environment.shape_class:
             logger.info("No run time limit set, use CTRL+C to interrupt")
 
-    if options.csv_prefix:
+    # `stats_csv_writer` is initialized with a StatsCSVFileWriter object if options.csv_prefix is "True"
```
cyberw, line 597: type hinting this makes it clear how confusing it is. maybe add a comment that it should be fixed some time?

locust/main.py
```diff
@@ -356,7 +360,7 @@ def kill_workers(children):
     locustfile_path = None if not locustfile else os.path.basename(locustfile)
 
     environment = create_environment(
-        user_classes,
+        selected_user_classes,
```
cyberw, line 363: why the rename?

locust/test/test_dispatch.py
```diff
@@ -3612,61 +3781,81 @@ class User5(User):
...
                                 self.assertEqual(
-                                    users_dispatcher._get_user_current_count(user_class.__name__),
+                                    users_dispatcher.get_current_user_count(user_class.__name__),
                                     user_class.fixed_count,
                                 )
 
 
-class TestRampUpDifferentUsers(unittest.TestCase):
-    def test_ramp_up_different_users_for_each_dispatch(self):
+@parameterized_class(PARAMETER_DISPATCHERS)
```
cyberw, line 3856: I think it makes more sense to duplicate this test rather than parameterizing it. Too many if statements for it to make sense as parameterized.

locust/test/test_dispatch.py
```diff
@@ -2214,15 +2335,17 @@ class User2(User):
...
         worker_node2 = WorkerNode("2")
 
         worker_nodes = [worker_node1, worker_node2]
 
-        users_dispatcher = UsersDispatcher(worker_nodes=worker_nodes, user_classes=user_classes)
+        users_dispatcher = self.user_dispatcher_class(worker_nodes=worker_nodes, user_classes=user_classes)
 
         # user count = 1
-        users_dispatcher.new_dispatch(target_user_count=1, spawn_rate=1)
+        users_dispatcher.new_dispatch(target_user_count=target_user_count(1), spawn_rate=1)
```
cyberw, line 2348: Uh. What? `target_user_count(1)` instead of just `1`? I dont understand it and I'm pretty sure if I did make the effort to understand it I wouldnt like it :)

cyberw:
Hmm.... I'd like to see an example locustfile using this as well as a mention of this in the documentation. And could you separate the type checks that you have added into a separate PR? 

Would this work together with the newly added `--config-users`/weight setting in the UI?

mgor:
I'll be gone on vacation for a week, I'll get back regarding your comments after that 🙂

cyberw:
Hmm. I only now got around to reviewing the actual dispatch code. Can you do something about the diff, it is kinda unreadable as is. When I think about it - does this dispatcher actually change anything if all users have fixed_count instead of weights? (because that is what I meant could be improved in the discussion we had before)

I think it is unlikely that we'll be able to merge this. Its just too much code for a niche case...

mgor:
Before spending time on resolving the comments, I think we should resolve this first :)

> When I think about it - does this dispatcher actually change anything if all users have fixed_count instead of weights? (because that is what I meant could be an improvement, in the discussion we had before)

Dispatch-logic-wise, without using the new `sticky_tag` there's not really a big difference. Usage wise the big difference is that you don't specify a total number of users to spawn, but rather you specify per user-type how many you want to spawn. Which was what I thought you meant with in #2522:

>  Not based on specifying weights and a total user count but on per-User type-numbers (like most other load test tools do it).

locust/stats.py
mgor: The `Total` / aggregated statistics is created with method set to `None`.
https://github.com/locustio/locust/blob/2ec2e70b18e98eb99b118b70e4229bfa04057710/locust/stats.py#L213
It was needed to solve a type issue for some of the sorting logic.

However, I see now that it should then probably be changed in `StatsEntry` as well, then,
https://github.com/locustio/locust/blob/2ec2e70b18e98eb99b118b70e4229bfa04057710/locust/stats.py#L291

cyberw:
Tbh, I dont have time to review this again, sorry. The diff is just too big.

I did have a short look and notice that it solves things in a way I wouldnt want anyway: Tagging workers at startup (due to different types of workers etc) might make sense, but having the master assign tags arbitrarily is an even more narrow use case.

Perhaps your underlying need could be solved by something else, like passing custom messages?

> Not based on specifying weights and a total user count but on per-User type-numbers (like most other load test tools do it).

Now I'm thinking this is probably solved best using `fixed_count` on workers and maybe adding a special value for `--users` that spawns the total fixed count automatically, so it doesnt need a special dispatcher. This was all I really wanted (except for the fact that the current dispatcher is a little slow with really large numbers of user types and worker count). Sorry if I wasted your time...

mgor:
Hm, ok.

Reason for the large diff is due to trying to avoid as little duplicated logic as possible, but sure, could probably rearrage it a bit.

Not sure what you mean by this:
> but having the master just assign tags arbitrarily is an even more rare use case.

We'll see how we can solve it then.. since we need a way to seperate/isolate user types.

cyberw:
A somewhat ugly way of working around could be by checking something like `self.runner.worker_id` in the task and having the workers behave differently (possibly just sleeping) depending on which worker we are on.

mgor:
I rather not :)

Would you accept a PR that makes the dispatcher logic more pluggable? E.g. specifying a dispatcher class in `Environment`?

mgor:
superseeded by #2606.

PR #2569, opened 2024-01-29
Modern UI as Default

andrewbaldwin44:
### Changes implemented
- Updated screenshots to be screenshots of the new UI
- Changed the `--modern-ui` flag to be `--legacy-ui` and updated the relevant tests to use this flag
- Updated the docs


cyberw:
Nice! Why are some of the tests run with legacy UI though? Ideally they should work in both, of course, but it is most important that they work in modern-ui :)

andrewbaldwin44:
Yeah it's because they are testing things using the legacy UI templates or the legacy UI HTML template. Even though we are not developing the legacy UI anymore, I figured we should keep them to ensure new developments to the modern UI don't cause regressions

cyberw:
Yea, testing some things with the old UI totally makes sense. But why wouldnt we want to test `--autostart` with modern UI?

andrewbaldwin44:
Oh the `--autostart` flag is testing the modern UI, it uses the modern UI by default now :) if it were not using the modern UI it would need the `--legacy-ui` flag

cyberw:
But test_autostart_wo_run_time is only run against legacy UI?

andrewbaldwin44:
Yeah you're right, but I believe it's because of this line: `self.assertIn('<body class="running">', response.text)`, which would fail in the modern UI. If you prefer I could change this so it just tests the modern UI, or I can add an additional test so it tests the modern UI as well

cyberw:
Hows about you change the test so that it works with modern UI but add a ”legacy” test (one of the more basic ones) that run with legacy ui?

andrewbaldwin44:
Ok I updated the autorun tests. They don't run against the legacy UI anymore, but I think these should be sufficient since we're not really testing the UI, but rather the output of the template arguments to ensure the state is set as "running"

PR #2570, opened 2024-01-29
docs(running-distributed): cores -> logical cores

tdadela:
“Cores” usually mean physical cores. I think adding “logical” makes the sentence more unambiguous.


cyberw:
Lgtm. Wont merge right now though, I want to make a clean release in a while with no other changes except for switching to modern UI by default.

PR #2571, opened 2024-01-29
docs(PoolManager): requests queuing

tdadela:


cyberw:
I think the block=True already gets the point across :)

tdadela:
After reading Locust docs for the first time, I googled for urllib docs because I wasn't sure.
(Of course, I don't know, maybe it's unclear just for me.)

cyberw:
If it could be rephrased into a single sentence, it could be an improvement. Overly verbose documentation isnt good either.

PR #2573, opened 2024-01-30
Fix Spwaning State Never Updates to Running

andrewbaldwin44:
- When the "spawning" state in locust changes to "running", the web UI never updates. This fixes this issue.
- Show user count while spawning or running


andrewbaldwin44:
@cyberw do you remember a flicker issue that was happening at some point, where if you reloaded the page it would briefly show the form for starting a test instead of the tabs? I'm sure I had a reason for not always setting updating the state but as of right now I can't see why

cyberw:
Yea, I think I remember something like that, but not 100%...

andrewbaldwin44:
Well we need to update the state so it doesn't always show as spawning. I tested with autostart, stopping, starting, editing, etc. and it all seems to work smoothly

PR #2583, opened 2024-02-05
Locustfile distribution from master to worker

cyberw:
Make workers get locustfile directly from master when locustfile argument is set to `-`


locust/argument_parser.py
```diff
@@ -201,9 +207,74 @@ def parse_locustfile_option(args=None) -> list[str]:
...
+        "--master-port",
+        type=int,
+        default=5557,
+        env_var="LOCUST_MASTER_NODE_PORT",
+    )
 
     options, _ = parser.parse_known_args(args=args)
 
+    if options.locustfile == "-":
+        # having this in argument_parser module is a bit weird, but it needs to happen early on
```
DennisKrone, line 236: Maybe i would have put this in a separate function just to prevent some bloat/make it more "digestible"

locust/runners.py
```diff
@@ -1033,6 +1033,31 @@ def client_listener(self) -> NoReturn:
                 # emit a warning if the worker's clock seem to be out of sync with our clock
                 # if abs(time() - msg.data["time"]) > 5.0:
                 #    warnings.warn("The worker node's clock seem to be out of sync. For the statistics to be correct the different locust servers need to have synchronized clocks.")
+            elif msg.type == "locustfile":
+                logging.debug("Worker requested locust file")
+                assert self.environment.parsed_options
+                filename = (
```
DennisKrone, line 1039: is this just in case the user didn't put .py at the end? Feels like this maybe should be done in the argument parser if possible

locust/test/test_main.py
```diff
@@ -2057,3 +2202,49 @@ def mytask(self):
                 self.assertEqual(status_code, 42)
             self.assertNotIn("Traceback", master_stderr)
             self.assertIn("failed to send heartbeat, setting state to missing", master_stderr)
+
+
+#     def test_interrupt_in_custom_message_handler(self):
```
DennisKrone, line 2207: A test that doesn't work with the new stuff? I'm guessing needs to be removed or rewritten?

DennisKrone (commented):
I'm kinda surprised that it works!

locust/runners.py
cyberw: The default value for locust file is "locustfile" (but gets expanded to "locustfile.py" implictly) which is why I support this. I _could_ change it to just "add .py if there is no .py at the end" which would support if someone says "locust -f module_name_without_py_at_the_end". Specifying a bare module name and then having the file transfered anyway, might be nice, but it is kinda weird, and exception logged later on (line 1048) might become confusing if locust just adds .py automatically. What do you think?

locust/test/test_main.py
cyberw: actually that was just accidentally commited. I'll remove it.

locust/argument_parser.py
cyberw: better now?

PR #2586, opened 2024-02-06
Fix detection of locustfile being named locust.py

cyberw:


cyberw:
fixed the underlying issue instead...

PR #2595, opened 2024-02-09
Allow getting locust files from http urls

DennisKrone:
-f will now accept http urls as locustfiles (for example from a remote git repo).


locust/test/test_load_locustfile.py
```diff
@@ -209,3 +211,18 @@ def test_locustfile_can_be_set_in_config_file(self):
                 ]
             )
             self.assertEqual("my_locust_file.py", options.locustfile)
+
+    def test_locustfile_from_url(self):
+        locustfiles = parse_locustfile_option(
```
cyberw, line 216: Why not call parse_options instead? (as that tests more). Not a big issue though...

locust/argument_parser.py
```diff
@@ -163,9 +163,8 @@ def download_file_from_url(url: str) -> str:
         sys.stderr.write(f"Failed to get locustfile from: {url}. Exception: {e}")
         sys.exit(1)
 
-    if response.ok:
-        with open(f"{tempfile.gettempdir()}/{url.rsplit('/', 1)[-1]}", "w") as locustfile:
-            locustfile.write(response.text)
+    with open(f"{tempfile.gettempdir()}/{url.rsplit('/', 1)[-1]}", "w") as locustfile:
```
cyberw, line 166: I think building the path can be expressed nicer using something like 

`os.path.join(empfile.gettempdir(), url.rsplit('/', 1)[-1])`

cyberw: I'll fix this in master :)

cyberw:
I wanna get 2.23 out there so I merged it :)

PR #2601, opened 2024-02-11
Modern UI: Add exception message column on the Exception Table

### LINKED ISSUE #2597: Modern UI: Add exception message column on the Exception Statistics
### Prerequisites

### Description

I the old UI it was much easier to find reason of exception:
image

In the new UI it's a bit hard, exceptionally when there are a lot of exception with the same exception class but different description:
image

andrewbaldwin44:
Fixes #2597

This adds the message column to the Exceptions Table. It will be visible in both the UI and in the HTML report
image


cyberw:
Cool stuff. We really need to work on building the web assets as part of the package build process instead of committing them :) Either by going back to setup.py or trying out pdm...

PR #2606, opened 2024-02-16
pluggable dispatcher logic

mgor:
superseeding #2568, coming from discussion in #2522.

this PR makes it possible to implement custom dispatcher logic and specifying the dispatcher class when creating an `Environment`.

a little unsure and not really satisfied with the name of the "interface" `UserDispatcherType`.
played around with `UserDispatcherBase` (but it's [discouraged](https://softwareengineering.stackexchange.com/questions/422970/does-using-the-word-base-in-a-class-name-indicate-abstraction)to use `Base`). `IUserDispatcher` doesn't feel pythonic. so any suggestion on a better name is welcomed :) my top choice would be to rename `UserDispatcher` to `WeightedUserDispatcher` (or similar) and have the interface changed to `UserDispatcher`, but at the same time I wanted to make as few changes as possible.

some methods and properties in the interface was added to make it easier to [compability] test implementations.

so minor changes in the current dispatcher logic so it uses the interface specified methods and properties.

### DIFF
--- a/locust/dispatch.py
+++ b/locust/dispatch.py
@@ -4,10 +4,9 @@ import contextlib
 import itertools
 import math
 import time
-from abc import ABCMeta, abstractmethod, abstractproperty
 from collections import defaultdict
 from operator import attrgetter
-from typing import TYPE_CHECKING, Generator, Iterator, overload
+from typing import TYPE_CHECKING, Generator, Iterator
 
 import gevent
 from roundrobin import smooth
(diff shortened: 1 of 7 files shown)


cyberw:
Idk, this still changes/adds a lot of code. Could you just subclass UsersDispatcher and override its methods insteas of introducing an abstract base class? (and not rename them. tbh I'm not sure about the reasoning for making them all private (_ prefix) but I'd rather not change it)

mgor:
sure, update with the only change in `locust.dispatch` being some type annotations coming soon.

changed the PR to a draft, since I'd like to test them out in our project first, which uses the branch in my fork that this PR also is based on.

fridgerator:
@mgor  Is there any documentation available to use this PR to allow sticky user types per worker?  Thanks!

mgor:
> @mgor  Is there any documentation available on how to use this code to allow sticky user types per worker?  Thanks!

This PR only allows custom dispatcher logic to be used, the actual logic that would allow for sticky user types is outside of locust. 

The implementation I'm using can be found here,

https://github.com/Biometria-se/grizzly/blob/158e715acd825070e62ddcd052da6f2d241d6197/grizzly/locust.py#L88

fridgerator:
Great, thanks for the response, cheers!

PR #2609, opened 2024-02-22
[Report][Modern-UI] HTML report is blank

### LINKED ISSUE #2603: [Report][Modern-UI] HTML report is blank
### Prerequisites

### Description

Run a test then open the HTML report -> Noticed that it is blank
Note: This bug occurs from 2.22.0, and did not occur on 2.21.0 
image

### Command line

locust -f SimpleWeb.py -u 100 -r 10 -t 30s --html=samplelocust.html

### Locustfile contents

```python3
from locust import FastHttpUser, HttpUser, between, constant_pacing, events, task
from loguru import logger


class QuickstartUser(FastHttpUser):
    wait_time = between(2, 5)


    host = "http://127.0.0.1:5000"
    # begin = time.time()

    @task()
    def get_tasks_1(self):
        res = None
        try:
            payload = {}
            headers = {"Cache-Control": "max-age=0, no-cache, no-store, must-revalidate"}
            res = self.client.get("/api/tasks", headers=headers, data=payload, name="Get Tasks")
        except Exception as exception:
            logger.error(exception)

    @task()
    def post_lewin(self):
        try:
            payload = {}
            headers = {"Cache-Control": "max-age=0, no-cache, no-store, must-revalidate"}
            self.client.post("/api/lewin", headers=headers, data=payload, name="Post Lewin")
        except Exception as exception:
            logger.error(exception)
```

### Python version

3.9.18

### Locust version

2.23.1

### Operating system

macOS 14.2.1 (23C71)

andrewbaldwin44:
Fixes #2603 

### Proposal
- I managed to get an error message that stated the API middleware was not properly setup, so I added react toolkits API middleware to the redux store
- Added an ErrorBoundary, which will allow us to catch any exceptions should they occur and then create an issue
image
- I needed to check that the auth page was still working as expected and I fixed the example


cyberw:
nice! is this something we could/should unit test?

andrewbaldwin44:
@cyberw We _could_ test the FallbackRender, but I'm not sure it's really worth it since it wouldn't really achieve anything other than testing a 3rd party (`react-error-boundary`)

Issue #2298, opened 2023-02-16
pyproject.toml support for locust configuration


johnnymetz:
### Is your feature request related to a problem? Please describe.

My project includes a lot of small configuration files. Many Python packages allow configurations to be included in `pyproject.toml` which helps reduce this config clutter and promotes consistency. See [PEP 518](https://peps.python.org/pep-0518/). Some of these packages include:

- [black](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#black-compatible-configurations)
- [isort](https://pycqa.github.io/isort/docs/configuration/config_files.html#pyprojecttoml-preferred-format)
- [pytest](https://docs.pytest.org/en/7.2.x/reference/customize.html#pyproject-toml)
- [mypy](https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file)

### Describe the solution you'd like

Support configuration in `pyproject.toml`. Would be great to look for the config there by default (the locust config files can still take precedence) so users don't have to pass `--config=pyproject.toml` - other projects support this.

### Describe alternatives you've considered

None

### Additional context

I'm happy to work on this is needed.

cyberw:
What would be required on locust side to support this?

alzex3:
@cyberw Hello! First, I want to thank you for maintaining Locust; it has been very helpful in my work. It would be really cool to add the ability for Locust to parse configurations from `pyproject.toml` instead of requiring a separate config file.

I have checked the code. Changes would require modifying [argument_parser.py](https://github.com/locustio/locust/blob/cc5e373ab48f402918b8e6f5197d1c97b689074c/locust/argument_parser.py). The Locust arguments parser uses the [configargparse](https://github.com/bw2/ConfigArgParse) lib, which already has a built-in TOML parser, so there shouldn't be a lot of code changes.

I would be really glad to contribute this feature if it's welcome. I'm waiting for your response, thanks!

cyberw:
A PR would be most welcome!

PR #2612, opened 2024-02-24
Implement pyproject.toml support for Locust configuration

alzex3:
**Issue Addressed:**

This PR implements support for TOML files as configuration files and sets [pyproject.toml](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec) as the default. It addresses the #2298 issue.

**Implementation Details:**
1. Used the built-in `CompositeConfigParser` with `TomlConfigParser` and `DefaultConfigFileParser`, which allow the [configargparse](https://github.com/bw2/ConfigArgParse) lib to automatically parse options from defined `.conf` and `.toml` files. 
2. Added `pyproject.toml` to `DEFAULT_CONFIG_FILES`. 
3. Introduced a new project dependency - [toml](https://github.com/uiri/toml), which is required for the `configargparse` TOML parser to work. 
4. Added tests to check the correct parsing of `.conf` and `.toml` configuration files.
5. Reworked the [Configuration File](https://github.com/uiri/toml) documentation section.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -24,7 +24,8 @@ Locust is configured mainly through command line arguments.
 Environment Variables
 =====================
 
-Options can also be set through through environment variables. They are typically the same as the command line argument but capitalized and prefixed with ``LOCUST_``:
+Options can also be set through environment variables. They are typically the same as the command line argument
+but capitalized and prefixed with ``LOCUST_``:
 
 On Linux/macOS:
 
@@ -44,17 +45,23 @@ On Windows:
 Configuration File
 ==================
 
-Options can also be set in a configuration file in the `config file <https://github.com/bw2/ConfigArgParse#config-file-syntax>`_
-format. 
+Options can also be set in a configuration file in the
+`config or TOML file format <https://github.com/bw2/ConfigArgParse#config-file-syntax>`_.
 
-Locust will look for ``~/.locust.conf`` and ``./locust.conf`` by default, and you can specify an 
-additional file using the ``--config`` flag.
+Locust will look for ``~/.locust.conf``, ``./locust.conf`` and ``./pyproject.toml`` by default.
+You can specify an additional file using the ``--config`` flag.
 
-Example:
+.. code-block:: console
 
-.. code-block::
+    $ locust --config custom_config.conf
 
-    # master.conf in current directory
+Here's a quick example of the configuration files supported by Locust:
+
+locust.conf
+--------------
+
+.. code-block:: ini
+
     locustfile = locust_files/my_locust_file.py
     headless = true
     master = true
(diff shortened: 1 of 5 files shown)


locust/argument_parser.py
```diff
@@ -23,7 +23,7 @@
 version = locust.__version__
 
 
-DEFAULT_CONFIG_FILES = ["~/.locust.conf", "locust.conf"]
+DEFAULT_CONFIG_FILES = ("pyproject.toml", "~/.locust.conf", "locust.conf")
```
cyberw, line 26: I think the order here matters, so put pyproject.toml last, so they are same order as they are documented :)

pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
     "pyzmq >=25.0.0",
     "geventhttpclient >=2.0.11",
     "ConfigArgParse >=1.5.5",
+    "toml >=0.10.2",
```
cyberw, line 20: Is the toml package really needed for configargparse to read toml files? The `toml` package look really old and unmaintained. If it is only needed to generate that test file you can just use a static text instead.

locust/argument_parser.py
cyberw: … or change the .rst..

pyproject.toml
alzex3: Yes, [toml](https://github.com/uiri/toml) is required for the configargparse TOMLParser to function. Although Python version 3.11 and later includes a built-in TOML library, it appears that configargparse is not actively maintained. If we want to avoid using toml, I can create our own TOMLParser, which will require [tomli](https://github.com/hukkin/tomli?tab=readme-ov-file#building-a-tomlitomllib-compatibility-layer) for Python versions earlier than 3.11.
Screenshot 2024-02-25 at 11 04 12
Screenshot 2024-02-25 at 11 55 38

locust/argument_parser.py
alzex3: The order is important, but it functions in the opposite way; the [configargparse](https://github.com/bw2/ConfigArgParse) reverses the list of configuration files internally.
Screenshot 2024-02-25 at 10 56 08

The DEFAULT_CONFIG_FILES are in the correct order, and I have defined the same in the updated documents.
`./pyproject.toml -> ~/.locust.conf -> ./locust.conf -> (file specified using --conf) -> env vars -> cmd args`.

pyproject.toml
cyberw: That sounds good, please do!

docs/configuration.rst
```diff
@@ -44,17 +45,23 @@ On Windows:
...
 ==================
 
-Options can also be set in a configuration file in the `config file <https://github.com/bw2/ConfigArgParse#config-file-syntax>`_
-format. 
+Options can also be set in a configuration file in the
+`config or TOML file format <https://github.com/bw2/ConfigArgParse#config-file-syntax>`_.
 
-Locust will look for ``~/.locust.conf`` and ``./locust.conf`` by default, and you can specify an 
-additional file using the ``--config`` flag.
+Locust will look for ``~/.locust.conf``, ``./locust.conf`` and ``./pyproject.toml`` by default.
```
cyberw, line 51: this is where I meant the order was confusing :)

locust/argument_parser.py
cyberw: I made another comment in the place where it was confusing. 

Anyways the global file should be overridden by local files so I'd like it to be ~/.locust.conf, ./locust.conf, ./pyproject.toml

docs/configuration.rst
cyberw: (but if you change it the way I suggested it _will_ be the right order :)

alzex3: Oh, now I see it. Thank you!

alzex3:
@cyberw Hello! I've replaced the toml lib and added a custom parser.

cyberw:
Awesome.

PR #2642, opened 2024-03-18
Dont throw an exception in FastHttpUser if a request is made with stream=True and the connection fails

### LINKED ISSUE #2640: `'NoneType' object has no attribute 'get'` when `stream=True` in `FastHttpSession.request`
### Prerequisites

### Description

I'm using the `request(..., stream=True)` in my task and occasionally, I get the following stack trace (slightly redacted):

```
'NoneType' object has no attribute 'get'
...
  File ".../locustfile.py", line XXX, in some_task
    with self.client.post("/some-path",
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/locust/contrib/fasthttp.py", line 284, in post
    return self.request("POST", url, data=data, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/locust/contrib/fasthttp.py", line 236, in request
    request_meta["response_length"] = int(response.headers.get("response_length") or 0)
```

The issue happens when there's a network connectivity problem, making `response` having no `response.headers` object.

Maybe a somewhat safer approach around the code
```
if stream:
    request_meta["response_length"] = int(response.headers.get("response_length") or 0)
``` 
in `FastHttpSession.request` is necessary?

### Command line

Executed via IDE Debugger (does not matter)

### Locustfile contents

```python3
class SimpleUser(FastHttpUser):
    @task(1)
    def some_task(self):
        with self.client.post("/some-path", stream=True) as resp:
            pass
```

### Python version

3.11

### Locust version

2.24.0

### Operating system

Linux

cyberw:
Fixes #2640

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -230,16 +230,18 @@ class FastHttpSession:
         if not allow_redirects:
             self.client.redirect_resonse_codes = old_redirect_response_codes
 
+        request_meta["response_length"] = 0  # default value, if length cannot be determined
+
         # get the length of the content, but if the argument stream is set to True, we take
         # the size from the content-length header, in order to not trigger fetching of the body
         if stream:
-            request_meta["response_length"] = int(response.headers.get("response_length") or 0)
+            if response.headers and "response_length" in response.headers:
+                request_meta["response_length"] = int(response.headers["response_length"])
         else:
             try:
-                request_meta["response_length"] = len(response.content or "")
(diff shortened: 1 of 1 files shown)


neiser:
@cyberw Thanks for that super quick fix. I was just about to open a PR also addressing this. Your code change does look more sound, though. I've another fix which I've already validated locally, see #2643

PR #2660, opened 2024-04-01
Use hasattr when processing parsed_options in Environment

mrunalkpatel96:
`hasattr` is safer and doesn't throw an exception if the attribute is missing from the `parsed_options` param when creating an instance of `Environment`.


cyberw:
Cool. But it would be a lot more concise using getattr with default value False, can you do that instead?

mrunalkpatel96:
> Cool. But it would be a lot more concise using getattr with default value False, can you do that instead?
> 
> And can you maybe add at test case that exposes the problem?

Sure -- thanks for the suggestion!

cyberw:
fixed in separate commit

Issue #2653, opened 2024-03-26
TaskSet's `_task_queue` should be a `collections.deque`?


bakhtos:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

While examining the source code to prepare the PR for #2651, I noticed that the `TaskSet` attribute `_task_queue` is a `list`, however insert/remove operations are performed both at the beginning and at the end.

Python has a dedicated [deque](https://docs.python.org/3/library/collections.html#collections.deque) for such operations, in particular:

_Deques support thread-safe, memory efficient appends and pops from either side of the deque with approximately the same O(1) performance in either direction.
Though [list](https://docs.python.org/3/library/stdtypes.html#list) objects support similar operations, they are optimized for fast fixed-length operations and incur O(n) memory movement costs for pop(0) and insert(0, v) operations which change both the size and position of the underlying data representation._

Should this be changed?

I could implement the change in the same pull request.

cyberw:
That makes a ton of sense to me. PR:s welcome!

cyberw:
Fixed in #2698

PR #2667, opened 2024-04-05
Update TaskSet to use deque

Hchanni:
Saw a comment in Issue #2653 and thought I'd give it a try, just updates the insertion to the first element to use `appendleft` and also updates the type of `_task_queue` to use the collections deque.

### DIFF
--- a/locust/user/task.py
+++ b/locust/user/task.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from locust.exception import InterruptTaskSet, MissingWaitTimeError, RescheduleTask, RescheduleTaskImmediately, StopUser
 
 import logging
+from collections import deque
 import random
 import traceback
 from time import time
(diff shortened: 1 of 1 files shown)


Hchanni:
Interesting, so I tried the test `python -m tox -e fail_fast_test_main` locally, and seems to fail even if I were to remove my changes. Any ideas?

cyberw:
Did you figure it out?

Hchanni:
> Did you figure it out?

Not yet, spinning up a new branch and messing with the tests, it’s weird cause the tests are failing even on a non altered master branch, for some reason the workers just don’t connect and it times our.

cyberw:
👍 let me know if you need help.

cyberw:
@Hchanni I tweaked your solution a little and made a new PR. Thanks for the inspiration :)

Hchanni:
Oh shoot thanks for getting the fix out @cyberw got wrapped up with some locust stuff at work haha and didn't get a chance to actually dig into it. I'll take a look at the PR I'm actually curious to see why mine was failing

PR #2669, opened 2024-04-07
Add an example that displays cache stats using the new UI.

### LINKED ISSUE #2644: Add an UI extension exemple that display proxy cache hit/miss
### Prerequisites

### Description

I just write a UI extension that count my CDN hit/miss per path.
@cyberw suggests on Slack that I should add this exemple to the code, so here I am 🙂.
I will push this as a PR soon, I will also add some details to the [extending-web-ui documentation](https://docs.locust.io/en/stable/extending-locust.html#extending-web-ui) base on my experience: migration to new ui, add deprecation on older examples, tips, ...

obriat:
Fixes #2644

Here's the result
ZgPof5tGyf

English is not my mother tongue so feel free to rewrite the documentation part.


docs/extending-locust.rst
```diff
@@ -123,9 +124,21 @@ as it involves also writing and including HTML and Javascript files to be served
...
 A working example of extending the web UI, complete with HTML and Javascript example files, can be found
-in the `examples directory <https://github.com/locustio/locust/tree/master/examples>`_ of the Locust
+in the `examples directory <https://github.com/locustio/locust/tree/master/examples/>`_ of the Locust
 source code.
 
+- `extend_modern_web_ui.py`: Display a table with content-length for each call.
+- `extend_modern_web_ui_cache_stats.py`: Display Varnish Hit/ Miss stats for each call. Could be easly
+extended to other CDN or cache proxies and gather other cache statistics such as cache age, control, ...  
+
+Since 2.22.0, the modern UI is the new default for the web UI, the `extend_web_ui` folder contains a
```
cyberw, line 134: We're deprecating the legacy UI shortly, you can just remove this from the docs.

obriat:
>     * I dont understand what this means (perhaps it is something that can be fixed?) `At the moment Locust does not show the corresponding menu entry on the default root path "/", however the menu entry appears properly if using the newly created path.`

The "Cache statistics" entry menu (`"extended_tabs": [{"title": "Cache statistics", "key": "name"}],`) only appears under the /extend route (`@extend.route("/extend")`), it takes me some time understand why it doesn't pop on the default path (`/`). It could save some time the new comers, no ?

obriat:
[cyberw](https://github.com/cyberw)
I add a image and some comments. I also check locally the doc rendering.

examples/extend_modern_web_ui_cache_stats.py
```diff
@@ -0,0 +1,209 @@
...
+        l.client.get("/")
+
+
+class WebsiteUser(HttpUser):
+    host = "http://httpbin.org"
+    wait_time = between(2, 5)
+    tasks = [MyTaskSet]
+
+
+# @see https://docs.acquia.com/acquia-cloud-platform/performance/varnish/headers
```
andrewbaldwin44, line 40: Maybe we can simply explain this in a comment or point to a blog post, rather than referencing a companies documentation?

examples/extend_web_ui/extend.py
```diff
@@ -9,9 +9,16 @@
 import os
 from html import escape
 from time import time
+from warnings import warn
 
 from flask import Blueprint, jsonify, make_response, render_template
 
+warn(
+    "This UI example is deprecated, please downgrade to Locust 2.21.0 or use --legacy-ui flag ; version=2.22.0",
```
andrewbaldwin44, line 17: Since this example would continue to work with the `--legacy-ui` flag, I think it would make more sense to suggest this as a solution rather than downgrading right?

obriat: "This UI example is deprecated, make sure to use --legacy-ui flag ; version=2.22.0" ?

examples/extend_modern_web_ui_cache_stats.py
obriat: True, I'll add a short comment, but I don't think that details about Varnish headers should be explain here.
I should also fix the link, here the correct one https://docs.varnish-software.com/tutorials/hit-miss-logging/

obriat:
>     * The docs do already state `You should now be able to start locust and browse to http://127.0.0.1:8089/added_page`. Reading this, the developer should already know to go to `/added_page`. The section is also called "Adding Web Routes". 
IMHO, the "Adding Web Routes" part is not clear enough, it doesn't mention that a menu entry  is created automatically with the path ans that it will only be visible on the new path along with other "core menu items" ( It's not unitive).
Shall I made a proposal in this PR?

>     * Since we already have an example showcasing how to extend the Locust UI, we could just call this example `web_ui_cache_stats`, or perhaps have these two extend examples together in a folder? We could then rename the old one to `legacy_extend_web_ui` for example
Ok for the first proposition.

examples/extend_web_ui/extend.py
cyberw: I've added a deprecation warning log when --legacy-ui is active and will probably remove it in 2.24, so lets not worry too much about it :)

cyberw:
>> `At the moment Locust does not show the corresponding menu entry on the default root path "/", however the menu entry appears properly if using the newly created path.`

>  I do however agree that it would make a lot more sense if the extended tabs / CSV files were available on the root path. I have some ideas on how this can be done that I will try to implement in a future ticket.

Considering the future plan of fixing this then I'm ok with the comment as is (it felt awkward, that's why I mentioned it)

cyberw:
Tests seemed to be flaky, so I re-ran them. I'm fine with merging now. Andrew?

andrewbaldwin44:
Looks good to me 👍

cyberw:
Thanks @obriat !

utipatos:
Hello, team. 

This example has issues when running tests with workers.  I suppose some workers update cache statistic extra time. 

Is it a known issue?

Pls take a look on the screenshots. Cache requests count is bigger than total:

image


image

obriat:
Oups, I’m pretty sure I tested it with workers, but there must be a bug, I’ll do some debugging to be sure.
Thanks for the report

utipatos:
sorry, @obriat. It might be some local issue when statistic page was updated with a little delay. Tried several times more and it wasn't reproduced for me anymore

obriat:
Ok, I think you got to wait for a complete stop of all workers to get consistent results, so all results are send to the UI.

PR #2684, opened 2024-04-19
Small cleanup: mypy type-ignore and dev deps

tdadela:
1. Removing deprecated mypy "# type: ignore" comments.
2. Pinning the same version of ruff in .pre-commit-config.yaml and tox.ini files.
3. Ruff 0.3.7 compatibility.


cyberw:
Gorgeous.

Issue #2683, opened 2024-04-19
Why defining run_time on worker node raises an error


Sawiq:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

Currently when worker node receives `run_time` parameter raises an error and it seems that stops working.

I thing that worker could simply ignore this parameter and at most log a warning. This would simplify configuration in case of 'local distributed' load, i.e. multiple worker processes on one machine for better CPU utilization.

Funnily enough it seems that this error is not raised, when `--config` option is used and file has non-standard name (like `worker.conf`)

Tested on v. 2.25.0 - latest on pip.

cyberw:
Yea, it really only made sense before config files were introduced. I'll change it to an info level log message.

PR #2685, opened 2024-04-20
Allow worker process to run with --run-time specified, just log a warning

Sawiq:
As discussed [here](https://github.com/locustio/locust/issues/2683)

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -433,8 +433,7 @@ See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-numb
 
     if options.run_time:
         if options.worker:
-            logger.error("--run-time should be specified on the master node, and not on worker nodes")
-            sys.exit(1)
+            logger.info("--run-time specified for a worker node will be ignored.")
         try:
             options.run_time = parse_timespan(options.run_time)
         except ValueError:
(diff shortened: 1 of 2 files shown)


cyberw:
Looks good apart from the type check fail (which probably isnt really your fault, but still needs to be fixed)

tdadela:
```py
    parts = timespan_regex.match(time_str)
    if not parts:
        raise ValueError("Invalid time span format. Valid formats: 20, 20s, 3m, 2h, 1h20m, 3h30m10s, etc.")
    parts = parts.groupdict()
```
The same variable name is used for storing two different types.
One of solutions is simply to rename **parts** variable.

Before this PR def parse_timespan(time_str): was untyped so mypy was skipping checking it: 
- "note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs"
- https://github.com/locustio/locust/blob/ca33b58ba9938c0ad68f9f2c9070da1da18b6c0a/pyproject.toml#L104

cyberw:
Yes that looks terrible. Renaming makes sense I think!

Sawiq:
I have removed reassignment and called `groupdict` inline within list comprehension.

PR #2686, opened 2024-04-21
refactor: Kullback-Leibler user generator

tdadela:
- simplified code
- added/fixed support for float weights
- yielding next value now has O(log n) complexity, instead of Θ(n) for roundrobin.smooth
- one dependency less


cyberw:
Btw, did you run the dispatch performance tests?

tdadela:
I rerun the same benchmark for this dummy implementation:
```py
cycle_fixed_gen = itertools.cycle([u.__name__ for u in fixed_users.values()])
cycle_weighted_gen = itertools.cycle([u.__name__ for u in self._user_classes if not u.fixed_count])
```
All results were above 11s. [So most time is consumed by other parts of code]
Mind you that for certain inputs, old implementation may be faster. Mostly due to "caching" in  the line 412:
```py
return itertools.cycle(gen() for _ in range(generation_length_to_get_proper_distribution))
```
I ran the following test a few times:
```py
if __name__ == "__main__":

    input_data = [(u, u.weight) for u in USER_CLASSES]
    ts = time.perf_counter()
    gen = _kl_generator(input_data)
    for _ in range(1_000_000):
        next(gen)
    instantiate_duration = time.perf_counter() - ts
    print(instantiate_duration * 1000)
```
(USER_CLASSES is exactly the same users list as in benchmark/dispatch.py)
On my hardware, it consistently takes less than 600 ms.
Equivalent version for roundrobin.smooth takes more than 9000 ms.
So, I think we have strong confidence that this change will not cause performance regression.
Furthermore, I believe that more important than performance gains are: code simplification, removal of weird normalization which led to issues (ex. #2662) and float weight support.

cyberw:
Great stuff! I’ll use it myself a couple times next week, and if I dont find any issues I’ll merge.

cyberw:
I noticed your branch is based on a pretty old commit. Can you rebase on latest master?

cyberw:
Thanks!

PR #2699, opened 2024-05-04
Httpsession requests typing

### LINKED ISSUE #2563: Pylance errors for catch_response and name parameters in self.client.get()
### Prerequisites

### Description

When using 'catch_response' and 'name' paramters in 'self.client.get()', Pylance reports 'No parameter named...' for each. However, the code does function properly as documented/expected. It's possible that this could be an issue with Pylance itself as well.

image

### Command line

locust -f pylance_bug.py

### Locustfile contents

```python3
"""Simple test case for Pylance error."""


class MyUser(HttpUser):
    """Simple test case for Pylance error."""

    @task
    def index(self) -> None:
        """Simple test case for Pylance error."""
        with self.client.get(url="/", catch_response=True, name="testcase") as response:
            if response.text != "Success":
                response.failure("Got wrong response")
```

### Python version

3.11.5

### Locust version

2.20.1

### Operating system

Red Hat Enterprise Linux 8.9

tdadela:
The aim of this PR is to provide better typing annotation support for `class HttpSession`.
This should also fix #2563.


tdadela:
This PR is still not finished (in draft). Requires adding similar annotations to FastHttpSession and more testing.
It would be nice and beneficial if you @cyberw could express if you are happy with this direction of changes.

cyberw:
I’ll have a look!

cyberw:
Is it possible to put all the override definitions in a if typing-block as well? I'd like to avoid the extra level of indirection at run time.

tdadela:
I'm not 100% sure, but according to my current understanding of Python typing ecosystem — no.
I'll do more research about it.

> I'd like to avoid the extra level of indirection at run time.

What is the main reason behind it?
Easier debugging / performance — necessity to create extra stack frame for additional function call / something else?


Source code of methods in the base class is very short:
```py
    def get(self, url, **kwargs):
        kwargs.setdefault("allow_redirects", True)
        return self.request("GET", url, **kwargs)

    def options(self, url, **kwargs):
        kwargs.setdefault("allow_redirects", True)
        return self.request("OPTIONS", url, **kwargs)

    def head(self, url, **kwargs):
        kwargs.setdefault("allow_redirects", False)
        return self.request("HEAD", url, **kwargs)

    def post(self, url, data=None, json=None, **kwargs):
        return self.request("POST", url, data=data, json=json, **kwargs)

    def put(self, url, data=None, **kwargs):
        return self.request("PUT", url, data=data, **kwargs)

    def patch(self, url, data=None, **kwargs):
        return self.request("PATCH", url, data=data, **kwargs)

    def delete(self, url, **kwargs):
        return self.request("DELETE", url, **kwargs)
```

What do you think about following possible implementation in Locust:
```py
    def get(self, url: str | bytes, **kwargs: Unpack[RESTKwargs]):  # type: ignore[override]
         kwargs.setdefault("allow_redirects", True)
        return self.request("GET", url, **kwargs)  # type: ignore[misc]
```
The number of function call would remain the same.

cyberw:
Yes, I’m mostly thinking about runtime performance. Why doesnt it work to define the wrappers in an if-block?

tdadela:
I think about it as defining types of method that de facto doesn't exist (in contrast to a function with the same name in a base class).
I doubt if type checkers were designed to handle such cases.
But I'm just guessing. I don't know.
I'm going to check it.

cyberw:
hi @tdadela did you have a chance to try it out? Or should I try it?

tdadela:
I will do it this weekend.

tdadela:
Sorry for the delay.
I tried defining HttpSession with annotations, inside if TYPE_CHECKING. It was ignored by type checkers. 

I see 3 options:
1. Keeping HttpSession as is. (no changes, closing PR)
2. Trying to define HttpSession with annotations in a .pyc file.
3. Calling self.requests directly inside HttpSession REST methods (proposition implemented in the last commit). This doesn't increase the number of function calls.

cyberw:
I like the version in the PR now. Have you tried it with vscode/pyright?

Maybe add a constraint to the typing_extensions dependency to only download it on Python versions <3.11?

0scarB:
Hi, I'm using **pyright** 1.1.338 in NeoVim. (With Python version 3.12.3). tdadela's version seems to work great. The `catch_response` and `name` kwargs weren't type hinted correctly before, now they are. Some screenshots:
image
image

Hope this helps. Thanks for the awesome project and great work tdadela!

locust/clients.py
```diff
@@ -187,6 +241,34 @@ def _send_request_safe_mode(self, method, url, **kwargs):
...
+        kwargs.setdefault("allow_redirects", True)
+        return self.request("OPTIONS", url, **kwargs)  # type: ignore[misc]
+
+    def head(self, url: str | bytes, **kwargs: Unpack[RESTKwargs]):  # type: ignore[override]
+        kwargs.setdefault("allow_redirects", False)
+        return self.request("HEAD", url, **kwargs)  # type: ignore[misc]
+
+    # These # type: ignore[misc] comments below are needed because data and json parameters are already defined in the
+    # RESTKwargs TypedDict. An alternative approach is to define another TypedDict which doesn't contain them.
+    def post(self, url: str | bytes, data: Any | None = None, json: Any | None = None, **kwargs: Unpack[RESTKwargs]):  # type: ignore[override, misc]
```
0scarB, line 260: There does seem to be a problem when using pyright after all. For HTTP methods where  `data=data` and/or `json=json` is passed to `self.request(...)` the return type becomes `Unknown`
image

I'm unsure what causes this behaviour. Explicitly providing the return type seems to fix it (see suggested changes).

```suggestion
    def post(self, url: str | bytes, data: Any | None = None, json: Any | None = None, **kwargs: Unpack[RESTKwargs]) -> (ResponseContextManager | Response | LocustResponse):  # type: ignore[override, misc]
```

Now the correct return type is shown:
image

locust/clients.py
```diff
@@ -187,6 +241,34 @@ def _send_request_safe_mode(self, method, url, **kwargs):
...
+        return self.request("POST", url, data=data, json=json, **kwargs)  # type: ignore[misc]
+
+    def put(self, url: str | bytes, data: Any | None = None, **kwargs: Unpack[RESTKwargs]):  # type: ignore[override, misc]
```
0scarB, line 263:
```suggestion
    def put(self, url: str | bytes, data: Any | None = None, **kwargs: Unpack[RESTKwargs]) -> (ResponseContextManager | Response | LocustResponse):   # type: ignore[override, misc]
```

locust/clients.py
```diff
@@ -187,6 +241,34 @@ def _send_request_safe_mode(self, method, url, **kwargs):
...
+        return self.request("PUT", url, data=data, **kwargs)  # type: ignore[misc]
+
+    def patch(self, url: str | bytes, data: Any | None = None, **kwargs: Unpack[RESTKwargs]):  # type: ignore[override, misc]
```
0scarB, line 266:
```suggestion
     def patch(self, url: str | bytes, data: Any | None = None, **kwargs: Unpack[RESTKwargs]) -> (ResponseContextManager | Response | LocustResponse):  # type: ignore[override, misc]
```

0scarB (changes requested):

tdadela:
@cyberw Do you prefer the last or the last but one version?
(I'll fix the stupid mistake in a minute)

Also, I've realized that  # type: ignore[misc] is no longer necessary, probably since calling self.request directly. I'll remove it later. (To not make the impression it's connected with the last commit).

tdadela:
> Have you tried it with vscode/pyright?

Yes. It works well.  (I hope I checked everything correctly.)

tdadela:
@0scarB Thank you very much for your support!
It's especially beneficial since I don't use some tools, e.g., pyright on daily bases (I'm not sure if I test everything correctly).

0scarB:
> @0scarB Thank you very much for your support! It's especially beneficial since I don't use some tools, e.g., pyright on daily bases (I'm not sure if I test everything correctly).

Tested the latest commit in my project. LGTM! The return types are now correct. Happy to help

cyberw:
Looks good to me too. @tdadela If you feel it is ready to go, move it out of draft and I'll merge it. Ideally I'd like to use the same typing for FastHttpSession but we can do that later on...

tdadela:
Today I tried to make return type argument dependent (catch_response) using `@overload`. Unfortunately, this makes the file super clutter.

> Ideally I'd like to use the same typing for FastHttpSession but we can do that later on...

Yup, soon in another PR.

PR #2709, opened 2024-05-11
Update Stats History on HTML Report

### LINKED ISSUE #2706: The Charts presentation in the report downloaded by locust is problematic
### Prerequisites

### Description

When I used the V1.0.0 version, the downloaded charts appeared fine
brznKr7O6H

When I used the newer version (2.27, 2.25), the downloaded chart was not fully displayed, and I tried the lower version 2.19, which had the same problem
WisXLzPws4
image

### Command line

 locust --tag post/pipelines

### Locustfile contents

```python3
    wait_time = constant(1)

    @task
    @tag("post/pipelines")
    def create_pipeline(self):
        self.client.post("/pipelines", json={
            "name": f"pressure_test_pipeline_{int(time.time())}",
            "version": "1.0.0",
            "title": "pressure_test_pipeline_title",
            "detailLink": "",
            "maintainer": "tester",
            "dag": "pipeline-bidsapp-bidsinfo",
            "isPublic": True
        })
```

### Python version

3.9

### Locust version

2.27

### Operating system

Windows11

andrewbaldwin44:
### The Issue

Because of `HISTORY_STATS_INTERVAL_SEC`, we will only have history starting after 5 seconds of run time. Unlike `request_stats` in `web.py`, we rely on history to display the charts in the HTML report. This is why the dashboard shows the results correctly in the dashboard but not in the report

### Proposal

Expose a new function `update_stats_history` that can be called when generating the HTML report. This will ensure that the stats history list is up to date before returning the report

image

Fixes #2706

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -10,7 +10,7 @@ from jinja2 import Environment, FileSystemLoader
 
 from . import stats as stats_module
 from .runners import STATE_STOPPED, STATE_STOPPING, MasterRunner
-from .stats import sort_stats
+from .stats import sort_stats, update_stats_history
 from .user.inspectuser import get_ratio
 
 PERCENTILES_FOR_HTML_REPORT = [0.50, 0.6, 0.7, 0.8, 0.9, 0.95, 0.99, 1.0]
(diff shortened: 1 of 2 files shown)


cyberw:
Nice!

cyberw:
Some time we should add automated tests for this kind of thing. But not necessarily today :)

andrewbaldwin44:
Definitely I think it's a good idea. I can look into it at some point with something like Playwright or Selenium.

Issue #2678, opened 2024-04-17
Add date to charts in web UI


kentmartin73:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

The charts produced by the UI are great for an overview of what happened when, but when doing load and stability tests that run for days, it is very difficult to see which day what happened. Adding the date to the X axis in addition to the time and/or the mouseover information would make the charts a lot easier to read in these circumstances

cyberw:
I think that sounds like a useful option, maybe even something that could be automatically enabled once the test runs for more than 24. But _someone_ needs to build it :)

alejandro-carmena:
I would also say that it could be quite useful being able to zoom in on the charts, also for having a better visibility in the tests that run for days.

andrewbaldwin44:
I think trying to detect when the test runs for more than 24hr to display the date adds too much overhead, not to mention adding the date to the x axis makes it quiet difficult to read. I would suggest displaying keeping the display on the x axis and add the full timestamp to the tooltip (on mouseover)
![image](https://github.com/locustio/locust/assets/59211596/29718932-b3ec-4ab3-8f39-88dfda1c3b25)
On the frontend it would be straightforward, we would just need to modify the stats and history in the backed to return a UTC timestamp instead of just the current time. What do you think @cyberw?

Also for zooming this is supported in echarts so I will add it :+1:

cyberw:
Sounds good!

JavierUhagon:
#2731 was merged, this should be closed, right? @cyberw

PR #2729, opened 2024-05-25
The time of the downloaded HTML report is not correct

### LINKED ISSUE #2691: The time of the downloaded html report is not correct
### Prerequisites

### Description

After I finish running with locust, I need to download the html of the report for subsequent report comparisons, but I realized that the time of the downloaded report is not correct

* Server time
image

* The timing of the charts is also correct
image

* The downloaded html reports the time incorrectly.
image

### Command line

locust -f 1_gloria_test.py 

### Locustfile contents

```python3
Nothing to do with content.
```

### Python version

3.11.0

### Locust version

2.26.0

### Operating system

amzn2023.x86_64

andrewbaldwin44:
### Proposed Changes

- Update the response from the server to provide a UTC timestamp as the start time and end time
- On the frontend, format the provided UTC date to locale time
image

Fixes #2691


cyberw:
This fixes #2691, right?

locust/util/date.py
```diff
@@ -0,0 +1,5 @@
+from datetime import datetime
```
cyberw, line 1: I think this fits better in html.py directly. it isnt super likely that it will be needed anywhere else anyway.

cyberw:
LGTM, apart from that one thing.

locust/util/date.py
andrewbaldwin44: I was thinking it might be useful for #2678, having a general utility to manage creating the timestamps would ensure we are consistent everywhere

cyberw: Ok if you foresee the need for it somewhere else too then it is fine!

PR #2731, opened 2024-05-25
Add date and zoom to charts in web UI

andrewbaldwin44:
### Proposal
- Stats history will now record a full UTC timestamp when recording history
- The frontend will keep a list of UTC timestamps for the charts to use
- The x-axis will format the timestamp in hours (same as before)
- The tooltip will format the timestamp as a locale string (with the date and hours, same as the HTML report)
- Add zoom functionality to the charts, which is a feature available in echarts
Recording 2024-05-25 at 19 03 44


cyberw:
Awesome stuff!

cyberw:
Can you maybe round the response times to .01ms precision?

andrewbaldwin44:
On the frontend we only record the chart data every 2 seconds and on the backend we record the history every 5 seconds. I'm not sure it would be useful to record

PR #2733, opened 2024-05-26
Mention installing Locust in Building the Docs

plaindocs:
Does what it says.


docs/developing-locust.rst
```diff
@@ -72,19 +72,28 @@ You can validate the whole project using tox:
 Build documentation
 ===================
 
-The documentation source is in the `docs/ <https://github.com/locustio/locust/tree/master/docs/>`_ directory. To build the documentation you first need to install the required Python packages:
+The documentation source is in the `docs/ <https://github.com/locustio/locust/tree/master/docs/>`_ directory. To build the documentation
 
-.. code-block:: console
+#. Fork Locust on `GitHub <https://github.com/locustio/locust/>`_ and install Lucust in editable mode:
```
cyberw, line 77: This is already mentioned a bit up the page. And while some people might want to contribute documentation updates only, most of them will have implemented some new feature which they then want to document, so going straight for building the docs is an uncommon use case :)

plaindocs: Yeah, agreed, but the docs build as is breaks if you don't have Locust installed.

cyberw: Maybe just reference that you need to do the thing it says at the start?

cyberw:
Great!

PR #2725, opened 2024-05-24
Poetry build system

mquinnfd:
## Switching to the Poetry build system

> [!TIP]
> This PR was worked on iteratively in `Draft` form, so there's a bit of history in the comments here.
>
> This is being suggested for integration to make building and bundling the front end a bit easier, comments welcome particularly if you have strong opinions on `poetry`

### In a Nutshell

This PR changes the build system used to package Locust over to [Poetry](https://python-poetry.org/).

The main use-case for this currently, was to allow the removal of the (snazzy new) dynamically generated front end bundle from the actual source code, and just let the build process create these for the package during the build.

#### Caveat

<details><summary>Details</summary>
<br>
<p>
I've tried to keep the package output and metadata like-for-like, because I havn't looked behind the curtain at the Locust release process, CI runners, etc. - it's likely this may need to change for that purpose.

There's plenty I am blissfully unaware of when it comes to maintaining this project, this is just a jumping off point for the sake of conversation - I have no particular love of Poetry, though I do use this day-to-day for projects currently. Alternatives noted were [PDM](https://github.com/pdm-project/pdm) (and now [Rye](https://rye-up.com/) by "the Ruff people") but I have no idea if these support the king of things we're doing here currently.
</p>
</details> 

---

### What

Adds the Poetry build system to the Locust package, this changes the build backend to Poetry's [PEP-517](https://peps.python.org/pep-0517/) compliant build back-end; [Poetry Core](https://github.com/python-poetry/poetry-core)

- Added Poetry build system
- Don't check in Poetry lock file, as we're often used as a library (worth discussion) - but do include it in generated builds, for provenance
- Poetry plugin [poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) does the job of [setuptools-scm](https://github.com/pypa/setuptools_scm/) to keep the version number in line with VCS info

### Why

Discussed with @cyberw here: https://github.com/locustio/locust/pull/2405#issuecomment-1806527834

- Allows for pre-build script, which allows us to build the front end on-demand and remove it from VCS
- While Poetry is far from perfect, it is at least [well maintained](https://github.com/python-poetry/poetry) by a large community
- Package size reduced by 10% due to not packaging tests *etc.* in the wheel (*were these necessary somehow? library users maybe?*)

---

### Try it Out

*Local dev installation*
<details><summary>Details</summary>
<br>
<code>make install</code>
<br><br>
<ul>
  <li>Installs project dependencies</li>
[...]

PR #2734, opened 2024-05-26
Docs: Import wiki to docs

plaindocs:
- added imports to TOC
- deleted Home.rst
- removed spurious html formatting `li`, `h4` etc
- fixed H1s
- rename/move files 
- actually look at content :-p
- migrated links to current content
- moved articles to discussion forum
- promoted "Increase my request rate/RPS"
- integrated installation troubleshooting

## Not done

- automation of examples
- Move "clear cookies" FAQ


cyberw:
A good start, but needs some work. 

`Increase my request rate/RPS` is great and really deserves a top level entry (with a link to the discussion forum for followup-questions). 

Same with installation issues (but it needs to be merged nicely with the current installation page somehow)

Articles need to be checked that they still exist (remove the oldest ones) and added as an entry in the discussion forum. It can still be a page in the docs, but maybe just a few titles and a link to that entry.

"Clear cookies, to make the next task iteration seem like a new user to my system under test" can be moved somewhere, maybe here: https://docs.locust.io/en/stable/writing-a-locustfile.html#examples (and we should list the examples there with a short description instead of just pointing to the examples/ directory - ideally by generating the list dynamically)

plaindocs:
@cyberw this is probably at the good enough stage, what do you think?

cyberw:
A few things:

Installation issues needs a proper introduction, probably even its own page, linked here 

> Install the package (check [the wiki](https://github.com/locustio/locust/wiki/Installation) if the installation fails)

And some of the titles (the ones that are error messages) should be formatted as code. It looks a bit weird atm:

image

There is some broken formatting here. And the last part about "Launch locust -f scripts/locustfile.py..." doesnt fit there, where did that come from?

image

cyberw:
There's an extra "the" here:

> If you need some help digging into server side problems, or you’re having trouble generating enough load to saturate your system, take a look at the [Increasing the request rate](https://locust--2734.org.readthedocs.build/en/2734/increasing-request-rate.html#increaserr).

Probably turning the sentence around to start with "having trouble generating enough load" fits better with the title of the page, but perhaps the title could be reworked as well?

cyberw:
Formatting: 

image

cyberw:
This screenshot is too small (maybe it has always been this way)

image

cyberw:
Random line breaks in the titles imported from the wiki:

image

cyberw:
Other than that, great stuff!

plaindocs:
I should have said, I'm trying to do the minimum viable import so we can merge this, so I'm concentrating more on importing the wiki as is rather than making sure all of the imported docs are in tip top shape, then we can iterate over it as part of the overhaul.

The code/title heading is in that psutil thing is gonna look weird either way unless we reword it. 
The `locust -f scripts/locustfile.py` is also as in the wiki. I've refactored that section (but what should the last code line read)

- Fixed the random line breaks
- Fixed 'the the'
- Fixed code formatting
- Fixed image width

cyberw:
> I should have said, I'm trying to do the minimum viable import so we can merge this, so I'm concentrating more on importing the wiki as is rather than making sure all of the imported docs are in tip top shape, then we can iterate over it as part of the overhaul.

Gotcha!

> The code/title heading is in that psutil thing is gonna look weird either way unless we reword it.
The locust -f scripts/locustfile.py is also as in the wiki. I've refactored that section (but what should the last code line read)

The "deploy on local environment" part of Installation in the wiki is confusing and broken. Just nuke it.

Still a few broken line endings from the import (but you can fix it later if you want)

image

cyberw:
oh, and formatting the "FAQ" page (if it should even exist at all, I'd prefer moving the entries there to other places, but there is no rush with that. What should be fixed is the explicit reference to the wiki :) (you can just remove it)

> If you have questions about Locust that are not answered here, please check [StackOverflow](https://stackoverflow.com/questions/tagged/locust), or post your question there. This wiki is not for asking questions but for answering them :)

plaindocs:
Agreed - migrating FAQ to other places usually makes more sense. But to do that properly I've have to take a proper look at the contents. :-D

cyberw:
Looks good now! I'll squash the commits, unless you want a plain merge?

plaindocs:
Whatever your normal is! Squashed is fine with me.

cyberw:
Btw, there's an open PR for changing the build/install process (#2725) so don't for some reason go off spending time documenting the current install procedure :)

PR #2741, opened 2024-05-27
Handle task weights

### LINKED ISSUE #2651: Use `random.choices` when selecting a task instead of `random.choice`
### Prerequisites

### Description

According to current implementation/documentation, if weights are given to tasks, tasks are stored in the list duplicated by the amount given as weight, and then the task is selected from the list by using Python's `random.choice`.
This, however, has two shortcomings:
- Inserting big weights can lead to unreasonably big datastructures
- Only integers can be weights

It should be fairly easy to rewrite the `gen_tasks_from_base_classes` function and how it's output is handled in `TaskSet` class to store the weights separately and pass them together with the list of tasks to `random.choices` in `gen_next_task`.

This will allow to store arbitrarily large weights without duplicating tasks in integer or float as well as potentially enable the functionality to modify task weights at runtime by modifying the list storing the weights.

I could try to implement this with a PR if this change is desired.

bakhtos:
Fixes #2651 

`tasks` attribute of `TaskSet`/`User` is now a mapping from tasks to their weights.
All other utility functions, like `tag`, `get_tasks_from_base_classes`, `filter_tasks_by_tags` take this into account in their recursion.

`gen_next_task` now uses `random.choices` to sample the tasks according to weights. Necessary cumulative weights are pre-computed in `__init__`.

The check of whether there are no tasks is now performed at `__init__`, instead of doing it on every call to `gen_next_task`.

Additionally, the `DefaultTaskSet` class is removed because it was only necessary to override two functions where `self.user` should be used instead of `self` when used directly inside a `User.` `User` now uses a `TaskSet` internally, and `TaskSet` now checks if it is being used by the `User` or nested in another `TaskSet` to make the correct choice.

Currently, this is a draft PR because only the backend itself is changed. Please let me know if the changes are understandable so far, and then I will proceed to change the tests and documentation to reflect this.

### DIFF
--- a/locust/user/inspectuser.py
+++ b/locust/user/inspectuser.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import inspect
-from collections import defaultdict
 from json import dumps
 
 from .task import TaskSet
@@ -65,17 +64,14 @@ def get_ratio(user_classes: list[type[User]], user_spawned: dict[str, int], tota
 
 def _get_task_ratio(tasks, total, parent_ratio):
     parent_ratio = parent_ratio if total else 1.0
-    ratio = defaultdict(int)
-    for task in tasks:
-        ratio[task] += 1
-
-    ratio_percent = {t: r * parent_ratio / len(tasks) for t, r in ratio.items()}
+    total_weight = sum(tasks.values())
 
     task_dict = {}
-    for t, r in ratio_percent.items():
+    for task, weight in tasks.items():
+        r = parent_ratio * (weight / total_weight)
         d = {"ratio": r}
-        if inspect.isclass(t) and issubclass(t, TaskSet):
-            d["tasks"] = _get_task_ratio(t.tasks, total, r)
-        task_dict[t.__name__] = d
+        if inspect.isclass(task) and issubclass(task, TaskSet):
+            d["tasks"] = _get_task_ratio(task.tasks, total, r)
+        task_dict[task.__name__] = d
 
     return task_dict
--- a/locust/user/task.py
+++ b/locust/user/task.py
@@ -121,8 +121,8 @@ def tag(*tags: str) -> Callable[[TaskT], TaskT]:
 
     def decorator_func(decorated):
         if hasattr(decorated, "tasks"):
-            # TODO Tasks should become a dict
-            decorated.tasks = list(map(tag(*tags), decorated.tasks))
+            tagged_tasks = map(tag(*tags), decorated.tasks.keys())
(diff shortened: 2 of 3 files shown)


bakhtos:
I see based on the failed tests already that if `TaskSet` is instantiated in `User` then the check of empty tasks in `__init__` will raise an error. There is a simple fix for that, will add it soon

cyberw:
Hi - looks like I reviewed your PRs out of order :)

I will do a full review once the build is green. 

I do get some errors out of the box running for example examples/locustfile.py.

```
[2024-05-27 19:06:22,011] Larss-MacBook-Pro/INFO/locust.runners: Ramping to 1 users at a rate of 1.00 per second
[2024-05-27 19:06:22,012] Larss-MacBook-Pro/INFO/locust.runners: All users spawned: {"QuickstartUser": 1} (1 total users)
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/Users/lars/git/locust/locust/user/users.py", line 189, in run_user
    user.run()
  File "/Users/lars/git/locust/locust/user/users.py", line 147, in run
    self._taskset_instance = TaskSet(self)
                             ^^^^^^^^^^^^^
  File "/Users/lars/git/locust/locust/user/task.py", line 293, in __init__
    raise Exception(
Exception: No tasks defined on TaskSet. Use the @task decorator or set the 'tasks' attribute
```

Also `DefaultTaskSet` is used in other places (like locust-plugins), so maybe we leave an implementation there for a while and just mark it deprecated?

cyberw:
Oh, and one more thing, would this change impact the "Current ratio" tab in the UI? (`http://0.0.0.0:8089/?tab=ratios`)

bakhtos:
@cyberw 

> I will do a full review once the build is green.

Okay, then I will work further on fixing the tests. Will take some time.

> Oh, and one more thing, would this change impact the "Current ratio" tab in the UI? (http://0.0.0.0:8089/?tab=ratios)

Does this get the data from this module? https://github.com/locustio/locust/blob/master/locust/user/inspectuser.py
If so, those functions are fixed internally to account for the dictionary of tasks to weights, but the interface and output should remain equivalent. If these are tested, once the build is fixed it should be ok.

cyberw:
Hi @bakhtos have you had any time to look at this? :)

bakhtos:
Hi @cyberw !

I started looking into fixing the tests for other PRs that I prepare here, and then I had a busy time at work.

Next week I am almost free from work, so I hope to finalize these PRs.

PR #2749, opened 2024-06-03
Add Logging to download_locustfile_from_master

### LINKED ISSUE #2748: Log a warning if getting locustfile from master takes more than 60s
### Prerequisites

### Description

https://github.com/locustio/locust/blob/1cf4f337bc19671661d3b2e2538efe75f2d64b22/locust/argument_parser.py#L250

andrewbaldwin44:
### Explanation

This one is a bit tricky, the argument parser happens before the log setup, so we don't have access to the same log levels as we would in the rest of the application. I also didn't want to have tons of messages logging for no reason, so I placed the messaging outside of the while loop. 

The problem with having the message outside of the while loop, is that we don't yet know if connecting to master failed or not (so it shouldn't _really_ be a warning). However, without logging being setup it seems we only have warning level logs. So the warning level is a bit of a lie, but it does at least provide some feedback that the master needs to be connected in order for the worker to start

Fixes #2748

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -6,7 +6,6 @@ from locust.rpc import Message, zmqrpc
 
 import ast
 import atexit
-import logging
 import os
 import platform
 import socket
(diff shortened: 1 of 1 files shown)


cyberw:
It is a bit awkward, but could you try instead launching a greenlet with 60s delay that is only responsible for printing the message, and interrupt it once a message is received (or skip printing the message if got_reply is true.

PR #2751, opened 2024-06-04
fix: external link syntax

alexravenna:


cyberw:
Thanks! But the work currently being done by @plaindocs in a separate PR makes this irrelevant.

PR #2758, opened 2024-06-10
Add proxy support for FastHttpUser

NicoAdrian:


cyberw:
Nice! While you could just provide your own clientpool to achieve the same thing, we're already allowing some parameters for convenience so I guess this is fine. I wont merge just yet, because I want to make a "clean" release with *only* the Poetry change in another open PR, but it should be soon (this week hopefully)

NicoAdrian:
> Nice! While I think you could just provide your own clientpool to achieve the same thing, we're already allowing some parameters for convenience so I guess this is fine.
> 
> I wont merge just yet, because I want to make a "clean" release with _only_ the Poetry change in another open PR, but it should be soon (this week hopefully)

With `HTTPClientPool` from `geventhttpclient` ? The thing is, I want all my users to use a different connection. Or I could just maybe instantiate it in my user's `__init__` ?

cyberw:
> Or I could just maybe instantiate it in my user's `__init__` ?

Yes, that what I meant :) But I don't expect people to do that, so this PR is fine.

PR #2766, opened 2024-06-20
URL Directory, and Multi-File Support for File Distribution

andrewbaldwin44:
### Proposal
1. Move some functions to utility files to permit re-usability and to help simplify the argument parser file
2. Update the logic for master sending locustfiles to the workers so that it can support URLs, directories, and multi-files
3. Re-use `parse_locustfile_paths` for workers receiving file paths from master

Opening as a draft, if this looks good I'll create some tests for it


cyberw:
What do you think @DennisKrone ?

cyberw:
Looks ok to me. Just a few things: parse_locustfile_paths_from_master isnt very descriptive of what it does (actually downloading files for instance). And needs type hints. Its a bit sad that it is a breaking protocol change.

cyberw:
Are you losing the ability to detect changes in the file contents?

locust/argument_parser.py
```diff
@@ -271,14 +252,27 @@ def wait_for_reply():
         sys.stderr.write(f"Got error from master: {msg.data['error']}\n")
         sys.exit(1)
 
-    filename = msg.data["filename"]
-    with open(os.path.join(tempfile.gettempdir(), filename), "w", encoding="utf-8") as locustfile:
-        locustfile.write(msg.data["contents"])
+    tempclient.close()
+    return msg.data
 
-    atexit.register(exit_handler, locustfile.name)
```
cyberw, line 278: Was this accidentally deleted?

andrewbaldwin44: No it's not needed anymore, we no longer open any files here so this `atexist` is no longer necessary. Instead we use re-use the same `parse_locustfile_paths` that is used in Master. And that function handles everything with reading the file and registering the `atexit`.

locust/argument_parser.py
```diff
@@ -271,14 +252,27 @@ def wait_for_reply():
...
+        filename = file_to_create["filename"]
+        file_contents = file_to_create["contents"]
+
+        with open(os.path.join(tempfile.gettempdir(), filename), "w", encoding="utf-8") as locustfile:
+            locustfile.write(file_contents)
+
+        return locustfile.name
+
+    locustfiles = [
+        create_locustfile(locustfile) if "contents" in locustfile else locustfile for locustfile in locustfiles
```
cyberw, line 272: I think this would be easier to read without the internal function and list comprehension? Also, parse_locustfile_paths_from_master is both long and not very descriptive, because it does all sorts of stuff not related to parsing of paths (and what "parsing a path" means is also unclear :)

locust/env.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         stop_timeout: float | None = None,
         catch_exceptions=True,
         parsed_options: Namespace | None = None,
+        parsed_locustfiles: list[str] = [],
```
cyberw, line 36: here's another dangerous default value :)

locust/runners.py
```diff
@@ -30,8 +30,17 @@
 from .dispatch import UsersDispatcher
 from .exception import RPCError, RPCReceiveError, RPCSendError
 from .log import get_logs, greenlet_exception_logger
-from .rpc import Message, rpc
-from .stats import RequestStats, StatsError, setup_distributed_stats_event_listeners
+from .rpc import (
+    Message,
+    rpc,
```
cyberw, line 35: Did you accidentally add a "magic" trailing comma here, or is your ruff not tuned to the same line length as the project? :)

locust/argument_parser.py
andrewbaldwin44: Yeah I agree maybe this was a bit over-engineered trying to make list comprehension work haha! I broke it down into a for loop now instead.

Naming is tough, perhaps this function is doing too much and that's why it's difficult to name. I went with `parse_locustfiles_from_master` to mimic the naming of the other function `parse_locustfile_paths`. But I agree, parsing could mean anything.

My one issue with breaking this function down more would be that then I feel like the argument parser is doing too much, so there's not much winning. But another option could be something like this:
```
locustfile_sources = download_locustfile_from_master(options.master_host, options.master_port).get("locustfiles", [])
locustfile_list = parse_locustfiles_from_master(locustfile_sources)
return parse_locustfile_paths(locustfile_list)
```
or even have a shared function call like this:
```
if options.locustfile == "-":
    if not options.worker:
        sys.stderr.write(
            "locustfile was set to '-' (meaning to download from master) but --worker was not specified.\n"
        )
        sys.exit(1)
    # having this in argument_parser module is a bit weird, but it needs to be done early
    locustfile_sources = download_locustfile_from_master(options.master_host, options.master_port).get("locustfiles", [])
    locustfile_list = parse_locustfiles_from_master(locustfile_sources)
else:
    locustfile_list = [f.strip() for f in options.locustfile.split(",")]
parsed_paths = parse_locustfile_paths(locustfile_list)

if not parsed_paths:
    note_about_file_endings = ""
    user_friendly_locustfile_name = options.locustfile

    if not options.locustfile.endswith(".py"):
        note_about_file_endings = "Ensure your locustfile ends with '.py' or is a directory with parsed_paths. "
    sys.stderr.write(
        f"Could not find '{user_friendly_locustfile_name}'. {note_about_file_endings}See --help for available options.\n"
    )
    sys.exit(1)

return parsed_paths
```

andrewbaldwin44: (I pushed the 2nd version in case you don't want to read all that, after thinking about it, I think it's easier to read)

locust/env.py
andrewbaldwin44: Yeah I thought mypy was getting upset with it not being a list but just the type was incorrect

locust/argument_parser.py
cyberw: Looks nice now. Maybe **download**_locustfiles_from_master is good? and a type hint to indicate that it returns a list of strings

PR #2772, opened 2024-06-25
A few minor FastHttpSession improvements

tdadela:
A few minor FastHttpSession improvements, including adding json as positional argument to `def post(self, url, data=None, json=None, **kwargs):` function.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -157,7 +157,7 @@ class FastHttpSession:
         headers: dict | None = None,
         auth=None,
         json: dict | None = None,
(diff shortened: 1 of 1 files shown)


scottARob:
any reason why you stopped converting to int for this? I assume performance reasons?

cyberw:
Idk why @tdadela did it, but it stops throwing away sub-ms precision and makes it more in line with HttpUser so its a good change I think :)

scottARob:
yeah thats fair to be honest. The new data type was breaking the data injection to influxdb on my side but was easily fixed. It probably is better to give the users the option to change the data type.

PR #2775, opened 2024-06-26
FastHttpSession requests typing 

### LINKED ISSUE #2563: Pylance errors for catch_response and name parameters in self.client.get()

tdadela:
The aim of this PR is to provide better typing annotation support for `class FastHttpUser`.
Fixes #2563. Follow-up to  #2699.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -12,12 +12,11 @@ import socket
 import time
 import traceback
 from base64 import b64encode
-from collections.abc import Generator
 from contextlib import contextmanager
 from http.cookiejar import CookieJar
 from json.decoder import JSONDecodeError
 from ssl import SSLError
-from typing import Any, Callable, cast
+from typing import TYPE_CHECKING, cast
 from urllib.parse import urlparse, urlunparse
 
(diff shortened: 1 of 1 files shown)


cyberw:
@DennisKrone Does this work with svs-locust's RestUser or does it introduce type check errors? I think this is a change for the better but want to be sure it doesnt break anything for you :)

PR #2777, opened 2024-06-27
new events for heartbeat and usage monitor

mgor:
when troubleshooting a problem where workers did not receive a heartbeat from master within `locust.runners.MASTER_HEARTBEAT_TIMEOUT` (not a locust problem) i created two new events which made it possible to easily update our statistics listener to send that information influxdb.

- `heartbeat_sent`: fired when master `sent` heartbeat to a worker
- `heartbeat_received`: fired when worker `received` said heartbeat
- `usage_monitor`: fired with information about current CPU and memory usage every `locust.runners.CPU_MONITOR_INTERVAL`

i did considered creating a custom `on_master_report` to add worker CPU and memory usage there, but that would leave out master. i also considered using the `heartbeat` event, since CPU and memory is part of the heartbeat RPC message sent, but that would also leave out master usage.

changes regarding imports organization and format is due to `ruff check` and/or `ruff format` complaining and failing the tests (in my fork).

### DIFF
--- a/examples/test_data_management.py
+++ b/examples/test_data_management.py
@@ -3,6 +3,10 @@
 #
 # 1. Locustfile parse time
 # 2. Locust start (init)
+# M1. CPU & memory usage
+# M2. master sent heartbeat to worker 1-N
+# M3. worker 1-N received heartbeat from master
+# (M* are repeated as long as locust is running)
 # 3. Test start
 # 4. User start
 # 5. Inside a task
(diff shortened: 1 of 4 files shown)


cyberw:
Looks very useful, but needs tests and an example (https://github.com/locustio/locust/blob/master/examples/test_data_management.py is where we dump all event examples I right now I guess :)

Have you considered using separate events (heartbeat_sent and heartbeat_received). To me that kind of makes more sense.

mgor:
ah, i was looking through the `locust/tests/` and `docs/` to see if there was something for events, guess i didn't look close enough.

i'll fix something!

examples/test_data_management.py
```diff
@@ -3,6 +3,10 @@
 #
 # 1. Locustfile parse time
 # 2. Locust start (init)
+# M1. CPU & memory usage
```
cyberw, line 6: What's "M"? :) Its kind of arbitrary where you put these, but I'd prefer having them after 5.

mgor: `M` as in "monitor events", i was abit unsure where to place them in the list and what to name them, it was `X` first, but then I saw that there already was one "numbered" with that.

I put them after 2 since they will start to appear after 2, I can for sure move them after 5, where I actually put them before I ran the test :)

cyberw:
Lgtm. Can you update the PR description to match the final solution?

PR #2778, opened 2024-06-27
update github actions to get rid of deprecation warnings

mgor:
update versions to get rid of:
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4, actions/cache@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

also includes `actions/setup-node@v3` and `borales/actions-yarn@v3`

with `actions/cache@v3` on windows image there would be an error, but the step wouldn't fail:
```
[Windows](https://github.com/mgor/locust/actions/runs/9692188762/job/26745040620#step:6:14)
Input required and not supplied: path
```

when updating to `actions/cache@v4` the step would also fail. to fix this the shell for the steps that redirects output to `$GITHUB_{OUTPUT,ENV}` was set to `bash`. since the default shell for windows runners is powershell, and `echo ... >> ...` isn't crossplatform compatible, and `bash` is available on the windows images.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,8 +26,8 @@ jobs:
           - { name: "3.9", python: "3.9", os: ubuntu-latest, tox: py39 }
 
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
       - name: Update pip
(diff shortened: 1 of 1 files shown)


mgor:
not sure why `py39` fails in print stats tests... saw that once in my other PR as well, but re-running it and it passed. doesn't seems like i have permission to re-run jobs though.

cyberw:
I re-ran it now. Is it always 3.9? I was already eager to drop 3.9 support to use nice language features that were introduced in 3.10 :P

But I might not want to merge this right now - https://github.com/locustio/locust/pull/2725 is quite a big change and I'd rather not make two changes in the build in the same release. For this reason I'm putting it in draft for now. Or is it urgent for you?

mgor:
the two times i've seen it fail it was `py39`, doesn't really have any good statistics for it :)

no, not urgent at all, it's just that i hate warning/errors and wanted it fixed ;) so that's fine

cyberw:
I'd love to do this update now! Want to update your PR?

cyberw:
I made a separate PR myself, but thanks for poking me with this :)

mgor:
Ops, sorry I missed your comment from two weeks ago.

Issue #2668, opened 2024-04-07
Custom Threshold Rules


ktchani:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

I've worked pretty extensively with the Locust framework. Locust is a fantastic bare bones (lightweight) framework. I really enjoy using it, and kudos to all the contributors. That being said, it seems to me (unless I'm missing something), that it is very pass/fail oriented. **It does not have out of the box solutions for setting/applying custom failure rules.** What is considered a failure is very different depending on the use case. Therefore, I believe that the pass/fail is insufficient.

As an example:
If I have a test that hits an endpoint: `some_example/resource?id=123`

There are no options to specify what my expectations of this endpoint are. If the endpoint responds in 2000 ms, I might very well consider that a critical failure. Currently, my options are as follows:
1. Look at the html report and see if there is any regression.
2. Save the raw json data to some data store, and keep historical records. There I can set up various tooling to alert me upon regression, etc.

**I suggest implementing an ability to:**
1. Define rules, which would take args like - attribute, operator, threshold. As an example: attribute=average response time, operator=greater than, threshold=2000ms.
2. Enable specific rules on specific tests. The idea here, is that there will be an output at the end of my test runs that I can then use automate things like sending notifications, retrying tests to see if it is transient, etc. 

A basic approach would be the extend the request method to accept an optional rule arg and process the rules downstream. The actual rule application would need to occur at the end of the test run since attributes could consist of multiple test run aggregations such as averages. I found it difficult to understand how the output json is generated. Ideally, this would just enhance that output.

cyberw:
Hi!

there are a couple features that at least solve parts of your problem:

1. Mark the response as failed based on response time: https://docs.locust.io/en/stable/writing-a-locustfile.html#validating-responses

2. Set exit code of the locust process based on checking some metric: https://github.com/SvenskaSpel/locust-plugins?tab=readme-ov-file#command-line-options

I do like your idea of rules/thresholds that would mark the request as failed (or maybe a third state: ”ok but with failed rule”) if it can be implemented cleanly, so if you or someone else would like to make a PR I would definitely consider it. Not sure it would be easy to do though..

ktchani:
1. Yea, I'm familiar with that functionality. It's great, but not really streamlined. Passing a rule should essentially execute something similar but ALSO save it to be a part of the JSON data at the end.
[...]

PR #2785, opened 2024-07-02
Add thresholds to each request

JavierUhagon:
This PR adds an argument to each request to add a threshold to it, if a request response time is over the threshold it'll be logged

This also adds a new stat in the webui (the threshold) that's hidden if there are no thresholds defined.
<details><summary>Images</summary>
<p>

image

image

</p>
</details> 

Related issue: #2668 

I want to know if @cyberw thinks this can be added to the core (and it's being done nicely) before I resume my work on it!


cyberw:
Hi! I do see the point in having this, but complicating HttpSession in for what is kind of a rare case doesnt make sense, it needs to be a subclass or something (side note: what are those print statements about? :)

JavierUhagon:
> Hi! I do see the point in having this, but complicating HttpSession in for what is kind of a rare case doesnt make sense, it needs to be a subclass or something (side note: what are those print statements about? :)

I think a subclass would make sense, yeah, this is just a draft, whenever the PR is ready for review I'll remove the print statements! :)

Kk, I'll work on that and @ you whenever it's ready!

JavierUhagon:
@cyberw I've added a new HttpSession subclass for this, and to use this there is a new HttpUser, if you think this can or should be done differently let me know!

cyberw:
Hi Javier! So sorry for the late response, things have been crazy lately. I understand the need for this, but the changes in stats.py are a little too big for a feature that will only rarely be needed. I think for now I'll have to suggest the workaround of marking requests as failed when they exceed your threshold https://docs.locust.io/en/stable/writing-a-locustfile.html#validating-responses

JavierUhagon:
Okay, that's fair enough, I'm not sure if you want me to close the PR or change my modifications and add an easy way to check the thresholds (even if we just fail the request!)

cyberw:
I'll close this. If you want to make it easier for other people to find the fail-request-on-bad-response time, maybe we could add it to one of the locustfile examples?

PR #2786, opened 2024-07-02
CLI argument for selecting `UserClasses` to run

### LINKED ISSUE #2745: Specifying users in CLI should use a keyword parameter
### Prerequisites

### Description

Currently, particular Users to run are specified on the command line like this

```
User classes:
  <UserClass1 UserClass2>
                        At the end of the command line, you can list User
                        classes to be used (available User classes can be
                        listed with --list). LOCUST_USER_CLASSES environment
                        variable can also be used to specify User classes.
                        Default is to use all available User classes
```

Doing this 

> at the end of the command line

as positional arguments is awkward if the command is big. The list of users should be a keyword CLI parameter like `--users UserClass1 UserClass2`, then it can be specified in any order of arguments in the command.

bakhtos:
Fixes #2745 

Argument parser now has an argument `--run-users` that is a list of UserClasses to choose.
Old version (simply listing classes without an arg) is still supported, even in the same call, but issues a warning that this method is deprecated

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -208,7 +208,7 @@ Usage: locust [options]
 
     locust -f my_test.py -H https://www.example.com
 
-    locust --headless -u 100 -t 20m --processes 4 --run-users MyHttpUser AnotherUser
+    locust --headless -u 100 -t 20m --processes 4 --user-classes MyHttpUser AnotherUser
 
 See documentation for more details, including how to set options using a file or environment variables: https://docs.locust.io/en/stable/configuration.html""",
     )
@@ -421,7 +421,7 @@ def setup_parser_arguments(parser):
         env_var="LOCUST_CONFIG_USERS",
     )
     parser.add_argument(
-        "--run-users",
+        "--user-classes",
         nargs="*",
         help="List of User classes to be used (available User classes can be listed with --list). LOCUST_USER_CLASSES environment variable can also be used to specify User classes. Default is to use all available User classes",
         default=os.environ.get("LOCUST_USER_CLASSES", "").split(),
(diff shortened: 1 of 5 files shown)


cyberw:
Can we comment out the deprecation warning for now? I do want to deprecate it, but not yet :)

On an unrelated note, I'm looking for endorsements/testimonials to put on our web page, could you say a few words about Locust?

Of course I want you to say what you think, but it could be something along the lines of "At the University of Oulu we use Locust to test the performance of our systems. Its flexible and developer friendly approach makes it the best tool for us."

bakhtos:
@cyberw I commented out the deprecation warning for now.

About the testemonial, you can email `alexander.bakhtin@oulu.fi` to discuss this.

locust/argument_parser.py
```diff
@@ -420,6 +420,12 @@ def setup_parser_arguments(parser):
         help="User configuration as a JSON string or file. A list of arguments or an Array of JSON configuration may be provided",
         env_var="LOCUST_CONFIG_USERS",
     )
+    parser.add_argument(
+        "--run-users",
```
cyberw, line 424: I think `--user-classes` is a better name

cyberw:
Can you update this test to use the new style: https://github.com/locustio/locust/blob/8a35d8de43797340b231161250550c7045f855fb/locust/test/test_main.py#L2205

locust/argument_parser.py
bakhtos: This creates an error since here https://github.com/locustio/locust/blob/137a269dccbb2fd0cb972aa25e5e4e40d74ee623/locust/test/test_load_locustfile.py#L169
you have short-form `u 100`, which gets parsed as `--u 100` instead of `-u 100 ` for some reason and so it gives an error because it cannot distinguish between `--users` and `--user-classes` if we use such a name

cyberw: Yea I figured what might cause an issue. I think --run-users is a bad name though. Can you think of something else?

bakhtos: The best way would have been changing `--users` to `--num_users` and then adding `--user-classes`, but this is not backward-compatible.

bakhtos:
@cyberw I have updated the test to use the long-form `users` in the config file, since config files only support long options according to documentation anyway (short options will be parsed to `--u`/`--r`/etc instead of `-u`/`-r`/etc, which will lead to ambiguous cli options)

PR #2788, opened 2024-07-04
Add Exporter as an Option for Locust Stats

andrewbaldwin44:
### Proposal
1. Add a new module `LocustExporter`, which will export stats to a given InfluxDB instance
2. Add a new flag `--exporter`, which will export the stats when active
3. Add documentation to explain the usage

### DIFF
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -156,7 +156,7 @@ A full example can be seen `in the auth example <https://github.com/locustio/loc
 Exporting Stats to InfluxDB
 ===========================
 
-You may wish to have more contorl over how the stats are aggregated and displayed. With the `--exporter` flag, Locust will
+You may wish to have more control over how the stats are aggregated and displayed. With the `--exporter` flag, Locust will
 export the stats to InfluxDB. Simply set the environment variables `INFLUXDB_URL`, `INFLUXDB_URL`, and `INFLUXDB_ORG` for your
 InfluxDB instance. For more on InfluxDB, see their `docs <https://docs.influxdata.com/influxdb/v2/>`.
 
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -153,6 +153,14 @@ authentication to the app should be granted.
 A full example can be seen `in the auth example <https://github.com/locustio/locust/tree/master/examples/web_ui_auth.py>`_.
 
 
+Exporting Stats to InfluxDB
+===========================
+
+You may wish to have more contorl over how the stats are aggregated and displayed. With the `--exporter` flag, Locust will
+export the stats to InfluxDB. Simply set the environment variables `INFLUXDB_URL`, `INFLUXDB_URL`, and `INFLUXDB_ORG` for your
+InfluxDB instance. For more on InfluxDB, see their `docs <https://docs.influxdata.com/influxdb/v2/>`.
+
+
 Run a background greenlet
 =========================
 
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -672,6 +672,12 @@ Typically ONLY these options (and --locustfile) need to be specified on workers,
         action="store_true",
         help="Prints the final stats in JSON format to stdout. Useful for parsing the results in other programs/scripts. Use together with --headless and --skip-log for an output only with the json data.",
     )
+    stats_group.add_argument(
+        "--exporter",
+        default=False,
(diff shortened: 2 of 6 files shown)


docs/extending-locust.rst
```diff
@@ -153,6 +153,14 @@ authentication to the app should be granted.
 A full example can be seen `in the auth example <https://github.com/locustio/locust/tree/master/examples/web_ui_auth.py>`_.
 
 
+Exporting Stats to InfluxDB
+===========================
+
+You may wish to have more contorl over how the stats are aggregated and displayed. With the `--exporter` flag, Locust will
+export the stats to InfluxDB. Simply set the environment variables `INFLUX_URL`, `INFLUX_TOKEN`, and `INFLUX_ORG` for your
```
cyberw, line 160: I'd prefer using the standard INFLUXDB_*-vars. 

Couldnt really find them properly documented but they are INFLUXDB_URL, INFLUXDB_TOKEN etc

docs/extending-locust.rst
```diff
@@ -153,6 +153,14 @@ authentication to the app should be granted.
 A full example can be seen `in the auth example <https://github.com/locustio/locust/tree/master/examples/web_ui_auth.py>`_.
 
 
+Exporting Stats to InfluxDB
+===========================
+
+You may wish to have more contorl over how the stats are aggregated and displayed. With the `--exporter` flag, Locust will
```
cyberw, line 159: contorl? :)

locust/argument_parser.py
```diff
@@ -672,6 +672,12 @@ def setup_parser_arguments(parser):
         action="store_true",
         help="Prints the final stats in JSON format to stdout. Useful for parsing the results in other programs/scripts. Use together with --headless and --skip-log for an output only with the json data.",
     )
+    stats_group.add_argument(
+        "--exporter",
+        default=False,
+        action="store_true",
+        help="Exports Locust stats to Influx",
+    )
```
cyberw, line 680: add `env_var="LOCUST_EXPORTER",`

locust/exporter.py
```diff
@@ -0,0 +1,141 @@
...
+        name,
+        response_time,
+        response_length,
+        exception,
+        context,
+        start_time=None,
+        url=None,
+        **kwargs,
+    ):
+        measurement = "request_failure" if exception else "request_success"
```
cyberw, line 119: I'm less versed then you are in InfluxDB, but shouldnt it be just _one_ measurement (request)? Wouldnt that make queries etc much easier?

andrewbaldwin44: It helps us to separate the failed requests versus the successful ones, for example, if we wanted to plot request_failures / s we could use a query like so:
```
from(bucket: v.bucket)
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_field"] == "response_time" and r["_measurement"] == "request_failure")
```

cyberw: Yes, but what about the (slightly more common) queries where we want *all* requests? At least that is how I would do it in SQL...

andrewbaldwin44:
```
from(bucket: v.bucket)
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_field"] == "response_time")
```
Would give back all requests
But yeah on the other hand, then we could also count errors like
```
from(bucket: v.bucket)
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_field"] == "exception")
```
So I think it's just a matter of preference

cyberw: Just reading the docs (https://docs.influxdata.com/influxdb/cloud/write-data/best-practices/schema-design/) I _think_ we should store a tag `success` (with true/false value). But lets chat about it on monday.

andrewbaldwin44: Yeah so based on the docs it definitely makes sense to just have the one measurement. Whether or not we want to have the tag `success`, we should decide if it's useful. Like I mentioned, since we could query for failures with `r["_field"] == "exception"`, we don't necessarily need a `success` field

cyberw: Don't we need it to be a tag for faster querying? Fields arent indexed, right?

andrewbaldwin44: Yup sorry! For some reason I was thinking exception was a tag but obviously it's not

Issue #2792, opened 2024-07-05
Locust --config-users incompatibility with --headless 


Hchanni:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

I'm trying to run `config-users` in my CLI using a JSON file that establishes a fixed user count for a class. However, it for some reason will only ramp to 1 user.

The JSON:

[
  {
    "user_class_name": "BasicHealthCheckScenario2",
    "fixed_count": 15
  },
]

### Command line

locust -f ./locustfiles/user_classes/myfile.py --headless --config-users config2.json --host <host>

### Locustfile contents

```python3
from locust import FastHttpUser, constant_throughput, task

class BasicHealthCheckScenario(FastHttpUser):
    wait_time = constant_throughput(1)

    @task
    def health(self):
       self.client.get("/health")
```

### Python version

3.9.6

### Locust version

2.29.1

### Operating system

Mac OS

Hchanni:
Figured out the issue is that if you don't define your user count with the `-u` it auto defaults to 1.

PR #2793, opened 2024-07-06
Log a message if total fixed_count is higher than number of users to spawn

cyberw:
Slightly relates to https://github.com/locustio/locust/issues/2792


andrewbaldwin44:
Nice! Tests are probably failing because it looks like the log message changed :)

Issue #2794, opened 2024-07-08
Build error with poetry build system


a18-kaushik:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

Building editable for locust (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building editable for locust (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [38 lines of output]
      Building front end...
      Traceback (most recent call last):
        File "C:\Users\Administrator\Desktop\Akshit\locust\pre_build.py", line 14, in <module>
          build()
        File "C:\Users\Administrator\Desktop\Akshit\locust\pre_build.py", line 10, in build
          subprocess.run(["make", "frontend_build"])
        File "C:\Python312\Lib\subprocess.py", line 548, in run
          with Popen(*popenargs, **kwargs) as process:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Python312\Lib\subprocess.py", line 1026, in __init__
          self._execute_child(args, executable, preexec_fn, close_fds,
        File "C:\Python312\Lib\subprocess.py", line 1538, in _execute_child
          hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      FileNotFoundError: [WinError 2] The system cannot find the file specified
      Traceback (most recent call last):
        File "C:\Python312\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
          main()
        File "C:\Python312\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Python312\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 273, in build_editable
          return hook(wheel_directory, config_settings, metadata_directory)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2801, opened 2024-07-10
Provide warning for local installs where yarn is not present

mquinnfd:
## Alter prebuild script to avoid `make` and check for `yarn`

This PR makes the pre-build script perform the same steps that the `make` command for building the project does, and provides the same error code where `yarn` is not present.

We try to avoid using `shell=True` and so need to split the command up

image

### DIFF
--- a/pre_build.py
+++ b/pre_build.py
@@ -1,8 +1,13 @@
 import os
 import subprocess
+from shutil import which
+from sys import exit
 
 
 def build() -> None:
+    if which("yarn") is None:
+        print("Locust requires the yarn binary to be available in this shell to build the web front-end.\nSee: https://docs.locust.io/en/stable/developing-locust.html#making-changes-to-locust-s-web-ui")
+        exit(1)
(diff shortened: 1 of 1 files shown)


pre_build.py
```diff
@@ -1,13 +1,21 @@
...
     if os.environ.get("SKIP_PRE_BUILD", "") == "true":
         print("Skipping front end build...")
         return
+    if which("yarn") is None:
+        print(
+            "Locust requires the yarn binary to be available in this shell to build the web front-end.\nSee: https://docs.locust.io/en/stable/developing-locust.html#making-changes-to-locust-s-web-ui"
+        )
+        exit(1)
     print("Building front end...")
-    subprocess.run(["make", "frontend_build"])
```
andrewbaldwin44, line 10: What's the reason for changing this? The make would run those two yarn commands right?

cyberw: It removes the dependency on make. Helps with some issues experienced here https://github.com/locustio/locust/issues/2794

Issue #2796, opened 2024-07-08
Windows Action Runs Wrong Version of Locust


andrewbaldwin44:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

When breaking the tests https://github.com/locustio/locust/compare/master...andrewbaldwin44:locust:bug/windows-action the Windows test action will not fail

![image](https://github.com/locustio/locust/assets/59211596/d8f5cc4d-b51c-4fd8-bd84-64233fbeeda5)

@mquinnfd This should be poetry related, any idea?

### Python version

3.12

### Locust version

2.29

### Operating system

Windows

mquinnfd:
Hey @andrewbaldwin44 I'll push something to replicate this, would be good to see the logs from the failing action 🧑‍💻

mquinnfd:
I think this is because...
- For Windows builds, we added a step to `build` & install the `locust` package, to put the current version on the `PATH`
- This is because `poetry install` did not appear to add `locust` to the `PATH` so caused tests to fail
- Running `poetry build` on the Windows runner, will mean that the front end build is run
- The version of `locust` used by the `fail_fast` tests will be the one that was build and installed as a package, not a dev install

So, in terms of what to do with this, the best way is probably to find the editable install location on the Windows runner, and add this to the `PATH` so the `fail_fast` tests can run. I avoided this initially because it might look a bit grim in the runner `steps` but it does make this build more predictable

mquinnfd:
Definitely fixable I'll just need to run this on my Windows machine to check the commands work

andrewbaldwin44:
Hey @mquinnfd, sorry about that I should have sent this link as well https://github.com/andrewbaldwin44/locust/actions/runs/9840186826/job/27163841986. The logs from the actions could have been seen on my fork but no matter :)

I think no frontend being built on Windows is not really an issue since we don't need the frontend for `fail_fast` right?

mquinnfd:
![image](https://github.com/locustio/locust/assets/82032454/7e54857a-e621-4ac4-9ac6-615d9b0a7993)
Didn't realise I could see your fork's actions, sorry

Yeah that's true - it was building this for the `build` step though, so tried removing it but it still doesn't fail the test.

It's a little bit weird, but I suppose the rub is that `poetry build` is overriding the output of `poetry install` here in ways which are not cool.

[...]

PR #2803, opened 2024-07-11
Fix tests on windows

mquinnfd:
## Ensuring Windows test runs use the correct version

Updates the GH Action tests for Windows/Linux, aiming to produce predictable failures of tests in controlled scenarios (i.e. using the correct Windows package).

Would recommend trying this out though, I still think there's some weird behaviour of some of these process-based tests, where the Windows vs. Linux tests do not operate or fail identically, perhaps this is intended.

I suspect this can be fixed more cleanly by sorting the `venv` mess in the pipeline and having the windows GH Action runner be in the correct, activated `venv` instead of doing this, so will raise a separate issue for that.

Addresses https://github.com/locustio/locust/issues/2796

Some testing of the breaking changes submitted in the issue by @andrewbaldwin44 
- Pass https://github.com/mquinnfd/locust/actions/runs/9889338165
- Fail https://github.com/mquinnfd/locust/actions/runs/9889352748
- Fail https://github.com/mquinnfd/locust/actions/runs/9889369296
- Pass https://github.com/mquinnfd/locust/actions/runs/9889376225

image

### Major
- Update the Windows-only step to create and install a binary version of the package, to use in the `tox`-based tests
- Add GH Actions matrix test value to allow front end build skipping for certain envs
- Pass the front end build skip env var to `tox` for fail fast tests

### Minor
- Updated steps in Developing Locust docs to correctly invoke `python`, not `pip3`
- Updated Developing Locust docs to be more consistent

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,16 +14,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {name: Windows, python: '3.12', os: windows-latest, tox: fail_fast_test_main}
-          # - {name: Mac, python: '3.12', os: macos-latest, tox: fail_fast_test_main}
-          - { name: "ruff", python: "3.11", os: ubuntu-latest, tox: "ruff" }
-          - { name: "mypy", python: "3.10", os: ubuntu-latest, tox: "mypy" }
+          - {name: Windows, python: '3.12', os: windows-latest, tox: fail_fast_test_main, skip_pre_build: "true" }
+          # - {name: Mac, python: '3.12', os: macos-latest, tox: fail_fast_test_main, skip_pre_build: "false" }
+          - { name: "ruff", python: "3.11", os: ubuntu-latest, tox: "ruff", skip_pre_build: "false" }
+          - { name: "mypy", python: "3.10", os: ubuntu-latest, tox: "mypy", skip_pre_build: "false" }
           # run some integration tests and abort immediately if they fail, for faster feedback
-          - { name: "Linux", python: "3.12", os: ubuntu-latest, tox: fail_fast_test_main }
-          - { name: "3.12", python: "3.12", os: ubuntu-latest, tox: py312 }
-          - { name: "3.11", python: "3.11", os: ubuntu-latest, tox: py311 }
-          - { name: "3.10", python: "3.10", os: ubuntu-latest, tox: py310 }
-          - { name: "3.9", python: "3.9", os: ubuntu-latest, tox: py39 }
+          - { name: "Linux", python: "3.12", os: ubuntu-latest, tox: fail_fast_test_main, skip_pre_build: "false" }
+          - { name: "3.12", python: "3.12", os: ubuntu-latest, tox: py312, skip_pre_build: "false" }
+          - { name: "3.11", python: "3.11", os: ubuntu-latest, tox: py311, skip_pre_build: "false" }
+          - { name: "3.10", python: "3.10", os: ubuntu-latest, tox: py310, skip_pre_build: "false" }
+          - { name: "3.9", python: "3.9", os: ubuntu-latest, tox: py39, skip_pre_build: "false" }
 
     steps:
       - uses: actions/checkout@v4
@@ -46,7 +46,7 @@ jobs:
       - name: Install poetry plugin
         run: python -m poetry self add "poetry-dynamic-versioning[plugin]"
       # Install test dependencies (tox) to the root (non-package install)
-      - name: Install dependencies
+      - name: Install test dependencies
         run: |
           python -m poetry install --no-root --only test
       - name: set full Python version in PY env var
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -60,11 +60,23 @@ jobs:
       - name: Install Yarn
         run: npm install -g yarn
       # Build and install the package if it's Windows, to service shell-based tests
-      - if: ${{ matrix.os }} == "windows-latest"
+      - name: Install binary on Windows
+        if: ${{ matrix.os == 'windows-latest' }}
+        env:
+          SKIP_PRE_BUILD: true
```
cyberw, line 66: Maybe this is no longer needed?

.github/workflows/tests.yml
```diff
@@ -14,16 +14,16 @@ jobs:
...
+          # - {name: Mac, python: '3.12', os: macos-latest, tox: fail_fast_test_main, skip_pre_build: "false" }
+          - { name: "ruff", python: "3.11", os: ubuntu-latest, tox: "ruff", skip_pre_build: "false" }
+          - { name: "mypy", python: "3.10", os: ubuntu-latest, tox: "mypy", skip_pre_build: "false" }
           # run some integration tests and abort immediately if they fail, for faster feedback
-          - { name: "Linux", python: "3.12", os: ubuntu-latest, tox: fail_fast_test_main }
-          - { name: "3.12", python: "3.12", os: ubuntu-latest, tox: py312 }
-          - { name: "3.11", python: "3.11", os: ubuntu-latest, tox: py311 }
-          - { name: "3.10", python: "3.10", os: ubuntu-latest, tox: py310 }
-          - { name: "3.9", python: "3.9", os: ubuntu-latest, tox: py39 }
+          - { name: "Linux", python: "3.12", os: ubuntu-latest, tox: fail_fast_test_main, skip_pre_build: "false" }
```
andrewbaldwin44, line 22: @mquinnfd I think we could skip the pre build for the "fail fast" test on Ubuntu as well right? Maybe to keep things consistent with the windows test and it would probably run faster?

docs/developing-locust.rst
```diff
@@ -11,19 +11,22 @@ Install Locust for development
...
     # clone the repo
     $ git clone git://github.com/<YourName>/locust.git
 
     # install the poetry build system
-    $ pip3 install poetry 
+    $ python -m pip install poetry 
 
     # install the dynamic versioning plugin for poetry
-    $ pip3 -m poetry self add "poetry-dynamic-versioning[plugin]"
+    $ python -m poetry self add "poetry-dynamic-versioning[plugin]"
```
cyberw, line 26: What will happen if this is not present? Can we make it optional, just to make the install easier for people? (maybe not now)

.github/workflows/tests.yml
```diff
@@ -60,11 +60,23 @@ jobs:
...
-          pip install --find-links=dist locust
-      - run: python -m poetry run tox -e ${{ matrix.tox }}
+          python -m poetry self add "poetry-plugin-export"
+          python -m poetry export --without-hashes --format=requirements.txt > requirements.txt
+          pip install -r requirements.txt
+          pip install poetry-core>=1.0.0
+          pip install locust --find-links=dist
+      - name: Run tox tests
+        env:
+          SKIP_PRE_BUILD: ${{ matrix.skip_pre_build }}
```
andrewbaldwin44, line 78: Will the string "false" really be evaluated as falsey? I would think you would have to do something such as `${{ matrix.skip_pre_build }} == "true"` in order for the statement to be truthy or falsey

cyberw:
I'm thinking maybe using `poetry ...` instead of `python -m poetry ...` might help with readability in docs and GH actions? 

Some people might have issues with poetry not having been placed in their path, but pip would/should warn them about that anyway...

Oh and same thing with pip**3**, just use `pip`.

.github/workflows/tests.yml
mquinnfd: In this case, the thing that checks the string is in `pre_build.py`:

```
os.environ.get("SKIP_PRE_BUILD", "") == "true"
```

Passing a boolean would be preferably, but you'd need to play cat and mouse between OS versions and how ENV vars are handled with respect to `"true" | True | true` etc. which is normally fairly painful

I could try and get this working without using the string if you want 🤔

docs/developing-locust.rst
mquinnfd: In terms of the plugin?

The default versioning will be used, which is not that big a deal really, we could omit this from this section of the docs if it's easier 👍 

You'll just end up with something like `0.0.1dev1` or similar as the package version, not a deal breaker if you're just wanting to build a dev version of the project

What do you think?

.github/workflows/tests.yml
mquinnfd: So...

Sort of!?

I found that building the front end (or not) actually does affect the fail-fast tests, I could probably find an example from earlier - some things give a totally different response code though (`200` vs `500`) so my current understanding is that the UI build is actually not optional for those tests currently 🤔

.github/workflows/tests.yml
mquinnfd: We're still doing a `poetry build` in here, so it will build the front end otherwise, we could of course tie this to the matrix value, so that it's picked up by any future Windows workflows (hopefully this step is removed soon though if I can figure out the `venv` stuff on the runner)

.github/workflows/tests.yml
andrewbaldwin44: No I don't have a problem with the strings I was just checking that it wasn't something that was missed since I didn't see any `== "true"` in the diff

.github/workflows/tests.yml
andrewbaldwin44: Ah I see and then I guess we just skip those tests if we're running on Windows

mquinnfd:
> I'm thinking maybe using `poetry ...` instead of `python -m poetry ...` might help with readability in docs and GH actions?
> 
> Some people might have issues with poetry not having been placed in their path, but pip would/should warn them about that when poetry is installed...
> 
> Oh and same thing with pip**3**, just use `pip`. If someone still has pip pointing to pip2, well, that's just sad :)

We could also just link to the `poetry` install docs, as a prerequisite 🤔

cyberw:
This is good enough for now I think. OK to merge?

.github/workflows/tests.yml
cyberw: Ah, I thought SKIP_PRE_BUILD was already set here.

cyberw: Perhaps `SKIP_PRE_BUILD: ${{ matrix.skip_pre_build }}` would make more sense?

mquinnfd:
> This is good enough for now I think. OK to merge?

I think it's definitely an improvement over what's in `master` currently, I have no problems with the merge as it stands, I'll raise some issues for any straggling issues

andrewbaldwin44:
Thanks so much for taking care of this @mquinnfd!

PR #2809, opened 2024-07-19
[Draft]: Fixing poetry tox tests on windows

mquinnfd:
Sticking this draft build up for visibility, current issue I'm working through on a minimum example repo is that the mock file context manager used by the tests seem to error out.

If I push something, this means you'll be able to see the actions though. This does give an idea about the cleanup of the GH test actions though.


cyberw:
When you switch to shell=True, dont you also have to switch from a list to a single string?
like this:
```
subprocess.Popen(["locust", "--help"])
subprocess.Popen("locust --help", shell=True, text=True)
```

Popen just ignores everything but the first element in shell=True mode

mquinnfd:
Thinking this is something to do with creating a `MasterRunner` process specifically, think the `shell=True` tests that implicitly do a `LocalRunner` are working OK (there are some `shell=True`) examples in there currently

I have a decent offline test bed for this issue now, so, hoping I can get some info on why the distributed tests with `shell=True` are no bueno

pre_build.py
```diff
@@ -14,9 +14,12 @@ def build() -> None:
         )
         exit(1)
     print("Building front end...")
-    subprocess.run(["yarn", "webui:install"])
-    subprocess.run(["yarn", "webui:build"])
-
+    if os.name == "nt":
+        subprocess.run(" ".join(["yarn", "webui:install"]), shell=True)
```
cyberw, line 18: Not sure if it relates to the issues you're having, but you probably wanna do subprocess.check_output instead of .run here, otherwise you're silently ignoring failures.

mquinnfd:
Closing to reincarnate as a cut-down PR

PR #2821, opened 2024-07-30
Update poetry windows tests

mquinnfd:
### Overview

Rights some wrongs currently in `master` relating to Windows integration tests.

In short - this works around some quirks of Poetry on Windows, with respect to the `PATH` and shell commands.

## Major

- Ensures that Windows tests run the current version of Locust
- Ensures that the prebuild script runs on the Windows shell (such that it can find `yarn`)
- Re-simplifies the test matrix

### DIFF
--- a/pre_build.py
+++ b/pre_build.py
@@ -15,12 +15,8 @@ def build() -> None:
         exit(1)
     print("Building front end...")
     try:
-        if os.name == "nt":
-            subprocess.check_output(" ".join(["yarn", "webui:install"]), shell=True)
-            subprocess.check_output(" ".join(["yarn", "webui:build"]), shell=True)
-        else:
-            subprocess.run(["yarn", "webui:install"])
-            subprocess.run(["yarn", "webui:build"])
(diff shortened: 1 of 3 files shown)


pre_build.py
```diff
@@ -14,8 +14,15 @@ def build() -> None:
...
         exit(1)
     print("Building front end...")
-    subprocess.run(["yarn", "webui:install"])
-    subprocess.run(["yarn", "webui:build"])
+    try:
+        if os.name == "nt":
+            subprocess.check_output(" ".join(["yarn", "webui:install"]), shell=True)
+            subprocess.check_output(" ".join(["yarn", "webui:build"]), shell=True)
+        else:
+            subprocess.run(["yarn", "webui:install"])
```
cyberw, line 22: Doesn't 
`subprocess.check_output("yarn webui:install", shell=True)` work on both platforms? I'd prefer that, to avoid the if-statement.

mquinnfd: It may do! Let me check

mquinnfd: Looks sound

Issue #508, opened 2017-01-06
Variance/Standard Dev. or something


danizen:
It would be nice to have the Variance/Standard Dev. in the request statistics CSV.   I know that the response time distribution CSV covers this nicely as well.

cgoldberg:
Pull Requests welcome

DaGeRe:
I've created a PR for this: https://github.com/locustio/locust/pull/2823

PR #2823, opened 2024-08-01
Add code to calculate standard deviation and write it to the CSVs

DaGeRe:
To see the deviation of measured values, it makes sense to also have a look at standard deviation besides mean, median and the percentiles (this has also been requested before: https://github.com/locustio/locust/issues/508). Therefore, I implemented it (using the formula from https://stats.stackexchange.com/questions/432256/how-would-i-calculate-the-standard-deviation-of-a-sample-given-the-sample-size, using the corrected sample variance with `n-1`).

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -4,6 +4,7 @@ import csv
 import hashlib
 import json
 import logging
+import math
 import os
 import signal
 import time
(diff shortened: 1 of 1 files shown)


cyberw:
Hi! Sorry for the slow response. I'm not sure this works in distributed mode? It needs a unit test & fixed linting/formatting.

PR #2836, opened 2024-08-05
Add PostgresUser to examples

guel-codes:
## Overview
This PR creates a new `database/` sub-dir inside of examples where database related files can be stored. Currently this only supports PostgreSQL. Ideally in the future we can have support for Mongo, MySQL, Redis, etc.

## Changes
- `examples/database/README.md`
   - This will store instructions on how to run the load test on various databases
- `examples/database/postgres_test.py`
   - This is the file where the PostgreSQL Client was created


cyberw:
Scratch that, I can see that psycopg3 has built in support for gevent! Awesome!

cyberw:
Can you split this into two files? One that contains a base `PostgresUser` marked with `abstract = True` and a separate one (the actual locustfile, maybe called postgres_ex.py) that imports PostgresUser from that file? 

Also, perhaps you could draw some inspiration from TimescaleListener in locust-plugins about how to configure the postgres connection? Or at least read them from class variables of the final User class defined in the actual locustfile.

guel-codes:
@cyberw I broke these files out and nested them inside of their own `database/postgresql/` directory. Let me know if you have any questions or preferences on how to proceed. Thanks.

examples/database/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Read the instruction below for your specific database
+
+## PostgreSQL
+### How to run the test
+- Prerequisites:
+    If you do not have them installed already
+    - `pip3 install psycopg3` - https://www.psycopg.org/psycopg3/docs/basic/install.html
+
+    - `pip3 install locust` - https://docs.locust.io/en/stable/installation.html
```
cyberw, line 11: I dont think you need to mention installing locust :)

examples/database/README.md
```diff
@@ -0,0 +1,31 @@
...
+    - PGPORT
+    - PGDATABASE
+    - PGUSER
+    - PGPASSWORD
+
+    so they can be picked up from `postgres_test.py`
+
+- Also, update the methods of the UserTasks class to run queries that will hit your database.
+
+#### Run test from browser:
```
cyberw, line 24: these things arent specific to PostgresUser, so I think you can skip these 2 sections as well.

examples/database/postgresql/base_postgres_user.py
```diff
@@ -0,0 +1,49 @@
...
+                    exception=e,
+                )
+                print(f"error: {e}")
+
+        return request_handler
+
+
+class PostgresUser(User):
+    abstract = True
+    client_class = PostgresClient
```
cyberw, line 46: I dont understand the purpose of this? When would PostgresUser have a different client class?

examples/database/postgresql/base_postgres_user.py
cyberw, line 1: I think this file should just be called `postgres_user`. You can move it into the `contrib` folder even (like fasthttp.py), that way it will be distributed as part of locust itself, and people (and you, in the example) can just do from `locust.contrib.postgres_user import PostgresUser`

examples/database/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Set your environment variables for: 
+    - PGHOST
+    - PGPORT
+    - PGDATABASE
+    - PGUSER
+    - PGPASSWORD
+
+    so they can be picked up from `postgres_test.py`
+
+- Also, update the methods of the UserTasks class to run queries that will hit your database.
```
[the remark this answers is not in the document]
guel-codes: Yeah sorry I brought this README over from my person repo where I shared with folks new to Locust and load testing so I wanted to over explain a little.

examples/database/postgresql/base_postgres_user.py
guel-codes: I actually love this pattern I will get it moved over

examples/database/postgresql/postgres_test.py
cyberw, line 1: Maybe skip the "database" subfolder and put it directly in `examples/postgres`? And name this file locustfile.py instead of postgres_test.py.

locust/contrib/postgres.py
```diff
@@ -0,0 +1,49 @@
...
+    return psycopg.connect(conn_string)
+
+
+def execute_query(conn_string, query):
+    db_conn = create_conn(conn_string)
+    return db_conn.cursor().execute(query)
+
+
+class PostgresClient:
+    def __getattr__(self, name):
```
cyberw, line 18: I dont think you need the whole `__getattr__` song and dance :)

Just define the execute_query method directly on the PostgresClient class.

locust/contrib/postgres.py
```diff
@@ -0,0 +1,49 @@
...
+
+import psycopg
+
+
+def create_conn(conn_string):
+    return psycopg.connect(conn_string)
+
+
+def execute_query(conn_string, query):
+    db_conn = create_conn(conn_string)
```
cyberw, line 13: Does this create a new connection on each query? That sounds very wasteful. Probably you wanna do that only when the client object is created.

locust/contrib/postgres.py
```diff
@@ -0,0 +1,52 @@
...
+        except Exception as e:
+            response_time = int((time.time() - start_time) * 1000)
+            events.request.fire(
+                request_type="postgres_failure",
+                name="execute_query",
+                response_time=response_time,
+                response_length=0,
+                exception=e,
+            )
+            print(f"error: {e}")
```
cyberw, line 38: I dont think we should print the exception :)

locust/contrib/postgres.py
```diff
@@ -0,0 +1,52 @@
+from locust import TaskSet, User, events
+
+import time
+
+import psycopg
+
+
+def create_conn(conn_string):
```
cyberw, line 8: I think this method is too small to make sense!

locust/contrib/postgres.py
```diff
@@ -0,0 +1,52 @@
...
+
+class PostgresClient:
+    def __init__(self, conn_string):
+        self.conn_string = conn_string
+        self.connection = create_conn(conn_string)
+
+    def execute_query(self, query):
+        start_time = time.time()
+        try:
+            cursor = self.connection.cursor()
```
cyberw, line 20: psycopg3 can do connection.execute(), no need to have the intermediate step of creating a cursor.

examples/postgres/locustfile.py
```diff
@@ -0,0 +1,36 @@
...
+    def run_update_query(self):
+        random_amount = random.randint(1, 12)
+        self.client.execute_query(
+            f"UPDATE loadtesting.invoice SET amount={random_amount} WHERE amount < 10",
+        )
+
+
+class PostgresLocust(PostgresUser):
+    tasks = [UserTasks]
+    min_wait = 0
```
cyberw, line 25: remove the wait_time stuff and just leave it as the default, to simplify the example.

You can skip the taskset and put the tasks directly under the user too.

cyberw:
Real close now, just a few last simplifications.

locust/contrib/postgres.py
guel-codes: Good call, it looks a lot cleaner calling the `psycopg.connect` directly

guel-codes:
@cyberw I made the updates for this. I think it is A LOT better than when it started. Thanks for all your patience as I made changes 🙏🏽

PR #2839, opened 2024-08-06
Prebuild UI in PyPi publish steps

mquinnfd:
# Overview

Ensures that `sdist` tarball contains web UI code, this is necessary for things such as the `brew` installation of Locust.

Example run of `sdist` build and tar unwrap with this added step:
https://github.com/mquinnfd/locust/actions/runs/10271769131

Example `sdist` build without web UI: 
https://pypi.org/project/locust/2.31.1/#files


cyberw:
Nice! If it works out I'll make a new release right away.

PR #2847, opened 2024-08-13
Web UI Should use Built-In Echarts Time Axis

andrewbaldwin44:
### Proposal
- The webui currently uses the 'category' type for the charts. We should use the built-in 'time' axis type instead, which will give us better looking time intervals, rather than random splits
- Update how chart history recorded is recorded in Locust to be formatted in the way echarts expects for time data ([time, data][] rather than data[])
-  Update tests and types

### Tested
- Ensure the mark line (between multiple test runs) continues to work as expected
- Ensure a chart from recorded history is rendered properly
- Ensure the HTML report is properly rendered

### Screenshots
image
image
image

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -904,18 +904,21 @@ def sort_stats(stats: dict[Any, S]) -> list[S]:
 
 def update_stats_history(runner: Runner) -> None:
     stats = runner.stats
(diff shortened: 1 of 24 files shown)


cyberw:
Nice! Ok to merge or is there a reason it is in draft?

PR #2858, opened 2024-08-17
Web UI as a Library NPM Release

andrewbaldwin44:
### Proposal
- Separate the useSwarmUI hook into three separate hooks. This will improve maintainability, performance, and testability. It will also allow for the hooks to be selectively imported and used in library mode as need be
- Add more exports
- Update logo to reduce hook calls and potentially allow for custom theming in the future
- Add steps to prerelease and release to deploy on NPM (tested in a separate repository so we should see these working as expected
- Export tabs as an object rather than an array in library mode for ease of use
- Add more exports from the locust-ui library
- Update locust-ui README
- Deployed a beta version of the locust UI on NPM: https://www.npmjs.com/package/locust-ui

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -220,6 +220,12 @@ jobs:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
           skip_existing: true
+      - uses: borales/actions-yarn@v5
+        name: Publish package on NPM
+        with:
+          cmd: publish --no-git-tag-version --non-interactive --new-version ${{ github.ref_name }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
     # The following is disabled because it has stopped working.
     #   - name: Tweet on release
     #     uses: infraway/tweet-action@v1.0.1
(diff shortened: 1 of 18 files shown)


.github/workflows/tests.yml
```diff
@@ -14,12 +14,22 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {name: Windows, python: '3.12', os: windows-latest, tox: fail_fast_test_main }
+          - {
```
cyberw, line 17: is this accidental or did you really change something?

.github/workflows/tests.yml
```diff
@@ -220,6 +230,16 @@ jobs:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
           skip_existing: true
+      - name: Set git config
+        run: |
+          git config --global user.email "github@action.com"
```
cyberw, line 235: these are weird :) I assume you had to put them there, but maybe explain the reason in a comment?

.github/workflows/tests.yml
andrewbaldwin44: It's my prettier setup that thinks this should be formatted differently :/

Issue #2857, opened 2024-08-16
Cleaning up the build process


mquinnfd:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

# Overview

After v1 of the migration to the `poetry` build system was completed, some kruft was left over and can now be tackled individually, and with care.

I'm using this issue to catalogue the things that should ideally be cleaned up and to contain discussion as well.

- Slim down the built docker image
- Introduce [poe](https://github.com/nat-n/poethepoet) to run pre-build hooks and remove janky pre-build hook
- Switch out the `poetry` installation in the pipeline to https://github.com/snok/install-poetry
- Resurrect the MacOS build
- Clean up use (or not use) of `poetry x` commands in the GH Actions pipeline
- Ensure caching of `venv` steps in build are doing what they're supposed to do
- Reduce GH Actions building of the project to the bare minimum

cyberw:
that all makes sense! Looking forward to your PR!

cyberw:
![image](https://github.com/user-attachments/assets/5e152f9d-e35e-4e60-b951-d8474b14b73c)

The more I hear about and try `uv` it looks like it will kick Poetry's ass :)

mquinnfd:
Ah man yeah `uv` looks great - it's really become more feature complete since I last looked at it ~6 months ago or so

I'm hoping _most_ of the build stuff we have is re-usable - it can use the `pyproject.toml` as it is for the most part, I suppose it's going to be the funkier stuff like the auto-scm versioning that might need some thinking (unless they have this now too :D)

I was actually working on a branch to clean up the pipeline which I'll put up now so you can at least have a look at it, maybe it's a reasonable interim step (which would make swapping for `uv` easier ultimately)

mquinnfd:
^ WIP branch added above, hey at least the tests are green.

mquinnfd:
I think betting against the Astral guys long term is probably going to be a bad idea :D I might start playing with `uv` again, even the docs look more polished

I know they also have [rye](https://rye.astral.sh/) which uses `uv` but seems to be more all-encompassing, not sure if this will follow the same path 🤔

mquinnfd:
I've been eyeing up `uv` for some internal projects just to try it out, I think from the POV of Locust builds, the main issues would be:
- Dynamic versioning [isn't currently supported ](https://github.com/astral-sh/uv/issues/6298)
[...]

PR #2873, opened 2024-08-26
Cleaning and improvements in the Build Pipeline

mquinnfd:
Overview:

Hey it's the draft PR guy again!

This is a WIP branch where I'm attempting to clean up the build pipeline as per  https://github.com/locustio/locust/issues/2857 - which is to say, implementing the hygiene that was omitted from the first iteration of the "get poetry working" work

The goal of this work is not only _jank reduction_, but:
- Build the package(s) once and once only
- Ship what we test
- Use caching and concurrency to speed things up as much as possible
- Make the pipeline easier to read and change

For @cyberw @andrewbaldwin44 and community 

### Major

- Updated build pipeline
  - Create distinct build, test and publish phases
  - Build packaged artifacts once, and use these in test and publishing phases
  - Added CI Docker image which picks up the built package instead of rebuilding it 
  - Attempts to consolidate publishing steps
  - Attempts to consolidate generation of build metadata 
- Added [poe](https://poethepoet.natn.io/index.html) to handle pre-builds
  - Eliminated renaming of wheels with multi-arch jank, and use of undocumented `poetry` feature
- Added a rudimentary test of the docker image
- Updated `tox` steps for integration testing to allow this to use an externally-built artifact
- Added the `poetry.lock` file to start caching dependency versions, but this could be reverted, there are tradeoffs I'll outline in the comments

### Minor

- Adds caching for NPM steps
- Brings back integration testing on MacOS
- Updated developing locust docs to be less specific about how to install `poetry`
- Added `tox` test to make sure the docs build works
- Updated comment on the pre-commit to point to the right `ruff` version marker
- Allowed `dist` to be picked up by the Docker context for the `ci` image


mquinnfd:
FYI this is mostly for consideration in the open, branch commits are dirty as I've been running remote builds a lot.

Things on my mind:
- I had split out the building, testing and packaging of the Docker image, but ended up consolidating this because multi-arch builds can't be easily exported and passed around with `buildx`
- I havn't adequately tested the publishing steps, to do this I'm probably going to make my own PyPi dummy repo and credentials, same with Dockerhub

mquinnfd:
On the subject of committing back the `poetry.lock` file

mquinnfd:
Subsequent builds will be quicker due to the addition of caching, I may re-run the actions for this ref for this reason, also looks like we got a slow runner.

mquinnfd:
Working on CI pipelines be like

image

cyberw:
Looking good! Let me know when it is ready for for final review!

mquinnfd:
> Looking good! Let me know when it is ready for for final review!

👍 no worries, I really just need to test the publication logic, which is hard to do without standing up some PyPi/Docker repo, so I'll do this as soon as I can to test it out on my fork then I'll flag this for review

mquinnfd:
So I tested out the publishing logic and fixed a few quirks to do with tag resolution etc. 
I now have some test projects set up that I can push to, will delete these repos once merged of course

Artifacts have been published successfully to:

- [PyPi](https://test.pypi.org/project/locust-test/)
- [DockerHub](https://hub.docker.com/repository/docker/mquinnfd/locust-test/general)
- [NPM](https://www.npmjs.com/package/locust-ui-test)

Would consider this in a review-able state:
- It's certainly better than what's there currently
- Main areas to check will be expected behaviour on merge, tag and merge-commit, which can all be tweaked

.github/workflows/tests.yml
```diff
@@ -5,85 +5,241 @@ on: [push, pull_request]
...
-          - { name: "ruff", python: "3.11", os: ubuntu-latest, tox: "ruff" }
-          - { name: "mypy", python: "3.10", os: ubuntu-latest, tox: "mypy" }
-          # run some integration tests and abort immediately if they fail, for faster feedback
-          - { name: "Linux", python: "3.12", os: ubuntu-latest, tox: fail_fast_test_main }
-          - { name: "3.12", python: "3.12", os: ubuntu-latest, tox: py312 }
-          - { name: "3.11", python: "3.11", os: ubuntu-latest, tox: py311 }
-          - { name: "3.10", python: "3.10", os: ubuntu-latest, tox: py310 }
-          - { name: "3.9", python: "3.9", os: ubuntu-latest, tox: py39 }
+          # Static analysis and utilities
+          - { name: "Verification Tests: Ruff", python: "3.11", os: ubuntu-latest, tox: "ruff" }
```
cyberw, line 120: I think you can drop the prefixes for the names. "Ruff/Mypy/Python 3.X" are specific enough, and I dont really know what a "verification test" is :P

cyberw: And if you really want to have prefixes, maybe "static analysis" is a better prefix?

Dockerfile.ci
```diff
@@ -0,0 +1,29 @@
+# This is the image pushed to Dockerhub, containing the built and tested Locust package
+
+# Stage 1: Install Locust package
+FROM python:3.11-slim AS base
+
+FROM base AS builder
+RUN apt-get update && apt-get install -y git
```
andrewbaldwin44, line 7: What's the reason for installing git here? Where is it used?

docs/developing-locust.rst
```diff
@@ -88,7 +91,7 @@ The documentation source is in the `docs/ <https://github.com/locustio/locust/tr
 
     .. code-block:: console
 
-        $ poetry install --with docs
+        $ poetry install --no-root --only docs
```
cyberw, line 94: Is the `--no-root` necessary?

mquinnfd: It saves having to install the project itself - it just sticks the dependencies required to build the docs into the `venv` only, which is quicker

Dockerfile.ci
mquinnfd: No idea! This is cruft from the historical docker image, I assumed the building of binaries required it

docs/developing-locust.rst
cyberw: Sure. I was just thinking the opposite is more rare (developing the docs without developing the project :)

Anyways, never mind :)

Dockerfile.ci
cyberw: It was needed so that we can determine the locust version name (it was used by setuptools_scm), e.g. for use with `locust --version`. Does it still work with the new way of building the docker image?

docs/developing-locust.rst
mquinnfd: It's a fair point! There are a fair few dependencies for building the docs - the alternative would be doing either:

`poetry install --with docs` <-- install/build everything (hopefully it's cached if you built it already) but add the docs dependencies

`poetry install --with dev,test,docs` up front ^^^ when originally installing the editable files and dev package, thus making the docs building available generally if you're developing the package

cyberw: I like `poetry install --with docs`

Dockerfile.ci
mquinnfd: I believe the Dynamic Versioning plugin for Poetry comes with whatever it needs 🤔 
--- but I'll check on the docker image, that a produced version is correct

On the CI, I did make a cut-down docker image which doesn't do any building at all, just takes the already-known-good built assets - the version of this is already resolved in the package (and can be seen in the test-docker-image step)
image

On the "normal" image, which users can still build end-to-end, I can check that the output of `locust --version` is correct inside the docker image 🤔

mquinnfd: It turns out the git binary is a necessity for resolving the version (understandably) even using `dunamai` etc. so I guess it gets left in, wanted to have the images as close to each other as possible.

mquinnfd: This does work with the local build as well, seems to get the correct version
image

.github/workflows/tests.yml
mquinnfd: Addressed in https://github.com/locustio/locust/pull/2873/commits/7d7bfc673d7b2676d3491d8f8218510d8d1bfa49

Dockerfile.ci
mquinnfd: Addressed in https://github.com/locustio/locust/pull/2873/commits/7d7bfc673d7b2676d3491d8f8218510d8d1bfa49

(had to allow `dist` in the docker context for the CI, so the local one has to remove it (or declare a wheel file name specifically)

docs/developing-locust.rst
mquinnfd: Addressed in https://github.com/locustio/locust/pull/2873/commits/7d7bfc673d7b2676d3491d8f8218510d8d1bfa49

cyberw:
I tried to improve the error message for the `--json` test that was failing https://github.com/locustio/locust/commit/4e29bf6f2ebf85d92dd24b5ec0654096bf226d84

cyberw:
the error went away on retry so 🎊

cyberw:
nooo... I didnt squash :-/

PR #2880, opened 2024-08-30
Add publishing dependency on build package step

mquinnfd:
### Overview

Adds a dependency on the build package step, from the POV of the publishing step.

Apparently, transitive dependencies are not acceptable for Github Actions job outputs, but there's no warning thrown about this during parsing, fun!


mquinnfd:
This should be good to go, I believe this fixes the build issue.

PR #2889, opened 2024-09-04
Update README.md

JonanOribe:
Hotfix on image URL


cyberw:
Thanks! 

Its not exactly perfect (screenshots are mixed dark/light, show the old UI, and are a little stretched), but its better than before at least :)

PR #2899, opened 2024-09-10
Add polling mechanism in tests

fletelli42:
Implemented polling mechanisms for various tests to improve reliability and remove arbitrary sleep delays

This commit introduces polling mechanisms in several automated tests across the Locust codebase. The polling replaces fixed time delays, ensuring tests wait for specific conditions or outputs before proceeding, which improves test reliability and execution speed.

Details of changes:

1. **test_custom_arguments**: Implemented polling to wait until the Locust web server is up on a specific port using `poll_until`.
2. **test_custom_exit_code**: Added polling to ensure the web server starts on a port, followed by proper shutdown validation.
3. **test_webserver**: Polls until the web server starts running on the designated port.
4. **test_percentile_parameter**: Uses polling to wait for the web server to start, removing the need for sleep delays.
5. **test_percentiles_to_statistics**: Added polling to check when the web server becomes available.
6. **test_invalid_percentile_parameter**: Implements incremental polling on stderr to validate the presence of the expected error message.
7. **test_webserver_multiple_locustfiles**: Polls until the web server starts, ensuring that multiple Locust files are loaded correctly.
8. **test_webserver_multiple_locustfiles_in_directory**: Polls until the server starts with multiple Locust files from a directory.
9. **test_webserver_multiple_locustfiles_with_shape**: Added polling to wait for the server to start and apply a load shape when multiple files are used.
10. **test_autostart_wo_run_time**: Polls for the web interface readiness when running Locust with the `--autostart` option but without a run time limit.
11. **test_autostart_w_run_time**: Waits for the web interface to be ready using polling, and then for Locust to quit automatically after the specified run time.
12. **test_run_autostart_with_multiple_locustfiles**: Polls until all users are spawned correctly using `--autostart` with multiple Locust files.
13. **test_autostart_w_load_shape**: Polls until the web interface is ready and the load shape is applied when using the `--autostart` option.
14. **test_autostart_multiple_locustfiles_with_shape**: Polls until the web interface is ready and applies the load shape with multiple Locust files and the `--autostart` option.
15. **test_web_options**: Polls until the Locust web server is running with the specified web host and port configurations.
16. **test_run_with_userclass_picker**: Polls until the web interface starts with the UserClass Picker enabled.
17. **test_error_when_duplicate_userclass_names**: Polls until the Locust process completes and checks for errors related to duplicate user class names.
18. **test_no_error_when_same_userclass_in_two_files**: Polls until the process completes to ensure that no error is raised when the same user class is used across two files.
19. **test_error_when_duplicate_shape_class_names**: Polls until the process completes to verify errors related to duplicate load shape class names.
20. **test_error_when_locustfiles_directory_is_empty**: Polls until the process completes to ensure an error is raised when the locustfiles directory is empty.
21. **test_graceful_exit_when_keyboard_interrupt**: Introduced polling to wait for the "Shape test starting" message, then sends a SIGINT signal and polls for a graceful shutdown, ensuring the test stops as expected and the correct messages appear in the output.

These changes improve the stability and accuracy of the test suite by eliminating arbitrary delays and waiting for specific system states before proceeding.

### DIFF
--- a/locust/test/test_main.py
+++ b/locust/test/test_main.py
@@ -102,22 +102,11 @@ class PopenContextManager:
     def __exit__(self, exc_type, exc_value, traceback):
         if self.process.poll() is None:
             try:
-                if platform.system() == "Windows":
-                    # Send CTRL_BREAK_EVENT to allow graceful shutdown
-                    os.kill(self.process.pid, signal.CTRL_BREAK_EVENT)
-
-                else:
-                    self.process.terminate()
-
-                self.process.wait(timeout=10)  # Wait for graceful termination
+                self.process.terminate()
+                self.process.wait(timeout=10)
             except subprocess.TimeoutExpired:
-                self.process.kill()  # Force kill if not terminated
-            except Exception as e:
-                print(f"Error terminating process: {e}")
-            finally:
-                if self.process.poll() is None:
-                    self.process.kill()
-                self.process.wait()  # Ensure process has fully terminated
+                self.process.kill()
+            self.process.wait()
         self.stdout_reader.join()
         self.stderr_reader.join()
 
--- a/locust/test/test_main.py
+++ b/locust/test/test_main.py
@@ -186,6 +186,7 @@ class PopenContextManager:
             universal_newlines=True,
             bufsize=1,
             **self.kwargs,
+            env={**os.environ, "PYTHONUNBUFFERED": "1"},
         )
 
         self.stdout_reader = gevent.spawn(self._read_output, self.process.stdout)
@@ -414,26 +415,25 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
         self.assertIn("--skip-log-setup", output)
 
     def test_custom_arguments(self):
-        """
-        Test that custom command-line arguments are correctly parsed and accessible.
-        """
         port = get_free_tcp_port()
         file_content = textwrap.dedent("""
             from locust import User, task, constant, events
-
             @events.init_command_line_parser.add_listener
             def _(parser, **kw):
                 parser.add_argument("--custom-string-arg")
-
             class TestUser(User):
                 wait_time = constant(10)
-
                 @task
                 def my_task(self):
                     print(self.environment.parsed_options.custom_string_arg)
         """)
 
-        with self.create_temp_locustfile(content=file_content) as temp_file_path:
+        with NamedTemporaryFile(mode="w+", delete=False, suffix=".py") as temp_file:
+            temp_file.write(file_content)
+            temp_file.flush()
+            temp_file_path = temp_file.name
+
+        try:
             args = [
                 sys.executable,
                 "-m",
@@ -446,12 +446,19 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
                 str(port),
             ]
 
-            with self.launch_locust(args) as manager:
-                # Verify the web interface is accessible by sending a POST request to /swarm
-                response = self.make_http_request(
-                    port=port,
-                    method="POST",
-                    path="/swarm",
+            with PopenContextManager(args) as manager:
+                if not wait_for_output_condition_non_threading(
+                    manager.process, manager.output_lines, "Starting Locust", timeout=30
+                ):
+                    self.fail("Timeout waiting for Locust to start.")
+
+                if not wait_for_output_condition_non_threading(
+                    manager.process, manager.output_lines, "Starting web interface at", timeout=30
+                ):
+                    self.fail("Timeout waiting for web interface to start.")
+
+                response = requests.post(
+                    f"http://127.0.0.1:{port}/swarm",
                     data={
                         "user_count": 1,
                         "spawn_rate": 1,
@@ -460,60 +467,68 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
                     },
                 )
 
-                self.assertEqual(response.status_code, 200, msg=f"Expected status code 200, got {response.status_code}")
+                self.assertEqual(response.status_code, 200)
 
-            # Combine all output lines for assertions
             combined_output = "\n".join(manager.output_lines)
-
-            # Assertions
             self.assertRegex(
                 combined_output, r".*Shutting down[\S\s]*Aggregated.*", "No stats table printed after shutting down"
             )
             self.assertNotRegex(
                 combined_output, r".*Aggregated[\S\s]*Shutting down.*", "Stats table printed BEFORE shutting down"
             )
-            self.assertNotIn("command_line_value", combined_output, "Command line value should not be in output.")
-            self.assertIn("web_form_value", combined_output, "Web form value should be present in output.")
+            self.assertNotIn("command_line_value", combined_output)
+            self.assertIn("web_form_value", combined_output)
+        finally:
+            os.unlink(temp_file_path)
 
     @unittest.skipIf(os.name == "nt", reason="Signal handling on Windows is hard")
     def test_custom_exit_code(self):
-        """
-        Test that Locust exits with a custom exit code as defined by event listeners.
-        """
         port = get_free_tcp_port()
         file_content = textwrap.dedent("""
             from locust import User, task, constant, events
-
             @events.quitting.add_listener
             def _(environment, **kw):
                 environment.process_exit_code = 42
-
             @events.quit.add_listener
             def _(exit_code, **kw):
                 print(f"Exit code in quit event {exit_code}")
-
             class TestUser(User):
                 wait_time = constant(3)
-
                 @task
                 def my_task(self):
                     print("running my_task()")
         """)
 
-        with self.create_temp_locustfile(content=file_content) as temp_file_path:
+        with NamedTemporaryFile(mode="w+", delete=False, suffix=".py") as temp_file:
+            temp_file.write(file_content)
+            temp_file.flush()
+            temp_file_path = temp_file.name
+
+        try:
             args = [sys.executable, "-m", "locust", "-f", temp_file_path, "--web-port", str(port)]
 
-            with self.launch_locust(args) as manager:
-                # Initiate graceful shutdown of Locust
-                self.terminate_locust(manager, shutdown_message="Shutting down (exit code 42)", timeout_shutdown=30)
+            with PopenContextManager(args) as manager:
+                if not wait_for_output_condition_non_threading(
+                    manager.process, manager.output_lines, "Starting Locust", timeout=30
+                ):
+                    self.fail("Timeout waiting for Locust to start.")
 
-            # Combine all output lines for assertions
+                if not wait_for_output_condition_non_threading(
+                    manager.process, manager.output_lines, "Starting web interface at", timeout=30
+                ):
+                    self.fail("Timeout waiting for web interface to start.")
+
+                manager.process.send_signal(signal.SIGTERM)
+
+                manager.process.wait(timeout=3)
+
             combined_output = "\n".join(manager.output_lines)
 
-            # Assertions
             self.assertIn("Shutting down (exit code 42)", combined_output)
             self.assertIn("Exit code in quit event 42", combined_output)
-            self.assertEqual(42, manager.process.returncode, "Locust did not exit with the expected exit code 42.")
+            self.assertEqual(42, manager.process.returncode)
+        finally:
+            os.unlink(temp_file_path)
 
     @unittest.skipIf(os.name == "nt", reason="Signal handling on Windows is hard")
     def test_webserver_multiple_locustfiles(self):
@@ -823,7 +838,7 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
 
         with (
             self.create_temp_locustfile(content=user_file_content) as mocked1,
-            self.create_temp_locustfile(content=shape_file_content) as mocked2,  # mocked2 is a string
+            self.create_temp_locustfile(content=shape_file_content) as mocked2,
         ):
             port = get_free_tcp_port()
 
@@ -831,7 +846,7 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
                 "-f",
                 mocked1,
                 "-f",
-                mocked2,  # Use mocked2 directly as it's a string
+                mocked2,
                 "--web-port",
                 str(port),
             ]
@@ -888,7 +903,7 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
 
     def test_default_headless_spawn_options(self):
         with mock_locustfile() as mocked:
-            proc = subprocess.Popen(
+            with PopenContextManager(
                 [
                     "locust",
                     "-f",
@@ -902,23 +917,29 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
                     "DEBUG",
                     "--exit-code-on-error",
                     "0",
-                    # just to test --stop-timeout argument parsing, doesnt actually validate its function:
                     "--stop-timeout",
                     "1s",
                 ],
-                stdout=PIPE,
-                stderr=PIPE,
-                text=True,
-                env=os.environ.copy(),
-            )
-            stdout, stderr = proc.communicate(timeout=4)
-            self.assertNotIn("Traceback", stderr)
-            self.assertIn('Spawning additional {"UserSubclass": 1} ({"UserSubclass": 0} already running)...', stderr)
-            self.assertEqual(0, proc.returncode)
+            ) as popen_ctx:
+                try:
+                    popen_ctx.process.wait(timeout=4)
+                except subprocess.TimeoutExpired:
+                    print("Process timed out")
+
+                all_output = "\n".join(popen_ctx.output_lines)
+
+                self.assertNotIn("Traceback", all_output)
+
+                # Check for the specific message
+                expected_message = 'Spawning additional {"UserSubclass": 1} ({"UserSubclass": 0} already running)...'
+                self.assertIn(expected_message, all_output)
+
+                # Check return code
+                self.assertEqual(0, popen_ctx.process.returncode)
 
     def test_invalid_stop_timeout_string(self):
         with mock_locustfile() as mocked:
-            proc = subprocess.Popen(
+            with PopenContextManager(
                 [
                     "locust",
                     "-f",
@@ -928,14 +949,14 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
                     "--stop-timeout",
                     "asdf1",
                 ],
-                stdout=PIPE,
-                stderr=PIPE,
-                text=True,
-                env=os.environ.copy(),
-            )
-            _, stderr = proc.communicate()
-            self.assertIn("ERROR/locust.main: Valid --stop-timeout formats are", stderr)
-            self.assertEqual(1, proc.returncode)
+            ) as popen_ctx:
+                popen_ctx.process.wait()
+
+                all_output = "\n".join(popen_ctx.output_lines)
+
+                self.assertIn("ERROR/locust.main: Valid --stop-timeout formats are", all_output)
+
+                self.assertEqual(1, popen_ctx.process.returncode)
 
     @unittest.skipIf(os.name == "nt", reason="Signal handling on Windows is hard")
     def test_headless_spawn_options_wo_run_time(self):
@@ -945,6 +966,9 @@ class StandaloneIntegrationTests(ProcessIntegrationTest):
         """
         with mock_locustfile() as mocked:
             args = [
+                sys.executable,
+                "-m",
+                "locust",
                 "-f",
                 mocked.file_path,
                 "--host",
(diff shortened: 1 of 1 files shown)


cyberw:
Cool stuff! Lets chat about it tomorrow!

.github/workflows/tests.yml
```diff
@@ -229,6 +229,8 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
+          cache-from: type=gha
```
cyberw, line 232: Interesting. Is this safe? Does it save a lot of time? Says here it is experimental, which makes me a little nervous :) https://docs.docker.com/build/ci/github-actions/cache/

locust/test/test_main.py
```diff
@@ -87,160 +111,197 @@ def test_custom_arguments(self):
...
+                self.assertIn("Exit code in quit event 42", stdout)
+                self.assertEqual(42, proc.returncode)
+            except PollingTimeoutError:
+                self.fail(f"Failed to start Locust process on port {port}")
+            finally:
+                proc.terminate()
 
     @unittest.skipIf(os.name == "nt", reason="Signal handling on windows is hard")
     def test_webserver(self):
+        port = get_free_tcp_port()  # Use a free port and pass it to Locust
```
andrewbaldwin44, line 210: Just a small nitpick here, I'm of the opinion that code should speak for itself. I think here and in a few other places we are repeating the same comment, and this comment doesn't really give any new information. `get_free_tcp_port` is already descriptive enough to tell us that we are getting a free port right?

locust/test/test_main.py
```diff
@@ -690,20 +997,72 @@ def my_task(self):
...
+                            return proc.poll() is not None
+
+                        # Wait for Locust to start
+                        poll_until(locust_started, timeout=10)
+
+                        # Wait for all users to be spawned
+                        poll_until(users_spawned, timeout=10)
+
+                        # Wait a bit to ensure we capture some task executions
+                        # gevent.sleep(2)
```
andrewbaldwin44, line 1032: If we don't need this gevent.sleep anymore I think we can remove it

locust/test/test_main.py
```diff
@@ -261,57 +322,108 @@ def my_task(self):
...
+                        line = proc.stderr.readline()
+                        if line:
+                            stderr_output.append(line)
+                        if "parameter need to be float" in line:
+                            return True
+                        if proc.poll() is not None:
+                            return False  # Stop if process terminates
+
+                # Use the polling mechanism to check stderr incrementally
+                poll_until(check_for_error_message, timeout=10)
```
andrewbaldwin44, line 388: No need to pass timeout=10 if the default timeout is 10 right?

locust/test/test_main.py
```diff
@@ -854,9 +1265,22 @@ def test_web_options(self):
...
+            try:
+
+                def server_ready():
+                    try:
+                        response = requests.get(f"http://127.0.0.1:{port}/", timeout=0.1)
+                        return response.status_code == 200
+                    except requests.RequestException:
+                        return False
+
+                poll_until(server_ready, timeout=5, poll_interval=0.1)
```
andrewbaldwin44, line 1277: Here as well, I think the default poll_interval is 0.1, so we shouldn't need to pass this argument

locust/test/test_main.py
```diff
@@ -1429,15 +1952,51 @@ def my_task(self):
...
+
+                while time.time() - start_time < timeout:
+                    ready, _, _ = select.select([proc.stderr], [], [], 0.1)
+                    if ready:
+                        line = proc.stderr.readline()
+                        buffer += line
+                        if expected_message in line:
+                            return True
+                    gevent.sleep(0.1)  # Prevent CPU hogging
+                return False
```
andrewbaldwin44, line 1969: For this `poll_for_message` function, could we use the top level `poll_until` with a callback that checks stderr for the expected message?

andrewbaldwin44:
Awesome work @fletelli42, this looks amazing!

locust/test/test_main.py
```diff
@@ -62,11 +67,16 @@ def web_interface_ready(port: int) -> bool:
 
 
 def all_users_spawned(proc: subprocess.Popen, output: list[str]) -> bool:
+    if proc.stderr is None:
+        raise ValueError("proc.stderr is None, cannot read from stderr")
```
cyberw, line 71: does this happen under any kind of normal circumstances? We don't need to check for things that should never, and this error message doesnt add much over the `AttributeError NoneType has no attribute 'readline'` that you would otherwise get anyway.

locust/test/test_main.py
```diff
@@ -1875,11 +1885,30 @@ def test_expect_workers(self):
...
+                        output += line + "\n"
+                        if "Waiting for workers to be ready" in line or "Gave up waiting for workers" in line:
+                            break
+            finally:
+                proc.terminate()
+                proc.wait(timeout=1)
+
+            self.assertTrue(
+                "Waiting for workers to be ready, 0 of 2 connected" in output
+                or "Gave up waiting for workers to connect" in output,
```
cyberw, line 1907: _both_ these strings should be in the output, not just one of them.

locust/test/test_main.py
```diff
@@ -1875,11 +1885,30 @@ def test_expect_workers(self):
...
+                output = ""
+                start_time = time.time()
+                while time.time() - start_time < 2:
+                    line = check_output()
+                    if line:
+                        output += line + "\n"
+                        if "Waiting for workers to be ready" in line or "Gave up waiting for workers" in line:
+                            break
+            finally:
+                proc.terminate()
```
cyberw, line 1902: the old test verifies that the process shuts down voluntarily, which is the most important feature of `--expect-workers-max-wait`, I think the new test kind of misses that?

locust/test/test_main.py
```diff
@@ -1875,11 +1885,30 @@ def test_expect_workers(self):
...
+                proc.wait(timeout=1)
+
+            self.assertTrue(
+                "Waiting for workers to be ready, 0 of 2 connected" in output
+                or "Gave up waiting for workers to connect" in output,
+                f"Expected output not found. Actual output: {output}",
+            )
+            self.assertNotIn("Traceback", output)
+            self.assertNotEqual(0, proc.returncode)
```
cyberw, line 1911: Why allow other return codes than 1?

.github/workflows/tests.yml
```diff
@@ -362,4 +362,4 @@ jobs:
           NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
         with:
           cmd: publish --no-git-tag-version --non-interactive --new-version ${{ github.ref_name }}
-          dir: locust/webui
+          dir: locust/webui
```
cyberw, line 365: accidental change? :)

locust/test/test_main.py
```diff
@@ -2681,15 +2681,11 @@ def my_task(self):
...
                     return "All users spawned" in line
 
-                print("Waiting for master to be ready...")
                 try:
                     poll_until(master_ready, timeout=60, sleep_time=0.1)
-                    print("Master is ready")
-                except PollingTimeoutError as e:
-                    print(f"Master readiness polling timed out: {e}")
+                except PollingTimeoutError:
                     raise
```
cyberw, line 2689: If you're not printing the error any more this is pretty much exactly the same as not having a try-except at all :)

locust/test/test_main.py
```diff
@@ -3,30 +3,128 @@
...
+        time.sleep(sleep_time)
+
+    raise PollingTimeoutError(f"Condition not met within {timeout} seconds")
+
+
 def is_port_in_use(port: int) -> bool:
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-        return s.connect_ex(("localhost", port)) == 0
+        try:
+            s.bind(("localhost", port))
```
cyberw, line 55: What's the advantage of checking this way?

locust/test/test_main.py
```diff
@@ -176,69 +251,114 @@ class TestUser(User):
...
-            from locust import User, task, constant, events
-            from locust import stats
-            stats.PERCENTILES_TO_CHART = [0.9, 0.4]
-            class TestUser(User):
-                wait_time = constant(3)
-                @task
-                def my_task(self):
-                    print("running my_task()")
-        """
+                from locust import User, task, constant, events
```
cyberw, line 325: Still a bit of whitespace changes going on here...

locust/test/test_main.py
```diff
@@ -176,69 +251,114 @@ class TestUser(User):
...
+                self.assertEqual(200, response.status_code)
+
+                proc.send_signal(signal.SIGTERM)
+                _, stderr = proc.communicate()
+
+                self.assertIn("Starting web interface at", stderr)
+                self.assertIn("Starting Locust", stderr)
+
+            except PollingTimeoutError as e:
+                self.fail(f"Polling failed: {str(e)}")
```
cyberw, line 356: Is this better than just not catching the exception?

locust/test/test_main.py
```diff
@@ -280,46 +412,104 @@ class TestUser(User):
...
+                stderr=PIPE,
+                text=True,
+                bufsize=1,  # Line-buffered output
+            )
+
+            try:
+                # Continuously read stderr for real-time updates
+                stderr_output = []
+
+                def check_for_error_message():
```
cyberw, line 430: for this case it is not worth polling, because the expected result is for the process to exit completely on its own, and the polling would make a failure harder to read.

locust/test/test_main.py
```diff
@@ -2348,7 +2348,7 @@ def t(self):
             output.extend(stdout.splitlines())
             output.extend(stdout_worker.splitlines())
 
-            self.assertIn("Shutting down (exit code 0)", "\n".join(output))
+            self.assertIn('Shutting down (exit code 0)', "\n".join(output))
```
cyberw: If you haven't already done so, I recommend installing the ruff plugin in vscode, that way you'll get formatting whenever you save the file. And set up the pre-commit hook as well, so any linting errors will be caught at commit-time: https://pre-commit.com/

locust/test/test_main.py
fletelli42: it's for windows compatibility

locust/test/test_main.py
```diff
@@ -3,30 +3,128 @@
...
+    if ready:
+        line = stdout.readline().strip()
+        output.append(line)
+
+    return output
+
+
+def read_nonblocking(proc: subprocess.Popen, output: list[str]) -> bool:
+    if platform.system() == "Windows":
+        if proc.stdout is None or proc.stderr is None:
```
cyberw, line 104: are you doing this just to keep the linter happy? if so, you can just do `assert proc.stdout` and do it before checking the platform, so you only need to do it once.

locust/test/test_main.py
```diff
@@ -230,17 +368,33 @@ def my_task(self):
...
+                proc.send_signal(signal.SIGTERM)
+                _, stderr = proc.communicate()
+
+                self.assertIn("Starting web interface at", stderr)
+                self.assertIn("Starting Locust", stderr)
+
+            finally:
+                proc.terminate()
+                proc.wait(timeout=5)
+                proc.terminate()
```
cyberw, line 392: Why kill it twice?

locust/test/test_main.py
```diff
@@ -750,13 +1024,11 @@ def tick(self):
             self.assertIn("Starting Locust", stderr)
             self.assertIn("Shape test starting", stderr)
             self.assertIn("Shutting down ", stderr)
-            self.assertIn("autoquit time reached", stderr)
-            # check response afterwards, because it really isn't as informative as stderr
+            self.assertIn("--run-time limit reached", stderr)
```
cyberw, line 1027: you've changed the text you're looking for here?

locust/test/test_main.py
```diff
@@ -1182,23 +1462,31 @@ def test_html_report_option(self):
...
-                stdout, stderr = proc.communicate()
+                _, stderr = proc.communicate()
 
                 self.assertIn("Locust is running with the UserClass Picker Enabled", stderr)
                 self.assertIn("Starting Locust", stderr)
                 self.assertIn("Starting web interface at", stderr)
 
+                if proc.poll() is None:
+                    proc.terminate()
+                    proc.communicate()
```
cyberw, line 1488: maybe use proc.wait(time=5) here instead, just to keep it consistent?

locust/test/test_main.py
```diff
@@ -1405,7 +1720,7 @@ def on_test_stop(environment, **kwargs) -> None:
                 class LoadTestShape(LoadTestShape):
                     def tick(self):
                         run_time = self.get_run_time()
-                        if run_time < 2:
+                        if run_time < 10:
```
cyberw, line 1723: This changes the test a little bit. Before, it was actually testing that the test exits, because tick() eventually returns None, now we kill it before that has a chance to happen.

locust/test/test_main.py
```diff
@@ -1425,19 +1740,47 @@ def my_task(self):
...
             stdout, stderr = proc.communicate()
-            print(stderr, stdout)
-            self.assertIn("Shape test starting", stderr)
-            self.assertIn("Exiting due to CTRL+C interruption", stderr)
+
             self.assertIn("Test Stopped", stdout)
-            # ensure stats printer printed at least one report before shutting down and that there was a final report printed as well
-            self.assertRegex(stderr, r".*Aggregated[\S\s]*Shutting down[\S\s]*Aggregated.*")
+
+            self.assertRegex(stderr, r".*Shutting down.*")
```
cyberw, line 1779: i guess you had to change this because the test run no longer has time to finish? I'd like to change it back, because it was an important verification :)

locust/test/test_main.py
```diff
@@ -1475,11 +1818,29 @@ def test_expect_workers(self):
                 stderr=PIPE,
                 text=True,
             )
-            _, stderr = proc.communicate()
-            self.assertIn("Waiting for workers to be ready, 0 of 2 connected", stderr)
-            self.assertIn("Gave up waiting for workers to connect", stderr)
-            self.assertNotIn("Traceback", stderr)
-            self.assertEqual(1, proc.returncode)
+
+            def check_output():
```
cyberw, line 1822: This function doesnt look very useful to me, and it is only used once. Its probably easier to read if you just call `proc.stderr.readline().strip()` directly

locust/test/test_main.py
```diff
@@ -2213,25 +2631,52 @@ def my_task(self):
...
+                poll_until(workers_shut_down, timeout=60, sleep_time=0.1)
+
                 worker_stdout, worker_stderr = worker_parent_proc.communicate()
-                assert False, f"worker never finished: {worker_stdout} / {worker_stderr}"
 
-            self.assertNotIn("Traceback", worker_stderr)
-            self.assertIn("Didn't get heartbeat from master in over ", worker_stderr)
-            self.assertIn("worker index:", worker_stdout)
+                self.assertNotIn("Traceback", worker_stderr)
+                self.assertIn("Got quit message from master, shutting down", worker_stderr)
```
cyberw, line 2667: Oh. I'm afraid you broke this test. It is for validating that if master is _killed_ (not politely shut down using `.terminate()`) , the workers will shut down.

locust/test/test_main.py
```diff
@@ -1766,34 +2128,58 @@ def t(self):
...
+                f"'Shutting down' not found in output:\n{''.join(output)}",
+            )
+            self.assertIn('All users spawned: {"User1": 1} (1 total users)', "\n".join(output))
+            self.assertNotIn("Traceback", "\n".join(output))
 
             self.assertEqual(0, proc.returncode)
             self.assertEqual(0, proc_worker.returncode)
+            self.assertEqual(0, proc_worker2.returncode)
+
+            for p in [proc, proc_worker, proc_worker2]:
```
cyberw, line 2175: Why is this necessary? We have already checked that their return code was zero, so the processes must already be dead :)

locust/test/test_main.py
```diff
@@ -1766,34 +2128,58 @@ def t(self):
...
-            self.assertNotIn("Traceback", stdout_worker2)
+            output = []
+
+            poll_until(lambda: all_users_spawned(proc, output), timeout=30)
+
+            stdout, stderr = proc.communicate(timeout=30)
+            stdout_worker, stderr_worker = proc_worker.communicate(timeout=10)
+            stdout_worker2, stderr_worker2 = proc_worker2.communicate(timeout=10)
+
+            output.extend(stdout.splitlines())
```
cyberw, line 2157: I don't think joining the output of all processes in a single string makes sense, it would be unnecessarily difficult to debug. Can you keep them separate?

locust/test/test_main.py
```diff
@@ -1872,36 +2258,64 @@ def t(self):
...
+            while time.time() < end_time:
+                read_nonblocking(proc, output)
+                if "Shutting down" in "\n".join(output):
+                    break
+                time.sleep(0.1)
+
+            stdout, stderr = proc.communicate(timeout=20)
+            stdout_worker, stderr_worker = proc_worker.communicate(timeout=5)
+
+            output.extend(stdout.splitlines())
```
cyberw, line 2296: Same here, I prefer to keep output of the processes separate.

locust/test/test_main.py
```diff
@@ -2213,25 +2631,52 @@ def my_task(self):
...
             try:
-                worker_stdout, worker_stderr = worker_parent_proc.communicate(timeout=7)
-            except Exception:
-                os.killpg(worker_parent_proc.pid, signal.SIGTERM)
+
+                def master_ready():
+                    line = master_proc.stderr.readline()
+                    return "All users spawned" in line
+
+                poll_until(master_ready, timeout=60, sleep_time=0.1)
```
cyberw, line 2650: default `timeout` and `sleep_time` should work here, no?

locust/test/test_main.py
```diff
@@ -1872,36 +2258,64 @@ def t(self):
...
+            self.assertTrue(
+                any("Shutting down" in line for line in output),
+                f"'Shutting down' not found in output:\n{''.join(output)}",
+            )
+            self.assertNotIn("Traceback", "\n".join(output))
 
             self.assertEqual(0, proc.returncode)
             self.assertEqual(0, proc_worker.returncode)
 
+            for p in [proc, proc_worker]:
```
cyberw, line 2310: same here, there's no way these processes still alive this point, no need to kill them.

locust/test/test_main.py
```diff
@@ -1425,19 +1740,47 @@ def my_task(self):
...
+                    gevent.sleep(0.1)
+                    return False
+
+                return condition
+
+            poll_until(condition_for_message("Shape test starting"), timeout=15)
+            gevent.sleep(1)
             proc.send_signal(signal.SIGINT)
+
+            poll_until(condition_for_message("Exiting due to CTRL+C interruption"), timeout=15)
```
cyberw, line 1773: No need to poll, `proc.communicate()` is easier to read (and faster)

locust/test/test_main.py
```diff
@@ -1425,19 +1740,47 @@ def my_task(self):
...
-                stderr=PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
                 text=True,
             )
-            gevent.sleep(1.9)
+
+            output = []
+
+            def condition_for_message(expected_message: str) -> Callable[[], bool]:
```
cyberw, line 1750: Hmm... I think this one got too complicated. I'm sure we're saving maybe ~0.5s run time, but its not worth replacing one line of code (`gevent.sleep(1.9)`) with 20.

locust/test/test_main.py
```diff
@@ -472,15 +690,31 @@ def my_task(self):
...
+                        proc.send_signal(signal.SIGTERM)
+
+                        stdout, stderr = proc.communicate(timeout=1)
+                        output.extend(stdout.splitlines())
+                        output.extend(stderr.splitlines())
+
+                        output_text = "\n".join(output)
+
+                        self.assertIn("Starting Locust", output_text)
+                        self.assertIn("All users spawned:", output_text)
```
cyberw, line 708: this assert is no longer needed, now that we're already polling for it.

locust/test/test_main.py
```diff
@@ -608,24 +842,30 @@ def test_autostart_wo_run_time(self):
...
             self.assertIn("Shutting down ", stderr)
             self.assertNotIn("Traceback", stderr)
+
             # check response afterwards, because it really isn't as informative as stderr
             d = pq(response.content.decode("utf-8"))
 
             self.assertEqual(200, response.status_code)
             self.assertIn('"state": "running"', str(d))
 
+            if proc.poll() is None:
```
cyberw, line 865: You've already done proc.communicate, so process is guaranteed to be dead already.

locust/test/test_main.py
```diff
@@ -647,21 +887,34 @@ def test_autostart_w_run_time(self):
...
-            # check response afterwards, because it really isn't as informative as stderr
-            d = pq(response.content.decode("utf-8"))
+                self.assertEqual(200, response.status_code)
 
-            self.assertEqual(200, response.status_code)
-            self.assertIn('"state": "running"', str(d))
+                def process_finished():
+                    return proc.poll() is not None
+
+                poll_until(process_finished, timeout=10)
```
cyberw, line 900: No need to poll for process finishing, that's what proc.communicate does anyway.

locust/test/test_main.py
```diff
@@ -803,26 +1077,23 @@ def my_task(self):
...
 
                 self.assertIn("Starting Locust", stderr)
                 self.assertIn("Shape test starting", stderr)
                 self.assertIn("Shutting down ", stderr)
                 self.assertIn("autoquit time reached", stderr)
-                # check response afterwards, because it really isn't as informative as stderr
-                self.assertEqual(200, response.status_code)
-                self.assertTrue(success, "got timeout and had to kill the process")
+
+                if proc.poll() is None:
```
cyberw, line 1092: proc.communicate() already called, no need to poll.

locust/test/test_main.py
```diff
@@ -144,20 +250,28 @@ def my_task(self):
...
+                    stderr=subprocess.PIPE,
                     text=True,
                 )
-                gevent.sleep(1)
+
+                poll_until(lambda: web_interface_ready("localhost", port), timeout=20)
+
             finally:
-                os.remove("locust.conf")
+                if os.path.exists("locust.conf"):
```
cyberw, line 266: this isn't necessarily an improvement. if locust.conf doesnt exist, something is really broken and we'd wanna crash ASAP :)

locust/test/test_main.py
```diff
@@ -3,30 +3,218 @@
...
+    if port is not None:
+        locust_command += ["--web-port", str(port)]
+
+    proc = subprocess.Popen(
+        locust_command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        bufsize=1,
+        text=True,
+        env={**os.environ, "PYTHONUNBUFFERED": "1"},
```
cyberw, line 194: This is probably unnecessary, because you're setting it on this line too https://github.com/locustio/locust/pull/2899/files#diff-81a8e34f7c9a9b06d20ff25a8b8d6f6fe983707f8ba66780659df4fe4fffa2aaR247

locust/test/test_main.py
```diff
@@ -69,24 +258,21 @@ def assert_run(self, cmd: list[str], timeout: int = 5) -> subprocess.CompletedPr
...
+            manager.proc.wait()
+
+        output = "\n".join(manager.output_lines).strip()
+
+        self.assertTrue(output.startswith("Usage: locust [options] [UserClass ...]"))
         self.assertIn("Common options:", output)
         self.assertIn("-f <filename>, --locustfile <filename>", output)
         self.assertIn("Logging options:", output)
-        self.assertIn("--skip-log-setup      Disable Locust's logging setup.", output)
+        self.assertIn("--skip-log-setup", output)
```
cyberw, line 270: it is far from being important, but why did you change this assertion? :)

locust/test/test_main.py
```diff
@@ -509,6 +798,7 @@ def tick(self):
                 stdout=PIPE,
                 stderr=PIPE,
                 text=True,
+                env=os.environ.copy(),
```
cyberw, line 788: this is essentially a no-op :) subprocess always inherits a copy of the current environ :)

locust/test/test_main.py
```diff
@@ -359,37 +359,80 @@ def my_task(self):
...
-            from locust import User, task, constant, events
-            class TestUser(User):
-                wait_time = constant(3)
-                @task
-                def my_task(self):
-                    print("running my_task()")
-        """)
+    def test_webserver_multiple_locustfiles(self):
+        with (
+            mock_locustfile(content=MOCK_LOCUSTFILE_CONTENT_A) as mocked1,
```
cyberw, line 364: I was not aware this syntax existed. Nice.

locust/test/test_main.py
```diff
@@ -29,6 +43,96 @@ def is_port_in_use(port: int) -> bool:
...
+        if platform.system() == "Windows":
+            creation_flags = subprocess.CREATE_NEW_PROCESS_GROUP
+
+        self.process = subprocess.Popen(
+            self.args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            bufsize=1,
+            creationflags=creation_flags,  # Windows-specific creation flag
```
cyberw, line 92: these comments are a bit excessive. and you can do it as a oneliner to avoid having to store another variable:
`creationflags=subprocess.CREATE_NEW_PROCESS_GROUP if platform.system() == "Windows" else 0`

locust/test/test_main.py
```diff
@@ -381,22 +699,27 @@ def test_default_headless_spawn_options(self):
...
-            )
-            stdout, stderr = proc.communicate(timeout=4)
-            self.assertNotIn("Traceback", stderr)
-            self.assertIn('Spawning additional {"UserSubclass": 1} ({"UserSubclass": 0} already running)...', stderr)
-            self.assertEqual(0, proc.returncode)
+            ) as popen_ctx:
+                try:
+                    popen_ctx.process.wait(timeout=4)
+                except subprocess.TimeoutExpired:
+                    print("Process timed out")
```
cyberw, line 710: Do you really want to catch this exception? Is it not more clear if you just allow it to crash?

PR #2900, opened 2024-09-10
fix: Add more worker info to LocalRunner for parity with others

tarkatronic:
This prevents the `custom_messages` example from failing when running locally with `AttributeError`s for `environment.runner.worker_count` and `environment.runner.clients`.

ref: https://github.com/locustio/locust/blob/7c85b7988c643ee8f688c8bb6919272c38fbcf7b/examples/custom_messages.py#L45-L48

### DIFF
--- a/examples/custom_messages.py
+++ b/examples/custom_messages.py
@@ -1,5 +1,5 @@
 from locust import HttpUser, between, events, task
-from locust.runners import MasterRunner, WorkerRunner
+from locust.runners import LocalRunner, MasterRunner, WorkerRunner
 
 import gevent
 
@@ -44,8 +44,12 @@ def on_test_start(environment, **_kwargs):
 
         worker_count = environment.runner.worker_count
         chunk_size = int(len(users) / worker_count)
+        if isinstance(environment.runner, LocalRunner):
+            workers = [environment.runner]
+        else:
+            workers = [environment.runner.clients]
 
-        for i, worker in enumerate(environment.runner.clients):
+        for i, worker in enumerate(workers):
             start_index = i * chunk_size
 
             if i + 1 < worker_count:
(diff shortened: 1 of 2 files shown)


cyberw:
Hmm... I understand the need in this example, but if I ask a LocalRunner how many workers/clients it has connected I'd expect it to be zero, not one, so this might add as much confusion as it solves...

tarkatronic:
> Hmm... I understand the need in this example, but if I ask a LocalRunner how many workers/clients it has connected I'd expect it to be zero, not one, so this might add as much confusion as it solves...

I can see that perspective. The way I view it though, when running locally, there is one worker: The runner itself. Nothing happens without a worker right? In fact, it does even have a worker: https://github.com/locustio/locust/blob/7c85b7988c643ee8f688c8bb6919272c38fbcf7b/locust/runners.py#L441

The real goal here is to simplify load testing code that I'm writing. Currently, my option is 
```python
        try:
            worker_count = environment.runner.worker_count
            workers = environment.runner.clients
        except AttributeError:  # For running locally
            worker_count = 1
            workers = [environment.runner]
```
Which feels like a hacky workaround for a bug.

tarkatronic:
This is also just making for a more complete picture, in light of this comment, directly above the change: https://github.com/locustio/locust/blob/7c85b7988c643ee8f688c8bb6919272c38fbcf7b/locust/runners.py#L436-L437

cyberw:
Hmm... Yea I guess the `worker_count = 1` makes sense in a way, but storing something in `self.clients` (which I think is what you were after, not `self.workers`, right?) is just too confusing for a runner that doesnt actually have any clients.

cyberw:
I get why this would make it easier for this particular use case, but I don't want LocalRunner to "lie" about its state. 

LocalRunner has no workers (so saying it has one is questionable) and settings self.workers to be a list containing itself could have very strange consequences down the line, if it were ever used by someone not aware of this feature. Its not even the same type (`WorkerNode`) as in MasterRunner...

tarkatronic:
@cyberw I've updated to remove `self.workers` (especially since that was the wrong name anyhow) and also fixed the example to work properly with `LocalRunner`

PR #2902, opened 2024-09-13
Add Error Message for Accessing Login Manager without --web-login

andrewbaldwin44:


locust/web.py
```diff
@@ -518,6 +514,16 @@ def update_user():
...
 
+    @property
+    def login_manager(self):
+        if self.web_login:
+            login_manager = LoginManager()
+            login_manager.init_app(self.app)
+            login_manager.login_view = "login"
+            return login_manager
+
+        raise AttributeError("The login_manager is only available with --web-login.\n")
```
andrewbaldwin44, line 525: We could also quit locust here I was just thinking maybe an AttributeError is a bit nicer

locust/web.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import mimetypes
 import os.path
+import sys
```
cyberw, line 8: I think this import was added accidentally? I wonder why it doesnt fail linting...

PR #2914, opened 2024-09-26
Url in template arg

fletelli42:


locust/web.py
```diff
@@ -656,6 +656,7 @@ def update_template_args(self):
             "users": users,
             "percentiles_to_chart": stats_module.PERCENTILES_TO_CHART,
             "percentiles_to_statistics": stats_module.PERCENTILES_TO_STATISTICS,
+            "api_base_url": self.environment.parsed_options.api_base_url,
```
andrewbaldwin44, line 659: Think this will break in locust core because there is no argument api_base_url. We also don't really need to affect / don't need locust core users to know about this about. I was thinking something along the lines of `self.template_args = { **self.template_args, ... }` and in the init we would have `self.template_args = {}`. This would allow us to further extend the template args in the future in case we need to

PR #2923, opened 2024-10-03
Decide if ipv6 can work

nc-marco:
**Summary of changes:**

Previously IPV6 was enabled if the kernel supported the IPV6 stack.  This was an issue in the distributed scenario when there is a network component between the master and worker nodes that does not support IPV6.  This is the case for example in AWS EKS cluster where the service resource does not support IPV6 even if the pods in the cluster do.  
To avoid attempting to communicate over IPv6 when it is not possible, some basic logic was added  which does the following:

1. Determine as before if the kernel supports IPv6, but uses the socket method, has_dualstack_ipv6().  This does not have the limitations of has_ipv6() of the socket library (and I imagine the reason why it was avoided) since it actually does a syscall to the kernel which will fail if the kernel does not support IPv6.  We chose to switch to this method (from that of urllib3) since we needed other methods from the basic python socket library.
2. Verify if the master hostname can be resolved as an IPv6 address.  This was deemed as a reasonable way to determine if the network environment supports IPv6 (outside of the OS where locust is running).  No admin would configure AAAA DNS records if the network isn't completely IPv6 ready.
3. If IP addresses are used instead of host names (for example in the bind on the server side), then determine if the address is a valid IPv6 address or not. 

The same checks are done independently on the server and client during socket creation and will require no changes in the use of locust.  For example, currently the bind address of '*' will cause the master node to listen on both IPv6 and IPv4 addresses.  This behaves the same as before while setting master_bind_host to an IPv4 address, such as '0.0.0.0', will prevent zmq from enabling IPv6 on the master since it should only listen on an IPv4 interface.

### DIFF
--- a/locust/rpc/zmqrpc.py
+++ b/locust/rpc/zmqrpc.py
@@ -1,10 +1,11 @@
 from locust.exception import RPCError, RPCReceiveError, RPCSendError
 from locust.util.exception_handler import retry
 
+from socket import IPPROTO_TCP, getaddrinfo, has_dualstack_ipv6
+
 import msgpack.exceptions as msgerr
 import zmq.error as zmqerr
 import zmq.green as zmq
-from socket import getaddrinfo, IPPROTO_TCP, has_dualstack_ipv6
 
 from .protocol import Message
 
--- a/locust/rpc/zmqrpc.py
+++ b/locust/rpc/zmqrpc.py
@@ -4,19 +4,19 @@ from locust.util.exception_handler import retry
 import msgpack.exceptions as msgerr
 import zmq.error as zmqerr
 import zmq.green as zmq
-from urllib3.util.connection import HAS_IPV6
+from socket import getaddrinfo, IPPROTO_TCP, has_dualstack_ipv6
 
 from .protocol import Message
 
 
 class BaseSocket:
-    def __init__(self, sock_type):
+    def __init__(self, sock_type, ipv4_only):
         context = zmq.Context()
         self.socket = context.socket(sock_type)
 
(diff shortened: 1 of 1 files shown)


cyberw:
Fix looks reasonable but we'd need to investigate what's wrong with macos/windows.

nc-marco:
That's very kind of you.  I will attempt to debug it on my own a bit. I
just haven't had the time.  Hopefully by the end of the week.  I have a
more experienced programmer that might give me a hand too. I have
unfortunately more of an infrastructure background.

nc-marco:
So, thanks to my friend's mac (I run linux of course), we were able to see what the issue was.  It was raising and resolver error.  I need to dig probably deeper to understand why it happens on macOS and windows and not Linux, but putting in a try/except block and now it passes on these OS's.  I verified it still does what I want it to do in EKS, so I'm happy with it.  I'll leave it up to you to decide if it is OK.

cyberw:
I think it looks OK! Any specific reason you renamed the socked module?

    import socket as csocket

nc-marco:
No strong feelings about that.  With the addition of gaierror, the list
after import ... was getting rather long I felt.  csocket, referring to the
traditional c socket, I thought would prevent the confusion with zmq
socket.

But as I said, I have no strong feelings.  If you prefer it as it was
before, or with a different alias, that's fine too.

cyberw:
maybe `from socket as import gaierror, has_dualstack_ipv6`?

nc-marco:
Sure.  Will do.

nc-marco:
Thx!

PR #2925, opened 2024-10-06
Add base url path param

### LINKED ISSUE #2909: Hard coded path make it impossible to host the UI on a path (instead of the domain root)
### Prerequisites

### Description

I am trying to host the locust UI on a path like this:
`www.mydomain.com/tests/locust`

But the full UI paths ([like this one ](https://github.com/locustio/locust/blob/7c85b7988c643ee8f688c8bb6919272c38fbcf7b/locust/webui/src/components/Reports/Reports.tsx#L18)) prevent me from doing that.

Can these paths be relative (with or without a configurable `<base>` url)?

Happy to work on it if this is of value/agreeable.

Grommash9:
Closes #2909

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -36,6 +36,7 @@ from .stats import (
 from .user.inspectuser import print_task_ratio, print_task_ratio_json
 from .util.load_locustfile import load_locustfile
 from .util.timespan import parse_timespan
+from .util.url import normalize_base_url
 
 # import external plugins if  installed to allow for registering custom arguments etc
 try:
@@ -168,6 +169,10 @@ def main():
 
     # parse all command line options
     options = parse_options()
+
+    # Normalize the base URL
+    options.base_url = normalize_base_url(options.base_url)
+    options.master_base_url = normalize_base_url(options.master_base_url)
 
     if options.headful:
         options.headless = False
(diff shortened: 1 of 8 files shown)


Grommash9:
WEB UI is working fine, I can run it, connect to it, I can have worker and tasks is running, I can see all the stats

But I can use any param for base url for worker now, it will be working anyway and it's confusing to me, because I am expecting it to work only if the both are same

Maybe somebody can double check it after me? I am not sure I am doing it in the right way

`
version: '3'

services:
  master:
    build:
      context: /home/oleksandr/vscode_projects/locust/
      dockerfile: Dockerfile
    ports:
     - "8089:8089"
    volumes:
      - ./:/mnt/locust
    command: -f /mnt/locust/locustfile.py --master -H http://master:8089 --base-url /locust
  
  worker:
    build:
      context: /home/oleksandr/vscode_projects/locust/
      dockerfile: Dockerfile
    volumes:
      - ./:/mnt/locust
    command: -f /mnt/locust/locustfile.py --worker --master-host master --base-url /locust

`

I am using docker file like that for my local tests

So I am confused what it's also working like that:

image

But I am expecting to get an error because I am not able to connect to master, or am I missing something?

Grommash9:
I we able to do it with tcp and ZMQ? 

I know we should do some changes here, but how?

image

cyberw:
"Base url" is not a good name, because that is what is used for HttpUser / -H parameter. Maybe ui_base_url?

Needs tests.

cyberw:
> I we able to do it with tcp and ZMQ?
> 
> I know we should do some changes here, but how?

Maybe I'm confused, but why would this relate to zmq at all?

Grommash9:
It's not only about UI, service will be running on that path, so workers should use it to be able to connect with master 

I am not sure so far are we able to do that or not

locust/main.py
```diff
@@ -491,20 +497,21 @@ def ensure_user_class_name(config):
         else:
             web_host = options.web_host
         if web_host:
-            logger.info(f"Starting web interface at {protocol}://{web_host}:{options.web_port}")
+            logger.info(f"Starting web interface at {protocol}://{web_host}:{options.web_port}{options.base_url}")
```
andrewbaldwin44, line 500: I think `url` doesn't actually make sense in this case, URL refers to the entire composition (e.g. protocol://domain:port/path?query=params), in this case `base_path` would be more accurate

locust/web.py
```diff
@@ -165,7 +166,7 @@ def send_assets(path):
 
             return send_from_directory(directory, path)
 
-        @app.route("/")
+        @app.route(f"{self.base_url}/")
```
andrewbaldwin44, line 169: If `self.base_url` is `/` (as it is by default) we will end up with `@app.route("//")` right?

PR #2944, opened 2024-10-16
LocustBadStatusCode without url param in fasthttp

swaalt:
a new function _verify_status was created and a LocustBadStatusCode class is sent as a response to the function which removes the url leaving only the code.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -586,6 +586,12 @@ class ErrorResponse:
     def raise_for_status(self):
         raise self.error
 
+class LocustBadStatusCode(ConnectionError):
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


cyberw:
👍 just add a test case to test_fasthttpuser (or add some validation it to some existing one) and format using ruff and I’ll merge.

cyberw:
Hmm... looks like _every_ Python version failed, that's a bit worrying, but may still be irrelevant.

Can you write the test this way instead? https://github.com/locustio/locust/blob/a1b0a047338dbd55db06f047ad4b85f60f3a2125/locust/test/test_http.py#L112

cyberw:
No I didnt mean the name, I meant the way of testing :)

PR #2947, opened 2024-10-20
Report name

### LINKED ISSUE #2931: Downloading report should provide a meaningful human name
### Prerequisites

### Description

Hi,
When downloading the report through the Web UI, the default name provided is `report_<timesatmp>.html`.

I suggest this default format: `<START DATE YYYY-MM-DD-HH:MM> - Locust - <SCRIPT NAME> - <HOSTNAME>.html`

This format could help sorting and finding previous reports for an human beeing

obriat:
- Add host, start time and scenario name to report filename (should be easier to find and sort)
- Add human friendly duration in html report
- Add some doc about dev and debugging

Close #2931


cyberw:
Can you do this without adding the humanfriendly dependency? Seems a bit much to add an entire package just to format one date...

There's already some info about debugging in running-in-debugger.rst, can you maybe add your info there instead (and link to it from developing-locust.rst)

@andrewbaldwin44 Any opinions on the TS stuff?

locust/web.py
```diff
@@ -318,17 +317,25 @@ def stats_report() -> Response:
             )
             if request.args.get("download"):
                 res = app.make_response(res)
-                res.headers["Content-Disposition"] = f"attachment;filename=report_{time()}.html"
+                host = self.environment.host or "no host"
+                res.headers["Content-Disposition"] = (
+                    f"attachment;filename=Locust - {format_safe_timestamp(self.environment.stats.start_time)} - "
+                    + f"{self.environment.locustfile} - {host}.html"
+                )
```
andrewbaldwin44, line 324: I don't know under which scenario there won't be a host, but maybe in the case that there isn't a host we could exclude it?

Perhaps also consider using an underscore to separate the words as we had before? `Locust_starttime_locustfile_host.html`? Spaces aren't necessarily friendly on all systems

andrewbaldwin44:
@obriat Could you also check that this works with the `--html` flag?

locust/web.py
obriat: I don't remember how I manage to get a null, maybe by going directly to the path without launching the scenario, but I really stumble on this case (which add a parenthesis at the end of the filename).

About the spaces vs underscore, as an old-timer I would agree, but at the current time I don't see any OS that doesn't manage spaces or utf-8 chars.

But at the end your have the final say.

obriat: Good point about the shell: go for underscores instead of spaces.
Ok for host, but it's really an edge case that should not occur. 
I'll made a new commit soon.

obriat:
> Can you do this without adding the humanfriendly dependency? Seems a bit much to add an entire package just to format one date...
> 
> There's already some info about debugging in running-in-debugger.rst, can you maybe add your info there instead (and link to it from developing-locust.rst)
> 
> @andrewbaldwin44 Any opinions on the TS stuff?
I'm not a Python expert, but does this package adds a overhead to the runners or just the report part? 
I'll move my doc to this page

JavierUhagon:
> 
> I'm not a Python expert, but does this package adds a overhead to the runners or just the report part?
> 

Hey there! Not a maintaner, but, I don't think the issues is adding a overhead, it's having to add a full package (that hasn't been updated since 2020) just to format a date for a report, I'd say that adding your own implementation would be better :)

You can always check how humanfriendly does it!
https://github.com/xolox/python-humanfriendly/blob/6758ac61f906cd8528682003070a57febe4ad3cf/humanfriendly/__init__.py#L402

It really isn't hard nor long!

locust/web.py
obriat: > `host = f" - {self.environment.host}" if host else ""`?

It should be `host = f"_{self.environment.host}" if self.environment.host else ""`, no ?

obriat:
> @obriat Could you also check that this works with the `--html` flag?
Hi,
I think that I response to every suggestions and fix failed tests.

About the `--html` flag, this option needs a filename as input (not mandatory but throw an exception on writing file). So what should we do ? 

- Keep the current behaviour?
- Use the default generated filename if no input is given (but detects if the input is a dir and not a full filename+path)?

locust/util/date.py
```diff
@@ -1,5 +1,194 @@
...
+    Format a timespan between two timestamps as a human readable string.
+    Taken from xolox/python-humanfriendly
+
+    :param start_unix_timestamp: Start timestamp.
+    :param end_unix_timestamp: End timestamp.
+
+    """
+    # Common time units, used for formatting of time spans.
+    time_units = (
+        dict(divider=1e-9, singular="nanosecond", plural="nanoseconds", abbreviations=["ns"]),
```
andrewbaldwin44, line 26: Sorry maybe it wasn't clear, but I think an argument for using our own solution, rather than the library, is because as you can probably see, the library handles a lot more cases than we probably need. In Locust's case, we probably never need nanoseconds, microseconds, weeks, or years. I think we could go with a simpler approach, for example:
```
def format_duration(start_time, end_time):
    time_diff = end_time - start_time

    days = time_diff.days
    seconds = time_diff.seconds
    hours = seconds // 3600
    minutes = (seconds % 3600) // 60
    seconds = seconds % 60

    return f"{days} days, {hours} hours, {minutes} minutes, {seconds} seconds"
```
(disclaimer: ChatGPT)

andrewbaldwin44:
Hey @obriat, for the `--html <filename>` flag, I just meant to double-check it's still working as expected, since it also uses the `get_html_report` function, but looks fine so no need to change anything :+1:

locust/util/date.py
obriat: Should I keep some methods in order to keep it human friendly (plural, no output for 0 values, ...)?

andrewbaldwin44: Good idea!
```
def format_duration(start_time, end_time):
    time_diff = end_time - start_time

    days = time_diff.days
    seconds = time_diff.seconds
    hours = seconds // 3600
    minutes = (seconds % 3600) // 60
    seconds = seconds % 60

    time_parts = [
        (days, "day"),
        (hours, "hour"),
        (minutes, "minute"),
        (seconds, "second")
    ]

    parts = [f"{value} {label}{'s' if value != 1 else ''}" for value, label in time_parts if value > 0]

    return ', '.join(parts) if parts else "0 seconds"
```
Think something like this would do the trick (disclaimer: haven't tested it)

obriat: Since I really want to keep the last "and" 😄 , here a working proposition: 
```python
def format_duration(start_time, end_time):
    seconds = end_time - start_time
    days = seconds // 86400
    hours = seconds // 3600
    minutes = (seconds % 3600) // 60
    seconds = seconds % 60

    time_parts = [(days, "day"), (hours, "hour"), (minutes, "minute"), (seconds, "second")]

    parts = [f"{value} {label}{'s' if value != 1 else ''}" for value, label in time_parts if value > 0]

    return " and ".join(filter(None, [", ".join(parts[:-1])] + parts[-1:])) if parts else "0 seconds"
```

0: 0 seconds
666: 11 minutes and 6 seconds
1332: 22 minutes and 12 seconds
1998: 33 minutes and 18 seconds
2664: 44 minutes and 24 seconds
3330: 55 minutes and 30 seconds
3996: 1 hour, 6 minutes and 36 seconds
4662: 1 hour, 17 minutes and 42 seconds
5328: 1 hour, 28 minutes and 48 seconds
5994: 1 hour, 39 minutes and 54 seconds
6660: 1 hour and 51 minutes
7326: 2 hours, 2 minutes and 6 seconds
7992: 2 hours, 13 minutes and 12 seconds

andrewbaldwin44: Maybe we can do something like so:
```
def format_duration(start_time, end_time):
    time_diff = end_time - start_time

    days = time_diff.days
    seconds = time_diff.seconds
    hours = seconds // 3600
    minutes = (seconds % 3600) // 60
    seconds = seconds % 60

    time_parts = [
        (days, "day"),
        (hours, "hour"),
        (minutes, "minute"),
        (seconds, "second"),
    ]

    parts = [
        f"{value} {label}{'s' if value != 1 else ''}"
        for value, label in time_parts
        if value > 0
    ]
    parts[-1] = "and " + parts[-1]

    return ", ".join(parts) if parts else "0 seconds"
```
To avoid iterating over the list multiple times and help keep things readable?

If you have time, maybe a couple of basic unit tests for the function would also be good?

obriat: It returns an error 😞 and I didn't find the way to fix it, so I push my version (which I tested with a loop).
I'll be happy to have some tips about unit tests, I'll add them if I had some more spare times

obriat: I add tests, could not find a way to use parametrize with class / self 😞

cyberw:
I tested --html just for fun :) It doesn't seem to have the duration? 

image

Maybe add a search for ' seconds' here to ensure it works https://github.com/locustio/locust/blob/eba774c8db030a103fd82bcdf26cef0d5376415e/locust/test/test_main.py#L1183

cyberw:
Btw, resolve the conflicts in launch.json whatever way you like, dont worry about my changes :)

obriat:
I’ll fix the conflict, a bit strange about the —html I’m pretty sure I tested it :(
Is there a way to make a automated test about it?

obriat:
> I tested --html just for fun :) It doesn't seem to have the duration?

"It works on my PC": 
```shell
git branch --show-current
report-name
poetry run locust -f examples/basic.py  -u 1 -H http://192.168.64.100:8089 --headless -u 1 -t 30s --html ~/headless-locust-report.html
grep -Eo '"duration": "[^"]*"'  ~/headless-locust-report.html
"duration": "29.067671298980713 seconds"
```
The rendered html is also correct  "During: 03/11/2024 23:01:02 - 03/11/2024 23:01:31 (29.067671298980713 seconds)"

(Yes, I know, I got a round problem 😄 )

cyberw:
Yea sorry, missing the seconds was my bad, I must have messed up rebuilding the UI.

Test looks good, but we dont have pytest as a dependency so looks like it is failing. I think we'll be going in the pytest direction in not too long a while, so feel free to add it to the dev dependencies (something like `poetry add --group dev pytest`)

obriat:
Done,  `pyproject.toml` changes are readable, but I don't know about the `poetry.lock` ones.

cyberw:
Sorry, it was the wrong group, it was supposed to be added to "test", not "dev". I'll fix it in master. Thanks!

Issue #2955, opened 2024-10-26
Allow to exclude some statistics from aggregation


tyge68:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am suggesting a new feature, not asking a question

### Description

Currently if our locust script track additional statistics via custom type event, it cannot be excluded from the aggregated line.

Ideally it should be possible to define via script somewhere (via annotation etc...) which metrics are ignored for the aggregation computation, or via ui (similarly to the columns selector).

The reason is that those metrics are mainly to track statistic like "cache missed etc..." detected from CDN response header and so shouldn't be part of the aggregated statistic line data.

**Note**: the solution is pretty trivial if we hack in the code as it's mainly to add an condition on this line to exclude specific method or name
https://github.com/locustio/locust/blob/master/locust/stats.py#L221

What is not trivial is to do it the locust way for me (using annotations or any means to make it configurable)
So if someone have more experience in that, it could be good to have some suggestions.

Possible options I see:
- having a command line config (which could be also part of the "locust.conf" file, environment variables)
- having a mean to set this up during script loading via "@events." syntax etc..

What could be filtered out from aggregation are then obviously the following 2 options:
- method matching regular expression
- name matching regular expression

AchilleasTsiamoulis:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

PR #2956, opened 2024-10-26
feat: #2955 Allow to exclude some statistics from aggregation

tyge68:
Here is a sample implementation allowing to set the `--exclude-from-aggregation` argument option.

`locust --exclude-from-aggregation "CUSTOM"`

**Note**:
- it can be also set using environment variable **LOCUST_EXCLUDE_FROM_AGGREGATION**
- it can be set via your init script under the `environment.stats` object property **exclude_from_aggregation**

Screenshot 2024-10-26 at 11 45 47


tyge68:
If anyone has ideas why Mypy step fails, the error seems unrelated to the change, at leat the line for which the error is triggered wasn't updated by this PR at all.

cyberw:
Does mypy (using 1.11.2) work for you locally?

tyge68:
@cyberw no it seems to be the same issue, it's really not clear why this tool is failing especially as those lines are not touched by this PR

tyge68:
@cyberw thanks for your feedback, I have updated it now and also changed to allow setting that via command arguments.
Still the same Mypy is happening which is really unclear. I am not too familiar with that tool and what is the purpose of it, why it fails etc... might be worth checking with Mypy experts.

cyberw:
Can you try grabbing this change: https://github.com/locustio/locust/commit/74d5de457fb68a26fe49670aec98e26feccead6b

It seems to have been a "real" issue with the type hints in the code, but I still don't understand why your PR exposed it :)

cyberw:
Hi, sorry I haven't been active here recently. I cant quite figure out the issue mypy is having with the type hints (I can't really tell if it is an issue in our code and just exposed by your changes or a "real" issue with your code) 

I don't mind this feature, but I don't really care either way it so I can't spend time on it.

Issue #2849, opened 2024-08-14
tag doesn't work fine


evanchaoli:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

With this simple locusfile:

```
from locust import HttpUser, task, tag


class RevisionApiUser(HttpUser):
    @tag('revision')
    @task
    def get_commit_list(self):
        print("revision")


class CBotApiUser(HttpUser):
    @tag('cbot')
    @task
    def get_commits(self):
        print("cbot")
```

Then run locust with a tag specified:

```
$ locust --tags cbot
[2024-08-14 15:05:43,870] C02GM2X6MD6R/ERROR/locust.user.task: No tasks defined on RevisionApiUser. Use the @task decorator or set the 'tasks' attribute of the User (or mark it as abstract = True if you only intend to subclass it)
Traceback (most recent call last):
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/user/task.py", line 335, in run
    self.schedule_task(self.get_next_task())
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/user/task.py", line 479, in get_next_task
    raise Exception(
Exception: No tasks defined on RevisionApiUser. Use the @task decorator or set the 'tasks' attribute of the User (or mark it as abstract = True if you only intend to subclass it)
```

From the UI, the test looks run properly, only `CBotApiUser` is executed.  However, the terminal keeps scrolling the error message.

### Command line

locust --tags cbot

### Locustfile contents

```python3
from locust import HttpUser, task, tag


class RevisionApiUser(HttpUser):
    @tag('revision')
    @task
    def get_commit_list(self):
        print("revision")


class CBotApiUser(HttpUser):
    @tag('cbot')
    @task
    def get_commits(self):
        print("cbot")
```

### Python version

3.9.19

### Locust version

locust 2.31.2 from /Users/chaol/venv/3.9/lib/python3.9/site-packages/locust (Python 3.9.19, OpenSSL 3.3.0)

### Operating system

MacOS

cyberw:
This is kind of intentional. You are spawning a user with no tasks (after filtering tags) which is usually a mistake, so we should log an error message (although I guess it would be enough to do it once, PRs fixing that would be welcome.)

[...]

PR #2957, opened 2024-10-28
--tags filter issue(#2849) fix 

vijay59122:


cyberw:
Fix ruff and tests first, then I'll have a look.

Issue #2852, opened 2024-08-15
Overwriting weight by config-users may lead to crash


evanchaoli:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

In the attached locustfile, there are no "weight" defined for both user classes, which works fine.

However, if I overwrite weights with config-users options, it will crash:

```
$ locust -u 2 --config-users '[{"user_class_name": "CBotApiUser", "weight": 0}]'
[2024-08-15 11:15:41,053] C02GM2X6MD6R/INFO/locust.main: Starting web interface at http://0.0.0.0:8089
init: RevisionApiUser, 1
init: CBotApiUser, 0
[2024-08-15 11:15:41,071] C02GM2X6MD6R/INFO/locust.main: Starting Locust 2.31.2
[2024-08-15 11:15:47,674] C02GM2X6MD6R/INFO/locust.runners: Ramping to 2 users at a rate of 1.00 per second
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/runners.py", line 544, in <lambda>
    lambda: self._start(user_count, spawn_rate, wait=wait, user_classes=user_classes)
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/runners.py", line 493, in _start
    for dispatched_users in self._users_dispatcher:
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/dispatch.py", line 133, in __next__
    users_on_workers = next(self._dispatcher_generator)
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/dispatch.py", line 168, in _dispatcher
    yield self._add_users_on_workers()
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/dispatch.py", line 300, in _add_users_on_workers
    for user in self._user_generator:
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/dispatch.py", line 388, in _user_gen
    yield next(weighted_users_gen)
  File "/Users/chaol/venv/3.9/lib/python3.9/site-packages/locust/dispatch.py", line 32, in _kl_generator
    heap = [(x * log2(x / (x + 1.0)), x + 1.0, x, name) for name, x in users]
[...]

PR #2971, opened 2024-11-07
Overwrite weight by config user before environment weight validation

tdadela:
Current order:
1. Removing user classes with weight zero. (_remove_user_classes_with_weight_zero()).
1. Overwriting weights of user classes by CL config.

New order:
1. Overwriting weights of user classes by CL config.
1. Removing user classes with weight zero. (_remove_user_classes_with_weight_zero()).

Related to: #2852.


cyberw:
Looks ok. Can you add some sort of test case that shows why this was needed/what benefit it provides?

cyberw:
`*poke*` ;)

PR #2977, opened 2024-11-09
Regenerate heartbeat by setting heartbeat_liveness instead of HEARTBEAT_LIVENESS

tdadela:
Currently, the heartbeat_liveness parameter is used only to initialize value self.heartbeat.
After receiving a heartbeat value is set to HEARTBEAT_LIVENESS: `c.heartbeat = HEARTBEAT_LIVENESS`.

I guess that original motivation to add this parameter was to give possibility to adjust how many missing heartbeats WorkerNode is considered missing.
To obtain this aim, value after regeneration should be equal to value passed to the constructor.

Another possible solution: complete removal of the heartbeat_liveness parameter from constructor.

Motivation behind change `self._local_worker_node = WorkerNode(id="local", heartbeat_liveness=HEARTBEAT_LIVENESS)` — default value of constructor parameter is evaluated during function creation but value HEARTBEAT_LIVENESS can change after that.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -439,7 +439,7 @@ class LocalRunner(Runner):
         self.client_id = socket.gethostname() + "_" + uuid4().hex
         self.worker_count = 1
         # Only when running in standalone mode (non-distributed)
-        self._local_worker_node = WorkerNode(id="local")
+        self._local_worker_node = WorkerNode(id="local", heartbeat_liveness=HEARTBEAT_LIVENESS)
         self._local_worker_node.user_classes_count = self.user_classes_count
 
         # register listener that's logs the exception for the local runner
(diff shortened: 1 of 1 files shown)


cyberw:
I like what you're doing, but would prefer we rewrite it so that we _compare_ the missing heartbeat count with HEARTBEAT_LIVENESS here: https://github.com/locustio/locust/blob/0a1a3d650f9568465e291463b5534e5277f4fd4c/locust/runners.py#L924 (instead of checking if it it below zero). Would that work?

Do you think you could do that? We should probably rename the variable to `.missed_heartbeats` and HEARTBEAT_LIVENESS to something like MAX_MISSED_HEARTBEATS or something...

cyberw:
`*another poke*` :)

PR #2978, opened 2024-11-09
Update ruff mypy

tdadela:
Unpin `poetry`:  `ruff` is pinned to one particular version to keep the same version as in .pre-commit-config.yaml. I don't see reasons for poetry to be pinned. (Especially since adding poetry.lock file to the repo).
Bump `ruff`: there were a lot of improvements over the last half year (including better detection of old style string interpolation #2974).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.3.7  # Pin the same version in the pyproject.toml file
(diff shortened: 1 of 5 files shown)


cyberw:
Pinning ruff/mypy was because new versions of those can cause a new build to fail with no source change. But its probably a smaller issue than accidentally having really old versions..

tdadela:
Makes sense.
Just to clarify and be sure that we are on the same:
`poetry install` uses the exact versions from  `poetry.lock`.
So someone will install newer version of mypy only if that person intentionally execute `poetry update` (or add new dependency or execute other similar command).

Issue #2987, opened 2024-11-15
No Tasks executed when mirgrating to config


vkuehn:
### Prerequisites

- [X] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [X] I am reporting a bug, not asking a question

### Description

locust --headless --users 10 --spawn-rate 1 -H http://localhost:3000
locustfile.py looks like that

from locust import HttpUser, task

class HelloWorldUser(HttpUser):
    @task
    def hello_world(self):
        self.client.get("/health")
but putting the same code in a local.py and this to a local.conf

locustfile = local.py
headless = true
master = true
expect-workers = 3
host = "http://localhost:3000"
users = 3
spawn-rate = 1
run-time = 1m

that command runs but does not bring back any results

locust --config local.conf

I have asked at stackeoverflow and read the documentation. No answer yet found so I guess it's a bug

### Command line

locust --config local.conf

### Locustfile contents

```python3
local.py

from locust import HttpUser, task

class HelloWorldUser(HttpUser):
    @task
    def hello_world(self):
        self.client.get("/health")

local.conf
locustfile = local.py
headless = true
master = true
expect-workers = 3
host = "http://localhost:3000"
users = 3
spawn-rate = 1
run-time = 1m

locust --config local.conf
```

### Python version

3.11.10

### Locust version

2.32.1

### Operating system

Mac OS 15.1 Build 24B83

Hchanni:
What Output do you get? You should be seeing some form of 3 workers waiting to get ready..

 I've tested with the same Locust version and Python version and have not been able to replicate.

andrewbaldwin44:
Hey @vkuehn, in your configuration file you have two arguments stated; `master=  true` - which sets locust to run in [distributed mode](https://docs.locust.io/en/stable/running-distributed.html), and `expect-workers = 3` - which is used in headless mode to wait for workers before starting the test (see [expect-workers](https://docs.locust.io/en/stable/running-distributed.html#expect-workers-number-of-workers)).

So in order for you test to run, you will either need to start three workers, or you can remove these flags and run in standalone. If you are still having trouble or have any questions, feel free to reach out to us on slack!

andrewbaldwin44:
Awesome! PRs are welcome if you think something is unclear in the docs!

PR #2990, opened 2024-11-18
Update configuration.rst to show minimalistic config example

vkuehn:
As dicussed in the issue #2987 a minimalistic  eample of config would prevent a reoccurence

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -66,10 +66,11 @@ locust.conf
 .. code-block:: ini
 
     locustfile = locust_files/my_locust_file.py
+    headless = true
     host = https://target-system
-    users = 1
-    spawn-rate = 
     run-time = 1m
+    spawn-rate = 
+    users = 1    
 
 Here's a quick example of the configuration files supported by Locust:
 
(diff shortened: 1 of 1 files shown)


cyberw:
> No Options are selected which would make Locust run e.g. headless or distributed

Adding an example is good, but headless and distributed options do exactly what they do on command line, so there's no need to mention them specifically, is there? 

I know they threw you off a little bit, but with an example, hopefully this is less likely to happen.

vkuehn:
@cyberw The reason to add this example was that starting with Locust comming from command line lead into a trap with the existing example., It took me while and help of an issue here to find out that the existing example would simple not work as it relies to TAG, MASTER..
so the intention was to add a most simple working example matching the locust.py example in the docs

cyberw:
I know what you're trying to achieve, but "No Options are selected which would make Locust run e.g. headless or distributed" makes no sense, because headless is completely valid in a config file and distributed settings _can_ be. Plus it is not grammatically correct so I'm not sure what you even mean...

vkuehn:
you are right headless should be in a config file.

cyberw:
I'm not saying it needs to be in the config file, I'm saying this sentence is misleading "No Options are selected which would make Locust run e.g. headless or distributed."

vkuehn:
sorry misunderstood that. Actually the whole sentence could be removed and just add a small example which would be useable when comming from the commandline examples straight

cyberw:
thx!

PR #3001, opened 2024-11-29
Number of requests lower than expected in web UI

### LINKED ISSUE #3000: Number of requests lower than expected in web UI
### Prerequisites

### Description

I have a test.js file which counts the amount of requests processed. I start the server with `node test.js`. A delay of 1 second for each response is added.
```
const http = require('http');

let requestCount = 0;

const server = http.createServer((req, res) => {
  setTimeout(() => {
    console.log(requestCount);
    res.writeHead(200, { 'Content-Type': 'text/plain' });
    res.end(`Request count: ${requestCount}`);
    requestCount++; // Increment the request counter
  }, 1000);
});

const PORT = 3000;
server.listen(PORT, () => {
    console.log(`Server running at http://localhost:${PORT}`);
});
```
When I start locust with only a single master node with one user, within 17 requests the amount differs between what Locust is saying (15) in the web UI and what I see in the web server logs (17). Even after clicking the "Stop" button and waiting several seconds the number in the UI stays on 15. When running `--headless` the number is correct.

### Command line

locust

### Locustfile contents

```python3


class HelloWorldUser(HttpUser):
    @task
    def hello_world(self):
        self.client.get("/")
```
```


### Python version

3.10

### Locust version

2.32.2

### Operating system

Ubuntu 22.04.5 WSL

andrewbaldwin44:
Fixes #3000 

### Issue
The problem here was that was that updating the stats was happening at the same time as the fetch for stat data. So the UI was being updated with stale data, and when the new stat data would come in, the state would be updated immediately, but not the rest of the UI. The solution is to wait for the new stat data to be fetched, then update the entire UI

### Screenshots
image
image

### DIFF
--- a/locust/webui/eslint.config.mjs
+++ b/locust/webui/eslint.config.mjs
@@ -1,15 +1,16 @@
+import { fixupPluginRules } from '@eslint/compat';
+import { FlatCompat } from '@eslint/eslintrc';
+import js from '@eslint/js';
+import typescriptEslint from '@typescript-eslint/eslint-plugin';
+import tsParser from '@typescript-eslint/parser';
+import _import from 'eslint-plugin-import';
+import prettier from 'eslint-plugin-prettier';
 import react from 'eslint-plugin-react';
 import reactHooks from 'eslint-plugin-react-hooks';
-import typescriptEslint from '@typescript-eslint/eslint-plugin';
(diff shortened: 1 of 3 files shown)


cyberw:
Love it. The tests don't tho :)

PR #3005, opened 2024-12-03
Add optional parameter parsing for HTML file name

ktchani:
Added a feature for adding optional parameters to the HTML report file name.

Here is an example:
`locust --headless -u 12 -r 10 -t 10s --host http://127.0.0.1:8000 --html "{u}_{r}_{t}_locust.html"`

This would generate a filename:
`12_10_10_locust.html`

I use locust a lot at work, and in order to keep track of various test conditions among all the report files, I would manually update the filename. With this change, the file names can be more indicative of what the test was for. I would also consider adding P50, avg, etc...

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -705,8 +705,27 @@ See https://github.com/locustio/locust/wiki/Installation#increasing-maximum-numb
         logger.info("Got SIGTERM signal")
         shutdown()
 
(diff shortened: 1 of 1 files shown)


cyberw:
Good idea! Please add documentation and a test case.

PR #3008, opened 2024-12-09
Sam/max run time

samuelebstein:


cyberw:
This doesnt make sense to me, too small a use case...

samuelebstein:
> This doesnt make sense to me, too small a use case...

Slack thread: https://locustio.slack.com/archives/C3NUJ61DJ/p1733854662612209

samuelebstein:
Closed this PR as you can use the `LOCUST_RUN_TIME` env variable or `--run-time` command when using `docker run` to prepopulate the parameter.  It can still be overriden in the webui after that. But at least with it prepopulated, the load test won't run indefinitely.

PR #3012, opened 2024-12-18
Fix issue where empty WebUI property is not parsed correctly

### LINKED ISSUE #3011: Cannot Update Custom options in the Web UI when Default Value is None
### Prerequisites

### Description

When custom options are set through the Locust WebUI, they consistently return `None` when accessed via `self.environment.parsed_options`. In this case, when running `lucust -f tasks.py --username test` the username is `test`. That's correct, but when I set this option through the Locust WebUI, it is `None` which is incorrect.

### Command line

lucust -f tasks.py

### Locustfile contents

```python3
@events.init_command_line_parser.add_listener
def _(parser: LocustArgumentParser):
    parser.add_argument('--user', '--username', dest='username', default=None, help='username')
```

### Python version

3.10

### Locust version

2.32.4

### Operating system

Linux

timhovius:
This resolves #3011

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -285,7 +285,7 @@ class WebUI:
                     if isinstance(parsed_options_value, bool):
                         parsed_options_dict[key] = value == "true"
                     elif parsed_options_value is None:
-                        parsed_options_dict[key] = parsed_options_value
+                        parsed_options_dict[key] = value or None
                     else:
                         parsed_options_dict[key] = type(parsed_options_dict[key])(value)
 


locust/web.py
```diff
@@ -285,7 +285,7 @@ def swarm() -> Response:
                     if isinstance(parsed_options_value, bool):
                         parsed_options_dict[key] = value == "true"
                     elif parsed_options_value is None:
```
andrewbaldwin44, line 287: Maybe we can do:
```
elif parsed_options_value is None and value is None:
  parsed_options_dict[key] = None
```
This way we still get the type casting from the else statement

andrewbaldwin44:
Hey @timhovius! Thanks for reporting and fixing this issue! If you have time, it would be great if you could also add a test case for this scenario?

cyberw:
Hi @timhovius ! Have you had any chance to look at making a test for this?

timhovius:
> Hi @timhovius ! Have you had any chance to look at making a test for this?

Yes I'm looking into this currently.

locust/web.py
timhovius: When the `parsed_options_value` is `None`, type checking does not make sense. I modified the code to catch this  case.

timhovius:
Actually it goes wrong when no default value was set. So I modified the tests to test this behavior and I added an extra test.

locust/web.py
andrewbaldwin44: Type checking indeed doesn't make sense if the `value` is `None`, but take for example if you have an argument with a default value of `None`:
```
parser.add_argument('--number', default=None)
```
Here the `parsed_options_value` will always be `None`, however in some cases the `value` will be a number, in which case keeping the type may be useful

locust/test/test_web.py
```diff
@@ -788,7 +788,59 @@ def _(parser, **kw):
...
+        parsed_options = parse_options(args=["--my-argument", "24"])
+        self.environment.user_classes = [MyUser]
+        self.environment.parsed_options = parsed_options
+        self.environment.web_ui.parsed_options = parsed_options
+        response = requests.post(
+            "http://127.0.0.1:%i/swarm" % self.web_port,
+            data={"user_count": 1, "spawn_rate": 1, "host": "", "my_argument": "42"},
+        )
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(42, my_dict["val"])
```
andrewbaldwin44, line 843: Nice! Thanks for adding tests!

PR #3022, opened 2025-01-08
Update api.rst

MasterKey-Pro:
Add iter_lines method to FastHttpSession for streaming responses


cyberw:
Please do this in the same PR as the other stuff :)

PR #3026, opened 2025-01-09
Move some argument parsing/validation from main.py to argument_parser.py and remove deprecated parameter --hatch-rate

ftb-skry:


locust/argument_parser.py
```diff
@@ -369,6 +371,66 @@ def parse_locustfile_option(args=None) -> list[str]:
...
+        def __call__(self, parser, namespace, values, option_string=None):
+            raise configargparse.ArgumentError(self, err_msg)
+
+    return ErrorRaisingAction
+
+
+# Definitions for some "types" to use with the arguments
+
+
+def timespan(time_str):
```
cyberw, line 389: Cool stuff! Probably wanna add a type hint for the return type at least?

If you do, does the hint carry over to when people do environment.parsed_options.stop_timeout? That'd be cool...

PR #3029, opened 2025-01-12
Update Dockerfile to use Python 3.12 as base

vejmoj1:
Python 3.13 is supported since version 2.32, but the base image was still based on 3.11.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN yarn webui:install --production --network-timeout 60000
 RUN yarn webui:build
 
 # Stage 2: Build Locust package
-FROM python:3.11-slim AS base
+FROM python:3.13-slim AS base
 
 FROM base AS builder
 RUN apt-get update && apt-get install -y git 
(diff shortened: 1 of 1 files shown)


cyberw:
3.11 was the latest supported in a couple of circumstances (e.g. AWS lambda), but it looks like they just released 3.12, so we could go to that. Are you solving a specific issue and would 3.12 be enough for you? 

We usually aim for the most popular version rather than the latest in the container image.

vejmoj1:
Ahh, I see. We just moved our code base to 3.13, and that policy was enforced for our entire Python code base, so we had an issue with it during our load test image builds, but I believe we can work around it the same way we did for Python 3.12 as we do not use any Python 3.13 specific feature.

Let me know, and I can adjust this PR to update to 3.12.

cyberw:
> Let me know, and I can adjust this PR to update to 3.12.

Great! Adjust it and then I’ll merge it soon!

PR #3048, opened 2025-02-18
FastHttpUser: Accept brotli and zstd compression encoding

kamilbednarz:
Hey, I was wondering if we can extend the locust defaults for Accept-Encoding header with `brotli` and `zstd` to get closer to what modern browsers do:
* Safari: gzip, deflate, br 
* Firefox: gzip, deflate, br, zstd
* Chrome: gzip, deflate, br, zstd

Underlying python requests libs handle them well, I tested this by overwriting the header.


cyberw:
Looks great! Will merge soon, just gotta get a release out with the build changes first.

PR #3082, opened 2025-03-21
Add LICENSE file for Locust Cloud and reference it from the main license

cyberw:
In the future, Locust Cloud files will be added under /proprietary instead of having them in a separate repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,9 @@
-The MIT License
-
 Copyright (c) 2009-2025, Carl Byström, Jonatan Heyman, Lars Holmberg
+
+* Content that resides under the "proprietary/" directory of this repository, is licensed under the license defined in "proprietary/LICENSE".
+* Content outside of the above mentioned directory is licensed as defined below.
+
+The MIT License
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/proprietary/LICENSE
+++ b/proprietary/LICENSE
@@ -1,0 +1,38 @@
+Copyright (c) 2024-present Locust Technologies Inc.
+
+This software and associated documentation files (the "Software") may only be
+used in production, if you (and any entity that you represent) have agreed to,
+and are in compliance with, the Locust Cloud Terms of Service, available
+at https://locust.cloud/terms, or other agreement governing the use of the Software,
+as agreed by you and Locust Technologies, and otherwise have a valid Locust Cloud
+subscription. Subject to the foregoing sentence, you are free to
+modify this Software and publish patches to the Software. You agree that Locust
+Technologies and/or its licensors (as applicable) retain all right, title and
+interest in and to all such modifications and/or patches, and all such modifications
+and/or patches may only be used, copied, modified, displayed, distributed, or otherwise
+exploited with a valid Locust Cloud subscription. Subject to the foregoing,
+it is forbidden to copy, merge, publish, distribute, sublicense,
+and/or sell the Software.
+
+This License applies only to the part of this Software that is not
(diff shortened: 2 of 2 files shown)


cgoldberg:
I know it doesn't break any licensing terms... but I think it's disappointing to have proprietary code mixed into this repo. I'd much rather have it completely separate so there is no confusion about what is or isn't open source. This kind of feels like a takeover of what was once a community based free open source project. I think this is kind of offensive to the community that built this project and also hinders future adoption.

cyberw:
Hi Corey! I dont have time to write a detailed response right now as I'm heading out.

The idea here is definitely not to take anything over. 

The only thing we'll be adding into the "proprietary" directory is the locust-cloud CLI tool for launching distributed runs on our infrastructure (and possibly, in the future, other things that relate to communicating with the commercial offering/SaaS solution), we're not removing or moving any code/existing features there.

We could _possibly_ maintain that in a separate repo and just have it as a dependency, but this way is easier to maintain and also more transparent (for the users of Locust Cloud) what is open source and what is not.

Also remember that the stuff we're doing in Locust Cloud is going to greatly benefit users of the open source project as well - its a win for everyone :)

I asked @heyman to weigh in when I made this PR, lets see what he thinks.

heyman:
To me it sounds more simple and clean to keep the proprietary code in a separate repository. 

Based on the license, it also seems like including it in the open-source repo might not be a good fit. For example:

> This software and associated documentation files (the "Software") may only be used in production, if [...]  and otherwise have a valid Locust Cloud subscription.

This suggests that there could be code within the open-source repository that users aren't even allowed to execute, which feels a bit off.

> This License applies only to the part of this Software that is not distributed as part of Locust Core

This sounds like the License invalidates itself if it's included in the Locust repository (depending on how you interpret "Locust Core").

mbeacom:
Hello everyone,

I share sentiments similar to @cgoldberg. Introducing these new license changes, especially by fragmenting the repository’s licensing structure, seems likely to encourage forking. It also risks confusing community members and potential contributors about what is and isn’t truly open source.

I’m not sure why the pull request for the latest license change (https://github.com/locustio/locust/pull/3019) is gone, but the commits (e.g., [944b7f8](https://github.com/locustio/locust/commit/944b7f822a53c8d058e9f05cb33e6cc3723b8763) and [522af43](https://github.com/locustio/locust/commit/522af43447b8fbe83443b3e4902130896e563df5)) reflect modifications to the effective license holders leading into these changes that could lead to long-term fragmentation. Of course, under standard open source norms, the original MIT license terms remain valid indefinitely for all contributions made prior to these changes.

On a personal note, I had hoped Locust might return to its community-driven roots, which attracted many of us to contribute in the first place. However, these recent changes appear to commercialize the project and move it closer to a BSL-like model. While the repository may remain “source available,” it’s arguably no longer purely open source under widely recognized definitions.

Regards

cyberw:
Hi everyone! 

I have taken your comments to heart and will not be making any changes to the license.

Instead we will make locust-cloud (the code used to communicate with the Locust Cloud service, currently in a private repo) public and available under an MIT license and add it as a dependency to Locust. That way Locust will not depend on anything that isn't open source.

I guess I underestimated how important the "purity" of the repo was for contributors, thinking that I could draw a line between the directories and add new things, without anyone feeling cheated.

I still think commercial backing has been _great_ for Locust, and in fact necessary for its long term survival. Since Locust Technologies was started, there have been security patches, UI improvements and a lot more deep work than just a small fix here or there (we've just started building proper support for asyncio, for instance). In the last year we made **27** releases of Locust, a much higher pace than before, when I was maintaining Locust on basically nights and weekends.

Best regards,
Lars

cgoldberg:
Thanks Lars... I think that's a great move.

PR #3101, opened 2025-04-18
Refactoring: Extract validate stats configuration from main function

insspb:
So, based on #3099 discussion let's begin refactoring of main function in step by step basis.

This PR includes the following changes:

1. stats module import is simplified in all project files, with exclusion to tests.
2. stats module parameters verification is moved from main function to the end of stats module itself.
3. Small typing adjustments in stats module.

### DIFF
--- a/locust/html.py
+++ b/locust/html.py
@@ -5,9 +5,8 @@ from itertools import chain
 from jinja2 import Environment as JinjaEnvironment
 from jinja2 import FileSystemLoader
 
-from . import stats as stats_module
+from . import stats
 from .runners import STATE_STOPPED, STATE_STOPPING, MasterRunner
-from .stats import sort_stats, update_stats_history
 from .user.inspectuser import get_ratio
 from .util.date import format_duration, format_utc_timestamp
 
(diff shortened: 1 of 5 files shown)


cyberw:
* I think `_stats` is unclear. Maybe skip importing stats module and just import `PERCENTILES_TO_CHART` if that is all we need? If you do want/need to rename the stats instance, maybe name it request_stats (as it is an instance of RequestStats)

* I like validate_stats_configuration, but I think only `main` should call `sys.exit`. Maybe have it raise an exception and catch it in main?

insspb:
> * I think `_stats` is unclear. Maybe skip importing stats module and just import `PERCENTILES_TO_CHART` if that is all we need? If you do want/need to rename the stats instance, maybe name it request_stats (as it is an instance of RequestStats)

Underscored variables is common practice to show variable as 'private'. Current PEP8 recommend trailing underscore, but in part with function attributes.

As about stats module imports, I do not like to import required objects from there, as usually many of them used and 'stats.' itself is only 6 characters addition.

> * I like validate_stats_configuration, but I think only `main` should call `sys.exit`. Maybe have it raise an exception and catch it in main?

I think better to make this as additional feature request in issues. Anyway stats module require redesign, as currently it expect user side patching. Better to move it to environment variables/console commands, then validation outside main.py with non sys exit behaviour will be required.

cyberw:
> Underscored variables is common practice to show variable as 'private'. Current PEP8 recommend trailing underscore, but in part with function attributes.

Its a function variable so it isn't really "private" and having both a `stats` and a `_stats` is confusing to a reader. I'm ok with renaming it `request_stats` and importing `stats` though.

> I think better to make this as additional feature request in issues. Anyway stats module require redesign, as currently it expect user side patching. Better to move it to environment variables/console commands, then validation outside main.py with non sys exit behaviour will be required.

Yea, ok. I guess there are already some other places (like argument_parser) that call sys.exit, so it isnt _terrible_.

insspb:
@cyberw Renamed to `request_stats`.

@cyberw I will continue work when https://github.com/locustio/locust/pull/3101 and https://github.com/locustio/locust/pull/3102 will be merged, as too many changes to same places makes merges difficult. Let me know if anything required from my side on this.

PR #3102, opened 2025-04-18
Refactoring: Extract locustfile content merger from main function

insspb:
Continue to work on #3099 

This PR includes the following changes:

1. Change of `load_locustfile` function output objects. This is internal API change, but still should be released as next version, for example 2.36.0
2. Move merge locustfiles objects to separate function in main.py
3. Fix old incorrect typing in `load_locustfile` function in load_locustfile.py. It is interesting why mypy do not complain earlier.

Note: I leaved TODO message. Incorrect typing of User.tasks attribute is one of the reason I looked at main, the second one is incorrect help work (for command line objects). But correct .tasks attribute type rewrite will trigger many errors in different places, and will require additional tests. So not included at this point. It is in long 'todo' list, that out of my several available days.

Note2: It is interesting that load_locustfile return class for user, and **instance** for `LoadTestShape`. It may be incorrect historical behaviour, but will be cleaner when more code will be refactored. At least at this point it is hard to catch this from first look (now typing will help).

Check commit messages for details

### DIFF
--- a/locust/main.py
+++ b/locust/main.py
@@ -99,8 +99,7 @@ def main():
     locustfiles = parse_locustfile_option()
 
     # Importing Locustfile(s) - setting available UserClasses and ShapeClasses to choose from in UI
-    user_classes: dict[str, locust.User] = {}
-    available_user_classes = {}
+    available_user_classes: dict[str, locust.User] = {}
     available_shape_classes = {}
     available_user_tasks = {}
     shape_class = None
@@ -121,7 +120,7 @@ def main():
         # Setting Available User Classes
         for key, value in file_user_classes.items():
             if key in available_user_classes.keys():
-                previous_path = inspect.getfile(user_classes[key])
+                previous_path = inspect.getfile(available_user_classes[key])
                 new_path = inspect.getfile(value)
                 if previous_path == new_path:
                     # The same User class was defined in two locustfiles but one probably imported the other, so we just ignore it
@@ -132,7 +131,6 @@ def main():
                     )
                     sys.exit(1)
 
-            user_classes[key] = value
             available_user_classes[key] = value
             available_user_tasks[key] = value.tasks or {}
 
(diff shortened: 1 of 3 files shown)


locust/main.py
```diff
@@ -93,53 +97,73 @@ def create_environment(
...
+    return available_user_classes, available_shape_classes, available_user_tasks, shape_class
+
+
+def main():
+    # find specified locustfile(s) and make sure it exists, using a very simplified
+    # command line parser that is only used to parse the -f option.
+    locustfiles = parse_locustfile_option()
+
+    # Importing Locustfile(s) - setting available UserClasses and ShapeClasses to choose from in UI
+    (
```
cyberw, line 161: Maybe we wanna scratch the `available_` prefix from all of these instead of adding it to user_classes?

locust/main.py
```diff
@@ -327,17 +351,19 @@ def kill_workers(children):
             if soft_limit < minimum_open_file_limit:
                 # Increasing the limit to 10000 within a running process should work on at least MacOS.
                 # It does not work on all OS:es, but we should be no worse off for trying.
-                resource.setrlimit(resource.RLIMIT_NOFILE, [minimum_open_file_limit, hard_limit])
+                limits = minimum_open_file_limit, hard_limit
+                resource.setrlimit(resource.RLIMIT_NOFILE, limits)
```
cyberw, line 355: not sure this is an improvement?

locust/main.py
cyberw: nvm, if that is what it is called in environment constructor then it make sense (at least unless we also change that)

cyberw:
Awesome stuff!

locust/main.py
insspb: This part will be refactored later anyway. At this point this line changed by mypy request, see failed pipelines earlier.

insspb:
@cyberw I will continue work when #3101 and #3102 will be merged, as too many changes to same places makes merges difficult. Let me know if anything required from my side on this.

cyberw:
Nice!

PR #3109, opened 2025-04-22
Fix yarn berry publish

andrewbaldwin44:

PR #3114, opened 2025-04-24
fix: remove unsupported syntax for yarn berry

schwannden:


cyberw:
Thanks for this! Can you remove the unrelated formatting changes? Might wanna do those changes later, but the git blame history will be messed up if we do it as part of an unrelated fix :)

schwannden:
> Thanks for this! Can you remove the unrelated formatting changes? (we probably wanna do those changes later, but the git blame history will be messed up if we do it as part of an unrelated fix :)

Hi @cyberw sure, but honestly I could not test this fix without more work (I need to setup my own npm auth secret, and try publish on my side). Would it be easier for you to verify this fix?

andrewbaldwin44:
I think we got it working now in https://github.com/locustio/locust/pull/3109. Super weird there's not a lot of great documentation on this but thanks for trying to help! :)

schwannden:
> I think we got it working now in #3109. Super weird there's not a lot of great documentation on this but thanks for trying to help! :)

Great! Sorry for the inconvenience, I was just about to look into this, and is glad that it is resolved. Yeah the documents are scarce...

PR #3122, opened 2025-04-28
Add locust exporter import

andrewbaldwin44:


locust/main.py
```diff
@@ -43,6 +44,15 @@
     locust_cloud = None
     if e.msg != "No module named 'locust_cloud'":
         raise
+try:
+    import locust_exporter  # pyright: ignore[reportMissingImports] # noqa: F401
+
+    locust_cloud_version = f" (locust_exporter {importlib.metadata.version('locust-exporter')})"
```
cyberw, line 50: this should be locust_exporter_version, eh?

cyberw: (and probably wanna update this line too https://github.com/locustio/locust/blob/0f3c6452fc1c101d62f0b1eaec4357edae280742/locust/main.py#L181)

PR #3123, opened 2025-04-28
Developer focused documentation

domik82:
Developer focused documentation created together with AI tooling. The goal is to allow both developers and AI tools figure out faster where is what in Locust repo. Hopefully documentation is valid at all places.  Mermaid graphs are a bit better compared to pure text in representing data flows. That is first step in my plan in enhancing Locust reporting.

You should have Mermaid plugin in IntelliJ / VSCode to properly see the graphs

### DIFF
--- a/docs-md/contributing_guide.md
+++ b/docs-md/contributing_guide.md
@@ -19,12 +19,14 @@ flowchart LR
 ```
 
 1. **Clone the repository**:
+
    ```bash
    git clone https://github.com/locustio/locust.git
    cd locust
    ```
 
 2. **Setup the environment using uv**:
+
    ```bash
    # Install uv if you don't already have it
    pip install uv
@@ -45,7 +47,8 @@ flowchart LR
    # uv sync --group build --group test
    ```
 
-   Alternatively, you can use Hatch directly (also used in the project):
+   Alternatively, you can use Hatch directly (also used in the project):ł
+
    ```bash
    # Install hatch if you don't have it
    pip install hatch
@@ -54,7 +57,8 @@ flowchart LR
    hatch run test:all
    ```
 
-4. **Install Web UI dependencies** (if working on UI changes):
+3. **Install Web UI dependencies** (if working on UI changes):
+
    ```bash
    cd locust/webui
    yarn install
@@ -81,6 +85,7 @@ flowchart TD
    - Create a new issue if needed, clearly describing the bug or feature
 
 2. **Create a branch**:
+
    ```bash
    git checkout -b feature/your-feature-name
    # or
@@ -93,6 +98,7 @@ flowchart TD
    - Add/update tests as appropriate
 
 4. **Run tests**:
+
    ```bash
    pytest
    ```
@@ -199,18 +205,21 @@ flowchart TD
 When working on the Web UI:
 
 1. **Run the development server**:
+
    ```bash
    cd locust/webui
    yarn dev
    ```
 
 2. **Watch for changes while running Locust**:
+
    ```bash
    cd locust/webui
    yarn watch
    ```
-   
+
 3. **Build for production**:
+
    ```bash
    cd locust/webui
    yarn build
@@ -221,17 +230,20 @@ For more details on Web UI development, see the [Developer Guide](developer_guid
 ## Building and Running
 
 1. **Running Locust from source**:
+
    ```bash
    python -m locust -f examples/basic.py
    ```
 
 2. **Building the Web UI**:
+
    ```bash
    cd locust/webui
    yarn build
    ```
 
 3. **Creating a distribution package**:
+
    ```bash
    pip install build
    python -m build
--- a/docs-md/contributing_guide.md
+++ b/docs-md/contributing_guide.md
@@ -1,0 +1,263 @@
+# Contributing to Locust
+
+This guide provides information for developers who want to contribute to the Locust project. It covers the development environment setup, coding standards, testing, and contribution workflow.
+
+## Development Environment Setup
+
+```mermaid
+flowchart LR
+    Clone[Clone Repository] --> Setup[Setup Environment]
+    Setup --> Install[Install Dependencies]
+    Install --> Dev[Start Development]
+    
+    subgraph "Environment Setup Steps"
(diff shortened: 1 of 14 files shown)


docs-md/modern_web_ui.md
```diff
@@ -0,0 +1,131 @@
...
+- Firefox
+- Safari
+
+It also provides responsive layouts for different screen sizes, from desktop monitors to tablet devices.
+
+## FAQ
+
+**Q: Can I still use the legacy UI?**
+
+A: Yes, you can use the `--legacy-ui` flag to switch back to the old UI. However, the legacy UI is deprecated and will be removed in a future version.
```
andrewbaldwin44, line 119: Hey this is actually outdated now, the legacy UI was removed in [2.28.0](https://github.com/locustio/locust/releases/tag/2.28.0)

docs-md/modern_web_ui.md
```diff
@@ -0,0 +1,131 @@
...
+
+Charts include zoom functionality and tooltips for detailed data exploration.
+
+## Custom Configuration
+
+### Auth Protection
+
+The Web UI can be protected with authentication using the `--web-login` flag. You can customize the login experience by:
+
+- Using HTTP Basic Authentication
```
andrewbaldwin44, line 87: HTTP Basic auth isn't supported anymore either

docs-md/contributing_guide.md
```diff
@@ -0,0 +1,275 @@
...
+    
+    subgraph "Environment Setup Steps"
+        Clone --> GitRepo[Github Repository]
+        Setup --> VirtualEnv[Python Virtual Environment]
+        Install --> DevDeps[Development Dependencies]
+        Install --> WebUI[Web UI Dependencies]
+    end
+```
+
+1. **Clone the repository**:
```
andrewbaldwin44, line 21: Most developers won't have write access to locust, maybe as a first step we can suggest forking the project?

docs-md/contributing_guide.md
```diff
@@ -0,0 +1,275 @@
...
+
+7. **Merge**:
+   - Once approved, your PR will be merged into the main branch
+
+## Coding Standards
+
+Locust follows these coding standards:
+
+1. **Code Formatting**:
+   - Python code is formatted with Black
```
andrewbaldwin44, line 123: In [2.21.0](https://github.com/locustio/locust/releases/tag/2.21.0) black was migrated to Ruff

docs-md/deployment_guide.md
andrewbaldwin44, line 1: While this information is all super useful, I think it seems a little out of scope as far as "developing locust".  These 3rd party cloud providers may also change their implementation at any given time. I'm not sure if we want to take the responsibility of keeping this up to date and ensuring the information is correct

docs-md/modern_web_ui.md
domik82: Also missed that.

docs-md/contributing_guide.md
domik82: Hmm thought I fixed it - good catch.

docs-md/deployment_guide.md
andrewbaldwin44: Yeah I think that might be best. I'm sure the code could be quite useful but IMO we shouldn't state it if we're not sure that it's correct. I think stating the options as you suggested would be great. Maybe a mention of [locust-swarm](https://github.com/SvenskaSpel/locust-swarm/) would also fit well here

cyberw: I think this stuff is really cool, and I don't mind linking it from the FAQ or something, but it needs to live somewhere else. Any code in the docs needs to be tested, because people will assume it is and waste a lot of time debugging if it isn't working.

The "Developing locust" page already exists https://github.com/locustio/locust/blob/master/docs/developing-locust.rst, so anything about actually contributing to Locust should be in that page. Probably doesn't need to be linked from the readme either, that one should be kept as short as possible and mainly focused on explaining the features of Locust rather than "how to".

As it stands, this PR is a little too big. Also it seems to contain a few accidental changes/reversions (https://github.com/locustio/locust/pull/3123/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5). Probably best if you could break it up into several smaller ones.

domik82: The change in the main Readme was intentional. People configure AI tooling like 'claude code'/cursor/windsurf to read readme to understand codebase. AI starts from here then loads rest of the files to memory (this is why I created project structure .md file). Claude suggests to create Claude.md file but I didn't wanted to push anything like that to this repo. 

I really can't convince myself to .rst since those don't render in the gitlab/github while .md files do that natively. The Index.rst doesn't work in IntelliJ - the plugin I have is not allowing for navigation while markdown does. 

This is why I decided to go with this direction. 

Since I'm using AI tools on a daily basis, I wanted to add Locust documentation to https://context7.com/ just like I added Wiremock few days ago -> https://context7.com/wiremock/wiremock.org but before doing it I wanted to make sure Locust documentation will be 'better suited for the AI needs'. This is why it's all in or it will be just current .rst files and I think development of Locust itself will still be hard - I know it because I would like to change something in reporting and I do struggle to understand how things connect to each other. 

Based on Context7 - AI Tooling can search for data and automatically propose snippets of the code. They automatically create RAG based on  .md, .mdx, .txt, .rst, .ipynb files

Let me know what you think about it.

cyberw: Ah, I didnt quite understand that this was mostly for machine consumption :) 

Is there a spec somewhere for how Context7 extracts documentation? I'd _really_ prefer to have only have one set of documentation if at all possible, as this stuff risks becoming out of date and feeding garbage to the AI instead. I don't think you're volunteering to be on the hook for maintaining all these documents going forward, or are you? :)

Perhaps we could add the /examples locustfiles, or maybe that is done automatically?

cyberw: If you want to make a much smaller version of this we can chat about it, but right now this is just too big to even review.

The principle of DRY applies even outside of program code :)

Issue #3089, opened 2025-04-04
export the results as a json file


kasir-barati:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am suggesting a new feature, not asking a question

### Description

Hi,

When I run `locust --help` I can see that we are allowed to pass `--json` as a flag to see the result of tests as a JSON, but what I need is to store that json in a file. Is it possible to implement this feature?

I know for csv you can pass a filename and it will store the result in that file but this is not the case for json which makes me wonder why not? It should be almost the same as csv IMO.

```bash
$ locust --help
# ...
Request statistics options:
  --csv <filename>      Store request stats to files in CSV format. Setting this option will generate three files: <filename>_stats.csv, <filename>_stats_history.csv and
                        <filename>_failures.csv. Any folders part of the prefix will be automatically created
  --csv-full-history    Store each stats entry in CSV format to _stats_history.csv file. You must also specify the '--csv' argument to enable this.
  --print-stats         Enable periodic printing of request stats in UI runs
  --only-summary        Disable periodic printing of request stats during --headless run
  --reset-stats         Reset statistics once spawning has been completed. Should be set on both master and workers when running in distributed mode
  --html <filename>     Store HTML report to file path specified. Able to parse certain tags - {u}, {r}, {t} and convert them to number of users, spawn rate and run time
                        respectively.
  --json                Prints the final stats in JSON format to stdout. Useful for parsing the results in other programs/scripts. Use together with --headless and --skip-log
                        for an output only with the json data.
# ...
```

cyberw:
I guess we could support a filename there as well. But you can just do `locust --json > myfile.json`

Cnoor0171:
It would nice to have the json argument accept a filename because redirecting the stdout to a json file requires that you use `--skip-log` to make sure the output is valid json. In a CICD pipeline, for example, you still want to show the relevant log statements, but then want to upload the final result file afterwards.

kasir-barati:
Thanks for adding `--json-file <filename>` option to Locust CLI 💮.

Just for anybody who wants to read the docs about this: https://docs.locust.io/en/stable/configuration.html#command-line-options

PR #3124, opened 2025-05-04
Add command line option to export json results into a file

ajt89:
#3089

### DIFF
--- a/locust/test/test_main.py
+++ b/locust/test/test_main.py
@@ -2067,47 +2067,6 @@ class SecondUser(HttpUser):
             self.assertEqual(dict, type(result["num_reqs_per_sec"]))
             self.assertEqual(dict, type(result["num_fail_per_sec"]))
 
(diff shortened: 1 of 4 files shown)


locust/test/test_main.py
```diff
@@ -2067,6 +2067,92 @@ def hello_world(self):
             self.assertEqual(dict, type(result["num_reqs_per_sec"]))
             self.assertEqual(dict, type(result["num_fail_per_sec"]))
 
+    def test_json_file_missing_arg(self):
```
cyberw, line 2070: this test is unnecessary, it just tests basic argparse functionality.

cyberw: Other than that, I'm good to merge this! Later might want to harmonize argument names with csv, but that is hard to do without breaking backwards compatibility.

ajt89: Sounds good, removed the test

PR #3136, opened 2025-05-09
Fixes #3134 - Improve the performance of the `/stats/requests` endpoint

### LINKED ISSUE #3134: p95 response time increases with the number of unique URLs
### Prerequisites

### Description

We have been investigating a performance issue in a service using Locust. The p95 response time would gradually creep up, indicating a serious issue with our service. The load test we created selected random elements from several high-cardinality lists and sent them as GET parameters: 

```python
        session_id = "test_" + str(random.randint(1, 10000))
        context = random.choice(contexts)
        product_id = random.choice(product_ids)

        self.client.get(
            f"/some-endpoint?session={session_id}&context={context}&product_id={product_id}",
            json=None,
        )
```

The p95 response time would gradually creep up:

Image

After a significant amount of debugging we realized that the issue appeared to be with Locust itself. If we refactor the test to use the params argument:

```python
     self.client.get(
            "/some-endpoint",
            params={
                "session": session_id,
                "context": context,
                "product_id": product_id,
            },
            json=None,
        )
```

Then the p95 response time as recorded by Locust remained stable:

Image

While I realize that the documentation points you towards using `params`, it's very surprising to see the behavior above. The service had no performance issues, the issue was within Locust itself. 

Inadvertently creating a very large set of URLs as part of a load test caused the reported *service* response time to increase, when in fact it remained stable throughout.

### Command line

locust -f loadtest.py

### Locustfile contents

```python3
from locust import FastHttpUser, between, task
import uuid

class User(FastHttpUser):
    wait_time = between(1, 5)

    @task
    def call(self):
        self.client.get(
            f"/some-endpoint?a={uuid.uuid4()}",
            json=None,
        )
```

### Python version

3.10.17

### Locust version

2.37.1

### Operating system

MacOS

orf:
With the following pathological locustfile, the following change hugely reduces the time it takes to call the `/stats/requests` endpoint.

```python
import uuid
from locust import FastHttpUser, between, task 

class User(FastHttpUser):
    wait_time = between(1, 5)

    @task
    def call(self):
        self.client.get(
            f"/some-endpoint?a={uuid.uuid4()}",
            json=None,
        )
``` 

With the existing code, the endpoint almost immediately takes >100ms to execute:

Screenshot 2025-05-09 at 14 04 18

Whereas with this branch, it stays around 50ms consistently:

Screenshot 2025-05-09 at 14 07 06

Fixes #3134

### DIFF
--- a/locust/web.py
+++ b/locust/web.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import csv
+import itertools
 import json
 import logging
 import mimetypes
@@ -8,7 +9,6 @@ import os.path
 from functools import wraps
 from html import escape
 from io import StringIO
-from itertools import chain
 from json import dumps
 from typing import TYPE_CHECKING, Any, TypedDict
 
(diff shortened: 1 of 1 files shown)


cyberw:
Love it! But leave some version of the comment in place, to explain why we're truncating.

orf:
Thanks @cyberw! I've updated the branch to retain the comment

PR #3154, opened 2025-06-06
Revert accidental removal of --json-file option

brtkwr:


cyberw:
👀 Can you tell where it got deleted?

brtkwr:
In this commit:: https://github.com/locustio/locust/commit/46c9a56b8bcf3f2e09910b99227361a29aa2b759 titled:

> bump min locust-cloud version to latest

PR #3168, opened 2025-07-10
Add milvus user class and example

zhuwenxing:


zhuwenxing:
@cyberw Could you please review this PR when you have a chance? Thanks!

cyberw:
It looks great. I’m trying to have a think about how to organize contributed extensions in general and possibly unit-test them. Will get back to you next week!

zhuwenxing:
> It looks great. I’m trying to have a think about how to organize contributed extensions in general and possibly unit-test them. Will get back to you next week!

sure, take your time.

examples/milvus/locustfile.py
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/env python3
```
cyberw, line 1: Can this example be simplified? Does introducing numpy as a dependency really make sense?

No need to print instructions manually on execution of __main__, rely on the arg help text or comments in code.

cyberw:
Hi! I made a first pass of comments. It would need to have some kind of unit tests too. Not sure if there is a milvus/grpc mock though.

locust/contrib/milvus.py
```diff
@@ -0,0 +1,407 @@
...
+from abc import ABC, abstractmethod
+from typing import Any
+
+import gevent.monkey  # noqa: E402
+from pymilvus import CollectionSchema, MilvusClient
+from pymilvus.milvus_client import IndexParams
+
+gevent.monkey.patch_all()
+
+import grpc.experimental.gevent as grpc_gevent  # noqa: E402
```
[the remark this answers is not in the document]
zhuwenxing: OK, I will place it at the beginning

zhuwenxing:
> Not sure if there is a milvus/grpc mock though.

There's a Mock in pymilvus's unit tests, but this is a Mock for pymilvus ORM. Now we're using the milvus client format, so we should need to write a new Mock

https://github.com/milvus-io/pymilvus/blob/master/tests/mock_milvus.py

examples/milvus/locustfile.py
cyberw: Great. Go ahead and remove the numpy stuff too. I like keeping examples as minimalist as possible.

cyberw:
> > Not sure if there is a milvus/grpc mock though.
> 
> There's a Mock in pymilvus's unit tests, but this is a Mock for pymilvus ORM. Now we're using the milvus client format, so we should need to write a new Mock
> 
> https://github.com/milvus-io/pymilvus/blob/master/tests/mock_milvus.py

Huh... well if we can't easily test the User class without needing all of milvus anyway it might not be worth testing. Just simplify the example and I'll have one last look.

cyberw:
Also, you can add it to the docs, like this: https://github.com/locustio/locust/blob/ca7e6ac14f6c98f81d144c278e9a997171d8e796/docs/api.rst#L44

zhuwenxing:
Hi, I've addressed all the issues you raised:
* Moved gevent monkey patch to the top before pymilvus imports
* Simplified example
* Removed numpy dependency, using `random.random()`
* Cleaned up README and added to API docs


and also added optional-dependencies in pyproject.toml:
```toml
[project.optional-dependencies]
milvus = ["pymilvus>=2.5.0"]
```

Is this the preferred approach for contrib dependencies? Other contrib modules (MongoDB/Postgres) don't currently declare their deps in pyproject.toml, should we standardize this?

cyberw:
> Is this the preferred approach for contrib dependencies? 

It is now :)

Great stuff!

cyberw:
~Hmm. Docs arent looking great (but there seems to have been an issue even earlier). Can you take a look?~

Never mind, I found the issue and I'm fixing it in a new PR.

PR #3169, opened 2025-07-14
Shrink report size

andrewbaldwin44:
### Proposal
- Remove escaping strings, which has not been needed since we switched to React, and allows us to remove the React Markdown package (saves ~100mb)
- Remove react-redux and rtk dependency from the report bundle (saves ~100mb)
- Aggressively tree-shake echarts (saves ~500mb)
- Use terser to minify the report bundle (saves ~60mb)

#### Before Build Size
```
dist/report.html  1,479.96 kB │ gzip: 488.19 kB
dist/assets/index-4J0tB7ST.js  1,728.94 kB │ gzip: 564.85 kB
```

#### After Build Size
```
dist/report.html  764.68 kB │ gzip: 251.14 kB
dist/assets/index-CGv5vUj8.js  1,057.03 kB │ gzip: 347.36 kB
```

#### Gains
- 715mb saved on report bundle size
- 672mb saved on locust webui bundle size
- **Downloaded HTML reports are now within 30mb of the old HTML report size** (old HTML report size ~730mb, new HTML report size ~760mb)


cyberw:
Nice! Build errors are unrelated and fixed in master, so I'm merging this!

PR #3181, opened 2025-07-25
Support for list based custom arguments

mickdwyer:
Add webui support for custom argparse options that use the action="append" to build a list of parameter values.


cyberw:
Nice!

PR #3183, opened 2025-07-29
Optimize Tests

amadeuppereira:


cyberw:
Looks nice! Once you have converted a few more tests, can you list some time improvement?

locust/test/test_main.py
```diff
@@ -102,29 +103,30 @@ def my_task(self):
...
-        proc.send_signal(signal.SIGTERM)
-        stdout, stderr = proc.communicate(timeout=3)
-        self.assertIn("Starting Locust", stderr)
-        self.assertRegex(stderr, r".*Shutting down[\S\s]*Aggregated.*", "no stats table printed after shutting down")
-        self.assertNotRegex(stderr, r".*Aggregated[\S\s]*Shutting down.*", "stats table printed BEFORE shutting down")
-        self.assertNotIn("command_line_value", stdout)
-        self.assertIn("web_form_value", stdout)
+            with TestProcess(
+                f"locust -f {file_path} --custom-string-arg command_line_value --web-port {port}",
+                self.fail,
```
cyberw, line 108: Can we have this as the default? If you have a hard time getting hold of the testcase instance (self) inside TestProcess, maybe try pytest.fail instead?

locust/test/test_main.py
```diff
@@ -205,13 +206,12 @@ def my_task(self):
...
-            gevent.sleep(SHORT_SLEEP)
-            proc.send_signal(signal.SIGTERM)
-            stdout, stderr = proc.communicate()
-            self.assertIn("Starting web interface at", stderr)
-            self.assertIn("Starting Locust", stderr)
-            self.assertIn("Shutting down (exit code 0)", stderr)
+            with TestProcess(f"locust -f {file_path}", self.fail, expect_return_code=0) as proc:
+                proc.expect("Starting Locust")
+                proc.expect("Starting web interface at")
+                # Ensure we do not trigger SIGIN when instantiating webui as it can result
```
cyberw, line 212: typo :)

locust/test/test_main.py
```diff
@@ -288,34 +288,25 @@ def my_task(self):
...
-                stdout, stderr = proc.communicate()
-                self.assertIn("Starting web interface at", stderr)
-                self.assertIn("Starting Locust", stderr)
-                self.assertIn("Shutting down (exit code 0)", stderr)
+                with TestProcess(
+                    f"locust -f {mocked1.file_path},{mocked2.file_path}", self.fail, expect_return_code=0
+                ) as proc:
+                    proc.expect("Starting Locust")
+                    proc.expect("Starting web interface at")
+                    gevent.sleep(0.1)
```
cyberw, line 299: why the final sleep and then no validations?

cyberw:
Looking really good, except for the failing tests :)
Did you have a look at hiding the helpers from the callstack?

```
def my_internal_helper():
    __tracebackhide__ = True
```

conftest.py
```diff
@@ -0,0 +1,3 @@
+import gevent.monkey
```
cyberw, line 1: I'm sure this is needed for some reason (because we dont import locust early enough?), but can you comment why?

amadeuppereira: we need gevent to be patched before spawning locust [here](https://github.com/amadeuppereira/locust/blob/9a74041690655e620ee990086176b8b4b8969e7e/locust/test/subprocess_utils.py#L67)

cyberw: Maybe just `import locust` directly from subprocess_utils? (and mention that you're doing it for the side effect of patching) 

I think this way we're patching twice, as soon as we're importing locust.

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,210 @@
...
+        self,
+        command: str,
+        on_fail: Callable[[str], None] = pytest.fail,
+        *,
+        expect_return_code: int | None = 0,
+        should_send_sigint: bool = True,
+        use_pty: bool = False,
+        join_timeout: int = 1,
+    ):
+        self._exitted = False
```
cyberw, line 36: I think the right spelling is to use one t :)

cyberw: Also, maybe it should be `_terminated` as it is only set from terminate()?

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        if isinstance(pattern, str):
+            regex = re.compile(pattern)
+        else:
+            regex = pattern
+
+        is_match: Callable[[re.Pattern, str], bool] = lambda pattern, line: pattern.search(line) is not None
+        return self._expect(regex, is_match, **kwargs)
+
+    def send_input(self, content: str):
+        if self.use_pty:
```
cyberw, line 206: Prefer to crash instead of silently ignoring if someone tries to send input when use_pty is false.

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        self.proc.send_signal(sig)
+        self._terminated = True
+
+    def wait(self, timeout=None):
+        return self.proc.wait(timeout=timeout or self.join_timeout)
+
+
+class PsutilTestProcess(TestProcess):
+    process_backend = psutil
+    proc: psutil.Popen  # type: ignore
```
cyberw, line 225: Hmm... this subclass feels weird, and process_backend isnt the most obvious thing. 

psutil.Process objects can be created from an existing subprocess though. So in the few/one case we need to list children the test case can just do:

        psutil.Process(proc.proc.pid).children(...)

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    __test__ = False
+    process_backend = subprocess
+    proc: subprocess.Popen[str]
+
+    def __init__(
+        self,
+        command: str,
+        on_fail: Callable[[str], None] = pytest.fail,
```
cyberw, line 29: I dont think we're passing on_fail anywhere anymore? removing it will simplify this class a lot.

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        if not IS_WINDOWS:
+            sig = signal.SIGINT
+        else:
+            # Signals are hard on Windows
+            sig = signal.SIGTERM
+
+        self.proc.send_signal(sig)
+        self._terminated = True
+
+    def wait(self, timeout=None):
```
cyberw, line 219: I think this method is not really needed. People can just call `proc.proc.wait()` with whatever timeout they want instead. 

Maybe we should call the variable used in tests `tproc`, or `tp` instead of proc, to not give the impression that it is a subprocess.Popen. Wdyt?

locust/test/subprocess_utils.py
```diff
@@ -0,0 +1,228 @@
...
+            buffer = self.stderr_output
+
+        if any(is_match(to_not_expect, line) for line in buffer):
+            self.on_fail(f"Found unexpected message: '{to_not_expect}'.")
+
+    def expect(self, output: str, **kwargs):
+        __tracebackhide__ = True
+
+        is_match: Callable[[str, str], bool] = lambda out, line: out in line
+        return self._expect(output, is_match, **kwargs)
```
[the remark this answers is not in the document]
amadeuppereira: you are right, it no longer makes sense!

Issue #3184, opened 2025-08-01
Failures table sorting is reset to ascending after a few seconds


KK-K-K-KK:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am reporting a bug, not asking a question

### Description

On the Failures tab, when I try to sort rows according to "# Failures" column in descending order, the sort is reset to ascending order within several seconds. Please keep the sorting while the table is updated during runs. Thanks a lot!

https://github.com/user-attachments/assets/f09fb42a-86c7-43fa-8b2d-5be8b75acc2f

### Command line

sudo docker run -p 10002:8089 -di -P --name locust_kevin locustenforce:update-task locust -f load_update.py --tags --master

### Locustfile contents

```python3
# basically any locustfile with a task that calls a request with several fail reasons will do.

import configparser
from logging import error, info, warning, debug
import time
from locust.user.task import task, tag
# from api.pyprotobuf.au import update_pb2
from locust import SequentialTaskSet, TaskSet, events
from locust.contrib.fasthttp import FastHttpUser
import uuid
import string
import random
import hashlib
import yaml
import gevent
from gevent.lock import Semaphore


#########################
###   TASKS & USERS   ###
#########################


class UpdateTask(SequentialTaskSet):

    @task
    class TrendMicroUpdateTask(SequentialTaskSet):
        def __get_server_ini(self):
            res = self.client.get("/rest/stellar-au/duplicate/protect/server.ini")
            return res

        @task
        @tag('trendmicro', "trendmicro-download")
        def get_component_path(self):
            self.downloadUrls = []
            res = self.__get_server_ini()
            if res is None or res.status_code != 200:
                debug('get_component_path: response is not 200.')
                return
            

        @task
        @tag("txone", "txone-download", "trendmicro", "trendmicro-download")
        def stop(self):
            self.interrupt()

    @task
    class WaitForNextUpdate(TaskSet):
        
        sleep_interval = 86400  # one day

        @task
        @tag("txone", "txone-download", "trendmicro", "trendmicro-download")
        def wait_till_next_update(self):
            time.sleep(UpdateTask.WaitForNextUpdate.sleep_interval)
            self.interrupt()

    
class ProtectUser(FastHttpUser):

    with open("env.yaml", 'r') as yaml_file:
        environment = yaml.safe_load(yaml_file)
    protect_env = environment["PROTECT"]
    host = protect_env['Host']
    host = f"https://{host}"
[...]

PR #3185, opened 2025-08-04
Fix useSortByField

andrewbaldwin44:
Recording 2025-08-04 at 18 15 55


cyberw:
Guessing this fixes #3184 :)

PR #3193, opened 2025-08-22
Update Dockerfile base to Python 3.13

adaamz:
Hi,
it was a long time ago since Python 3.13 was released (back in October 2024). So maybe it's time to release locust Docker image with Python 3.13 maybe.

Feel free to comment if I should change some Python version in CI or elsewhere.

Thanks for review.


cyberw:
Looks good! There's the tiniest risk that it could cause issues with some downstream images, so I wont merge this just yet, but give me a couple weeks!

Issue #3203, opened 2025-09-05
Documentation around distributed locustfile (-f -) is lacking


gregersn:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am reporting a bug, not asking a question

### Description

The documentation says 
> The -f - argument tells Locust to get the locustfile from master instead of from its local filesystem. This feature was introduced in Locust 2.23.0.

But it does not mention that this will not work if you follow the suggested ways of structuring test code into multiple files. 

Documentation should either be updated to mention this, or there should be ways of distributing all required code to run tests. 

### Command line

locust -f - --worker --master-host <host> --process -1

### Locustfile contents

```python3
N/A
```

### Python version

N/A

### Locust version

N/A

### Operating system

N/A

cyberw:
I agree, we could improve the docs at the very least (serving multiple files is a bit thorny), that's why it was never implemented.

Updated the docs in https://github.com/locustio/locust/commit/6422848afd072032005f2eafa9ed4db11238708a

cyberw:
Maybe we _could_ give an error message if multiple files were specified on master, but I dont know if that is a priority.

PR #3231, opened 2025-10-02
docs: add distributed mode limitations documentation

1234-ad:
## Description
This PR addresses issue #3203 by adding comprehensive documentation about the limitations of using `-f -` (distributed locustfile) with multiple files.

## Changes
- Added new documentation file `docs/distributed-limitations.rst`
- Clearly explains why `-f -` doesn't work with multiple files
- Provides practical workarounds for multi-file test structures
- Includes examples of both problematic and working approaches

## Why this change?
Issue #3203 highlighted that the current documentation mentions `-f -` works for "single locustfiles" but doesn't clearly explain the implications for users with multi-file test structures. This leads to confusion and `ModuleNotFoundError` when users try to import from other modules.

## Benefits
- **Prevents user frustration** - Clear upfront explanation of limitations
- **Provides solutions** - Multiple workaround strategies
- **Improves developer experience** - Users can make informed architectural decisions
- **Reduces support burden** - Fewer duplicate issues about import errors

## Documentation Structure
The new file covers:
1. Clear explanation of the single-file limitation
2. Examples of what doesn't work and why
3. Multiple workaround strategies
4. Example of proper single-file structure

This addresses the maintainer's comment that "serving multiple files is a bit thorny" by documenting the current limitations while providing practical alternatives.

### DIFF
--- a/docs/distributed-limitations.rst
+++ b/docs/distributed-limitations.rst
@@ -1,0 +1,63 @@
+.. _distributed-limitations:
+
+===============================
+Distributed Mode Limitations
+===============================
+
+When using distributed mode with ``-f -`` (getting locustfile from master), there are important limitations to be aware of:
+
+Single File Only
+================
+
+The ``-f -`` feature only works with single locustfiles. If your test setup uses multiple files or modules, this approach will not work.
+
+**This will NOT work:**
+
+.. code-block:: bash
+
+    # Multiple files - will fail
+    locust -f locustfile.py -f common.py --worker --master-host <host>
+    
+    # Imports from other modules - will fail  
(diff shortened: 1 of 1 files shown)


cyberw:
It is nice that you provide all this info but it is a little too verbose. Specifically the code example is not necessary, and I think it something only <1% of locust users will care about so it doesnt deserve its own top level page in the docs.

Maybe add it under the FAQ and link it from the note that is already there next to the introduction of `-f -` ("The -f - argument tells Locust to get the locustfile from master instead of from its local filesystem. This only works for single locustfiles.")

PR #3236, opened 2025-10-10
loosen dependency on gevent

bollwyvl:
Congratulations on 2.41.6!

This PR loosens the top pin for `gevent` to allow for using versions with a number of [bug fixes](https://github.com/gevent/gevent/compare/25.8.1...25.9.1).


cyberw:
Cool! Yea the issue was fixed in 25.8.2. We could probably relax it all the way to <26.0.0

PR #3237, opened 2025-10-12
Added readme badge for supported python versions

Nirzak:
Added readme badge for supported python versions. This version information will be fetched from pypi project's classifier information. We usually mention this information on repo's pyproject.toml file.


cyberw:
We had this before, until it started having issues. But it is worth another try!

PR #3242, opened 2025-10-16
Fallback to check for Toml without tool.locust heading for  Issue#3225

andrewasche:


cyberw:
Not sure what this really does, but if you can fix the linting and test issues I'll have a look!

andrewasche:
It looked like the documentation has examples both with and without a toml section called tool.locust, but in the ticket it is referencing the a configuration without tool.locust section. It seems like it wasn't working properly without the tool.locust section in the config file. It I tried to make it check for a file without a tool.locust config file

cyberw:
Can you add a test case that this change fixes?

andrewasche:
I added a test case

andrewasche:
I don't know if this makes a difference, but all checks pass

cyberw:
The test you added passes even without the code change. I did make a change in the area fixing another issue though, maybe that also sorted it? https://github.com/locustio/locust/pull/3257

andrewasche:
That might have fixed this downstream of your change. Thanks for letting me know

PR #3244, opened 2025-10-17
Avoid using most recent python-requests because it may introduce performance issues

cyberw:


jclerman:
Hi, I just noticed the effect of this when my project, which has a development-dependency on `locust`, no longer was compatible with the latest version of `requests` (v 2.32.5 at the time of writing).

Can you provide any context on specific issues that led to this change?  Is this expected to be a long-standing limitation, or is it temporary? Thanks!

sgerrand:
👋 @cyberw: Do you have a timeframe for how long this constraint on `requests` will be maintained? This is currently blocking the upgrade of some projects to Python 3.14.

cyberw:
Hi! I don't have a timeline right now, but understand it is definitely more than a little inconvenient... Contributions would be most welcome. 

I think we should probably disable ssl verification by default (FastHttpUser already does this), and/or implement SSLContext caching (like requests had from 2.32.0 to 2.32.4). https://github.com/psf/requests/pull/6667

Issue #3221, opened 2025-09-11
grpc load testing doesnt scale with number of users


geraldstanje:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am reporting a bug, not asking a question

### Description

hi,

for testing a grpc server with multiple connections - is there anything i need to do to have multiple connections?
i currently looked at https://github.com/locustio/locust/blob/master/examples/grpc/grpc_user.py

i modified this code - but rpc for 1, 2, 5, 10 users is the same - so i wonder why.

i also check cpu usage - im not even using 50% right now.

### Command line

-

### Locustfile contents

```python3
to be added
```

### Python version

3.10

### Locust version

latest

### Operating system

linux

akilesh1706:
@geraldstanje I can work on this issue. Can I get it assigned as part of Hacktober fest 2025?

sarika-03:
hey @cyberw 
can you assign this to me ?

sonianuj287:
Hi @cyberw @geraldstanje , please take a look to my changes. 
[https://github.com/locustio/locust/pull/3246](https://github.com/locustio/locust/pull/3246)

Thanks :)

PR #3246, opened 2025-10-18
grpc load testing

sonianuj287:
related issue - [https://github.com/locustio/locust/issues/3221](https://github.com/locustio/locust/issues/3221)
## Changes Made:
1. Fixed Response Time Measurement in grpc_user.py
Problem: The original code calculated response_time after firing the request event, which could lead to incorrect timing
Solution: Captured the response_time value before firing the event, ensuring accurate measurement
Impact: Response times are now correctly measured and reported
2. Added Proper Resource Cleanup in grpc_user.py
Problem: gRPC channels were never closed when users stopped, leading to resource leaks
Solution: Added an on_stop() method to properly close the gRPC channel
Impact: Better resource management and cleaner test execution
3. Increased Server Capacity in hello_server.py
Problem: The example server had max_workers=10, which limited throughput to ~10 RPS when each request takes 1 second
Solution: Increased max_workers to 100
Impact: The server can now handle up to 100 concurrent requests, allowing RPS to scale with user count

## Root cause 
Issue was likely the server-side bottleneck (only 10 worker threads) combined with the 1-second sleep in each request handler. With only 10 workers, the server could only handle 10 concurrent requests, resulting in approximately 10 RPS regardless of the number of Locust users.


sonianuj287:
Hi @cyberw, please review the PR and suggest me changes -

cyberw:
1. I don't think this makes any real difference? Your solution might be cleaner, but it should realistically have no performance impact.
2. Sure, that makes sense I guess.
3. This is good. @geraldstanje , was your issue with the example server or something else?

Remove the comments you added, I dont think they make sense. And please ignore the intermittent "traceback" issue in tests. We should fix it, but not in this PR and probably not by ignoring it but rather by fixing it.

cyberw:
You have some randomly added (locust_output.json) and deleted files (uv.lock)?

cyberw:
This is too messy. Open a new PR that just increases the thread count (one line change only), that I can merge.

sonianuj287:
> You have some randomly added (locust_output.json) and deleted files (uv.lock)?



> This is too messy. Open a new PR that just increases the thread count (one line change only), that I can merge.

sure, will do

PR #3250, opened 2025-10-26
Add option to force the use of IPv4

MichaelFoleyFZ:
Locust works great in scalable container environments like ECS which has different ways to allow for the workers to talk to master nodes. One option is Service Connect/Envoy Proxy, this adds a host file entry for a service name for both IPv4 and IPv6 based comms. Except while the OS might have an Ipv6 address the added entry can't be resolved and so the connection fails. 

See https://github.com/aws/amazon-ecs-service-connect-agent/issues/120 for a more detailed rundown of the findings for this

This PR  allows you to force IPv4 as the preferred option instead of the current setup which prefers IPv6 if the provided hostname resolves to an IPv6 address.

This can be configured using the --master-ipv4-only argument and will apply to both master and workers

### DIFF
--- a/docs/running-distributed.rst
+++ b/docs/running-distributed.rst
@@ -76,6 +76,17 @@ to localhost)
 
 Optionally used together with ``--worker`` to set the port number of the master node (defaults to 5557).
 
+``--master-ipv4-only``
+----------------------
+
+Optionally used together with ``--worker`` or ``--master`` to force the use of IPv4 (defaults to preferring Ipv6)
+
+.. note::
+    If a hostname is provided in ``--msater-host`` and this can be resolved via getaddrinfo to an IPv6 address
+    this will be preferred by default. Some environments might have an IPv6 hosts file entry in place which
+    doesn't work but is difficult to remove so this allows you to pick.
+
+
 ``--master-bind-host <ip>``
 ---------------------------
 
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -250,9 +250,9 @@ See documentation for more details, including how to set options using a file or
     return parser
 
 
-def download_locustfile_from_master(master_host: str, master_port: int) -> str:
+def download_locustfile_from_master(master_host: str, master_port: int, ipv4_only: bool) -> str:
     client_id = socket.gethostname() + "_download_locustfile_" + uuid4().hex
-    tempclient = zmqrpc.Client(master_host, master_port, client_id)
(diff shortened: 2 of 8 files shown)


cyberw:
In general, I see the need for this feature, so I like the PR. However:

* There's some arbitrary whitespace changes in the docs. Nice change, but should be a separate commit to not pollute the git blame history (also your commit message is strange - probably wanna lead with what you've actually changed, not with "Locust works great..." :)
* I think the Locust worker zmq Client will fall back to ipv4 automatically. Can you try it in your environment? That would make the implementation a bit simpler.

MichaelFoleyFZ:
> In general, I see the need for this feature, so I like the PR. However:
> 
> * There's some arbitrary whitespace changes in the docs. Nice change, but should be a separate commit to not pollute the git blame history (also your commit message is strange - probably wanna lead with what you've actually changed, not with "Locust works great..." :)

fixed up the whitespace removal, my vscode setup code was configured to do it automatically and have fixed up the git commit. Should have been the PR description 

> * I think the Locust worker zmq Client will fall back to ipv4 automatically. Can you try it in your environment? That would make the implementation a bit simpler.

The client was the one that was causing my issue, the worker was the one that was resolving the IPv6 address to get to the master.  The use_ipv4 function seems to be used by both client and server as its performed in the BaseSocket class, it probably doesn't need to be in the master config since you can control it with bind, but since they both had a consistent setup I figured it would be easier to say you could just set the ipv4 only flag on both to ensure its consistent.

cyberw:
Did you try using `export ZMQ_IPV6=0`? That seems to be the standard way for disabling ipv6 in zmq.

MichaelFoleyFZ:
> Did you try using `export ZMQ_IPV6=0`? That seems to be the standard way for disabling ipv6 in zmq.

Set the the env var and updated to use the hostname but looks like it doesn't apply 

```
[2025-10-27 22:34:33,241] ip-10-0-30-5/INFO/locust.runners: Failed to connect to master master.devops-hyena-loadtest:5557, retry 15/60.
```

 I'm guessing if there is an explicit config on the socket that wins? 
 https://github.com/locustio/locust/blob/2d21010ebb2064d0ff02b17b8a9c4e9138e8ec1a/locust/rpc/zmqrpc.py#L22C3-L22C4

Would it be better to remove the IPv6 handling in locust if zeromq offers those env vars?

cyberw:
> Would it be better to remove the IPv6 handling in locust if zeromq offers those env vars?

I think it might be. At the very least we can change that line of code to also respect the zeromq env var?

PR #3252, opened 2025-10-31
fix: MQTT client_id and protocol not passed down to Client

ionutab:
When migrating the MqttClient from Locust Plugins I accidentally modified the call to the parent constructor.

https://github.com/SvenskaSpel/locust-plugins/pull/224/commits/f1845b924bc435751c6c0d18c244f8b78d4e86f9#diff-94fac3dc568e54a6e6f7dc07955bfe16912cea237d2dfcc98f54d97d9d556071L115

### DIFF
--- a/locust/contrib/mqtt.py
+++ b/locust/contrib/mqtt.py
@@ -101,7 +101,7 @@ class MqttClient(mqtt.Client):
         if not client_id:
             client_id = f"locust-{_generate_random_id(16)}"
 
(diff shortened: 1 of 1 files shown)


ionutab:
should I add a test to ensure that the protocol is passed on correctly? 
I'm not sure how I would implement it. 
Would it be worth it?
Thanks

cyberw:
No need for a test, but probably nice to add an example for how to use it!

cyberw:
I just realized that you were only correcting a mistake in moving it, in which case no need for anything more :)

PR #3253, opened 2025-10-31
remove setuptools from dependencies

dotlambda:
It isn't used anywhere.


cyberw:
Good catch.

cyberw:
Well... setuptools is still an indirect dependency, from gevent so it doesnt make any significant difference.

```
gevent==24.11.1
├── zope.event [required: Any, installed: 5.0]
│   └── setuptools [required: Any, installed: 80.9.0]
├── zope.interface [required: Any, installed: 6.3]
│   └── setuptools [required: Any, installed: 80.9.0]
└── greenlet [required: >=3.1.1, installed: 3.2.1]
```

bollwyvl:
> still an indirect dependency

welp, it looks like `setuptools` has been/will be removed from: 
- the latest version of [`zope.interface`](https://github.com/zopefoundation/zope.interface/compare/8.0...8.0.1)
- the next release of [`zope.event`](https://github.com/zopefoundation/zope.event/compare/6.0...master)

hopefully all this gets resolved before the _next_ breaking `setuptools` change: the seemingly monthly new warnings are tiresome and confusing to users.

PR #3264, opened 2025-11-10
Creates dependabot file to keep gh actions updated

jairhenrique:


cyberw:
Cool stuff. Does it open a PR or what?

jairhenrique:
> Cool stuff. Does it open a PR or what?

@cyberw yes, dependabot open pull requests with updates. We can configure it for other stuffs too, like python, node etc. https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories

PR #3265, opened 2025-11-12
Enables FURB ruff lint

jairhenrique:
https://docs.astral.sh/ruff/rules/#refurb-furb


cyberw:
Cool stuff! I'll fix mypy in master.

PR #3268, opened 2025-11-16
adding mqtt user feature that works around the paho mqtt 340 connections limit issue

ionutab:
Hello,

There's a limitation caused by the use of `select` in the [paho.mqtt.python](https://github.com/eclipse-paho/paho.mqtt.python) library.

https://github.com/eclipse-paho/paho.mqtt.python/issues/697

The purpose of this PR is to provide an implementation of `MqttClient` that has an overriden `_loop` method that uses `selectors` instead of `select`.

The implementation was initially posted on the solution above, by @j04n-f

Please advise on:
🤔 I'm not sure about the `Experimental` name.
📌 Code placement in a separate file.
🧀 If any specific tests are desired and or necessary.
🩻 What better example I could provide.

After the implementation is confirmed I will add documentation and README file updates.

Thanks!

### DIFF
--- a/locust/contrib/mqtt.py
+++ b/locust/contrib/mqtt.py
@@ -332,8 +332,8 @@ class MqttClient(mqtt.Client):
     def _loop(self, timeout: float = 1.0) -> MQTTErrorCode:
         """Override the parent's _loop method to optionally use selectors.
 
-        When use_loop_selectors is True, this uses a selector-based implementation
-        for better performance. Otherwise, it falls back to the parent's implementation.
+        When use_loop_selectors is True, this uses a selector-based implementation that allows more than 340 connections.
+        Otherwise, it falls back to the parent's implementation.
         """
         if self._use_loop_selectors:
             return self._loop_selectors(timeout)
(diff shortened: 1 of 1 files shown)


cyberw:
> Code placement in a separate file.

Maybe make it a flag in MqttUser instead (passed on to MqttClient triggering a replace of the _loop method in its constructor)? If this is a good fix, maybe it should even be the default?

Perhaps we should log a warning if someone tries to launch >340 connections with the default setup?

ionutab:
Good ideas. I'll see what I can do. Thanks!

ionutab:
I've made the update.
Need to test it.

I'm not sure where to add the validation for the 340 warning message.
Is it in the `start` method in the `runners.py` file?

ionutab:
commenting this so that it does not get closed.

cyberw:
Hey, sorry for the slow response, I must have misunderstood and thought you were going to do something more.

I guess we can merge this as-is. But isnt it better to _always_ ues the custom loop implementation?

cyberw:
@ionutab ?

ionutab:
Hi,
To be honest, I'm a little hesitant to set `_use_loop_selectors ` to `True` in this pull request in order not to break things for other users.
In the long term the correct solution would be to always use the custom loop implementation,  indeed.
I would be more confident to set it as the default in a couple of weeks, maybe by then others will have used the implementation as well.
How does that sound?

PR #3275, opened 2025-11-17
Add OpenTelemetry support

amadeuppereira:
This also configures auto instrumentation for `HttpUser` by instrumenting `requests` package

### DIFF
--- a/locust/opentelemetry.py
+++ b/locust/opentelemetry.py
@@ -7,21 +7,20 @@ logger = logging.getLogger(__name__)
 
 
 def setup_opentelemetry():
-    traces_exporters = [e.strip().lower() for e in os.getenv("OTEL_TRACES_EXPORTER", "otlp").split(",") if e.strip()]
-    metrics_exporters = [e.strip().lower() for e in os.getenv("OTEL_METRICS_EXPORTER", "otlp").split(",") if e.strip()]
-
-    traces_exporters = [e for e in traces_exporters if e != "none"]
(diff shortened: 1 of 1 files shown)


locust/opentelemetry.py
```diff
@@ -0,0 +1,148 @@
...
+        tracer_provider = _setup_tracer_provider(resource, traces_exporters)
+        trace.set_tracer_provider(tracer_provider)
+
+    if metrics_exporters:
+        meter_provider = _setup_meter_provider(resource, metrics_exporters)
+        metrics.set_meter_provider(meter_provider)
+
+    _setup_auto_instrumentation()
+
+    logger.info("OpenTelemetry configured!")
```
cyberw: debug level

pyproject.toml
```diff
@@ -57,6 +57,12 @@ dependencies = [
 milvus = ["pymilvus>=2.5.0"]
 mqtt = ["paho-mqtt>=2.1.0"]
 dns = ["dnspython>=2.8.0"]
+otel = [
+    "opentelemetry-sdk>=1.38.0",
+    "opentelemetry-exporter-otlp-proto-grpc>=1.38.0",
```
cyberw: should we also install http? if we do, we can also remove the slightly complicated handling of ImportError.

In general, we dont need to support an incomplete installation (missing urllib3 instrumentation or whatnot). Just crashing is cleaner.

PR #3282, opened 2025-11-19
Avoid duplicated action run on pull requests

jairhenrique:
Screenshot 2025-11-19 at 14-43-01 Bump the vite group in _locust_webui with 6 updates by dependabot bot · Pull Request #3271 · locustio_locust

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,11 +2,15 @@ name: Tests
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


cyberw:
There was a reason for this (running builds in locustio/locust on all branches before opening a PR), but since I'm pretty much the only one pushing directly to locustio I think this is an improvement. Thanks :)

PR #3293, opened 2025-11-25
Fix Toml Parser Being Called on Conf Files

andrewbaldwin44:
### Proposal
- Adds a class to detect if the TOML or default config file parser should be used and parse the file accordingly
- Update docs


cyberw:
:(
 
http://github.com/locustio/locust/actions/runs/19674964715/job/56353599726?pr=3293

PR #3327, opened 2026-01-08
Fix: Resolve response time data loss and type checking issues in distributed mode

MasterKey-Pro:
Core Fixes:
Add the use_response_times_cache parameter to StatsEntry.unserialize() to properly initialize the cache when the master node deserializes worker node data
Pass use_response_times_cache=True in on_worker_report() to ensure response time cache consistency
Type System Improvements:
Modify the parameter type of the StatsEntry.__init__() method to str | None
Modify the StatsHolder protocol method to str | None to support covariance
Modify the parameter type of sort_stats() to Mapping[Any, S] to support covariant types
Modify the return type of CSVWriter.writerow() to Any
Modify the parameter type of the avg() function to Sequence[float | int] to avoid list covariance issues
Initialize the variable k in median_from_dict() to prevent unbound variable warnings
Modify on_worker_report() to use entry.method or "" to ensure type safety
Modify _failures_data_rows() to perform direct sorting and avoid protocol incompatibility

### DIFF
--- a/locust/test/test_stats.py
+++ b/locust/test/test_stats.py
@@ -23,7 +23,6 @@ import os
 import re
 import time
 import unittest
-from collections import OrderedDict
 from io import StringIO
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


locust/stats.py
```diff
@@ -43,7 +43,7 @@
 
 class CSVWriter(Protocol):
     @abstractmethod
-    def writerow(self, columns: Iterable[str | int | float]) -> None: ...
+    def writerow(self, row: Iterable[str | int | float | None]) -> Any: ...
```
cyberw: each entry in the Iterable is a column, so columns is a better name, row is less specific.

cyberw:
I have some comments, but before I do a full review, can you add a test case that this fixes? 

I'm still not sure what the impact of the issue is and a test case would help clarify.

MasterKey-Pro:
Supplements the StatsEntry deserialization with the use_response_times_cache parameter for full-scenario test cases.

cyberw:
That's too much instead :) I just want a single test that exposes the problem you were having, no need to test implementation details. Are you using AI?

MasterKey-Pro:
> That's too much instead :) I just want a single test that exposes the problem you were having, no need to test implementation details. Are you using AI?

1、 Resubmitted a simplified version of the test case
2、 Found that response time recording was abnormal on different servers, and didn't know when it would be fixed, so I used AI to try to fix it

cyberw:
Did you use AI to generate the fix? If so, I don't have time to review it, sorry.

I like that you reduced the number of tests, but that test doesn't actually show the problem of data loss, does it?

PR #3354, opened 2026-02-19
Add Qdrant support

Anush008:
## Description

This PR introduces `locust/contrib/qdrant.py` and examples to demonstrate Locust load testing on a [Qdrant](https://qdrant.tech/) server.

The dependencies are optional and gated behind a `qdrant` extra.


cyberw:
Nice! Typically User classes provide the option to set configuration (like host/url or collection name) as class variables instead of having to override `__init__` for regular use cases. Maybe you can take the time to do that?

cyberw:
Nice! 🚀

Issue #2005, opened 2022-02-09
Allow to customize rounded_response_time


mariadb-DmitryVolkov:
I understand the intention beyond rounding response time in stats.py but it would be great if locust allows redefining this logic.

cyberw:
That would be at least moderately hard to do, so it is probably something you'd have to do yourself (I'm not against such a feature, but I dont know who else would build it)

If you want exact metrics, maybe https://github.com/SvenskaSpel/locust-plugins/tree/master/locust_plugins/timescale is something for you.

mariadb-DmitryVolkov:
Thanks, @cyberw - this is great. Eventually, I think I will use this plugin, yes. But I am looking for an intermediate solution as well. 
I see that stats.py has a bunch of variables like PERCENTILES_TO_REPORT, why not introduce yet another variable for precision?

cyberw:
> Thanks, @cyberw - this is great. Eventually, I think I will use this plugin, yes. But I am looking for an intermediate solution as well.
> I see that stats.py has a bunch of variables like PERCENTILES_TO_REPORT, why not introduce yet another variable for precision?

I’m not against it, someone just needs to build it :)

Issue #255, opened 2015-03-16
Rounding of percentiles (premature optimization?)


rnewson:
I noticed that all my percentile values appear to be rounded. On investigation, it's deliberate (commit 7f0b2d67) for bandwidth reasons.

Is it true that sending 58000 is more efficient than sending 58459? Is it true enough to warrant discarding this much data?

heyman:
Hi!

No, this was made out of necessity and not prematurely. 

It's not that sending the integer 58000 takes up less space than 58459. We construct a dict whose keys are the response time, and the values are the number of requests that are completed in that response time. 

So let's say we have a test running whose response times are between 200 ms and 20k ms. If you run that test long enough (without the rounding), you're going to have a dict that contains almost 20k keys, and dicts that have at least a couple of thousand keys are going to be sent from each slave node to the master node every third second (and you might have many hundreds of slave nodes). Compare this to the current version (when we do rounding) where the dict wouldn't have more than 310 keys.

Also, I can't see any load testing scenario where the lost precision of the rounding would matter.

rnewson:
ok, I see how that reduces the size of the dict, though it raises further questions (for another time).

Perhaps the reporting of percentiles should 'x' out the values to more clearly demonstrate that the values have been rounded? I first starting looking at it when I saw all my stats rounded.

heyman:
The reasoning behind it is that +/-10ms rounding is okay for for requests between 100 ms and 10k ms, and +/-100 is OK for >1k < 10k, etc. There might be cases where you would benefit from knowing that the 95th percentile is exactly 58459 ms rather than between 57500 and 58499, but if they exist they should be extremely rare.

I don't think we should "X out" the zeroes (since that might complicate further processing of the data), but it could probably be a good idea to document this behaviour better.

rehevkor5:
I am not sure that the histogram bucketing approach is necessarily a problem, although I believe there are probably ways to implement a fully accurate report at the end without worrying about streaming data to a master node during test execution. And I think that the UI should communicate some kind of accuracy estimate, such as "+/- X" where X is some value.

[...]

PR #3373, opened 2026-03-13
Extract response time bucketing into an overridable function

thessem:
Locust rounds response times into histogram buckets in `StatsEntry._log_response_time`. The rounding logic is not overridable, which means users who need different bucketing granularity have no clean way to change it. This PR extracts it into a module-level function so it can be replaced at runtime, the same way `PERCENTILES_TO_CHART` can be.

Our (MongoDB) use-case has us sometimes measuring things in the high microseconds/very low milliseconds range all the way up to the low seconds range. To get the needed range we report microseconds to Locust (I'm aware this is a horrible hack, sorry). For tests that need to cover both extremes we end up with too many buckets because the rounding cuts off at `round(response_time, -3)`. Extracting the function gives us an avenue to fix that.

I expect usage to look something like:

``` python
import locust.stats
from math import floor, log10

def sigfig_bucket(response_time):
    """Bucket to 3 significant figures."""
    if response_time == 0:
        return 0
    return int(round(response_time, -int(floor(log10(abs(response_time)))) + 2))

locust.stats.bucket_response_time = sigfig_bucket
```

I annotated the argument as `int | float` since this is now a public interface and it is easier to later narrow the arg type than to broaden it without breaking user implementations. Happy to drop the `float` if you'd prefer to keep it strict. I'd like to discuss plumbing `float` through the stack at some point separately.

The docs update might be more than needed and I haven't tested how it renders. Happy to trim that.

Fixes [\#2005](https://github.com/locustio/locust/issues/2005) (closed as stale). Related to [\#255](https://github.com/locustio/locust/issues/255).

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -310,12 +310,12 @@ You can replace the bucketing function to change this behaviour:
 .. code-block:: python
 
     import locust.stats
+    from math import floor, log10
 
-    def my_bucket_function(response_time):
+    def my_bucket_function(response_time: int | float) -> int:
         """Example: bucket to 3 significant figures."""
         if response_time == 0:
             return 0
-        from math import floor, log10
         return int(round(response_time, -int(floor(log10(abs(response_time)))) + 2))
 
     locust.stats.bucket_response_time = my_bucket_function


cyberw:
Looks great. Made some minor comments, happy to merge if you can have a look at those.

cyberw:
Nice!

PR #3379, opened 2026-03-21
Add locust-otel Docker image with OpenTelemetry dependencies

### LINKED ISSUE #3309: Docker image does not have OTEL libraries installed
### Prerequisites

### Description

When I enable OTEL via the instructions at https://www.locust.cloud/blog/observable-load-testing/ I get the following error when starting the container with the `--otel` flag:
[2025-12-05 18:53:37,231] e4744a99b8a5/ERROR/locust.opentelemetry: OpenTelemetry SDK is not installed, opentelemetry not enabled. Run 'pip install locust[otel]'

### Command line

docker run locustio/locust:latest --host https://host.docker.internal:49942 --users 10 --spawn-rate 2 --run-time 300s --locustfile /mnt/locust/Load --processes -1 --otel

### Locustfile contents

```python3
Issue triggers before needing a file
```

### Python version

whatever is in the container

### Locust version

2.42.6

### Operating system

linux

eyupcanakman:
The official Docker image does not include the OpenTelemetry exporter libraries, so running with `--otel` fails with an import error.

Added a `Dockerfile.otel` that installs the `[otel]` extras and publishes a separate `locustio/locust-otel` image. A CI test job verifies the OTel packages import correctly before publish.

Fixes #3309

### DIFF
--- a/docs/running-in-docker.rst
+++ b/docs/running-in-docker.rst
@@ -42,4 +42,3 @@ official Locust docker image as a base image::
 
     FROM locustio/locust
     RUN pip install some-package # some dependency you need
-    RUN pip install locust[otel] # for OpenTelemetry support


cyberw:
Niiice! Would it be possible to chain the build to reduce duplication? So the other file becomes just something like this (not tested)

FROM locustio/locust
RUN pip install "$(echo dist/*.whl)[otel]"

eyupcanakman:
@cyberw The Dockerfile already chains from the base image. It does FROM ${BASE_IMAGE} and installs the wheel with [otel] extras on top. The COPY ./dist is needed because the wheel isn't on PyPI at build time.

Fixed the CI failure and removed the otel line from the base image example. The otel build was using Buildx, which runs in an isolated BuildKit container that can't see images loaded into the host Docker daemon via load: true. Switched to plain docker build for that step.

Issue #1152, opened 2019-11-15
When locust exits the current RPS is outputted instead of the total RPS


heyman:
When locust exits the current request/s (an average for the last 10 seconds) and failures/s is outputted instead of the *total* requests/s for the whole test run (or since the last stats reset). It makes more sense to output the total stats.

PR #3384, opened 2026-03-30
fix: use total_rps instead of current_rps in HTML report and navbar stats

### LINKED ISSUE #3355: HTML report uses current_rps instead of total_rps for req/s column
### Prerequisites

### Description

The HTML report generated via `--html` uses `current_rps` (a ~10-second
sliding window average) for the **req/s** column instead of `total_rps`
(total requests / total test duration). This makes the HTML report
inconsistent with the **final console summary**, which correctly uses
`total_rps` since the fix in #1152.

Since the HTML report is a post-test artifact representing the
**entire test run**, it should use `total_rps` — not a stale snapshot
of whatever the last ~10 seconds of the test looked like.

### How the inconsistency arises

**`html.py` calls `to_dict()` without any `current` flag:** [Code Link](https://github.com/locustio/locust/blob/027c2cf4d68598876c263ffdea5149703039bfc7/locust/html.py#L83)

```
"requests_statistics": [stat.to_dict() for stat in requests_statistics],
```

### Command line

locust -f locustfile.py --headless -u 5 -r 5 -t 30s --html repro_report.html --only-summary

### Locustfile contents

```python3
from locust import HttpUser, task, between


class ReproUser(HttpUser):
    wait_time = between(0.1, 0.2)
    host = "https://httpbin.org"

    @task
    def get_ip(self):
        self.client.get("/ip")
```

### Python version

3.12

### Locust version

2.32

### Operating system

Linux

juliosuas:
## Summary

Fixes #3355.

The stats API endpoint used `current_rps` (a ~10-second sliding-window average) to populate `total_rps` and `total_fail_per_sec` in the report payload. This made the HTML report inconsistent with the final console summary, which correctly uses `total_rps` (total requests ÷ elapsed time) since #1152.

## Root Cause

```python
# locust/web.py line 486
report["total_rps"] = total_stats["current_rps"]         # ← wrong
report["total_fail_per_sec"] = total_stats["current_fail_per_sec"]  # ← wrong
```

## Fix

```python
report["total_rps"] = total_stats["total_rps"]           # ✓ aggregate rate
report["total_fail_per_sec"] = total_stats["total_fail_per_sec"]    # ✓
```

Since the HTML report is a post-test artifact representing the entire run, it should always show the aggregate rate — not whatever the last ~10 seconds happened to look like.

### DIFF
--- a/locust/test/test_web.py
+++ b/locust/test/test_web.py
@@ -130,6 +130,20 @@ class TestWebUI(LocustTestCase, _HeaderCheckMixin):
         self.assertEqual("Aggregated", data["stats"][1]["name"])
         self.assertEqual(1, data["stats"][1]["num_requests"])
 
+    def test_html_report_uses_total_rps_not_current_rps(self):
+        self.stats.log_request("GET", "/test", 100, 1000)
+        self.stats.log_request("GET", "/test", 120, 1200)
+        response = requests.get("http://127.0.0.1:%i/stats/requests" % self.web_port)
+        self.assertEqual(200, response.status_code)
+
+        data = json.loads(response.text)
+        self.assertIn("total_rps", data)
+        self.assertIn("total_fail_per_sec", data)
+
+        total_entry = data["stats"][-1]
+        self.assertEqual(total_entry["total_rps"], data["total_rps"])
+        self.assertEqual(total_entry["total_fail_per_sec"], data["total_fail_per_sec"])
+
     def test_stats_cache(self):
         self.stats.log_request("GET", "/test", 120, 5612)
         response = requests.get("http://127.0.0.1:%i/stats/requests" % self.web_port)
--- a/locust/web.py
+++ b/locust/web.py
@@ -483,8 +483,13 @@ class WebUI:
             total_stats = _stats[-1]
 
             if _stats:
-                report["total_rps"] = total_stats["current_rps"]
-                report["total_fail_per_sec"] = total_stats["current_fail_per_sec"]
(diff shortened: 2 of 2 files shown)


juliosuas:
Added some context: I traced this through `web.py` line 486 where `report['total_rps'] = total_stats['current_rps']` was using the sliding-window metric. The fix simply uses `total_stats['total_rps']` and `total_stats['total_fail_per_sec']` instead — both are already present in the stats dict from `to_dict()`. The frontend's `HtmlReport.tsx` already had `totalRps` in its table structure (line 29), so this was purely a backend data source mismatch. Happy to add a test if that would help.

cyberw:
Nice! The fix looks good. No need for the code comments, and would love to see a test case too! Funny how I wasn’t able to reproduce it..

juliosuas:
Thanks for the feedback, @cyberw! I've pushed two changes:

1. **Removed the inline comments** — agree the code is self-explanatory without them
2. **Added a test** `test_html_report_uses_total_rps_not_current_rps` that logs two requests, hits `/stats/requests`, and asserts the report-level `total_rps` matches the Aggregated entry's value (total_requests / elapsed_time) rather than the sliding-window metric

The test follows the same pattern as the existing `test_stats()` so it should fit right in. Let me know if you'd like any adjustments.

cyberw:
Thanks, great stuff!

abhi2196:
Thanks @juliosuas for the fix!

juliosuas:
It was a pleasure to contribute to the project! 😄

Issue #3207, opened 2025-09-05
Python 3.13: KeyError: 'name' on request_meta in ResponseContextManager.__exit__


tanveebadheka:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am reporting a bug, not asking a question

### Description

**Background:**
[Appian](https://appian.com/about/explore/overview) is a low-code software platform that lets businesses build custom applications using a visual, drag-and-drop interface. It helps companies automate their processes by connecting different systems. 

[Appian Locust](https://appian-locust.readthedocs.io/en/latest/what_is_appian_locust.html) is an open source wrapper library around Locust for load testing Appian ([repository](https://gitlab.com/appian-oss/appian-locust)). 

Appian Locust was on Python 3.10.7 until May 2025, when we upgraded it to Python 3.13.0. We have recently received reports of users running into this Key Error, notably when running their Appian Locust tests with more than 1 user.

**Bug description:** 
  `KeyError: 'name' in ResponseContextManager.__exit__` when running Appian Locust Tests in Python 3.13.X with multiple users
```[2025-09-05 09:58:54,420] lxwdn4qpqy/ERROR/locust.user.task: 'name'
Traceback (most recent call last):
  File "/Users/tanvee.badheka/.local/share/virtualenvs/appian-locust-foSro4o6/lib/python3.13/site-packages/locust/user/task.py", line 355, in run
    self.execute_next_task()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/tanvee.badheka/.local/share/virtualenvs/appian-locust-foSro4o6/lib/python3.13/site-packages/locust/user/task.py", line 388, in execute_next_task
    self.execute_task(self._task_queue.popleft())
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tanvee.badheka/.local/share/virtualenvs/appian-locust-foSro4o6/lib/python3.13/site-packages/locust/user/task.py", line 502, in execute_task
    task(self.user).run()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/Users/tanvee.badheka/.local/share/virtualenvs/appian-locust-foSro4o6/lib/python3.13/site-packages/locust/user/task.py", line 340, in run
    self.on_start()
    ~~~~~~~~~~~~~^^
[...]

Issue #3050, opened 2025-02-19
Python 3.13: `KeyError: 'name'` on request_meta in ResponseContextManager.__exit__ after GC


ktosiek:
### Prerequisites

- [x] I am using [the latest version of Locust](https://github.com/locustio/locust/releases/)
- [x] I am reporting a bug, not asking a question

### Description

After upgrade to Python 3.13 I've started seeing errors when running locust (I've added the traceback at the end). It seems if a GC happens within `with self.client.get(..., catch_response=True)` then somehow ResponseContextManager.request_meta gets cleared.

Changing `ResponseContextManager.__init__` to copy the `response.__dict__`, instead of just grabbing it, seems to help. I'm not sure how that change affects the rest of the project.

## The Traceback on failure:
```
[2025-02-19 07:59:11,514] SDUPC07/ERROR/locust.user.task: 'name'
Traceback (most recent call last):
  File "/home/wsl/.cache/uv/archive-v0/s81GZnyMdf_txEZnLJlMJ/lib/python3.13/site-packages/locust/user/task.py", line 340, in run
    self.execute_next_task()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/wsl/.cache/uv/archive-v0/s81GZnyMdf_txEZnLJlMJ/lib/python3.13/site-packages/locust/user/task.py", line 373, in execute_next_task
    self.execute_task(self._task_queue.popleft())
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsl/.cache/uv/archive-v0/s81GZnyMdf_txEZnLJlMJ/lib/python3.13/site-packages/locust/user/task.py", line 490, in execute_task
    task(self.user)
    ~~~~^^^^^^^^^^^
  File "/home/wsl/locust-gc-bug-demo/locustfile.py", line 10, in get_json
    with self.client.get("/json", catch_response=True) as get_json:
         ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsl/.cache/uv/archive-v0/s81GZnyMdf_txEZnLJlMJ/lib/python3.13/site-packages/locust/clients.py", line 368, in __exit__
    self.url = self.request_meta["name"]
               ~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'name'
```

### Command line

[...]

PR #3391, opened 2026-04-05
fix: break reference cycle in ResponseContextManager (fixes #3388)

### LINKED ISSUE #3388: FastHttpUser crashes with TypeError: on_request() missing 4 required positional arguments on Python 3.13+
### Prerequisites

### Description

`ResponseContextManager.__init__` in fasthttp.py shares a dict between two objects using `self.__dict__ = response.__dict__`. This, combined with `request_meta["response"] = response`, creates a reference cycle. On Python 3.13+, the GC collects this cycle mid-request and clears request_meta to {}. When the with-block exits, events.request.fire(**{}) is called with no arguments, triggering the TypeError.

The bug is intermittent and more likely with large responses because gevent yields to the GC during I/O.

**Root cause**

`ResponseContextManager.__init__` in locust/contrib/fasthttp.py:

```
def __init__(self, response, request_event, request_meta, catch_response):
    self.__dict__ = response.__dict__   # <- REFERENCE share, not a copy
    self.request_meta = request_meta    # <- stored in the shared dict
```
And in FastHttpSession.request():
```
request_meta = {
    ...
    "response": response,   # <- response object stored here
}
return ResponseContextManager(response, self.request_event, request_meta, catch_response)
# After this return, 'response' local variable goes out of scope
```

**Proposed fix**

Copy `response.__dict__` instead of sharing it:
```
def __init__(self, response, request_event, request_meta, catch_response):
    self.__dict__.update(response.__dict__)  # copy, not reference share
    try:
        self._cached_content = response._cached_content
    except AttributeError:
        pass
    self._request_event = request_event
    self.request_meta = request_meta
    self._catch_response = catch_response
```

**Related**

Issue [#3050](https://github.com/locustio/locust/issues/3050) 
Issue [#3207](https://github.com/locustio/locust/issues/3207) 

### Command line

locust -f locustfile.py --headless -u 1 -r 2 -t 60s

### Locustfile contents

```python3
from locust import task, FastHttpUser

class POCUser(FastHttpUser):
    host = "https://api.openalex.org"

    @task
    def get_response(self):
        with self.client.get(
            "/works?search=machine+learning&per-page=1",
            name="works",
            catch_response=True,
        ) as response:
            response.json()  # large response -> gevent I/O -> GC fires-> TypeError
```

### Python version

3.14.0 (also reproducible on 3.13.x per issues #3050 and #3207)

### Locust version

2.43.3

### Operating system

macOS (also Linux)

kaiisfree:
## Summary

- **Root cause:** `ResponseContextManager.__init__` used `self.__dict__ = response.__dict__`, which shares (not copies) the underlying dict object. Combined with `request_meta["response"] = response`, this creates a reference cycle. Python 3.13+ incremental GC collects this cycle mid-request, clearing `request_meta` to `{}`, causing `TypeError` in `events.request.fire(**{})`.
- **Fix:** Changed `self.__dict__ = response.__dict__` to `self.__dict__.update(response.__dict__)`, which copies the attributes into a new dict, breaking the reference cycle.
- **Impact:** This crash only manifests on Python 3.13+ due to changes in the garbage collector (PEP 703 / incremental GC). On older Python versions the cycle existed but was rarely collected at a problematic time.

## Related issues

- Fixes #3388
- Also addresses the root cause behind #3050 and #3207, which reported similar `TypeError` / `request_meta` clearing symptoms

## Test plan

- [x] Added unit test `test_response_context_manager_does_not_share_dict` verifying that `ResponseContextManager.__dict__` is not the same object as `response.__dict__` after construction
- [ ] Verify existing `TestFastHttpCatchResponse` tests still pass
- [ ] Test with Python 3.13+ under load to confirm the `TypeError` crash no longer occurs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -710,7 +710,7 @@ class ResponseContextManager(FastResponse):
         # copy data from response to this object
         # I wanted to change this to the approach from the HttpUser's ResponseContentManager.from_request,
         # but it was such a mess
-        self.__dict__ = response.__dict__
+        self.__dict__.update(response.__dict__)
         try:
             self._cached_content = response._cached_content
         except AttributeError:
--- a/locust/test/test_fasthttp.py
+++ b/locust/test/test_fasthttp.py
@@ -806,6 +806,20 @@ class TestFastHttpCatchResponse(WebserverTestCase):
         self.assertEqual(0, self.num_success)
 
 
+    def test_response_context_manager_does_not_share_dict(self):
+        """Test that ResponseContextManager copies response.__dict__ instead of sharing it.
+        Sharing the dict creates a reference cycle that Python 3.13+ GC can collect
+        mid-request, clearing request_meta and causing TypeError (fixes #3388).
+        """
+        response = self.user.client.get("/ultra_fast")
+        from locust.contrib.fasthttp import ResponseContextManager
+        request_meta = {"request_type": "GET", "name": "/test", "context": {}, "response": response,
+                        "exception": None, "start_time": 0, "url": "/test", "response_time": 0, "response_length": 0}
+        rcm = ResponseContextManager(response, self.environment.events.request, request_meta, True)
+        # After construction, rcm.__dict__ must NOT be the same object as response.__dict__
+        self.assertIsNot(rcm.__dict__, response.__dict__,
+                         "ResponseContextManager should copy, not share, response.__dict__")
+
 class TestFastHttpSsl(LocustTestCase):
     def setUp(self):
         super().setUp()


cyberw:
Please add a test case that causes the crash. And remove the one you added, it seems unnecessary.

Probably won't merge this because it is AI, but maybe...

antonkislinskiy:
Hey @cyberw, this PR seems stale...

 here's a test that reproduces the actual crash path. It simulates what the GC does (clearing request_meta via the shared dict) and asserts the TypeError that users hit:

`
class TestResponseContextManagerDictIsolation(WebserverTestCase):
    #Regression test for issue #3388: shared __dict__ between response and
    #ResponseContextManager lets GC clear request_meta mid-request

    def setUp(self):
        super().setUp()

        class MyUser(FastHttpUser):
            host = "http://127.0.0.1:%i" % self.port

        self.user = MyUser(self.environment)

    def test_request_meta_survives_response_dict_clear(self):
        response = self.user.client.get("/ultra_fast")
        request_meta = {
            "request_type": "GET",
            "name": "/test",
            "context": {},
            "response": response,
            "exception": None,
            "start_time": 0,
            "url": "/test",
            "response_time": 0,
            "response_length": 0,
        }
        rcm = ResponseContextManager(response, self.environment.events.request, request_meta, True)
        """
        Simulate what the GC does when it collects the reference cycle:
        it clears the dict that both objects shared, wiping request_meta.
        With the fix, rcm has its own dict so this only affects response.
        response.__dict__.clear()

        With the old code (self.__dict__ = response.__dict__), this would
        raise KeyError because request_meta was wiped along with everything else.
        """
        self.assertEqual(rcm.request_meta["request_type"], "GET")
`

Want me to open a separate PR ?

cyberw:
Merged nearly identical PR (#3398)

PR #3399, opened 2026-04-25
Add AI-optimized documentation (llms.txt)

nk-tedo-001:
## Summary

  Adds a Sphinx extension that auto-generates `llms.txt` and `llms-full.txt` during the
  normal HTML docs build, following the [llms.txt standard](https://llmstxt.org/).

  - `llms.txt` — lightweight index with links and descriptions for all doc pages
  - `llms-full.txt` — complete documentation in a single Markdown file (~210KB)

  Both files are derived artifacts (like HTML) generated from the same RST sources —
  **no content duplication**. This addresses the concern from [PR #3123](https://github.com/locustio/locust/pull/3123)
  which was closed because it introduced a separate `docs-md/` directory with duplicated content.

  Closes #3372

  ## How it works

  A Sphinx extension (`docs/_ext/llms_txt.py`) hooks into the `build-finished` event and:
  1. Walks the toctree structure to build the `llms.txt` index
  2. Converts each document's doctree to Markdown using `sphinx-markdown-builder`
  3. Concatenates all pages into `llms-full.txt`
  4. Writes both files to the HTML output directory

  Changelog, FAQ, and further-reading pages are excluded to keep files focused on actionable docs.

  ## Files changed

  | File | Change |
  |------|--------|
  | `docs/_ext/llms_txt.py` | New Sphinx extension (~150 lines) |
  | `docs/ai-docs.rst` | New doc page describing the AI-optimized files |
  | `docs/conf.py` | Register the extension |
  | `docs/index.rst` | Add ai-docs to toctree |
  | `pyproject.toml` | Add `sphinx-markdown-builder` to docs deps |

  ## Post-merge action

  Configure RTD redirects so AI tools can find files at the standard paths:
  - `/llms.txt` → `/en/stable/llms.txt`
  - `/llms-full.txt` → `/en/stable/llms-full.txt`

  ## Test plan

  - [x] `sphinx-build -b html docs/ docs/_build/` succeeds (4 pre-existing warnings only)
  - [x] `docs/_build/llms.txt` follows llmstxt.org spec (H1, blockquote, H2 sections with links)
  - [x] `docs/_build/llms-full.txt` contains all expected pages in Markdown (~5600 lines)
  - [x] Excluded pages (changelog, faq, further-reading) are absent from generated files
  - [x] `docs/_build/ai-docs.html` renders correctly

PR #3400, opened 2026-04-25
Embed client-side chart snapshot in downloaded reports

MichaelSpece:
## Summary
- Update `Download Report` so it captures a client-side SwarmCharts PNG from the existing dashboard chart history and injects it into the downloaded HTML report.
- Keep the existing Charts-tab ECharts toolbar export as the standalone PNG path; do not add a duplicate Reports-tab PNG button.
- Strip duplicated chart history from the downloaded report payload once the PNG snapshot is embedded, while letting HtmlReport render either injected PNG or report-time charts.
- Keep live chart points aligned with report-history chart points by sourcing RPS, failures/s, and average response time from the aggregate current stats row.

## Testing
- `node .yarn/releases/yarn-4.12.0.cjs lint`
- `node .yarn/releases/yarn-4.12.0.cjs type-check`
- `node .yarn/releases/yarn-4.12.0.cjs vitest run src/hooks/tests/useFetchStats.test.tsx --silent=false`
- `node .yarn/releases/yarn-4.12.0.cjs test --run`
- `git diff --check`

## Review note
This addresses Andrew's question about the existing chart toolbar: the separate Reports-tab PNG action was removed. The PR now changes the report download path so the saved HTML report includes a snapshot of the chart data already held by the UI.

The final commit also fixes a preexisting live-chart sampling mismatch discovered while validating the downloaded report image. The top-level `total_rps` fields intentionally represent total-run summary values, but chart history uses aggregate current stats. Since this PR now captures the live chart into the report, the chart-point source needs to match report history. I kept that fix here because it directly affects the report snapshot, but it can be split out if maintainers prefer.

### DIFF
--- a/locust/webui/src/components/LineChart/LineChart.tsx
+++ b/locust/webui/src/components/LineChart/LineChart.tsx
@@ -34,6 +34,8 @@ interface ILineChartProps<ChartType extends IBaseChartType> extends ILineChart<C
   xAxis?: XAXisComponentOption;
   grid?: GridComponentOption;
   isDarkMode?: boolean;
+  onChartReady?: (chart: ECharts) => void;
+  chartGroup?: string;
 }
 
 echartsUse([
@@ -60,10 +62,17 @@ export default function LineChart<ChartType extends IBaseChartType>({
   scatterplot,
   shouldReplaceMergeLines = false,
   isDarkMode = false,
+  onChartReady,
+  chartGroup = 'swarmCharts',
 }: ILineChartProps<ChartType>) {
   const [chart, setChart] = useState<ECharts | null>(null);
 
   const chartContainer = useRef<HTMLDivElement | null>(null);
+  const onChartReadyRef = useRef(onChartReady);
+
+  useEffect(() => {
+    onChartReadyRef.current = onChartReady;
+  }, [onChartReady]);
 
   useEffect(() => {
     if (!chartContainer.current) {
@@ -94,10 +103,11 @@ export default function LineChart<ChartType extends IBaseChartType>({
     };
     window.addEventListener('resize', handleChartResize);
 
-    initChart.group = 'swarmCharts';
-    connect('swarmCharts');
+    initChart.group = chartGroup;
+    connect(chartGroup);
 
     setChart(initChart);
+    onChartReadyRef.current?.(initChart);
 
     return () => {
       dispose(initChart);
--- a/locust/webui/src/components/Reports/Reports.test.tsx
+++ b/locust/webui/src/components/Reports/Reports.test.tsx
@@ -89,7 +89,7 @@ describe('Reports', () => {
 
   test('injects a client-side charts PNG when downloading the report from chart data', async () => {
     const reportHtml =
-      '<html><body><script>window.templateArgs = {"is_report":true}</script><script type="module" src="/assets/report.js"></script></body></html>';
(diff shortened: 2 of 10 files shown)


MichaelSpece:
CI note before marking this ready for review: the remaining failing Linux/Python checks appear unrelated to this PR.

They fail in `locust/test/test_main.py::TelemetryTests::test_otel_flag`, matching failures seen on current `master` and in PR #3399. A recent `master` commit also changed that exact test (`increase join timeout for --otel test cases`), and the WebUI-specific check for this change passes.

Relevant checks for this PR are green: `test_webui`, Ruff, Mypy, Docs, Docker, and Windows.

andrewbaldwin44:
Hey @MichaelSpece, welcome to Locust and thanks for your willingness to contribute! 

Maybe a question before I review the code, the charts tab already has a toolbar where you can click a button to download the charts as a .png, would this not suffice for your use-case?

MichaelSpece:
Thanks for the question. You're right: the existing chart toolbar already exports the connected chart group as one combined PNG, so this version is duplicating that capability rather than solving the report-download problem.

The original intent was to incorporate the chart image into the downloaded HTML report, using the chart data already accumulated in the GUI and avoiding an extra server transfer. This PR does not do that yet: the PNG is downloaded separately and is not included in the full report.

I'm going to revise the approach: remove the duplicate Reports-tab PNG action and instead make the `Download Report` flow generate a client-side chart snapshot from the existing chart history, inject it into the downloaded HTML report, and download that final report. I'll move the PR back to draft while making that change.

MichaelSpece:
Thanks again for pointing out the existing chart toolbar export, @andrewbaldwin44. I've pushed the revised approach now: the duplicate Reports-tab PNG action is gone, and `Download Report` captures the existing chart history as a client-side snapshot and embeds it into the downloaded HTML report instead.

Checks are green on `faf6349`, including `test_webui`, Ruff, Mypy, Docs, Docker, Windows, and ReadTheDocs.

MichaelSpece:
I pushed one small follow-up commit, `6bd22e1`, after validating the report snapshot path.

It fixes a preexisting live-chart sampling mismatch: the live chart was taking RPS/failures/s/average response time from top-level total-run fields, while report history samples those points from the aggregate current stats row. That matters here because this PR now captures the live chart into the downloaded report.

The navbar/report summary still uses the total-run values; only chart points now use the same aggregate current stats source as report history. I kept the fix in this PR because it affects the report PNG fidelity, but I can split it into a separate PR if preferred.

Local checks on the latest branch passed: lint, type-check, focused `useFetchStats` regression, full webui test suite, and `git diff --check`.

andrewbaldwin44:
Hi @MichaelSpece,

Sorry I'm still not very clear as to what the goal of this PR is, what problem are we trying to solve?

From the first iteration I saw a new link in the downloads tab to download a PNG snapshot of the charts. I asked if this link was necessary since we can already download these PNG's at any time from the report tab, or indeed from the HTML report itself.

image

I think replacing the intractable charts with static PNG images is a step backwards, especially if we aren't giving users an option between a "static" or "intractable" report.

Could you please open an issue as a bug or feature request, that way we can all understand the problem and agree on what should be implemented?

MichaelSpece:
Thanks, that makes sense. I opened #3407 to step back and agree on the user need/desired report behavior before pushing this implementation further.

I also converted this PR back to draft for now. The key point I took from your comment is that replacing interactive charts with a static image would be a regression unless that is an explicit, agreed report mode. I'll wait for the issue discussion before changing the PR further.

cyberw:
I think we can close this, right?

MichaelSpece:
Yes, that's fine with me. Thanks for the review and discussion here.

I've turned my attention toward `aiolocust`, so I'm happy to close this PR. #3407 can remain the place to revisit the report behavior if there's interest later.

PR #3401, opened 2026-04-29
fix: allow passing arbitrary OpenAI SDK arguments to OpenAIUser

zhangjiabo1107:
Currently, passing `base_url` or other OpenAI SDK parameters to `OpenAIUser` crashes because User.__init__ receives unexpected kwargs. This PR extracts OpenAI-specific kwargs and maps Locust's host attribute to OpenAI's `base_url`.

example:
```python
from locust.contrib.oai import OpenAIUser
from locust import run_single_user, task


class OpenRouterUser(OpenAIUser):
    def __init__(
        self,
        *args,
        **kwargs,
    ):
        kwargs["base_url"]="https://openrouter.ai/api/v1"
        kwargs["default_headers"] = {
            "X-OpenRouter-Title": "Locust",
            "HTTP-Referer": "https://locust.io/",
        }
        super().__init__(*args, **kwargs)

    @task
    def t(self):
        with self.client.rename_request("mini"):  # here's how to rename requests
            self.client.responses.create(
                model="gpt-4o-mini",
                instructions="You are a coding assistant that speaks like it were a Monty Python skit.",
                input="How do I check if a Python object is an instance of a class?",
            )


if __name__ == "__main__":
    run_single_user(OpenRouterUser)

```

And the [websuite](https://openrouter.ai/logs) shown:
image

Related documentation: https://openrouter.ai/docs/quickstart#using-the-openai-sdk


zhangjiabo1107:
```
locust/contrib/oai.py:59: error: Incompatible types in assignment (expression has type "str", variable has type "None")  [assignment]
locust/contrib/oai.py:73: error: "OpenAIClient" gets multiple values for keyword argument "request_event"  [misc]
locust/contrib/oai.py:73: error: "OpenAIClient" gets multiple values for keyword argument "user"  [misc]
```

This issue is pre-existing and not introduced by this PR. I'd prefer to address it separately to keep the scope of this change focused.

cyberw:
Happy to merge, but I'd actually prefer having the typing fix in the same PR if it is ok with you?

zhangjiabo1107:
Hi there, thanks for the review .

However, after diving deeper into the actual implementation of locust.contrib.oai, I've realized that the current architecture—which relies on httpx response event_hooks—is fundamentally flawed when it comes to streaming requests (stream=True). Since streaming is arguably the most critical use case for LLM load testing, I don't think it's worth patching the typing issues on an implementation that is structurally broken for its main purpose.

Just to summarize the underlying issues I found with the current httpx hook approach:
1. **It Blocks Streams & Ruins TTFB**: The request_end hook explicitly calls response.read(). For streaming requests, this blocks the entire response until all chunks are fully downloaded. It completely destroys the streaming nature and makes it impossible to measure the most important metric: Time To First Byte (TTFB).
2. **Crashes on SSE Responses**: OpenAI streaming uses Server-Sent Events (SSE) format (data: {...}\n\n). Calling response.json() in the hook directly raises a JSONDecodeError and crashes the test.
3. **Fails to Count Streaming Tokens**: Token usage in streams isn't located in a top-level JSON response. It requires parsing the final chunk (with stream_options={"include_usage": True}), which the current hook cannot handle at all.

Given these severe architectural limitations, fixing the mypy types feels like putting a band-aid on a broken implementation. A proper OpenAI load testing client shouldn't intercept raw httpx responses; instead, it should wrap the OpenAI Python SDK methods directly to handle generators properly and fire custom events for TTFB and Token metrics.

Because of this, I've decided not to pursue this fix and will be closing this PR. Honestly, I think locust.contrib.oai might need a complete redesign (or a clear disclaimer that it does not support streaming) in the future.

Thanks again for your time and guidance!

cyberw:
Sorry to hear that. Yea, its more a PoC than a proper implementation that works for all use cases. If you’re interested in remaking it in a way that works for you, let me know!

perhaps is would be easier to do using openai’s aiohttp based implementation and https://github.com/cyberw/aiolocust

PR #3403, opened 2026-04-30
Add first seen / last seen timestamps to failure stats

### LINKED ISSUE #3402: Show when each failure was first and last seen in the Failures tab
### Prerequisites

- [x] I am suggesting a new feature, not asking a question

## Problem

The Failures tab shows *what* errors occurred and *how many times*, but not *when*. During a load test I often want to know if an error started during ramp-up or only at full load, whether it's still happening, and which of several listed errors is the most recent.

## Proposal

Add **First Seen** and **Last Seen** columns to the Failures tab (and to the `failures.csv` download and the HTML report, which reuses the same component). Local time in the UI, UTC ISO in the CSV.

## Fits the existing model

`StatsEntry` already tracks `start_time` and `last_request_timestamp` per request. This mirrors that pattern on `StatsError`. In distributed mode the master merges worker reports with `min(first_seen)` / `max(last_seen)`, the same approach `StatsEntry.extend()` uses.

## Status

I have a working local implementation (~100 line diff, backend + frontend + 2 tests). Happy to open a PR if this direction sounds good.

Image

tugkanboz:
Closes #3402.

## What

Adds **First Seen** and **Last Seen** to each `StatsError`, surfaced in:

- The Web UI **Failures** tab (local time)
- `failures.csv` download (UTC ISO, two new columns at the end)
- The standalone HTML report (same `FailuresTable` component)

## How

`StatsError.occurred()` records `time.time()` — sets `first_seen` once, refreshes `last_seen` on every occurrence. This mirrors how `StatsEntry` already tracks `start_time` and `last_request_timestamp`.

In distributed mode the master merges incoming reports with `min(first_seen)` / `max(last_seen)`, matching the pattern in `StatsEntry.extend()`. Wire format keeps timestamps as floats (no parse cost on merge); the CSV writer formats them via the existing `format_utc_timestamp` helper. Older workers that don't send the new fields are handled via `data.get(...)`.

## Tests

Two new tests in `test_stats.py`:

- `test_error_first_seen_and_last_seen` — first_seen sticks, last_seen advances
- `test_error_master_merge_extends_first_and_last_seen` — distributed merge picks the widest window


locust/stats.py
```diff
@@ -815,7 +840,18 @@ def on_worker_report(client_id: str, data: dict[str, Any]) -> None:
             if error_key not in stats.errors:
                 stats.errors[error_key] = StatsError.unserialize(error)
             else:
-                stats.errors[error_key].occurrences += error["occurrences"]
+                existing = stats.errors[error_key]
+                existing.occurrences += error["occurrences"]
+                incoming_first = error.get("first_seen")
+                incoming_last = error.get("last_seen")
+                if incoming_first is not None:
```
cyberw: I think this is a nice place to use the walrus operator :)

if incoming_first := error.get("first_seen"):

cyberw (commented):
Looks great! One small wish from my side, but everything else looks good!

locust/stats.py
tugkanboz: Good call, applied!

PR #3405, opened 2026-05-01
fix(fasthttp): handle zlib.error for truncated gzip streams under high load

jorgetamayo21:
## Problem
Under high load, servers may close the connection before finishing 
sending the gzip-encoded body. This results in a truncated stream 
that zlib cannot decompress, raising an unhandled `zlib.error` that 
propagates out of `FastHttpSession.request()`.

Affected stacktrace:
    File "geventhttpclient/useragent.py", line 188, in unzipped
        return zlib.decompress(bodystr, 16 + zlib.MAX_WBITS)
    zlib.error: Error -5 while decompressing data: incomplete or truncated stream

## Fix
Add `zlib.error` to the existing exception handler alongside 
`HTTPParseError` so truncated gzip responses are recorded correctly 
in Locust's statistics instead of propagating as unhandled exceptions.

## Notes
Same pattern as the Brotli decompression failure — the existing 
handler already does the right thing, `zlib.error` just wasn't 
included.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -10,6 +10,7 @@ import re
 import socket
 import time
 import traceback
+import zlib
 from base64 import b64encode
 from contextlib import contextmanager
 from http.cookiejar import CookieJar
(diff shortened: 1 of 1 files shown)


cyberw:
Nice, but doesnt it make more sense to just add it to the list in FAILURE_EXCEPTIONS?

jorgetamayo21:
@cyberw done.

PR #3409, opened 2026-05-10
unify ruff in pyproject.toml and pre-commits

even-even:
Hello
In this PR:
1) run ruff check and format from pyproject.toml lib version (like run yarn)
2) up ruff version for latest
3) fix ruff warnings
4) minor fix doc about run ruff


cyberw:
great stuff!

PR #3415, opened 2026-05-21
Handle custom web routes without UI

puneetdixit200:
Fixes #3396.

## Summary
- add a falsey no-op web UI during init on worker/headless nodes so documented route decorators do not crash when no real web UI exists
- keep `environment.web_ui` restored to `None` after init when there is no real UI
- add a process-mode regression test that registers a custom route, verifies it on the master UI, and starts the test
- make the empty locustfile directory test wait for its expected process exit instead of sending SIGINT during teardown

## Tests
- `UV_CACHE_DIR=$PWD/.uv-cache uv run pytest locust/test/test_main.py::StandaloneIntegrationTests::test_error_when_locustfiles_directory_is_empty locust/test/test_main.py::DistributedIntegrationTests::test_processes_custom_web_route -q`
- `UV_CACHE_DIR=$PWD/.uv-cache uv run pytest locust/test/test_main.py::DistributedIntegrationTests::test_processes_custom_web_route locust/test/test_main.py::DistributedIntegrationTests::test_processes locust/test/test_main.py::StandaloneIntegrationTests::test_exception_in_init_event -q`
- `UV_CACHE_DIR=$PWD/.uv-cache uv run ruff check .`
- `UV_CACHE_DIR=$PWD/.uv-cache uv run ruff format --check`
- `UV_CACHE_DIR=$PWD/.uv-cache uv run mypy locust/`
- `git diff --check`

PR #3416, opened 2026-05-24
docs: guard custom web routes without web UI

puneetdixit200:
## Summary
- Update the custom web route example to guard `environment.web_ui` before registering Flask routes.
- Clarify that workers, headless mode, and `--processes` do not have a Web UI in every process.

Docs-only replacement requested in #3415.

## Validation
- `git diff --check`

AI assistance was used to prepare this docs-only replacement; I reviewed the final diff.

### DIFF
--- a/docs/extending-locust.rst
+++ b/docs/extending-locust.rst
@@ -108,11 +108,14 @@ to the Flask app instance and use that to set up a new route::
 
     @events.init.add_listener
     def on_locust_init(environment, **kw):
-        @environment.web_ui.app.route("/added_page")
-        def my_added_page():
-            return "Another page"
+        if environment.web_ui:
+            @environment.web_ui.app.route("/added_page")
+            def my_added_page():
+                return "Another page"
 
 You should now be able to start locust and browse to http://127.0.0.1:8089/added_page. Note that it doesn't get automatically added as a new tab - you'll need to enter the URL directly.
+The ``if environment.web_ui`` guard is needed when running workers, headless mode, or ``--processes``,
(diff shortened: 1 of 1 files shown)


puneetdixit200:
CI check: the current red `Tests` jobs do not appear to come from this docs-only diff (`docs/extending-locust.rst`).

What I checked:
- This PR: Python 3.11 and 3.13 fail in `StandaloneIntegrationTests.test_error_when_locustfiles_directory_is_empty` with return code `-2` instead of `1`.
- Latest `master` Tests run `26369517649` at `d90bc90a700b429ba64c84c432e80a3dfcd30ecf` fails the same empty-directory test on Python 3.10, 3.11, 3.13, and 3.14.
- This PR's Python 3.14 job instead hit `TestHttpSession.test_event_measure` with a timing value just above the assertion window; that is also outside the docs file changed here.

All non-matrix checks I saw for this PR are green, including Ruff, Mypy, Docs Build, Linux, Windows, Docker image tests, and ReadTheDocs.

puneetdixit200:
The remaining failures look unrelated to this docs-only change. Python 3.11 and 3.13 fail in StandaloneIntegrationTests.test_error_when_locustfiles_directory_is_empty because the subprocess exits with -2 instead of the expected 1 during cleanup, and Python 3.14 fails a tight timing assertion in TestHttpSession.test_event_measure at 1.596 ms versus the 1.5 ms upper bound. I tried to rerun the failed jobs, but GitHub requires repository admin rights for that.

cyberw:
Agree, dont worry about those failures.

PR #3425, opened 2026-06-17
fix: reject partially-matched timespan strings in parse_timespan

JSap0914:
## Bug

`parse_timespan` uses `re.match()` which anchors only at the **start** of the string. The regex pattern has every group optional (`?`), so it always produces a match object — even when the remaining input contains unrecognised characters. Strings like `"1h2m3"`, `"30m45"`, or `"2h3m5s6"` were silently accepted and the trailing digits discarded, returning a wrong second count instead of raising `ValueError`.

```python
>>> parse_timespan("1h2m3")  # should raise ValueError
3720                           # silently returns 1h2m, ignoring the trailing '3'
>>> parse_timespan("30m45")  # should raise ValueError
1800                           # silently returns 30m, ignoring the trailing '45'
```

## Fix

After matching, verify the matched span covers the **entire** input:

```python
if not parts or parts.group(0) != time_str:
    raise ValueError(...)
```

One-character change in `locust/util/timespan.py`.

## Verification

```
pytest locust/test/test_util.py -v
# 5 passed (4 existing + 1 new regression test)
```

### DIFF
--- a/locust/test/test_util.py
+++ b/locust/test/test_util.py
@@ -17,6 +17,14 @@ class TestParseTimespan(unittest.TestCase):
         self.assertEqual(7200, parse_timespan("2h"))
         self.assertEqual(3787, parse_timespan("1h3m7s"))
 
+    def test_parse_timespan_rejects_trailing_junk(self):
+        # Strings with digits after a valid pattern but missing a unit suffix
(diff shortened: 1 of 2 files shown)


cyberw:
Thanks!

PR #3432, opened 2026-06-29
Add perf ruff rule

even-even:
Hello

In PR i added https://docs.astral.sh/ruff/rules/#perflint-perf rule and fix warnings


cyberw:
nice! only one of those fixes was in any kind of hot path, but I dont mind the other ones either

PR #3434, opened 2026-06-29
fix: use warnings.warn(DeprecationWarning) in parse_options instead of print

### LINKED ISSUE #3433: parse_options() emits deprecation warning via print() instead of warnings.warn(DeprecationWarning)
## Bug description

`parse_options()` in `locust/argument_parser.py` signals its own deprecation with a bare `print()` call instead of `warnings.warn(msg, DeprecationWarning, stacklevel=2)`.

## Consequences

- The warning goes to **stdout** instead of stderr, breaking scripts that capture stdout.
- Python's `-W ignore::DeprecationWarning` flag **cannot suppress** this warning.
- `warnings.catch_warnings(record=True)` (used in test suites) never sees it.
- Every call prints the warning again; Python's once-per-call-site deduplication does not apply.
- The caller's file/line does not appear in the warning text.

## Reproduction

```python
import warnings
from locust.argument_parser import parse_options

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter('always')
    parse_options(args=['-f', 'locustfile.py'])

# Expected: len(caught) == 1 and caught[0].category is DeprecationWarning
# Actual:   len(caught) == 0  (warning was print()-ed to stdout instead)
print(len(caught))  # prints 0
```

## Expected behaviour

`parse_options()` should call `warnings.warn(msg, DeprecationWarning, stacklevel=2)` so the warning participates in Python's standard warning system, can be suppressed with `-W`, and is caught by `catch_warnings()`.

## Locust version

master (d473f655)

JSap0914:
Closes #3433

## Bug

`parse_options()` signalled its own deprecation with a bare `print()` call.
This caused the warning to be sent to **stdout** instead of stderr, made it impossible to suppress with `-W ignore::DeprecationWarning`, prevented callers from intercepting it with `catch_warnings()`, and caused it to appear on every call instead of once per call-site.

## Fix

Replace `print("Warning: ...")` with `warnings.warn(msg, DeprecationWarning, stacklevel=2)`.

## Verification

New test `TestParser.test_parse_options_emits_deprecation_warning` in `locust/test/test_parser.py`:

- **RED** (before fix): `catch_warnings(record=True)` captures 0 warnings — the message only appeared in captured stdout.
- **GREEN** (after fix): exactly one `DeprecationWarning` is captured by `catch_warnings`.

```
locust/test/test_parser.py::TestParser::test_parse_options_emits_deprecation_warning PASSED
```

All other `TestParser` and `test_util.py` tests continue to pass.

### DIFF
--- a/locust/argument_parser.py
+++ b/locust/argument_parser.py
@@ -13,6 +13,7 @@ import os
 import platform
 import socket
 import sys
+import warnings
 import tempfile
 import textwrap
 from collections import OrderedDict
@@ -910,7 +911,12 @@ def get_parser(default_config_files=DEFAULT_CONFIG_FILES) -> LocustArgumentParse
 
 
 def parse_options(args=None) -> configargparse.Namespace:
-    print("Warning: This function is deprecated and may be removed in a future version")
+    warnings.warn(
+        "parse_options() is deprecated and may be removed in a future version. "
+        "Use get_parser().parse_args() instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     parser = get_parser()
     parsed_opts = parser.parse_args(args=args)
     if parsed_opts.stats_history_enabled and (parsed_opts.csv_prefix is None):
--- a/locust/test/test_parser.py
+++ b/locust/test/test_parser.py
@@ -2,10 +2,12 @@ import locust
 from locust.argument_parser import (
     get_parser,
     parse_locustfile_paths,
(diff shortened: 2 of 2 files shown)


cyberw:
Why are you calling parse_options? I'd be more inclined to delete the function instead of fixing it :) It is untested nowadays. Use `get_parser().parse_args(args=args)` instead.

PR #3453, opened 2026-07-08
Bump Webui Deps

andrewbaldwin44:


cyberw:
I happened to be merging lots of python dependency PRs right now, so the github actions runner might be overwhelmed...

andrewbaldwin44:
Yeah something timed out, I re-ran and it passes now

cyberw:
Thanks a bunch! :)

PR #3469, opened 2026-07-20
Update swarm state with extraOptions

### LINKED ISSUE #3462: [Feature-Request][Web UI] Preserve Custom Parameters/Options values during edit
### Prerequisites

- [x] I am suggesting a new feature, not asking a question

### Description

Hello, it would be great if the custom values we defined on Web UI during test is preserved on Edit menu just like `Number of users` and `Ramp up`. Thank you in advance

Related PR/Issue:
-  #2531

yukha-dw:
Fixes #3462

Update the swarm state with updated `extraOptions` so that edit and start new forms will be provided with correct default values

On start, we can configure custom parameters:
image
Then, on Edit menu, the custom parameters are preserved
image

### DIFF
--- a/locust/webui/src/components/SwarmForm/SwarmForm.tsx
+++ b/locust/webui/src/components/SwarmForm/SwarmForm.tsx
@@ -31,7 +31,7 @@ import { useSelector } from 'redux/hooks';
 import { swarmActions } from 'redux/slice/swarm.slice';
 import { IRootState } from 'redux/store';
 import { ICustomInput } from 'types/form.types';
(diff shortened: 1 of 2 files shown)


yukha-dw:
Note:
I don't like to use placeholder here, but i am not sure how to do it without it
```TS
            defaultValue: (inputData as Record<string, any>)[key] ?? option.defaultValue,
```
https://github.com/yukha-dw/locust/blob/3e06ffcc222c8eb6c7ede794948c68ad59817177/locust/webui/src/components/SwarmForm/SwarmForm.tsx#L156

PR #3473, opened 2026-07-21
Fix response time percentiles being deflated by requests logged with response_time=None

### LINKED ISSUE #3472: get_response_time_percentile counts None (async) requests in the denominator, deflating all reported percentiles
### Prerequisites

- [x] I am using the latest version of Locust (reproduced on pip `locust==2.46.0` and on master `1878c01`)

### Description

Logging any request with `response_time=None` — the documented pattern for async / fire-and-forget requests (`num_none_requests` is a first-class field on `StatsEntry`, its docstring says "typically async requests") — deflates every reported response-time percentile from p50 through p99.9. `median_response_time` and `avg_response_time` stay correct on the same data.

**Root cause** (`locust/stats.py`, line numbers at master `1878c01`):

`StatsEntry.get_response_time_percentile` (stats.py:607) passes `self.num_requests` as the denominator:

```python
return calculate_response_time_percentile(self.response_times, self.num_requests, percent)
```

`self.num_requests` includes `num_none_requests`, but `self.response_times` — the histogram the percentile walks — holds only non-None samples (`_log_response_time`, stats.py:391-394, returns early on `None`). Inside `calculate_response_time_percentile` (stats.py:161-179) the inflated denominator pushes the descending nearest-rank loop past the correct bucket toward the minimum. Each None request effectively counts as an infinitely fast one. The two sibling metrics subtract the correction:

```python
# stats.py:430  avg
float(self.total_response_time) / (self.num_requests - self.num_none_requests)
# stats.py:438  median
median_from_dict(self.num_requests - self.num_none_requests, self.response_times)
```

The percentile alone omits `- self.num_none_requests`. The `calculate_response_time_percentile` docstring itself says `num_requests` "could be derived from response_times", i.e. it expects the non-None sample count.

### Steps to reproduce (no locust process needed)

```python
from locust.stats import StatsEntry, RequestStats
e = StatsEntry(RequestStats(False), "/api", "GET")
for rt in range(1, 101):
    e.log(rt, 0)      # 100 real requests, 1..100 ms
for _ in range(100):
    e.log(None, 0)    # 100 async requests (response_time=None)
print(e.get_response_time_percentile(0.5))  # actual: 1   expected: ~51
print(e.median_response_time)               # 50 (correct: excludes None)
```

With 100 real samples (1..100 ms) plus 100 None requests, p50 collapses to **1 ms** (vs 51 without None — a ~50x deflation), p95 reads 91 instead of 96, p99 reads 99 instead of 100. I checked the numbers against numpy `np.percentile` (methods lower/nearest/higher) and against a reimplementation of the same descending nearest-rank algorithm fed the correct denominator; without None requests the shipped code matches both. The wrong values reach the console percentile table (`StatsEntry.percentile`, stats.py:653), CSV export (`_percentile_fields`, stats.py:1035), and the JSON / web UI (`to_dict`, stats.py:687). Aggregation through `RequestStats.total` (the distributed path) carries the same error.

### Expected vs actual

- **Expected:** percentiles computed over the non-None sample count, consistent with `median`/`avg`.
- **Actual:** None requests are counted in the denominator, deflating every percentile.

### Suggested fix (matches median/avg)

```python
# stats.py:607
return calculate_response_time_percentile(
    self.response_times, self.num_requests - self.num_none_requests, percent)
```

The sliding-window cousin `get_current_response_time_percentile` (stats.py:609, windowed denominator at stats.py:647) shares the defect and the `CachedResponseTimes` namedtuple (stats.py:116) stores no `num_none_requests`; a complete fix caches it and subtracts its windowed diff too.

### Environment

- Locust 2.46.0 (also reproduced on master `1878c01`)
- Python 3.14.6
- macOS (Darwin); platform-independent (pure stats arithmetic)

---
*Found via property-based & differential bug-hunting, part of an effort to scale PBT (DepTyCheck-based) testing across the OSS ecosystem.*

*If this is intended / by-design: I'm really sorry, please just close it — no need to flag or ban me. I'm trying to scale property-based testing across the whole ecosystem and my publishing agents may have gotten this one wrong. I read every issue and follow up on each.*

haskiindahouse:
Fixes #3472

`get_response_time_percentile` passed `self.num_requests` as the denominator, but `self.response_times` only holds non-None samples, so every request logged with `response_time=None` deflated the percentiles (`avg_response_time` and `median_response_time` already subtract `num_none_requests`).

- `get_response_time_percentile`: subtract `num_none_requests`, same as avg/median
- `get_current_response_time_percentile`: `CachedResponseTimes` now also carries `num_none_requests` (with a default, so existing two-field constructions keep working) and the sliding-window denominator gets the same correction
- tests mirror `test_percentile` and `test_get_current_response_time_percentile` with None requests added; both fail without the fix

### DIFF
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -114,7 +114,9 @@ response time percentile
 CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW = 10
 
 CachedResponseTimes = namedtuple(
-    "CachedResponseTimes", ["response_times", "num_requests", "num_none_requests"], defaults=(0,)
+    "CachedResponseTimes",
+    ["response_times", "num_requests", "num_none_requests"],
+    defaults=(0,),  # num_none_requests is optional for backwards compatibility
 )
 
 PERCENTILES_TO_REPORT = [0.50, 0.66, 0.75, 0.80, 0.90, 0.95, 0.98, 0.99, 0.999, 0.9999, 1.0]
--- a/locust/stats.py
+++ b/locust/stats.py
@@ -113,7 +113,9 @@ response time percentile
 """
 CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW = 10
 
-CachedResponseTimes = namedtuple("CachedResponseTimes", ["response_times", "num_requests"])
+CachedResponseTimes = namedtuple(
+    "CachedResponseTimes", ["response_times", "num_requests", "num_none_requests"], defaults=(0,)
+)
 
 PERCENTILES_TO_REPORT = [0.50, 0.66, 0.75, 0.80, 0.90, 0.95, 0.98, 0.99, 0.999, 0.9999, 1.0]
 
@@ -604,7 +606,11 @@ class StatsEntry:
 
         Percent specified in range: 0.0 - 1.0
         """
-        return calculate_response_time_percentile(self.response_times, self.num_requests, percent)
+        # response_times only holds non-None samples, so None (async) requests
+        # must be excluded from the denominator, just like in avg/median
+        return calculate_response_time_percentile(
+            self.response_times, self.num_requests - self.num_none_requests, percent
+        )
 
     def get_current_response_time_percentile(self, percent: float) -> int | None:
         """
@@ -644,7 +650,7 @@ class StatsEntry:
             # for that timeframe
             return calculate_response_time_percentile(
                 diff_response_time_dicts(self.response_times, cached.response_times),
-                self.num_requests - cached.num_requests,
+                (self.num_requests - self.num_none_requests) - (cached.num_requests - cached.num_none_requests),
                 percent,
             )
         # if time was not in response times cache window
@@ -669,6 +675,7 @@ class StatsEntry:
         self.response_times_cache[t] = CachedResponseTimes(
             response_times=copy(self.response_times),
             num_requests=self.num_requests,
+            num_none_requests=self.num_none_requests,
         )
 
         # We'll use a cache size of CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW + 10 since - in the extreme case -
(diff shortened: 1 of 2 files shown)


locust/stats.py
```diff
@@ -113,7 +113,9 @@ def resize_handler(signum: int, frame: FrameType | None):
 """
 CURRENT_RESPONSE_TIME_PERCENTILE_WINDOW = 10
 
-CachedResponseTimes = namedtuple("CachedResponseTimes", ["response_times", "num_requests"])
+CachedResponseTimes = namedtuple(
+    "CachedResponseTimes", ["response_times", "num_requests", "num_none_requests"], defaults=(0,)
```
cyberw: can you add a comment about what `defaults=(0,)` means?

haskiindahouse: i leave com, is it ok?

cyberw: Awesome, thanks!

cyberw:
Hmm... Now I already merged it, but do you think it might make more sense to _not_ keep CachedResponseTimes backwards compatible? It is a very internal thing, so it isn't like we're breaking anyone's tests...

haskiindahouse:
> Hmm... Now I already merged it, but do you think it might make more sense to _not_ keep CachedResponseTimes backwards compatible? It is a very internal thing, so it isn't like we're breaking anyone's tests...

hm... I searched through codebase and yeah, you're right...

we can make the field required (I think it's actually safer) + silent default to zero is bad practice, totally agree with you

what i need to do?

cyberw:
It's a simple fix, I can do it, just wanted to confirm :)

cyberw:
https://github.com/locustio/locust/commit/963ad2a86ae10b24906b638b1d6f1c71d6652a17

PR #3483, opened 2026-07-26
[FEAT] : add active user metric as an observable gauge

RamachandraBhardwaj:
We can use Observable gauge (async instrument) rather than a normal gauge and having dispersed calls of users gauge being updated everywhere. This will have a callback that gets executed at particular time interval.

Verified the usage (increase and decrease) via a small test using  `LoadTestShape`  and console metrics exporter


RamachandraBhardwaj:
@cyberw please let me know if this design is okay ? We can further discuss about it

cyberw:
Let me know when you've added the test case!

karanpathak:
@cyberw, @RamachandraBhardwaj: Which issue does this PR address?

RamachandraBhardwaj:
> @cyberw, @RamachandraBhardwaj: Which issue does this PR address?

Hey @karanpathak ! This isn't specifically an issue but rather an observability side feature, to have number of users count in place with open telemetry. Please feel free to go through it. Thanks !

karanpathak:
Thanks for clarifying @RamachandraBhardwaj.

PR #3490, opened 2026-08-02
Fix filename unbound and print() in locustfile distribution

reachsridhard:
## Description

Fix two bugs in locustfile distribution and one logging inconsistency in RPC code.

### [runners.py](cci:7://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/runners.py:0:0-0:0) — `filename` possibly unbound in locustfile distribution error handler

When `locustfile_list` is empty (e.g. locustfile path is a directory with no `.py` files) or an exception occurs before the `for` loop assigns `filename`, the `except` block references `filename` which would raise `NameError`, crashing the master's [client_listener](cci:1://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/runners.py:973:4-1013:47) greenlet.

Additionally, `filename` was being reassigned with `os.path.basename(filename)` inside the loop, so if `f.read()` failed after the reassignment, the error message would show the basename instead of the full path.

### [zmqrpc.py](cci:7://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/rpc/zmqrpc.py:0:0-0:0) — `print()` instead of `logging.error()`

The [ipv4_only](cci:1://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/rpc/zmqrpc.py:66:4-75:20) method used `print()` for error logging, which bypasses logging configuration and can interfere with output parsing in headless mode. Replaced with `logging.error()` for consistency with the rest of the codebase.

## Changes

- [locust/runners.py](cci:7://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/runners.py:0:0-0:0): initialize `filename = None` before the try block; use `os.path.basename(filename)` directly in the dict instead of reassigning `filename`
- [locust/rpc/zmqrpc.py](cci:7://file:///Users/satyasivasundarsalagrama/open-source/locust/locust/rpc/zmqrpc.py:0:0-0:0): replace `print()` with `logging.error()`; add `import logging`

### DIFF
--- a/locust/rpc/zmqrpc.py
+++ b/locust/rpc/zmqrpc.py
@@ -1,3 +1,5 @@
+import logging
+
 from locust.exception import RPCError, RPCReceiveError, RPCSendError
 from locust.util.exception_handler import retry
 
@@ -69,7 +71,7 @@ class BaseSocket:
             if str(csocket.getaddrinfo(host, port, proto=csocket.IPPROTO_TCP)).find("Family.AF_INET6") == -1:
                 return True
         except gaierror as e:
-            print(f"Error resolving address: {e}")
+            logging.error(f"Error resolving address: {e}")
             return False
         return False
 
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1067,6 +1067,7 @@ class MasterRunner(DistributedRunner):
                     if os.path.isdir(locustfile_option):
                         locustfile_list.extend(get_abspaths_in(locustfile_option, extension=".py"))
 
+                filename = None
                 try:
                     locustfiles: list[str | dict[str, str]] = []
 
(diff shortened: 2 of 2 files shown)


cyberw:
Is this a real issue? I have no issues running locust against a dir without python files.

```
locust -f bar
Could not find any locustfiles in directory 'bar'
```

reachsridhard:
> Is this a real issue? I have no issues running locust against a dir without python files.
> 
> ```
> locust -f bar
> Could not find any locustfiles in directory 'bar'
> ```

@cyberw 

You're right, the empty directory case is already handled earlier. The filename = None initialization is defensive and unlikely to be triggered in practice. The more meaningful fix in this PR is the print() to logging.error() change in zmqrpc.py, which bypasses logging configuration. I can drop the runners.py changes if you'd prefer to keep this PR focused on just the logging fix, or close this PR entirely if the print() change isn't worth a separate PR either.

cyberw:
I think neither really make a lot of sense. Simplicity is key in the examples, that’s why it is a print rather than logging.

PR #3492, opened 2026-08-06
Fix FastHttpUser streaming response_length using wrong header key

reachsridhard:
# Fix: FastHttpUser streaming requests report `response_length=0`

## Summary

Fixes an issue where `FastHttpUser` with `stream=True` always reports `response_length=0`, even when the response includes a valid `Content-Length` header.

## Problem

`FastHttpSession.request()` incorrectly checks for `"response_length"` in the HTTP response headers:

```python
if stream:
    if response.headers and "response_length" in response.headers:
        request_meta["response_length"] = int(response.headers["response_length"])
```

`The correct header is `content-length`.

## Fix

Update the header lookup to use `content-length`:

```python
if stream:
    if response.headers and "content-length" in response.headers:
        request_meta["response_length"] = int(response.headers["content-length"])
```

This matches the existing behavior in `HttpSession` and correctly reports response sizes for streaming requests.

### DIFF
--- a/locust/contrib/fasthttp.py
+++ b/locust/contrib/fasthttp.py
@@ -273,8 +273,8 @@ class FastHttpSession:
         # get the length of the content, but if the argument stream is set to True, we take
         # the size from the content-length header, in order to not trigger fetching of the body
         if stream:
-            if response.headers and "response_length" in response.headers:
-                request_meta["response_length"] = int(response.headers["response_length"])
+            if response.headers and "content-length" in response.headers:
+                request_meta["response_length"] = int(response.headers["content-length"])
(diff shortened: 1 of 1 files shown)


cyberw:
Hi! You’re using AI to write this, right? You need to shorten/focus the description of the PR. I dont have time to read all that repetitive stuff.

the fix might be good, but the description basically looks like spam.

reachsridhard:
@cyberw 

Thanks for the feedback! I used AI to help structure the description, and I agree it ended up being more verbose and repetitive than necessary. I’ve shortened the description and kept it focused on the fix.

PR #3493, opened 2026-08-12
Fix KeyError crash in spawning_complete handler when worker is missing

reachsridhard:
# Fix `KeyError` in `spawning_complete` for missing workers

## Summary

Fixes a `KeyError` in the `spawning_complete` handler when a worker has already been removed from `self.clients` due to missed heartbeats.

## Problem

Unlike the `spawning` handler, `spawning_complete` accesses the worker without checking whether it still exists:

```python
self.clients[msg.node_id].state = STATE_RUNNING
self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
```

If the worker was removed and later sends `spawning_complete`, this raises a `KeyError` and can crash the master's `client_listener` greenlet.

## Fix

Add the same `KeyError` handling already used by the `spawning` handler:

```python
try:
    self.clients[msg.node_id].state = STATE_RUNNING
    self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
except KeyError:
    logger.warning(
        f"Got spawning_complete message from unknown worker {msg.node_id}. Asking worker to quit."
    )
    self.server.send_to_client(Message("quit", None, msg.node_id))
```

Unknown workers are now logged and asked to quit instead of crashing the listener.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1147,7 +1147,9 @@ class MasterRunner(DistributedRunner):
                     self.clients[msg.node_id].state = STATE_RUNNING
                     self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
                 except KeyError:
-                    logger.warning(f"Got spawning_complete message from unknown worker {msg.node_id}. Asking worker to quit.")
+                    logger.warning(
+                        f"Got spawning_complete message from unknown worker {msg.node_id}. Asking worker to quit."
+                    )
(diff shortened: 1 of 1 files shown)


cyberw:
Nice, but I prefer checking `if msg.node_id in self.clients:` instead of catching an exception, as it is more clear and has no risk of accidentally catching some other keyerror

PR #3501, opened 2026-08-21
Fix master stopping test prematurely when a missing worker quits

reachsridhard:
# Fix premature test stop when a missing worker quits

## Summary

Fixes an issue where `MasterRunner` can stop the entire test when a missing worker sends a `"quit"` message while other workers are still running.

## Problem

The `"quit"` handler currently checks:

```python
if self.worker_count - len(self.clients.missing) <= 0:
    logger.info("The last worker quit, stopping test.")
    self.stop()
```

However, `worker_count` already excludes missing workers:

```python
return (
    len(self.clients.ready)
    + len(self.clients.spawning)
    + len(self.clients.running)
)
```

Subtracting `len(self.clients.missing)` again can incorrectly make the count reach zero and stop the test while healthy workers are still running.

## Fix

Remove the redundant subtraction:

```diff
- if self.worker_count - len(self.clients.missing) <= 0:
+ if self.worker_count <= 0:
```

This is consistent with the existing fix in `heartbeat_worker()` for the same worker-counting issue.

### Testing

Added `test_quit_from_missing_worker_does_not_stop_test_if_other_workers_are_running` to verify that a "quit" message from a missing worker does not stop the test when another worker is still running.

### DIFF
--- a/locust/runners.py
+++ b/locust/runners.py
@@ -1159,7 +1159,7 @@ class MasterRunner(DistributedRunner):
                     logger.info(
                         f"Worker {msg.node_id!r} (index {self.get_worker_index(msg.node_id)}) quit. {len(self.clients.ready)} workers ready."
                     )
-                    if self.worker_count - len(self.clients.missing) <= 0:
+                    if self.worker_count <= 0:
                         logger.info("The last worker quit, stopping test.")
                         self.stop()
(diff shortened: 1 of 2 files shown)


cyberw:
Looks good, thanks!

PR #3505, opened 2026-08-31
chore: up mypy to 2.3.1 and fixes warnings

even-even:
Hello

In PR:
1) up mypy to 2.3.1 (latest)
2) up typing-exceptions to 4.16 (for supporting latest mypy)
3) fixes warns (example - https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides)


cyberw:
Having to ignore so many warnings is unfortunate. Long term I want to switch to Pyright. That would be a bigger change ofc, but it kind of makes it not worth merging updates to mypy at the moment.

even-even:
> Having to ignore so many warnings is unfortunate. Long term I want to switch to Pyright. That would be a bigger change ofc, but it kind of makes it not worth merging updates to mypy at the moment.

Ok, I'm closing this PR.
Thanks for answer.

PR #3508, opened 2026-09-10
Compute Med. with the same rank as the 50% percentile

### LINKED ISSUE #3503: "Med." column and "50%" percentile report different values for identical data (lower- vs upper-median conventions)
## Summary

Locust computes its two headline 50%-of-request metrics with two different median conventions:

- `median_response_time` (console "Med." column, CSV/HTML fields) uses `median_from_dict`, a **lower-median** rule.
- `get_response_time_percentile(0.5)` / `calculate_response_time_percentile` (percentile rows/tables reporting "50%") implements an **upper-median** rule.

For any even total number of requests spanning more than one response-time bucket, both numbers are shown at the same time and disagree - so at least one always-reported statistic is wrong relative to any consistent definition of median.

Static-analysis finding based on reading `master`; not executed here.

## Location

- File: `locust/stats.py`
- Functions: `calculate_response_time_percentile` (lines ~161-179) and `median_from_dict` (lines ~815-826)
- Relevant code:

```python
def calculate_response_time_percentile(response_times, num_requests, percent):
    num_of_request = int(num_requests * percent)
    processed_count = 0
    for response_time in sorted(response_times.keys(), reverse=True):
        processed_count += response_times[response_time]
        if num_requests - processed_count <= num_of_request:
            return response_time
    return 0
```

```python
def median_from_dict(total, count):
    pos = (total - 1) / 2
    for k in sorted(count.keys()):
        if pos < count[k]:
            return k
        pos -= count[k]
    return k
```

## Problem

With `response_times = {100: 1, 200: 1}` (two requests):

1. Median path (`median_from_dict(2, {...})`): `pos = 0.5`; first ascending key `100` satisfies `0.5 < 1` -> returns **100** (lower median).
2. Percentile path (`calculate_response_time_percentile(..., 2, 0.50)`): `num_of_request = int(2*0.5) = 1`; descending keys `[200, 100]`; after key `200`, `processed_count=1` and `num_requests - processed_count = 1 <= 1` -> returns **200** (upper median).

Both are rendered concurrently: `StatsEntry.median_response_time` (stats.py ~line 438) feeds the summary "Med." column, while `get_response_time_percentile(0.5)` (called from stats.py lines ~609/~649) feeds the reported 50% percentile in console output, CSV (`stats.csv`/`percentiles.csv`) and the HTML report.

## Trigger / Reproduction

Based on source reading: run any load test that ends with an **even** number of requests whose rounded response-time buckets fall into >= 2 distinct buckets, e.g. exactly two requests of 100 ms and 200 ms. Console summary shows `Med. 100` while the percentile table shows `50% 200`.

## Expected Behavior

Both displayed statistics should derive from one convention (whichever is chosen - lower, upper, or nearest-rank median) so that `Med.` equals the reported `50%` value for the same aggregation window.

## Actual Behavior

They diverge whenever the request count is even across bucket boundaries; with large bucket widths (buckets grow up to 1000 ms steps at higher latencies per `bucket_response_time`/`round_response_time`), the gap between the two reported numbers can reach an entire bucket.

## Impact

Users compare `Med.` against the `50%` column to sanity-check their load test; the built-in inconsistency makes such checks contradictory and erodes trust in the reported statistics. Automated consumers of both CSV fields receive two different "medians" for the same data.

## Suggested Direction

Route the 50% percentile through `median_from_dict` (or make `median_from_dict` delegate to `calculate_response_time_percentile` with `percent=0.5`), keeping a single implementation of the convention; document the chosen rule in the docstrings.

## Evidence

- The two functions implement provably different order statistics for even-sized samples (traced above).
- Both feed user-visible outputs simultaneously via `StatsEntry.median_response_time` and `PERCENTILES_TO_REPORT`-driven percentile reporting.

gyanu2507:
Fixes #3503.

Console/CSV `Med.` used `median_from_dict`, which walked buckets ascending with `(n-1)/2`. The percentile table's `50%` uses `calculate_response_time_percentile(..., 0.5)`, which walks descending with `int(n * 0.5)`. Two requests at 100 ms and 200 ms showed `Med. = 100` and `50% = 200`.

`median_response_time` now calls that percentile helper directly, so both numbers use the upper-median rank. Odd counts were already the same. Existing even-sample tests that encoded the lower median are updated.


cyberw:
Interesting. Sounds like a problem worth fixing. But remove the median_from_dict function while you are at it, now that it is just a one-liner.

gyanu2507:
Removed `median_from_dict`. `median_response_time` now calls `calculate_response_time_percentile(..., 0.5)` directly, so Med. and the 50% column share one rank implementation.

PR #3509, opened 2026-09-11
Fix negative --users causing infinite hang in headless mode

reachsridhard:
# Fix infinite hang with negative `--users`

## Summary

Fixes `locust --headless -u -1 -r 1` hanging indefinitely.

Negative user counts were not validated, causing `UsersDispatcher` to enter an infinite dispatch loop because the current user count could never reach the negative target.

## Fix

* Added `ge_zero` validator for `--users`, while keeping `-u 0` valid.
* Added a `ValueError` guard in `UsersDispatcher.new_dispatch()` for negative target counts, covering programmatic API usage and custom `LoadTestShape`s.


cyberw:
I agree that it should be fixed, but just crash with an error message. 2-3 lines of code at most. No need for unit tests either as it is not a big issue.

Khael20:
ieo

reachsridhard:
@cyberw 

Sure, trimmed it down to just the ValueError guard in UsersDispatcher.new_dispatch(). Removed the argparse validator and the unit tests.

thnx!!

PR #3510, opened 2026-09-13
Add wait_time.poisson() for Poisson-distributed task arrivals

whr313703:
constant_throughput() produces a fixed pace: exactly N task runs per second, so every user is perfectly evenly spaced. Real users are better modelled as a Poisson process - a stable average rate with randomly distributed intervals, which naturally produces bursts and quiet periods.poisson(rate) returns wait times sampled from an exponential distribution with mean 1/rate seconds, so tasks arrive as a Poisson process with an average rate of rate. It is the random counterpart of constant_throughput: use the former to reproduce realistic traffic, the latter for a deterministic baseline.Also exported from the locust package, documented alongside the other wait_time functions, and covered by tests for both the sampled distribution and the argument validation.


cyberw:
Very nice!

PR #3516, opened 2026-09-16
Don't let a spawning_complete from an unknown worker kill client_listener

### LINKED ISSUE #3515: MasterRunner.client_listener dies permanently with KeyError on spawning_complete from an unknown worker
## Summary

An unguarded dict lookup in the `spawning_complete` branch of `MasterRunner.handle_message` raises `KeyError` when a worker that has already been removed from `self.clients` sends that message. The exception propagates out of `client_listener`, which is the greenlet that processes **all** inbound worker messages, and kills it for the lifetime of the process.

The master does not exit. The web UI keeps serving, the ZMQ port keeps accepting TCP connections, and the process sits at near-zero CPU — but it never again acknowledges a worker. Every worker retries for `CONNECT_RETRY_COUNT * CONNECT_TIMEOUT` (5 minutes), exits 255, restarts, and fails again. The UI reports:

> You can't start a distributed test before at least one worker process has connected

This looks like a networking or DNS problem and is quite hard to attribute without finding the traceback, because the master is still up and still answering HTTP.

## The code

`locust/runners.py`, current `master` (a9d20d6):

```python
case "spawning":
    try:
        self.clients[msg.node_id].state = STATE_SPAWNING
    except KeyError:
        logger.warning(f"Got spawning message from unknown worker {msg.node_id}. Asking worker to quit.")
        self.server.send_to_client(Message("quit", None, msg.node_id))
case "spawning_complete":
    # a worker finished spawning (this happens multiple times during rampup)
    self.clients[msg.node_id].state = STATE_RUNNING          # <-- unguarded
    self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]   # <-- unguarded
```

The adjacent `spawning` case already handles exactly this situation. It was guarded in d73612f9 ("Fix unknown worker spawning message", released in 2.15.0), and that commit's diff shows `spawning_complete` sitting directly underneath with the identical exposure, untouched:

```diff
             elif msg.type == "spawning":
-                self.clients[msg.node_id].state = STATE_SPAWNING
+                try:
+                    self.clients[msg.node_id].state = STATE_SPAWNING
+                except KeyError:
+                    logger.warning(f"Got spawning message from unknown worker {msg.node_id}. Asking worker to quit.")
+                    self.server.send_to_client(Message("quit", None, msg.node_id))
             elif msg.type == "spawning_complete":
                 self.clients[msg.node_id].state = STATE_RUNNING
                 self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
```

So this is the same defect, one case over, that has simply never been hit often enough to get reported.

## How a worker becomes unknown

`heartbeat_worker` marks a client `STATE_MISSING` once its heartbeat count goes negative, and prunes it from `self.clients` entirely once it reaches `HEARTBEAT_DEAD_INTERNAL`. A worker can be pruned and then still have an in-flight or subsequent `spawning_complete` on the wire — the same race the `spawning` guard was written for. `on_worker_report` guards it too, which is why the log shows `Discarded report from unrecognized worker` for the same node id immediately before the crash.

## Observed

Locust 2.46.5, Python 3.11.16, distributed mode on Kubernetes: 1 master, 90 worker pods each running `--processes 8`, so 720 worker processes against one master. The master pod was replaced, all workers were orphaned at once, and on reconnect:

```
INFO/locust.runners: Discarded report from unrecognized worker locust-sf-web-worker-6db4c68ffd-lx7bc_1c66b78ad27042d2af9212aa5ef1698b
INFO/locust.runners: Discarded report from unrecognized worker locust-sf-web-worker-6db4c68ffd-lx7bc_1c66b78ad27042d2af9212aa5ef1698b
  File "src/gevent/greenlet.py", line 900, in gevent._gevent_cgreenlet.Greenlet.run
  File "/usr/local/lib/python3.11/site-packages/locust/runners.py", line 1015, in client_listener
    self.handle_message(client_id, msg)
  File "/usr/local/lib/python3.11/site-packages/locust/runners.py", line 1146, in handle_message
    self.clients[msg.node_id].state = STATE_RUNNING
  File "/usr/local/lib/python3.11/site-packages/locust/runners.py", line 640, in __getitem__
    return self._worker_nodes[k]
KeyError: 'locust-sf-web-worker-6db4c68ffd-lx7bc_1c66b78ad27042d2af9212aa5ef1698b'
<Greenlet at 0xfffeeacf4b80: <bound method MasterRunner.client_listener of <locust.runners.MasterRunner object>>> failed with KeyError

CRITICAL/locust.runners: Unhandled exception in greenlet: <Greenlet ... MasterRunner.client_listener ...>
```

Note the `Discarded report from unrecognized worker` lines for the same node id immediately before — the client was already gone, and the guarded handler dealt with it correctly while the unguarded one did not.

After that line the master logged nothing further. State from outside the process:

- `/stats/requests` served normally: `{"state": "stopped", "workers": []}`
- TCP connect to the master's port 5557 from a worker pod: succeeds
- master container CPU: 4m (idle)
- workers: `Failed to connect to master ...:5557, retry 29/60.` indefinitely

So from every external signal the master is healthy; only the message loop is gone.

## Why it is worth fixing even though the trigger is a race

1. **The failure is permanent and silent.** One `KeyError` on one message from one stale worker takes the master out for good. There is no restart of the greenlet and no non-CRITICAL indication; the process keeps serving HTTP.
2. **The blast radius is the whole run.** Every worker then burns its full 5-minute reconnect budget and exits 255.
3. **Probability scales with worker count.** With `--processes` the number of worker identities is pods x processes, and any event that orphans them all at once (a master restart, a rolling update, a node drain) produces a burst of prune-then-reconnect churn. At ~720 identities we hit it on essentially every master replacement; at 90 we had not seen it.
4. **The fix already exists in the file**, one case above.

## Suggested fix

Mirror the `spawning` guard. PR to follow.

```python
case "spawning_complete":
    # a worker finished spawning (this happens multiple times during rampup)
    try:
        self.clients[msg.node_id].state = STATE_RUNNING
        self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
    except KeyError:
        logger.warning(f"Got spawning_complete message from unknown worker {msg.node_id}. Asking worker to quit.")
        self.server.send_to_client(Message("quit", None, msg.node_id))
```

A broader hardening would be to wrap `handle_message` so that no single malformed or stale message can kill `client_listener`, since the greenlet is a single point of failure for the entire master. Happy to follow up with that separately if you would prefer it over (or in addition to) the targeted guard.

## Environment

| | |
|---|---|
| locust | 2.46.5 (also present on `master` @ a9d20d6) |
| python | 3.11.16 |
| gevent | 25.5.1 |
| pyzmq | 27.2.0 |
| OS | Linux, aarch64 (container) |
| mode | distributed, 1 master / 90 worker pods x `--processes 8` = 720 worker processes |

ergdevops:
Fixes #3515

## Problem

`MasterRunner.handle_message` indexes `self.clients` unguarded in the `spawning_complete` branch:

```python
case "spawning_complete":
    # a worker finished spawning (this happens multiple times during rampup)
    self.clients[msg.node_id].state = STATE_RUNNING
    self.clients[msg.node_id].user_classes_count = msg.data["user_classes_count"]
```

A worker that `heartbeat_worker` has already pruned from `self.clients` can still have a `spawning_complete` in flight. The lookup then raises `KeyError`, which propagates out of `handle_message` and out of `client_listener` — the greenlet that processes **every** inbound worker message. Nothing restarts it.

The master does not exit. It keeps serving the web UI and keeps accepting TCP on the ZMQ port, but never acknowledges another worker. Workers burn their full `CONNECT_RETRY_COUNT * CONNECT_TIMEOUT` budget, exit, restart, and fail the same way, while the UI reports:

> You can't start a distributed test before at least one worker process has connected

Because every external signal says the master is healthy, this is hard to attribute without finding the traceback.

## Why this one and not its neighbour

The adjacent `spawning` case has handled exactly this since d73612f9 ("Fix unknown worker spawning message", released in 2.15.0). That commit's own diff shows `spawning_complete` sitting directly underneath with identical exposure, untouched:

```diff
             elif msg.type == "spawning":
-                self.clients[msg.node_id].state = STATE_SPAWNING
+                try:
+                    self.clients[msg.node_id].state = STATE_SPAWNING
+                except KeyError:
+                    logger.warning(f"Got spawning message from unknown worker {msg.node_id}. Asking worker to quit.")
+                    self.server.send_to_client(Message("quit", None, msg.node_id))
             elif msg.type == "spawning_complete":
                 self.clients[msg.node_id].state = STATE_RUNNING
```

This applies the same guard, with the same wording and the same `quit` response, so the two branches behave consistently.

## Test

The added test reproduces the production failure rather than just the exception: it registers a worker, deletes it from `master.clients` the way `heartbeat_worker` prunes a dead one, sends `spawning_complete` from that now-unknown id, and then asserts the master still processes a subsequent `client_ready`.

Without the fix it fails with the real symptom:

```
KeyError: 'fake_client'
<Greenlet ... MasterRunner.client_listener ...> failed with KeyError
AssertionError: 'fake_client2' not found in <locust.runners.WorkerNodes object>
  : master stopped accepting workers after the unknown-worker message
```

With the fix:

```
1 passed
```

Full `TestMasterRunner` suite: `39 passed`.

## How it was hit

Distributed run on Kubernetes, locust 2.46.5: 1 master and 90 worker pods each running `--processes 8`, so ~720 worker processes. Replacing the master pod orphaned all of them at once, and the resulting prune-then-reconnect burst triggered it reliably. At 90 worker identities we had never seen it, which is probably why this has gone unreported — the probability scales with the number of worker identities, and `--processes` multiplies them.

## Note

A broader hardening would be to wrap `handle_message` so no single stale or malformed message can kill `client_listener`, since that greenlet is a single point of failure for the whole master. I kept this PR to the targeted guard to match the existing pattern, but I'm happy to follow up with the broader change if you'd prefer it.


cyberw:
Nice!

ergdevops:
thank you @cyberw  for reviewing and merging, could you please help with a release as well as its blocking some of our tests.

cyberw:
Done!

PR #3517, opened 2026-09-20
Fix web UI user count not persisting in parsed_options

saiyaswanth883:
## Summary
- Store the swarm form user count in `parsed_options.num_users` so a page refresh shows the last submitted value instead of falling back to 1.
- Assert this in the existing web UI swarm test.

## Test plan
- [ ] Start Locust with the web UI, set Number of users to something other than 1, start the test, refresh the page, confirm the form still shows that value.
- [ ] `pytest locust/test/test_web.py::TestWebUI::test_swarm_host_value_specified`


saiyaswanth883:
Related bug report: https://github.com/locustio/locust/issues/3518

PR #3519, opened 2026-09-20
chore: add RUF ruff rules

even-even:
Hello

In this PR:
1) add https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf ruff rules
2) fixes warns


cyberw:
cool stuff! couple of things:

to me, `next(iter(self.stats.errors.values())` reads worse than `list(self.stats.errors.values())[0]`

also, `assert (foo := resp.js["foo"])` should probably be fixed instead of adding `# noqa: RUF018`!

even-even:
> cool stuff! couple of things:
> 
> to me, `next(iter(self.stats.errors.values())` reads worse than `list(self.stats.errors.values())[0]`
> 
> also, `assert (foo := resp.js["foo"])` should probably be fixed instead of adding `# noqa: RUF018`!

done
1) ignore next() - RUF015
2) delete valrus - fix RUF018

